PR #1207, opened 2019-07-27
Set a strict Content-Security-Policy and reorganize things to make it work

startling:
This may be risky to deploy -- I *think* it works, but I've only tested in dev. Sorry that I won't be around to help!

https://trello.com/c/I120p7Ck/1474-get-disciplined-about-where-we-load-resources-from-add-a-content-security-policy-to-protect-from-xss

This sets a Content-Security policy that restricts what javascript and other resources get loaded. This means that, if we've done it correctly, XSS is basically impossible! This is a huge security improvement and, even though it adds some complexity, I think it's worth it.

+ Set a CSP policy that, in production, prevents loading Javascript from anywhere other than `darklang.com`, `static.darklang.com`, and prevents connecting to anywhere other than  `darklang.com`, `static.darklang.com`, and Pusher.
+ Disable pusher's internal operational stats reporting, [it's optional](https://support.pusher.com/hc/en-us/articles/360020083153) and loading a [JSONP script breaks CSP.](https://storage.googleapis.com/pub-tools-public-publication-data/pdf/45542.pdf)
+ Hash the dynamically-created Javascript constants before sticking them in the html, and add that hash to the HTTP header so that it's allowed by the policy. 
+ Instead of fetching scripts and concatenating them with the Rollbar config on the frontend to dynamically create a Javascript file and start a WebWorker, do those fetches on the backend and create a webworker from the URL. This is the biggest piece of added complexity, I think. I also put these at `darklang.com/a/webworker.js` and `darklang.com/a/fetcher.js`, which is an awkward place to put them, but I'm not sure what would be better (`darklang.com/a/lizzie/fetcher.js?`).
+ Move the `onKeyDown` and `onKeyUp` properties away from the HTML and into the `appsupport.js`, so that it's part of a trusted.

This might need a little more work before merging -- it _is_ risky to deploy, and also it might be worth picking a better route (n.b., will need to update the nginx configuration to make it go to the backend instead of to the external site), and an integration test [is failing in a strange way](https://19999-75570901-gh.circle-artifacts.com/1/root/project/artifacts/rundir/videos/pressing_up_doesnt_return_to_start.mp4) -- so I marked it DO-NOT-MERGE so no one idly merges it.

---

- [x] Trello link included
- [x] Discussed goals, problem and solution.
- [x] Information from this description is also in comments
  - [ ] No useful information
  - [x] Screenshots aren't useful
- [ ] Intended followups are trelloed.
  - [x] No followups
- [ ] Reversion plan exists
  - [ ] Unnecessary
- [x] Tests are included (required for regressions)

Reviewer checklist:
- Product:
  - [ ] PR matches stated goal and Trello ticket.
  - [ ] Out-of-scope product changes have been explained.
  - [ ] I pulled the branch and tested out the feature.
- User facing:
  - [ ] Existing stdlib and language semantics are unchanged.
  - [ ] Existing granduser HTTP responses are unchanged.
  - [ ] All existing canvases should continue to work.
  - [ ] New features are documented in the User Manual or Trello filed.
- Engineering:
  - [ ] Tests are included or unnecessary (required for regressions).
  - [ ] Functions and variables are well-named and self-documenting.
  - [ ] Comments have been added for all explanations in PR review comment.
  - [ ] Serialization format changes look good and have been double-checked and tested against local prodclone.
  - [ ] Unneeded code has been removed.


pbiggar:
Getting rid of the client

PR #2031, opened 2020-02-25
Add ppx-deriving dev dependency to esy.json

jceipek:
While getting autocomplete and type information working in my editor again, I discovered two issues:

1. The esy version we are using is incompatible with our node version (see https://github.com/esy/esy/pull/1018)
Screen Shot 2020-02-24 at 10 00 36 AM

2. We didn't tell esy about our dependency on `ppx-deriving`.
Screen Shot 2020-02-24 at 10 14 14 AM

The first issue is annoying to fix -- I couldn't update `esy` without it complaining that it needed a version of OCaml that is incompatible with bucklescript (bucklescript relies on a patched version of OCaml version 4.06) but the patch is super simple, so for now I'm adding it locally:

```
Module._findPath = function(request, paths, isMain) {
+    paths = paths || [];
```

(In prior versions of node, `paths` could not be `null` but now it can)

The second problem requires a change to `esy.json`. This PR adds the missing dependency.


pbiggar:
Closing this as going to switch away from OCaml instead.

PR #2536, opened 2020-06-06
Add support to Visual Studio Codespaces

Maxim-Filimonov:
**Problem:**
My machine is quite old and unable to compile darklang without spinning all fans like crazy for 40 minutes and crashing after that.
**Documentation:**
- [Codespaces](https://code.visualstudio.com/docs/remote/codespaces)
- Not sure where to put docs on how to use this for other devs. Readme?
**Solution:**
Using the power of virtualization we can get any kind of machine. With the latest additions to VSCode and Visual studio, the setup to be productive in a remote environment is minimal. 
With this setup
```
scripts/builder --compile --watch --test
```
Can be run in either:
- A docker container which removes the need to have python3 and fswatch on the host machine.
- For my case in codespace - it will use the same config.

**Limitations:**
- It's required to clone the GitHub repo using **https** not via ssh. As far as I understand the reason is that for now codespaces don't yet support proxying of ssh keys into the codespace.
- To be investigated how does it work with dnsmasq - this is only initial setup. Will be doing more PR's which will add port forwarding and etc.

- [N/A ] Trello link included
- [x ] Discussed goals, problem and solution
- [ x] Information from this description is also in comments
  - [ ] No useful information
  - [x ] Screenshots aren't useful
  - [ ] No followups
- [ ] Reversion plan exists
  - [ ] Standard git revert is fine
- [ ] Tests are included (required for regressions)
- [ ] Specs (docs/trello) are linked in code 
  - [x ] No spec exists

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,73 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+# Note: You can use any Debian/Ubuntu based image you want. 
+FROM ubuntu:18.04@sha256:3235326357dfb65f1781dbc4df3b834546d8bf914e82cce58e6e6b676e23ce8f
+
+# This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
+# property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
+# will be updated to match your local UID/GID (when using the dockerFile property).
+# See https://aka.ms/vscode-remote/containers/non-root-user for details.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+USER root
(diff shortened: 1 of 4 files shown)


Maxim-Filimonov:
This is probably obsolete due to MS switch to github codespaces which has different configuration.

pbiggar:
Thanks for looking at this! Since you first looked, I started using VSCode remote containers and so fixed a few things that I think would simplify this too.

Maxim-Filimonov:
@pbiggar ah nice as this repo is now public GitHub code spaces should be free to use here.

pbiggar:
Cool. If you or anyone else tries it out, would love to know how it goes or if we could improve it!

Maxim-Filimonov:
I just tried to create code space but it didn't work.
The [image](https://github.com/darklang/dark/blob/4b80a28e28a5ef4d5ec0fb41e53aca9e372e8f35/.devcontainer/devcontainer.json#L8) being used doesn't seem to be valid image on docker hub.

Does this needs to change to dark-base/

pbiggar:
This is the docker project: https://hub.docker.com/repository/docker/darklang/dark-rust. So I guess it needs to be darklang/dark-rust?

Maxim-Filimonov:
Yes, I believe that would work.

pbiggar:
Thanks @Maxim-Filimonov, I'll make the change and follow up.

PR #2571, opened 2020-06-09
Implement function description using BlankOrs

jceipek:
https://trello.com/c/btt9BT48/3058-re-implement-docstring-spec

This makes user function descriptions editable and displays function descriptions for user functions and package manager functions. This is a new attempt at https://github.com/darklang/dark/pull/2394 using `blankOr`s instead. Spec is here: https://www.notion.so/darklang/MVP-for-User-function-description-1c2d6540d3f94e52b45935e11d679b50

## Before this PR:
Screen Shot 2020-06-09 at 2 14 33 PM

## Without a Description typed:
Screen Shot 2020-06-09 at 2 12 20 PM

## With a description typed:
Screen Shot 2020-06-09 at 2 13 39 PM

## Package Manager function:
Screen Shot 2020-06-09 at 2 18 19 PM

# Relative to the previous PR
- The implementation extends `blankOr`s and autocomplete instead of introducing a new `cursorState`
- The visuals look closer to the spec thanks to @sydNoteboom
- Only User Function descriptions with Read/Write permissions can be edited
- The integration test asserts that the model has updated and checks the initial state as a precondition.

# Quirks
- This uses autocomplete and changes a central integration point of autocomplete in order to commit on every keystroke. I haven't been able to find any negative consequences, but they likely exist. The fact that this uses autocomplete at all seems problematic, because descriptions are not autocompletable and will likely never be, but I don't know how to bypass it properly.
- Clicking in a multiline blankOr misplaces the caret. This is because blankOrs in general are divs that become input boxes only when you click in them, and we use some hacky math to place the caret during the transition from div to input, which relies on assumptions such as single-line, non-wrapping input. That entire process is unnecessary but fixing it is out of scope IMO.
- [x] Trello link included
- [x] Discussed goals, problem and solution
- [x] Information from this description is also in comments
  - [ ] No useful information
- [x] Before/after screenshots are included
  - [ ] Screenshots aren't useful
  - [ ] No followups
- [ ] Reversion plan exists
  - [x] Standard git revert is fine
- [x] Tests are included (required for regressions)
- [x] Specs (docs/trello) are linked in code 
  - [ ] No spec exists

### DIFF
--- a/backend/test/test_other_libs.ml
+++ b/backend/test/test_other_libs.ml
@@ -942,7 +942,7 @@ let t_list_stdlibs_work () =
   check_dval
     "List::dropWhile works"
     (DList [Dval.dint 3; Dval.dint 4])
-    (exec_ast'
+    (exec_ast
        (fn
           "List::dropWhile"
           [ list [int 1; int 2; int 3; int 4]
@@ -950,7 +950,7 @@ let t_list_stdlibs_work () =
   check_dval
     "List::dropWhile stops upon false"
     (DList [Dval.dint 5; Dval.dint 2; Dval.dint 2])
-    (exec_ast'
+    (exec_ast
        (fn
           "List::dropWhile"
           [ list [int 1; int 5; int 2; int 2]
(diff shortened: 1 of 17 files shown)


pbiggar:
Getting rid of the client

PR #2611, opened 2020-06-18
Remove expr type parameter from dval

pbiggar:
Another cleanup in the long line of converting from `expr` to `FluidExpression.t`. 
Now, the old `expr` is only used in serialization.

- [ ] Trello link included
- [ ] Discussed goals, problem and solution
- [ ] Information from this description is also in comments
  - [ ] No useful information
  - [ ] Screenshots aren't useful
  - [ ] No followups
- [ ] Reversion plan exists
  - [ ] Standard git revert is fine
- [ ] Tests are included (required for regressions)
- [ ] Specs (docs/trello) are linked in code 
  - [ ] No spec exists


pbiggar:
> I am trying to review this, but struggling to understand the SerializationFormat module? I see some things got moved there and then are concretized in Types, but others (like `op`) look like they got copied into both? Can you just explain what's going on a bit?

Serialization_format is now the only place where types get serialized (all the types derive `bin_io` there) and none of the types in other files too (minor exception in Package_manager.ml). So now we have the serialized types in one place, and then we can ignore them everywhere else (by which i mean we can change the contents of types.ml without worrying about accidentally triggering a serialization format change).

The types are duplicated because they are different types (in some cases I could directly reuse them, but I couldn't figure out how to get the constructors to be in the `Types` namespace without copying them directly.

pbiggar:
Things I'm going to test before merging:

- does every canvas load
- test that patterns match (esp nested patterns)
- manually test analysis

Issue #2585, opened 2020-06-10
Fix tab ordering


pbiggar:
Tab ordering has gotten to be a little bit of a mess, and also we have changed our goals a little since we created it.

`Tab` should move to the next editable space, even if filled in. `Shift-tab` should do the opposite.

pbiggar:
Closing as it doesn't apply to darklang-next

PR #2615, opened 2020-06-21
Make tab through all editable fileds in fluid editing

br1anchen:
This PR tries to address #2585 
simplescreenrecorder-2020-06-21

I am not sure if I understand fluid editing state fully from the description in the issue, there are still some times editable fields are not in focus if the first time cursor is in the spec header.


pbiggar:
Will test this out later and have a look at the code. However, it looks like integration tests are failing due to this?

br1anchen:
@pbiggar Close this in favor of #2658 , which is caused by default branch changes, my fork is not considered as fork to make pull request any more.
Have to re-clone dark repo

Issue #2424, opened 2020-05-29
Expand the query compiler


pbiggar:
Dark's query compiler supports a subset of builtin functions from the full language. Those functions can be seen by reading the source in `backend/libbackend/sql_compiler`.

To add new functions, look at `binop_to_sql` - those are functions that map directly from Dark's builtin types and standard library to Postgres 9.6's builtins. Functions which are easy to add will also have a direct mapping.

More complex functions need to be handled in `lambda_to_sql` directly.

Query compiler functions should all be tested. There are pretty good tests of basic functionality, but could do with more tests. If you want to start small, adding a test for an existing function is very welcome.

To work on this, make a new issue saying which function you'd like to implement.

pbiggar:
Closing this as the explanation is wrong due to updates, and we don't super want to expand the query compiler at the moment as our tech choices are up in the air.

PR #2617, opened 2020-06-22
support Date::hour_v1 in query compiler and allow appending of Dark arg in unary_ops

### LINKED ISSUE #2616: Add Date::hour_v1 support for query compiler
The query compiler isn't configured to handle Dark's Date::hour_v1 function.

cooleydw494:
I've added Date::hour_v1 to the unary_op_to_sql function, mapping the Dark function to respective Postgres function (date_part('hour', ...).
This will allow the Db::query family of functions to compile the aforementioned Date::hour_v1 function successfully when applied to DB record date values.

I had to add the ability to pass the Dark unary op argument (in this case a Date) as the last argument as opposed to the first where it is evaluated in lambda_to_sql. This probably isn't the most elegant solution and likely this will need to become even more sophisticated later on for unary Dark ops that translate to postgres functions with more than 2 args.

Closes #2616

### DIFF
--- a/backend/libbackend/sql_compiler.ml
+++ b/backend/libbackend/sql_compiler.ml
@@ -85,28 +85,30 @@ let binop_to_sql (op : string) : tipe_ * tipe_ * tipe_ * string =
       error2 "This function is not yet implemented" op
 
 
-let unary_op_to_sql op : tipe_ * tipe_ * string * string list =
+let unary_op_to_sql op : tipe_ * tipe_ * string * string list * bool =
   (* Returns a postgres function name, and arguments to the function. The
    * argument the user provides will be inserted as the first argument. *)
   match op with
   | "Bool::not" ->
-      (TBool, TBool, "not", [])
+      (TBool, TBool, "not", [], false)
   (* Not sure if any of the string functions are strictly correct for unicode *)
   | "String::toLowercase" | "String::toLowercase_v1" ->
-      (TStr, TStr, "lower", [])
+      (TStr, TStr, "lower", [], false)
   | "String::toUppercase" | "String::toUppercase_v1" ->
-      (TStr, TStr, "upper", [])
+      (TStr, TStr, "upper", [], false)
   | "String::length" ->
       (* There is a unicode version of length but it only works on bytea data *)
-      (TStr, TInt, "length", [])
+      (TStr, TInt, "length", [], false)
   | "String::reverse" ->
-      (TStr, TStr, "reverse", [])
+      (TStr, TStr, "reverse", [], false)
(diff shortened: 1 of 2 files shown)


backend/libbackend/sql_compiler.ml
```diff
@@ -85,28 +85,30 @@ let binop_to_sql (op : string) : tipe_ * tipe_ * tipe_ * string =
       error2 "This function is not yet implemented" op
 
 
-let unary_op_to_sql op : tipe_ * tipe_ * string * string list =
+let unary_op_to_sql op : tipe_ * tipe_ * string * string list * bool =
```
pbiggar, line 88: We try to avoid `bool`s as arguments, and use variants instead. So here you could have a type

```
type position = Front | Back
```

And then return `(TBool , TBool, "bot", [], Back)`

pbiggar:
Code looks great, apart from the bool. Ready to merge after the change I suggested.

Are you planning to add the other `Date` functions too? That would be great!

cooleydw494:
Yep, definitely planning on adding other Date functions. I see myself working on the query compiler a lot in general, albeit slow. A good way to focus on one thing while putting together how the language works, etc. Plus, super relevant to my interests. Being able to perform complex, efficient queries can replace so much application logic!

backend/libbackend/sql_compiler.ml
cooleydw494: Sounds good! Any advice on an appropriate place to define the type? Not sure I understand the right context to put it in.

pbiggar:
> Yep, definitely planning on adding other Date functions. I see myself working on the query compiler a lot in general, albeit slow. A good way to focus on one thing while putting together how the language works, etc. Plus, super relevant to my interests. Being able to perform complex, efficient queries can replace so much application logic!

great! Unrelated to this PR, but it occurs to me that if we can move the per-function info from the query compiler into the functions definitions, I think it would be more long-term maintainable. Dunno whether it could be done nicely though.

cooleydw494:
> > Yep, definitely planning on adding other Date functions. I see myself working on the query compiler a lot in general, albeit slow. A good way to focus on one thing while putting together how the language works, etc. Plus, super relevant to my interests. Being able to perform complex, efficient queries can replace so much application logic!
> 
> great! Unrelated to this PR, but it occurs to me that if we can move the per-function info from the query compiler into the functions definitions, I think it would be more long-term maintainable. Dunno whether it could be done nicely though.

I like it. Yeah, I'm not sure how that would look in practice, but conceptually I'm with you. I'll keep it in the back of my mind as I work through some other stuff.

PR #2639, opened 2020-06-27
support Date::hour_v1 in query compiler and allow appending of Dark arg in unary_ops

### LINKED ISSUE #2616: Add Date::hour_v1 support for query compiler

cooleydw494:
I've added Date::hour_v1 to the unary_op_to_sql function, mapping the Dark function to respective Postgres function (date_part('hour', ...).
This will allow the Db::query family of functions to compile the aforementioned Date::hour_v1 function successfully when applied to DB record date values.

I had to add the ability to pass the Dark unary op argument (in this case a Date) as the last argument as opposed to the first where it is evaluated in lambda_to_sql. This probably isn't the most elegant solution and likely this will need to become even more sophisticated later on for unary Dark ops that translate to postgres functions with more than 2 args.

Closes #2616

### DIFF
--- a/backend/libbackend/sql_compiler.ml
+++ b/backend/libbackend/sql_compiler.ml
@@ -11,6 +11,10 @@ let error str = raise (DBQueryException str)
 
 let error2 msg str = error (msg ^ ": " ^ str)
 
+type position =
+  | First
+  | Last
+
 let binop_to_sql (op : string) : tipe_ * tipe_ * tipe_ * string =
   let allInts str = (TInt, TInt, TInt, str) in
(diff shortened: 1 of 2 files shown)


cooleydw494:
@pbiggar I went with First/Last as opposed to Front/Back, hope that is ok. Also let me know if the placement of the variant definition is bad, wasn't totally sure where to put it, and can fix either of those very quickly now that my repo is fixed.

pbiggar:
This looks great! Will merge once CI completes.

PR #2658, opened 2020-06-28
Make tab through all editable fields in body

br1anchen:
Fixes #2615 

find editable fields in body to loop through

tab-through-2020-06-28-215001

Manually tested and integration does not fail

### DIFF
--- a/client/src/forms/Entry.ml
+++ b/client/src/forms/Entry.ml
@@ -413,6 +413,15 @@ let submitACItem
           then NoChange
           else Model.updateErrorMod (Error.set error)
       | None ->
+          let selectOrMove =
+            match move with
+            | GotoNext ->
+                Many
+                  [ AutocompleteMod (ACSetQuery "")
+                  ; Selection.enterNextEditable m tlid id ]
+            | _ ->
+                Select (tlid, STID id)
+          in
           let wrap ops next =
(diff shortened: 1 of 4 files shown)


pbiggar:
This works great for code, nice!

However, it brings a lot of jankiness to editing non-code (DBs, handler headings). For example:



The first tab doesn't move into the next blankOr, it instead moves to selecting mode (a mode which I want to get rid of, but haven't done yet).

The second tab goes into the next blankOr, however it also replaces the text with the previous text, instead of keeping the same text. This is a weird piece of code: the text is stored in `m.complete.value` and it's stateful - it needs to be updated (with an `ACSetQuery`) call as the tabs move around.

I also see some jank in HTTP headers, and moving between HTTP headers and code bodies. Most of those bugs were there before, but I'm also seeing the cursor focus on the wrong place. Can you test that out - I'd be happy to merge this once it's strictly better than the current code, even if it doesn't solve all problems.

br1anchen:
Thanks for testing it out. Good to know other cases to test with. Will try to fix in these scenarios you mentioned.

pbiggar:
> Thanks for testing it out. Good to know other cases to test with. Will try to fix in these scenarios you mentioned.

Thanks! 😊

pbiggar:
We're getting rid of the client so closing this.

PR #2664, opened 2020-06-30
Hi, this is my first PR: added tests to List::isEmpty

ajroberts0417:
Just adding some tests


pbiggar:
Wonderful, thanks!

PR #2685, opened 2020-07-05
Add info about needing python for vim+merlin integration

fmilani:
The editor-setup doc instructions on integrating merlin is missing information if the contributor is using vim or neovim.

I added that information to the doc


pbiggar:
Great, thanks! Can you sign the CLA again (i fixed a typo and it decided everyone should sign again 😠)?

PR #2692, opened 2020-07-05
Add String::replaceAll to sql compiler

### LINKED ISSUE #2691: Add String::replaceAll support to the query compiler
The query compiler should be able to handle String::replaceAll. Part of #2424

cooleydw494:
I've added String::replaceAll to the lambda_to_sql function, mapping to the postgres function "replace", and wrote a test to verify it works
This will allow the Db::query family of functions to compile the aforementioned String::replaceAll function successfully when applied to DB record values.

I think likely there should be another function to contain the logic of functions with 3 arguments (like unary and binop functions) but for now I just really want the functionality. We've also discussed evolving where the logic for each function is stored, but I'm still just focusing on expanding support within the existing infrastructure (what I'm capable of right now).

Closes #2691

### DIFF
--- a/backend/libbackend/sql_compiler.ml
+++ b/backend/libbackend/sql_compiler.ml
@@ -319,8 +319,16 @@ let rec lambda_to_sql
   | EFnCall (_, "String::contains", [lookingIn; searchingFor], NoRail) ->
       (* strpos returns indexed from 1; 0 means missing *)
       "(strpos(" ^ lts TStr lookingIn ^ ", " ^ lts TStr searchingFor ^ ") > 0)"
-  | EFnCall (_, "String::replaceAll", [lookingIn; searchingFor; replaceWith], NoRail) ->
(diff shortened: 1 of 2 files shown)


cooleydw494:
Btw, sorry for the extra commits that aren't relevant to the task at hand. I keep forgetting to format and check for changes to merge in before I commit

pbiggar:
Looks great!

Could you add:
- a test where it doesn't replace anything (eg search for "xxx")
- a test where it replaces more than once

Should be good to go with those!

cooleydw494:
> Looks great!
> 
> Could you add:
> 
> * a test where it doesn't replace anything (eg search for "xxx")
> * a test where it replaces more than once
> 
> Should be good to go with those!

done!

cooleydw494:
hmmm, I can't see details on the integration tests and they are passing locally. Let me know what I can do. No rush, likely a tomorrow thing for me.

PR #2698, opened 2020-07-06
Add simple returns type for function reference

### LINKED ISSUE #2687: Function reference boxes should show the return type
The currently show parameter names and types, but not return types.

image

br1anchen:
Add return type in function reference, to fix #2687 

Retrieve user function and package function return type, and display it as separated row in reference.

function-reference-return-type


br1anchen:
The failing in integration tests seems unrelated. Did all integration tests locally, all passed. Is there any issue on circleci config for this `ax wait count for analysis exceeded: expected true to be falsy` test?

pbiggar:
Wonderful, thanks!

That's a flaky test, apologies!

PR #2705, opened 2020-07-07
Add checking duplicated record filed name in fluid editor

### LINKED ISSUE #2686: Highlight an error when the same key is used twice in a dict/record
When the same key/field name appears twice in a dict/record, we should show an error. I think the live value should display an error, and the field names should be underlined in red.

image

br1anchen:
Adding `fluid-error` for duplicated record filed name, to fix #2686 

Checking record filed names to see duplicated tokens, and add `fluid-error` to them.

highlighted-duplicated-field-name

### DIFF
--- a/client/src/fluid/FluidEditorView.ml
+++ b/client/src/fluid/FluidEditorView.ml
@@ -106,6 +106,22 @@ let toHtml (p : props) : Types.msg Html.html list =
     | _ ->
         false
   in
+  let isDuplicatedFiledName ti =
+    match ti.token with
+    | TRecordFieldname {fieldName; recordID; _} ->
+        let exprAst = FluidAST.toExpr p.ast in
+        Fluid.recordFields recordID exprAst
+        |> Option.andThen ~f:(fun fields ->
+               let count =
+                 fields
+                 |> List.filter ~f:(fun (v, _) -> v = fieldName)
+                 |> List.length
+               in
+               if count > 1 then Some count else None)
+        |> Option.is_some
+    | _ ->
+        false
+  in
   (* Gets the source of a DIncomplete given an expr id *)
   let sourceOfExprValue id =
     if FluidToken.validID id
@@ -244,7 +260,7 @@ let toHtml (p : props) : Types.msg Html.html list =
             && (* This expression is the source of its own incompleteness. We
             only draw underlines under sources of incompletes, not all
             propagated occurrences. *)
-            sourceId = Some (p.tlid, analysisId)
+            (sourceId = Some (p.tlid, analysisId) || isDuplicatedFiledName ti)
           in
           let isNotExecuted = exeFlow ti = CodeNotExecuted in
           (* Unfade non-executed code if the caret is in it,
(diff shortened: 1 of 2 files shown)


pbiggar:
Nice fix!

There's an issue with performance here (which you weren't to know about, I'm adding it to the docs here https://github.com/darklang/docs/pull/177). The main loop that turns tokens into html is very performance sensitive, and we don't want to do anything where a token does an AST search. I think if you pre-processed it (outside the loop but within the view code is fine, it doesn't need to be saved on the model or anything), this would be fine.

Specifically, I'd suggest going through the AST once per toHtml call, and store the IDs of any duplicates. Then look those IDs up when looping.

br1anchen:
Thanks @pbiggar for the performance info. I will try to see how to restructure this logic. Ideally I would like to have `Map<recordID, fieldName>`  for duplicated field name look up. But did not find a good way to retrieve `ast` from correct context scope.
Will try to see how to get correct `ast` outside of `toHtml` loop I guess.

pbiggar:
> Thanks @pbiggar for the performance info. I will try to see how to restructure this logic. Ideally I would like to have `Map<recordID, fieldName>` for duplicated field name look up. But did not find a good way to retrieve `ast` from correct context scope.

Yeah, sounds right!

> Will try to see how to get correct `ast` outside of `toHtml` loop I guess.

To be clear, it's fine to do this in `toHTML`, just not within the loop (from the `List.map`).

client/src/fluid/FluidEditorView.ml
```diff
@@ -62,7 +62,8 @@ let viewPlayIcon (p : props) (ti : FluidToken.tokenInfo) : Types.msg Html.html =
       Vdom.noNode
 
 
-let toHtml (p : props) : Types.msg Html.html list =
+let toHtml (p : props) (duplicatedRecordFields : (ID.t * StrSet.t) list) :
```
br1anchen, line 65: @pbiggar I end up with `duplicatedRecordFields: List<Record, Set<DuplicatedFieldName>>`, would like to use `Map` instead of `List`, but I am not so fluent in `ocaml`, the overhead to convert `exprAst` to `Map` is a bit much.
Also passing `duplicatedRecordFields` as second argument than inside `props` to reduce footprint for this fix.

pbiggar: Yeah, understandable. Maps are much more complex than they need to be, though the next version of tablecloth has fixes for this, so I think the best bet is to merge as is, and improve when/if we see a performance problem or when there's an opportunity to fix this later.

pbiggar:
This looks great, I'll merge it as is.

Two followups come to mind, but I think they don't make sense to block this on them:
- show a message indicating what the exact problem is
- adding a test that this renders correctly (we don't currently have any rendering tests, so this would be the first one - the idea is a unit test that checks the structure of the rendered HTML-like OCaml values, as opposed to an integration test)

client/src/fluid/FluidEditorView.ml
br1anchen: Cool, good to know :+1:

PR #2717, opened 2020-07-09
Fix playbutton z-index with dropdown

### LINKED ISSUE #2676: Play button is shown over autocomplete
In the image, you can see that a gray triangle is showing over the "i" in `Uuid`. It should be behind it.

image

br1anchen:
Change `playbutton` z-index to `100` to be lower than `dropdown`.
Fixes #2676 

Change `z-index`, since these elements are not siblings and positioned `absolute`.

play-button-position


pbiggar:
Awesome, thanks!

PR #2722, opened 2020-07-10
Allow update rebounded variable name

### LINKED ISSUE #2712: Renaming a variable doesn't rename in the right-hand-side of a shadowing definition
In the gif, you can see that when the 2nd definition is called `origin`, renaming the first definition doesn't work. What should happen is that the uses of `origin` within the `if`-expression would change when the first definition of `origin` is renamed.

br1anchen:
Allow to update variable name even it has been rebounded, to fix #2712 

Currently, it is not allow to rename rebounded variable. Enable this might lead to other unexpected behaviour, eg if new variable name is same for other variable than rebounded variable name, then all referenced places will be marked as updated variable.

rename-shadow-definition-2020-07

The original implementation seems disable rename rebounded variable intentionally, is to enable it has been well thought about in this case?

### DIFF
--- a/client/test/fluid_expression_test.ml
+++ b/client/test/fluid_expression_test.ml
@@ -214,4 +214,27 @@ let run () =
         "right partials with diff exprs"
         (rightPartial "++" (str "foo"))
         (rightPartial "++" (str "bar")) ;
+      describe "renameVariableUses" (fun () ->
+          test "with variable used in right hand side" (fun () ->
+              let exprAst = let' "a" (var "x") (E.newB ()) in
+              expect
+                ( E.renameVariableUses ~oldName:"x" ~newName:"x1" exprAst
+                |> FluidPrinter.eToTestString )
+              |> toEqual "let a = x1\n___") ;
+          test "with rebound let expression" (fun () ->
+              let exprAst = let' "x" (int 6) (let' "a" (var "x") (E.newB ())) in
+              expect
+                ( E.renameVariableUses ~oldName:"x" ~newName:"x1" exprAst
+                |> FluidPrinter.eToTestString )
+              |> toEqual "let x = 6\nlet a = x\n___") ;
+          test
+            "ith rebound let expression and used as variable in right hand side"
+            (fun () ->
+              let exprAst =
+                let' "x" (fn "Int::add" [var "x"; int 6]) (E.newB ())
+              in
+              expect
+                ( E.renameVariableUses ~oldName:"x" ~newName:"x1" exprAst
+                |> FluidPrinter.eToTestString )
+              |> toEqual "let x = Int::add x1 6\n___")) ;
       ())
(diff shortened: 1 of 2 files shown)


pbiggar:
Nice find! Will comment inline on the fix in a moment.

```
let x1 = 6
let a = b
let c = d
```

Can you add tests for all the permutations of the above code, where one or more of a, b, c, and d are x1?

libshared/FluidExpression.ml
```diff
@@ -457,9 +457,7 @@ let rec updateVariableUses (oldVarName : string) ~(f : t -> t) (ast : t) : t =
   | EVariable (_, varName) ->
       if varName = oldVarName then f ast else ast
   | ELet (id, lhs, rhs, body) ->
-      if oldVarName = lhs (* if variable name is rebound *)
-      then ast
-      else ELet (id, lhs, u rhs, u body)
+      ELet (id, lhs, u rhs, u body)
```
pbiggar, line 460: For reference, in `ELet`, `lhs` is `x`, `rhs` is `a + b`, `body` is `z + 4`

```
let x = a + b
z + 4
```

So if `oldVarName = lhs` is true then `oldVarName` is rebound in `body` and so there should not be a rename, in which case the old code was correct. But, even if `oldVarName = lhs`, that doesn't apply in `rhs`, in which case the old code is wrong and the new code is correct.

So I think the correct code is probably:

```
if oldVarName = lhs
then ELet(id, lhs, u rhs, body)
else ELet(id, lhs, u rhs, u body)
```

br1anchen:
@pbiggar Updated with tests now.

client/test/fluid_expression_test.ml
```diff
@@ -214,4 +214,27 @@ let run () =
         "right partials with diff exprs"
         (rightPartial "++" (str "foo"))
         (rightPartial "++" (str "bar")) ;
+      describe "renameVariableUses" (fun () ->
+          test "with variable used in rightPartial" (fun () ->
```
pbiggar, line 218:
```suggestion
          test "with variable used in right hand side" (fun () ->
```

client/test/fluid_expression_test.ml
```diff
@@ -214,4 +214,27 @@ let run () =
...
+                |> FluidPrinter.eToTestString )
+              |> toEqual "let a = x1\n___") ;
+          test "with rebound let expression" (fun () ->
+              let exprAst = let' "x" (int 6) (let' "a" (var "x") (E.newB ())) in
+              expect
+                ( E.renameVariableUses ~oldName:"x" ~newName:"x1" exprAst
+                |> FluidPrinter.eToTestString )
+              |> toEqual "let x = 6\nlet a = x\n___") ;
+          test
+            "with rebound let expression and used as variable in rightPartial"
```
pbiggar, line 231:
```suggestion
            "with rebound let expression and used as variable in right hand side"
```

pbiggar:
Looks great, thanks! Will merge once tests are done again.

FYI: a "right partial" is different than the "right hand side". The "right hand side" is the `y` in `let x = y`. A rightPartial is a special construct used when adding an infix (such as `+`) after a completed expression (before it gets turned into a completed `BinOp`).

PR #2723, opened 2020-07-12
Replace start and end quotes for live value copy text

### LINKED ISSUE #2684: Do not include quotes when copying a string live value
For the program `"hello world"`, the live value shows `"hello world"` as well. If we copy this using the "copy" button/icon, the clipboard gets the value `"hello world"` (including quotes). This is frustrating when you're trying to calculate a value to paste somewhere.

So when copying live values, don't include the quotes.

br1anchen:
Replace start and end quotes for live value copy, to fix #2684 

Replace start and end quotes for live value through `copy-value` button.

### DIFF
--- a/client/src/app/Main.ml
+++ b/client/src/app/Main.ml
@@ -2022,6 +2022,7 @@ let update_ (msg : msg) (m : model) : modification =
       in
       Many [SetClipboardContents (copyData, e); mod_; toast]
   | ClipboardCopyLivevalue (lv, pos) ->
+      let lv = lv |> Regex.replace ~re:(Regex.regex "(^\")|(\"$)") ~repl:"" in
       Native.Clipboard.copyToClipboard lv ;
       ReplaceAllModificationsWithThisOne
         (fun m ->


client/src/fluid/FluidView.ml
```diff
@@ -25,7 +26,11 @@ let viewCopyButton tlid value : msg Html.html =
     ; ViewUtils.eventNoPropagation
         "click"
         ~key:("copylivevalue-" ^ value ^ TLID.toString tlid)
-        (fun m -> ClipboardCopyLivevalue (value, m.mePos)) ]
+        (fun m ->
```
br1anchen, line 29: Doing the escape here instead of inside `ClipboardCopyLivevalue` because of its usage in [ViewSidebar](https://github.com/darklang/dark/blob/main/client/src/canvas/ViewSidebar.ml#L615) to intentionally set `quotes` for hash.

pbiggar:
Looks great! Will take a closer look today, thanks :)

client/src/fluid/FluidView.ml
pbiggar: I don't see a great reason for ViewSidebar to do that, but also I would say the ViewSidebar use case is odd as it reuses something clearly meant for copying the live value. I think it would be better to move this to the ClipboardCopyLiveValue handler instead and ignore the ViewSidebar thing.

pbiggar:
Code looks good, would be happy to merge either after moving it to the ClipboardCopyLivevalue handler. If you felt strongly about keeping the PR as is, I could also see the case to merge as is.

client/src/fluid/FluidView.ml
br1anchen: Good to know. Move the changes to `ClipboardCopyLivevalue` handler now.

br1anchen: Good to know. Move the changes to `ClipboardCopyLivevalue` handler now.

br1anchen: Good to know. Move the changes to `ClipboardCopyLivevalue` handler now.

br1anchen:
@pbiggar updated now.

PR #2725, opened 2020-07-13
Fix parameter positions after function rename

### LINKED ISSUE #2682: Put parameters in the right place after function rename
We have two types of HttpClient request functions - 3 arity and 4 arity. 3-arity ones have no body (like GET); 4-arity ones do (like POST).

Problem is, if I have HttpClient::get, and I edit it to be HttpClient::post, the args are now incorrect - it goes from (url, params, headers) to (url,  body, params, headers), meaning that the params-dict is now in the body, the headers-dict is now the params, and the new headers arg is incomplete.

It'd be nicer if this edit moved the args around accordingly, so I'd have (url, body=blank, params, headers).

There is code to decide what happens to function arguments when we rename. It should be improved to:
- if there is a matching argument (same name and type), it should put it in the correct position
- if not, but the current position has the same type, keep it there
- otherwise, move it to a separate `let` definition (this is currently done if the position does not match both name and type)

br1anchen:
Original bug caused by using old parameter `index` as new parameter `index` for aligned parameters. This PR is to fix #2682 

Using `compareParams` to check two different parameters matched case `Aligned` and `TypeAndPositionMatched` to keep parameter.

args-rename-fun-2020-07-13-23032

### DIFF
--- a/client/src/fluid/Fluid.ml
+++ b/client/src/fluid/Fluid.ml
@@ -1785,6 +1785,11 @@ let rec mergeExprs (e1 : fluidExpr) (e2 : fluidExpr) : fluidExpr * caretTarget =
       (e1, caretTargetForEndOfExpr' e1)
 
 
+type compareParamsResult =
+  | CPAligned
+  | CPTypeAndPositionMatched
+  | CPNothingMatched
+
 let replacePartialWithArguments
     (props : props) ~(newExpr : E.t) (id : ID.t) (ast : FluidAST.t) :
     FluidAST.t * caretTarget =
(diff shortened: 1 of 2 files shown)


client/src/fluid/Fluid.ml
```diff
@@ -1902,8 +1911,36 @@ let replacePartialWithArguments
...
+                             , Some (_, _, expr, _) ) ->
+                               Some (name, tipe, expr, index)
+                           | _ ->
+                               None )
+                       | None, None ->
+                           None
+                     in
+                     match matchedNewParam with
+                     | Some (name, tipe, expr, index) ->
+                         Left (Some (name, tipe, expr, index))
```
pbiggar, line 1941: I think this could be folded into the previous match statement.

br1anchen: Good point, changed.

client/src/fluid/Fluid.ml
```diff
@@ -1901,11 +1913,39 @@ let replacePartialWithArguments
...
                let oldParams = getFunctionParams oldName count existingExprs in
+               (* Divide old existing params to matched and mismatched lists*)
                let matchedParams, mismatchedParams =
-                 List.partition oldParams ~f:(fun p ->
-                     List.any newParams ~f:(isAligned p))
+                 List.partitionMap oldParams ~f:(fun p ->
+                     (* compare old param with all new param,
+                      * to check if there is any aligned or type and position matched,
+                      * needs to compare with all new params
+                      * because same old param could align and type matched with multiple new params *)
```
pbiggar, line 1922:
```suggestion
                     (* Split out the "aligned" parameters (parameters from the old fnCall that match the new fnCall) *)
```

client/src/fluid/Fluid.ml
```diff
@@ -1848,12 +1853,19 @@ let replacePartialWithArguments
     then Rail
     else NoRail
   in
-  let isAligned p1 p2 =
+  (* Compare two params function,
+   * params are aligned if both name and type are the same,
+   * Params are type and position matched if both type and position are the same *)
```
pbiggar, line 1858:
```suggestion
  (* Compare two parameters, params are aligned if both name 
    * and type are the same, or if both type and position are the same *)
```

pbiggar:
This is great, thanks! Will merge once CI finishes.

PR #2730, opened 2020-07-14
Move caret to text of end when paste in front of open quote

### LINKED ISSUE #2675: Caret placement issue on paste before an open quote
When the caret in front of a quote, pasting successfully pastes but puts the carat in the wrong position.

One could also argue that it shouldn't paste.

br1anchen:
Put caret at the end of text if paste in front of open quote, to fix #2675 

paste-before-open-quote-2020-07

### DIFF
--- a/client/src/fluid/Fluid.ml
+++ b/client/src/fluid/Fluid.ml
@@ -5415,15 +5415,14 @@ let pasteOverSelection (data : clipboardContents) (astInfo : ASTInfo.t) :
           astInfo |> ASTInfo.setAST newAST |> moveToCaretTarget caretTarget
       | EString (id, str), _, Some {astRef = ARString (_, SPOpenQuote); offset}
         ->
-          (* Paste into a string, to take care of newlines.
-           * Note: pasting before an open quote
-           * places the caret one unit left of the text end *)
           let replacement =
             E.EString (id, String.insertAt ~insert:text ~index:(offset - 1) str)
           in
           let newAST = FluidAST.replace ~replacement id ast in
           let caretTarget =
-            CT.forARStringOpenQuote id (offset + String.length text)
+            if offset = 0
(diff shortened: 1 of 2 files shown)


client/src/fluid/Fluid.ml
```diff
@@ -5417,13 +5417,17 @@ let pasteOverSelection (data : clipboardContents) (astInfo : ASTInfo.t) :
         ->
           (* Paste into a string, to take care of newlines.
            * Note: pasting before an open quote
-           * places the caret one unit left of the text end *)
```
br1anchen, line 5420: It seems like putting `caret one unit left of the text end` is intended. But not sure what it is for?

pbiggar: That seems like it's commenting on the bug you're fixing (as opposed to documenting expected behaviour), so you can just remove the comment.

client/src/fluid/Fluid.ml
```diff
@@ -5417,13 +5417,17 @@ let pasteOverSelection (data : clipboardContents) (astInfo : ASTInfo.t) :
...
+           * places the caret at the text end *)
           let replacement =
             E.EString (id, String.insertAt ~insert:text ~index:(offset - 1) str)
           in
           let newAST = FluidAST.replace ~replacement id ast in
           let caretTarget =
-            CT.forARStringOpenQuote id (offset + String.length text)
+            match offset with
```
pbiggar, line 5426: I think an if statement here would be better, since we don't expect to have other offsets.

br1anchen: Right, makes sense.

br1anchen:
@pbiggar Changed as you suggest and add test for this paste string case.

pbiggar:
Wonderful, thank you!

Issue #2466, opened 2020-06-01
Add List::interpose to the stdlib


pbiggar:
`List::join "sep" ["a", "b", "c"]` -> `["a", "sep", "b", "sep", "c"]`

s0kil:
Already merged

pbiggar:
Thanks!

PR #2740, opened 2020-07-17
Add List::interpose to the stdlib

PosaLusa24:
Issue #2466: Standard library lacks a list interpose/join function.

To add List::interpose to the stdlib, along with the required tests.

List::interpose tests completed successfully.


pbiggar:
This looks great, thank you! I'm going to take a quick glance to see about the parameter ordering here, and I'll get back to you again shortly.

pbiggar:
Can you make the list the first argument? That way we can pipe to it.

Thanks!

PosaLusa24:
Sure! That'd be much better.

pbiggar:
Amazing, thanks so much!

PR #2742, opened 2020-07-19
String::toList_v1 test

kberridge:
My first pull request!  Adds a test for String::toList_v1.  As far as I can tell this doesn't have tests yet.  There are some string tests in test_other_libs and others in test_string_libs, I added this in test_string_libs, not sure where you would prefer it though.


kberridge:
Knew I was missing something!  OK, I *swear* I saw the first version of this test pass when I first ran it locally.  But toList clearly returns DCharacters, so I must have been missing something.

pbiggar:
Awesome, thank you!

Issue #2653, opened 2020-06-28
Make the Dark repo work on Windows


pbiggar:
Folks have tried and failed to run the Dark dev container on Windows.

Write a guide for windows users, and make any code changes necessary to make it a good experience.

kberridge:
I got Dark running on Windows in WSL2 (Windows Subsystem for Linux) this weekend.  Here's some information on setting it up.

- It does NOT work on WSL1, because docker only works in WSL2
- WSL2 is available starting in Windows 10 Version 2004, which is rolling out now
- Instructions on setting up WSL2 are here: https://docs.microsoft.com/en-us/windows/wsl/install-win10#update-to-wsl-2
- Instructions on setting up Docker for WSL2 are here: https://docs.docker.com/docker-for-windows/wsl/
- After that, following the instructions in Dark's README works fine with no additional steps or modifications needed

pbiggar:
Awesome, thanks! Can you add these as notes in the README?

pbiggar:
This looks like it's working

PR #2744, opened 2020-07-21
add notes on running dark on Windows in WSL2

kberridge:
add notes on running dark on Windows in WSL2, from #2653


pbiggar:
Wonderful, thanks!

Issue #2465, opened 2020-06-01
Add List::interleave to the standard library


pbiggar:
`List::interleave [1, 2, 3] [4,5,6]` -> `[1,4,2,5,3,6]`

s0kil:
Can be closed?

pbiggar:
Thanks!

PR #2749, opened 2020-07-23
Add List::interleave to the standard library

kberridge:
Adds List::interleave to the standard library. #2465 

I tried multiple different takes on the description and settled with this very verbose one as a starting point.  I failed to find any example descriptions from other standard libraries to work from.

The "List::interleave works (mixed types)" test probably isn't necessary as it's kind of a given I think, I just wanted to verify it worked as I expected.  Let me know if you want it removed.

### DIFF
--- a/backend/libexecution/liblist.ml
+++ b/backend/libexecution/liblist.ml
@@ -379,6 +379,28 @@ let fns : fn list =
               fail args)
     ; preview_safety = Safe
     ; deprecated = false }
+  ; { prefix_names = ["List::interleave"]
+    ; infix_names = []
+    ; parameters = [par "as" TList; par "bs" TList]
+    ; return_type = TList
+    ; description =
+        "Returns a new list with the first value from `as` then the first value from `bs`, then the second value from `as` then the second value from `bs`, etc, until one list ends, then the remaining items from the other list."
+    ; func =
+        InProcess
+          (function
+          | _, [DList l1; DList l2] ->
+              let rec f l1 l2 =
+                match l1 with
+                | [] ->
+                    l2
+                | x :: xs ->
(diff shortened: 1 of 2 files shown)


backend/libexecution/liblist.ml
```diff
@@ -379,6 +379,28 @@ let fns : fn list =
               fail args)
     ; preview_safety = Safe
     ; deprecated = false }
+  ; { prefix_names = ["List::interleave"]
+    ; infix_names = []
+    ; parameters = [par "as" TList; par "bs" TList]
+    ; return_type = TList
+    ; description =
+        "Returns a new list with the first value from `as` then the first value from `bs`, then the second value from `as` then the second value from `bs`, etc, until one list ends, then the remaining items from the other list."
```
pbiggar, line 387: Can you use the docstring format from here? https://github.com/darklang/dark/blob/main/docs/writing-docstrings.md

backend/test/test_other_libs.ml
```diff
@@ -932,6 +932,54 @@ let t_list_stdlibs_work () =
...
+    (exec_ast (fn "List::interleave" [list []; list [int 4; int 5; int 6]])) ;
+  check_dval
+    "List::interleave works (b empty)"
+    (DList [Dval.dint 1; Dval.dint 2; Dval.dint 3])
+    (exec_ast (fn "List::interleave" [list [int 1; int 2; int 3]; list []])) ;
+  check_dval
+    "List::interleave works (both empty)"
+    (DList [])
+    (exec_ast (fn "List::interleave" [list []; list []])) ;
+  check_dval
```
pbiggar, line 970: I think removing this is probably better as Dark isn't really supposed to have mixed type lists :)

pbiggar:
Wonderful, thanks!

Issue #2660, opened 2020-06-28
Only show functions which are allowed in the sql compiler, within a DB::query lambda


pbiggar:
We know at compile time which functions can be run in the query compiler, so only allow those functions to appear in the autocomplete.

I think a approach direction for this would be to move the compilation code functions within to the function definitions for functions which can be used. That would also make them more robust. Then, that will be sent to the client to be used in FluidAutocomplete.

Then in the Fluidautocomplete, we can look to see if we're in a query block, and limit the functions accordingly.

pbiggar:
Done

PR #2769, opened 2020-08-02
add is_supported_in_query to function metadata, filter to supported f…

kberridge:
For issue #2660 

Only a subset of dark's builtin functions are supported by Db::query, but autocomplete doesn't know that today.  

This PR updates FluidAutocomplete so only supported functions are suggested when inside Db::query.

The function metadata sent to the client includes if the function is supported by Db::query.  Then FluidAutocomplete checks if we are inside a Db::query statement in the AST and if so filters down to only functions supported by Db::query before applying the normal complete behavior.


client/src/fluid/FluidAutocomplete.ml
```diff
@@ -328,12 +328,32 @@ let secretToACItem (s : SecretTypes.t) : fluidAutocompleteItem =
...
+
+let filterToDbSupportedFns (tl : toplevel) ti functions =
+  if not (isInQuery tl ti) then
+    functions
+  else
+    functions
+    |> List.filter ~f:(fun f -> 
+      match f with
+      | FACFunction fn -> fn.fnIsSupportedInQuery
+      | _ -> false)
```
pbiggar, line 348: I think you need to allow other things, such as pipes and variables (I don't know the full list, the SQL compiler source will probably know).

client/src/fluid/FluidAutocomplete.ml
```diff
@@ -328,12 +328,32 @@ let secretToACItem (s : SecretTypes.t) : fluidAutocompleteItem =
   let asDval = DStr (Util.obscureString s.secretValue) in
   FACVariable (s.secretName, Some asDval)
 
+let isInQuery (tl : toplevel) ti =
```
pbiggar, line 331: There's a possibility that this will be slow and cause problems in rendering, but I think it's fine (typically the things that are slow do an AST traversal per token, and I think this just does one per keystroke, which is probably fine).

backend/libbackend/api.ml
```diff
@@ -178,7 +179,14 @@ let functions ~username =
          ; return_type = Dval.tipe_to_string v.return_type
          ; preview_safety = v.preview_safety
          ; infix = List.mem ~equal:( = ) v.infix_names k
-         ; deprecated = v.deprecated })
+         ; deprecated = v.deprecated
+         ; is_supported_in_query =
```
pbiggar, line 183: This function looks weird, and I'm not sure what it does. Two `List.exists` seems suspect.

I'd suggest make Sql_compiler.compilerSupportedFns a Set (`Tc.StrSet` would do I think) and then just lookup whether it's in there, instead of a list traversal.

kberridge: Updated to a set.  There is still the List.exists because fn.prefix_names is a list.  I didn't see a better key to use on fn, so I wrote this so if *any* prefix name is in the supported list, the fn is supported.  I haven't found an example of a fn with more than one prefix name yet though.

Can you point me at docs or something for StrSet?  I found tc.ml, which includes Tablecloth, but Tablecloth's docs don't match what seems to be available (based on error messages).  And the ocmal set docs I found didn't seem the same either, so I'm not sure what this thing actually is!  I really need to get merlin working in vscode, I feel like I'm flying blind a lot of the time and maybe that would help.

client/src/fluid/FluidAutocomplete.ml
kberridge: pipes and variables are working because they are included in separate lets (keywords and varnames respectively I think) and concat'd together in generateExprs.  In this context I'm pretty sure we will only ever see FACFunctions.

backend/libbackend/api.ml
pbiggar: Ah, that's helpful. Yes, there's only ever one prefix_names, afaik. We preprocess it out later so it's actually kinda useless.

You can see some uses of StrSet in client/src/analysis/Sync.ml.

pbiggar: I also don't have merlin setup and it bothers me 😭

kberridge:
@pbiggar finally found the time to get back to this.  When I wrote the tests I found that in addition to filtering the functions, many of the "keywords" are also not supported, so those are filtered now as well.

client/test/fluid_ac_test.ml
```diff
@@ -518,5 +520,51 @@ let run () =
...
+          test "includes variables" (fun () ->
+              expect (valid |> List.filter ~f:isLiteral |> List.map ~f:AC.asName)
+              |> toEqual ["true"; "false"; "null"]) ;
+          test "includes keywords" (fun () ->
+              expect (valid |> List.filter ~f:isKeyword |> List.map ~f:AC.asName)
+              |> toEqual ["let"; "|>"]) ;
+          test "includes constructors" (fun () ->
+              expect
+                (valid |> List.filter ~f:isConstructor |> List.map ~f:AC.asName)
+              |> toEqual ["Just"; "Nothing"; "Ok"; "Error"]) ;
```
kberridge, line 567: @pbiggar the last question I have is are these constructors supported by queries/do they make sense in a query?

pbiggar: They are not currently supported. It seems they don't have support in `Sql_compiler.partially_evaluate` or in `Sql_compiler.lambda_to_sql` so that means they aren't supported at all.

kberridge: filtered out!  I think this is good to go?

pbiggar: Wonderful. I hope to review this on Tuesday!

client/src/fluid/FluidAutocomplete.ml
```diff
@@ -328,27 +328,32 @@ let secretToACItem (s : SecretTypes.t) : fluidAutocompleteItem =
...
+  | None ->
+      false
   | Some ast ->
-    FluidAST.ancestors (FluidToken.tid ti.token) ast
-    |> List.find ~f:(function FluidExpression.EFnCall (_, name, _, _) -> if name = "DB::query_v4" then true else false | _ -> false)
-    |> Option.is_some
+      FluidAST.ancestors (FluidToken.tid ti.token) ast
+      |> List.find ~f:(function
+             | FluidExpression.EFnCall (_, name, _, _) ->
+                 if name = "DB::query_v4" then true else false
```
pbiggar, line 341: There are a number of different DB::query functions so this won't support them all. I looked for calls to `User_db.query` in https://github.com/darklang/dark/blob/main/backend/libbackend/libdb2.ml:

- DB::query_v4
- DB::queryWithKey_v3
- DB::queryOne_v3
- DB::queryOne_v4
- DB::queryOneWithKey_v3
- DB::queryCount

Perhaps there should be an attribute for whether it's a query compiler function too? Otherwise we can just use a list.

pbiggar (changes requested):
Thanks for your patience on this review! Looks like it's just about ready to go over the line. Let me know if you won't have time and I can finish it out.

Thanks!

client/src/fluid/FluidAutocomplete.ml
kberridge: I wasn't sure what you meant by attribute here so I went with a list for starters.  (And sorry about the stupid commit message, I did a git amend and didn't noticed it overwrote the message until I'd already pushed it :\ )

For the "attribute" were you thinking of something like adding another field to the function_metadata record, maybe something like is_query_function and setting that based on a list defined in Sql_compiler like we did for compilerSupportedFns?

pbiggar: > For the "attribute" were you thinking of something like adding another field to the function_metadata record, maybe something like is_query_function and setting that based on a list defined in Sql_compiler like we did for compilerSupportedFns?

Yes, exactly like that! And then in the future we can move it into the function definitions in libdb2, etc.

client/src/fluid/FluidAutocomplete.ml
```diff
@@ -329,16 +329,57 @@ let secretToACItem (s : SecretTypes.t) : fluidAutocompleteItem =
...
+    |> List.map ~f:(fun fn ->
+           match fn with
+           | FACFunction fn ->
+               if fn.fnIsQuery then Some fn.fnName else None
+           | _ ->
+               None)
+    |> List.filter ~f:Option.is_some
+    |> List.map ~f:(fun x -> match x with Some fnName -> fnName | _ -> "")
+  in
+  let isQueryFn name = queryFnNames |> List.any ~f:(fun q -> q = name) in
```
kberridge, line 344: This works, it's probably fine, but I don't like it!

The AST's EFnCall has the function name, but doesn't have the actual fn record (I don't think?). So we have to look up the fn record to check the fnIsQuery field. I did this by filtering down to the query functions first, then checking each ancestor function name against that list. So at least we're not searching the entire functions list for each ancestor name. But, we're still filtering down all the functions to query functions every time we generate the autocomplete list. I didn't figure out a place to do that just once.

pbiggar: OK, I think you're right here. Let's go back to the old way.

I plan to add the fn to the EFnCall at some point, but the old way is certainly simpler until that happens.

kberridge: reverted!

pbiggar:
Works great, thanks!

PR #2795, opened 2020-09-11
Update to OCaml 4.08 & Core v0.12

pbiggar:
Update to OCaml 4.08 - this is for the new `let` syntax.


pbiggar:
Closing this as going to switch away from OCaml instead.

PR #2821, opened 2020-10-05
Downgrade to 0.4.18

pbiggar:
Upgrading to 0.4.19 coincides with the scheduler starting to misbehave, so let's
revert to make sure it's unrelated.


pbiggar:
Not needed

PR #2859, opened 2020-11-04
Updates the readme so that the dig command is formatted correctly.

andredublin:
Updates the readme so that the dig command is formatted correctly.

Updating the readme file


pbiggar:
Thank you!

PR #2864, opened 2020-11-13
Add test for Int::greaterThan

Athinanarof:
"Hi - this is my first pull request. I noticed that Int::greaterThan didn't have any tests so I added one."


pbiggar:
Wonderful, thank you!

PR #2868, opened 2020-11-19
LibInt.fs update.

Athinanarof:
I'm adapting old LibInt Ocalm functions to F#.

I tested each function by using the unit test.

I can always revert these changes, I will ask Paul if there is information related to this that I need to be aware of.

No, I'm not sure where those comments need to be, but I will add them as soon as I know where those need to be.


fsharp-backend/src/LibExecution/StdLib/LibInt.fs
```diff
@@ -286,177 +256,135 @@ let fns : List<BuiltInFn> =
...
+            Value(errStr("The second param" + a.ToString() + "is a float, but only works on Ints."))
+          | _, [ DStr a; _ ] ->
+            Value(errStr("The first param" + a.ToString() + "is a string, but only works on Ints."))
+          | _, [ _; DStr _ as a ] ->
+            Value(errStr("The second param" + a.ToString() + "is a string, but only works on Ints."))
+          | args ->
+              incorrectArgs ())
+      ; sqlSpec = NotYetImplementedTODO
+      ; previewable = Impure
+      ; deprecated = ReplacedBy(fn "" "" 0) }
```
pbiggar, line 339: Can you fill in the names of the replacement functions for deprecated ones?
```suggestion
      ; deprecated = ReplacedBy(fn "Int" "random" 1) }
```

fsharp-backend/tests/testfiles/int.tests
```diff
@@ -1,12 +1,25 @@
...
+Int.lessThanOrEqualTo_v0 10 1 = false
+Int.sqrt_v0 4 = 2.0
+Int.sqrt_v0 100 = 10.0
+Int.toFloat_v0 2 = 2.0
+Int.toFloat_v0 955656 = 955656.0
+Int.subtract_v0 10 9 = 1
+Int.subtract_v0 88 9 = 79
+Int.multiply_v0 8 8 = 64
+Int.multiply_v0 5145 5145 = 26471025
+
```
pbiggar, line 25: I think you could test `Int::random_v0` like this:

```
List.range 1 100 |> List.map (\i -> Int.random 0 2) |> List.all (\i ->  i <= 1 && i >= 0) = true
```

pbiggar (changes requested):
This looks great, thanks! There's a bunch of small things in the comments that are applicable to most of the tests (esp that we don't need warnings for strings in most cases)

fsharp-backend/tests/testfiles/int.tests
```diff
@@ -1,12 +1,25 @@
...
+//Int.min_v0 5 6 = 5
+//Int.negate_v0 -5 = 5
+//Int.negate_v0 5 = -5
+//Int.remainder_v0 5 0 = Error "`divisor` must be non-zero"
+Int.greaterThanOrEqualTo_v0 6 1 = true
+Int.greaterThanOrEqualTo_v0 6 8 = false
+Int.lessThanOrEqualTo_v0 6 8 = true
+Int.lessThanOrEqualTo_v0 10 1 = false
+Int.sqrt_v0 4 = 2.0
+Int.sqrt_v0 100 = 10.0
```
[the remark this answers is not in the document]
Athinanarof: It is returning a DFloat 9.949874371, should we return a DInt?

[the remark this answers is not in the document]
pbiggar: Oh, it's supposed to return a float! Ok, that's fine, thanks.

fsharp-backend/src/LibExecution/StdLib/LibInt.fs
```diff
@@ -107,13 +107,13 @@ let fns : List<BuiltInFn> =
...
           | _, [ (DFloat _ as a); _ ] ->
-            Value(errStr("The first param" + a.ToString() + "is a float, but only works on Ints. Use Float::add to compare Floats or use Float::truncate to truncate Floats to Ints."))
+            Value(errStr($"The first argument ({a}) is a float, but Int:add only works on Ints. Use Float::add to compare Floats or use Float::truncate to truncate Floats to Ints."))
           | _, [ _; DFloat _ as a ] ->
-            Value(errStr("The second param" + a.ToString() + "is a float, but only works on Ints. Use Float::add to compare Floats or use Float::truncate to truncate Floats to Ints."))
+            Value(errStr($"The second argument ({a}) is a float, but Int:add only works on Ints. Use Float::add to compare Floats or use Float::truncate to truncate Floats to Ints."))
           | _, [ DStr a; _ ] ->
-            Value(errStr("The first param" + a.ToString() + "is a string, but only works on Ints."))
+            Value(errStr($"The first argument ({a}) is a string, but Int:add only works on Ints."))
           | _, [ _; DStr _ as a ] ->
```
pbiggar: Can you switch to `DStr a` rather than `DStr _ as a`? Otherwise the error message will say `DStr "string"` in it.

pbiggar:
Looking good. There's still two unresolved comments; ready to merge after that.

pbiggar:
Great, thank you!

As a followup can you:
- add an infix operator for Int::subtract_v0
- add a [test for Int::random](https://github.com/darklang/dark/pull/2868#discussion_r527244547)

PR #2884, opened 2020-12-04
Liblist

Athinanarof:
ListLib functions implemented

Functions Implementation and testing.

LibList functions Implementation and tested.

I created the test cases for each function.

Revert branch merged.

### DIFF
--- a/fsharp-backend/src/LibExecution/StdLib/LibList.fs
+++ b/fsharp-backend/src/LibExecution/StdLib/LibList.fs
@@ -112,24 +112,18 @@ let fns : List<BuiltInFn> =
 //     ; sqlSpec = NotYetImplementedTODO
 //       ; previewable = Pure
 //     ; deprecated = NotDeprecated }
-//   ; { name = fn "List" "last" 0
-//
-//     ; parameters = [Param.make "list" TList]
-//     ; returnType = TAny
-//     ; description =
-//         "Returns the last value in `list`. Returns null if the list is empty."
-//     ; fn =
-//
-//           (function
-//           | _, [DList []] ->
-//               DNull
-//           | _, [DList l] ->
-//               List.last_exn l
-//           | args ->
-//               incorrectArgs ())
-//     ; sqlSpec = NotYetImplementedTODO
-//       ; previewable = Pure
-//     ; deprecated = ReplacedBy(fn "" "" 0) }
+    { name = fn "List" "last" 0
+      parameters = [Param.make "list" (TList varA) ""]
+      returnType = TAny
+      description =
+        "Returns the last value in `list`. Returns null if the list is empty."
+      fn =
+        (function
+        | _, [ DList l ] -> List.tryLast l |> Option.defaultValue DNull |> Value
+        | args -> incorrectArgs ())
+      sqlSpec = NotYetImplementedTODO
+      previewable = Pure
+      deprecated = ReplacedBy(fn "" "" 0) }
 //   ; { name = fn "List" "last" 1
 //
 //     ; parameters = [Param.make "list" TList]
--- a/fsharp-backend/src/LibExecution/StdLib/LibList.fs
+++ b/fsharp-backend/src/LibExecution/StdLib/LibList.fs
@@ -118,29 +118,26 @@ let fns : List<BuiltInFn> =
         "Returns the last value in `list`. Returns null if the list is empty."
       fn =
         (function
-        | _, [ DList l ] -> List.tryLast l |> Option.defaultValue DNull |> Value
+        | _, [ DList l ] ->
+            if l.Length > 0 then List.last l else DNull
+            |> Value
         | args -> incorrectArgs ())
       sqlSpec = NotYetImplementedTODO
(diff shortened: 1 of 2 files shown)


fsharp-backend/src/LibExecution/StdLib/LibList.fs
```diff
@@ -66,29 +66,24 @@ let fns : List<BuiltInFn> =
...
-//     ; sqlSpec = NotYetImplementedTODO
-//       ; previewable = Pure
-//     ; deprecated = NotDeprecated }
+    { name = fn "List" "tail" 0
+      parameters = [Param.make "list" (TList varA) ""]
+      returnType = TList varA
+      description =
+        "If the list contains at least one value, returns `Just` a list of every value other than the first. Otherwise, returns `Nothing`."
+      fn =
+        (* This matches Elm's implementation, with the added benefit that the error rail
```
pbiggar, line 70: Can you change these comments to use `//` instead of `(*`?

fsharp-backend/src/LibExecution/StdLib/LibList.fs
```diff
@@ -45,59 +45,51 @@ let fns : List<BuiltInFn> =
...
+         * means you don't need to handle unwrapping the option
+         * unless the passed list is truly empty (which shouldn't happen in most practical uses). *)
+        (function
+        | _, [ DList l ] ->
+            (match List.tryLast l with
```
pbiggar, line 75: TryLast is very slow, it recurses to the end of the list. I think you could use `if List.length l > 0` instead.

fsharp-backend/tests/testfiles/list.tests
```diff
@@ -73,3 +73,9 @@ List.filter_v0 [-20; 5; 9;] (fun x -> x < 1) = [-20]
 List.all_v0 [0; 1; 2; 3;] (fun elem -> elem > 0) = false
 List.all_v0 [1; 2; 3;] (fun elem -> elem > 0) = true
 List.all_v0 [0; -1; -2] (fun elem -> elem < 0) = false
+List.empty_v0 = []
+List.tail_v0 [10, 20, 30, 40] = [20, 30, 40]
+List.tail_v0 [] = Nothing
+List.head_v2 [1, 2, 3] = 1
+List.head_v2 [] = Nothing
+List.last_v0 [1000, 2000, 3000] =  3000
```
pbiggar, line 81: Can you add a test for last on the empty list?

Can you add a List.last_v1 that behaves like List.head_v2?

pbiggar (changes requested):
Nice job. A couple small changes and we're good to go.

fsharp-backend/src/LibExecution/StdLib/LibList.fs
```diff
@@ -118,29 +118,26 @@ let fns : List<BuiltInFn> =
...
+        | _, [ DList l ] ->
+            if l.Length > 0 then List.last l else DNull
+            |> Value
+        | args -> incorrectArgs ())
+      sqlSpec = NotYetImplementedTODO
+      previewable = Pure
+      deprecated = ReplacedBy(fn "" "" 0) }
+    { name = fn "List" "last" 1
+      parameters = [Param.make "list" (TList varA) ""]
+      returnType = varA
```
pbiggar, line 127: `varA` should be `TOption varA`

fsharp-backend/src/LibExecution/StdLib/LibList.fs
```diff
@@ -118,29 +118,26 @@ let fns : List<BuiltInFn> =
...
+      description = "Returns the last value in `list`, wrapped in an option (`Nothing` if the list is empty)."
+      fn =
+        (function
+        | _, [DList l] ->
+            if l.Length > 0 then List.last l else DOption None
```
pbiggar: I think this fn can just be `Value (DOption (List.tryLast l))`, right?

fsharp-backend/tests/testfiles/list.tests
```diff
@@ -79,4 +79,6 @@ List.tail_v0 [] = Nothing
 List.head_v2 [1, 2, 3] = 1
 List.head_v2 [] = Nothing
 List.last_v0 [1000, 2000, 3000] =  3000
-List.last_v0 [] =  null
+List.last_v0 [] = null
+List.last_v1 [1, 2, 3] =  3
```
pbiggar, line 83: Should be `Some 3`, no?

pbiggar (changes requested):
Looks nice, thanks! A few more small issues and we can merge.

fsharp-backend/src/LibExecution/StdLib/LibList.fs
Athinanarof: Right!

fsharp-backend/tests/testfiles/list.tests
Athinanarof: It seems it is Just 3.

pbiggar: Perfect.

PR #2893, opened 2020-12-11
Libstring fs

Athinanarof:
Translate LibString.fs functions.

Functions implemented and tested.

Functions implemented and tested.

Multiple test cases included for each function

Revert changes


Athinanarof:
Hi Paul!
I noticed the functions toFloat, toInt, toUppercase are very similar, but I'm not sure if I need to wrap some of them in other structures. Please let me know.

I have two questions: 

1. how do I compare in the .tests files an Error result? I tried something like Error "string content"
image

2. For the split function, in the test case String.split_v0 "34564" "" = [ "34564" ], I think we should return ["3", "4", "5", "6", "4"], right?

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -138,25 +138,23 @@ let fns : List<BuiltInFn> =
...
-//      ; deprecated = ReplacedBy(fn "" "" 0) }
+    { name = fn "String" "toInt" 0
+      parameters = [Param.make "s" TStr ""]
+      returnType = TInt
+      description = "Returns the int value of the string"
+      fn =
+        (function
+        | _, [DStr s] ->
+            ( try
+                bigint (s |> int)
```
pbiggar: You can't coerce to ints or you'll lose data. If you add a test case for a really large int (> 64 bits), you'll see.

As it happens, I implemented something similar for another test (just merged to main). You want System.Numerics.BigInteger.Parse.

fsharp-backend/tests/testfiles/string.tests
```diff
@@ -55,7 +55,8 @@ String.replaceAll_v0 "abcABCcbaCBA" "b" "x" = "axcABCcxaCBA"
 // String.split_v0 "hello😄world" "😄" = [ "hello", "world" ] // String split works on strings with emoji + ascii
 // String.startsWith_v0 "a string" " s" = false // startsWith works
 // String.startsWith_v0 "a string" "a s" = true // startsWith works
-// String.toInt_v0 "1" = 1 // string_toint_works
+String.toInt_v0 "1" = 1
+String.toInt_v0 "55555555" = 55555555
```
pbiggar, line 67: Can you add a test for very very large numbers, negative numbers, and zero?

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -155,24 +158,24 @@ let fns : List<BuiltInFn> =
...
-//             ( try DResult (ResOk (DInt (Dint.of_string_exn utf8)))
-//               with e ->
-//                 error_result
-//                   ("Expected a string with only numbers, got " ^ utf8) )
-//         | args ->
-//             incorrectArgs ())
-//      ; sqlSpec = NotYetImplementedTODO
-//      ; previewable = Pure
-//      ; deprecated = NotDeprecated }
+    { name = fn "String" "toInt" 1
```
pbiggar, line 159: I added this, it's on main now. You can just replace this (same notes about int/bigint data loss as before.

fsharp-backend/tests/testfiles/string.tests
```diff
@@ -62,6 +62,11 @@ String.toInt_v0 "-11111" = -11111
 String.toInt_v0 "+333" = 333
 String.toInt_v0 "-11111" = -11111
 String.toInt_v1 "1" = 1
+String.toFloat_v0 "1.5" = 1.5
```
pbiggar, line 73: Can you add a test for a large float (larger than would fit in 32 bits)? The code looks right, I just like to handle edge cases.

fsharp-backend/tests/testfiles/string.tests
```diff
@@ -67,6 +67,11 @@ String.toFloat_v0 "0.0" = 0.0
 String.toFloat_v0 "-0.5" = -0.5
 String.toFloat_v0 "+0.5" = 0.5
 String.toFloat_v0 ".5" = 0.5
+String.toFloat_v0 "1.5" = 1.5
```
pbiggar, line 78: I think you mean these to be `_v1`

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -193,24 +193,25 @@ let fns : List<BuiltInFn> =
...
-//               with e ->
-//                 error_result
-//                   "Expected a string representation of an IEEE float" )
-//         | args ->
-//             incorrectArgs ())
-//      ; sqlSpec = NotYetImplementedTODO
-//      ; previewable = Pure
-//      ; deprecated = NotDeprecated }
+      deprecated = ReplacedBy(fn "String" "toFloat" 1) }
+    { name = fn "String" "toFloat" 1
```
pbiggar, line 190: This isn't the right implementation. Look at the OCaml code. It's supposed to return a float wrapped in a Result.

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -195,8 +195,8 @@ let fns : List<BuiltInFn> =
       previewable = Pure
       deprecated = ReplacedBy(fn "String" "toFloat" 1) }
     { name = fn "String" "toFloat" 1
-      parameters = [Param.make "s" TStr]
-      returnType = TResult
+      parameters = [Param.make "s" TStr ""]
+      returnType = TFloat
```
pbiggar, line 192: This should be `TResult(TFloat, TString)`

fsharp-backend/tests/testfiles/string.tests
```diff
@@ -73,10 +73,17 @@ String.toFloat_v0 "-0.5" = -0.5
 String.toFloat_v0 "+0.5" = 0.5
 String.toFloat_v0 ".5" = 0.5
 String.toList_v1 "abc" = [ 'a', 'b', 'c'] // String.toList_v1 works
-// String.toLowercase_v1 "ABCDEF" = "abcdef" // String.toLowercase_v1 works for ASCII range
-// String.toUppercase_v1 "abcdef" = "ABCDEF" // String.toUppercase_v1 works for ASCII range
+String.toLowercase_v0 "ABCDEF" = "abcdef" // String.toLowercase_v0 works for ASCII range
+String.toLowercase_v0 "AB323CDEF" = "ab323cdef"
+String.toLowercase_v0 "sánchez" = "sánchez"
```
pbiggar, line 86: Can you add a lowercase test for "ŻÓŁW"?

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -319,45 +309,47 @@ let fns : List<BuiltInFn> =
...
-//      { name = fn "String" "slugify" 1
+    { name = fn "String" "slugify" 0
+      parameters = [ Param.make "string" TStr "" ]
+      returnType = TStr
+      description = "Turns a string into a slug"
+      fn =
+        (function
+        | _, [ DStr s ] ->
+
+            let to_remove = @"[^a-z0-9\s-]"
```
pbiggar, line 331: The `to_remove`s here look pretty different. Why is that?

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -319,45 +309,47 @@ let fns : List<BuiltInFn> =
...
+            let spaces = @"[-\s]+"
+
+            let objRegex (pattern : string)
+                         (input : string)
+                         (replacement : string)
+                         =
+              Regex.Replace(input, pattern, replacement)
+
+            s
+            |> fun s -> objRegex (to_remove) (s) ("")
```
pbiggar, line 339: You don't need parens around the arguments here.

fsharp-backend/tests/testfiles/string.tests
```diff
@@ -52,6 +52,7 @@ String.replaceAll_v0 "abcABCcbaCBA" "b" "x" = "axcABCcxaCBA"
 // String.slice_v0 "abcd" 2 6 = "cd"
 // String.slice_v0 "abcd" 3 2 = ""
 // String.slice_v0 "abcd" 5 6 = ""
+String.slugify_v0 "  m@y  'super'  really- excellent *uber_ amazing* ~very  ~ \"clever\" thing: coffee😭! " = "my-super-really-excellent-uber-amazing-very-clever-thing-coffee" // Slugify works
```
pbiggar, line 55: This works too well for slugify_v0 I think. the goal is to have it continue to work exactly the same as in OCaml. Can you check the output of the ocaml version on darklang.com?

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -406,16 +406,27 @@ let fns : List<BuiltInFn> =
...
-//      ; parameters = [Param.make "string" TStr]
-//      ; returnType = TStr
-//      ; description = "Reverses `string`"
-//      ; fn =
-//         (function
-//         | _, [DStr s] -> DStr (Unicode_string.rev s) | args -> incorrectArgs ())
-//      ; sqlSpec = NotYetImplementedTODO
-//      ; previewable = Pure
-//      ; deprecated = NotDeprecated }
+    { name = fn "String" "reverse" 0
```
pbiggar, line 407: I don't think this works right - in the OCaml version we used a unicode reversing function. Here you use a character reversing function. I think this will break for unicode and similar. Can you add tests for it?

I'm sure there's like a String.reverse() method which does this right.

pbiggar (changes requested):
Thanks! There's a bunch of changes needed still - watch out for unicode vs non-unicode!

pbiggar:
> Hi Paul!
> I noticed the functions toFloat, toInt, toUppercase are very similar, but I'm not sure if I need to wrap some of them in other structures. Please let me know.

They should work exactly the same as the OCaml ones. If you don't know what the OCaml version returns, try it out on darklang.com or darklang.localhost:8000/a/dark (username "dark", password "what").

> I have two questions:
> 
> 1. how do I compare in the .tests files an Error result? I tried something like Error "string content"
>    image

I think you should be able to do `Error "string content"` just fine. That should create a `DResult (Error "string content")` which seems to be what you want.

> 2. For the split function, in the test case String.split_v0 "34564" "" = [ "34564" ], I think we should return ["3", "4", "5", "6", "4"], right?

Check out what happens when you try this on darklang.com and emulate it.

fsharp-backend/src/LibExecution/StdLib/LibString.fs
Athinanarof: There is a rev(string) function, I just tried it, but it does not work properly with Unicode. After reading some documentation, I have implemented ParseCombiningCharacters method to get each character or surrogate index and then work with them. Let me know what do you think :)

fsharp-backend/tests/testfiles/string.tests
Athinanarof: This is the result, 
image

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -150,7 +150,7 @@ let fns : List<BuiltInFn> =
         (function
         | _, [ DStr s ] ->
             (try
-              bigint (s |> int) |> DInt |> Value
+              s |> System.Numerics.BigInteger.Parse |> DInt |> Value
```
gusty, line 154: Here you can use `System.Numerics.BigInteger.TryParse` instead of doing a try-catch over the "unsafe" `Parse` function.

You can do:

```
match System.Numerics.BigInteger.TryParse with
| true , v -> v |> DInt |> Value
| false, _ -> Value(errStr ("Expected a string with only numbers")))
```

Alternatively, you can use something like FSharpPlus generic `tryParse` and map it with `Option.map` or stuff like that, if you want to use the combinators approach instead.

fsharp-backend/tests/testfiles/string.tests
```diff
@@ -62,6 +62,11 @@ String.toInt_v0 "-11111" = -11111
 String.toInt_v0 "+333" = 333
 String.toInt_v0 "-11111" = -11111
 String.toInt_v1 "1" = 1
+String.toFloat_v0 "1.5" = 1.5
+String.toFloat_v0 "0.0" = 0.0
+String.toFloat_v0 "-0.5" = -0.5
+String.toFloat_v0 "+0.5" = 0.5
```
pbiggar, line 91: great idea to test `+0.5` as well!

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -419,15 +420,13 @@ let fns : List<BuiltInFn> =
       fn =
         (function
         | _, [ DStr s ] ->
-            let append (reversedString : char list) (ch : char) : char list =
-              ch :: reversedString
+            let reverseString str =
+                StringInfo.ParseCombiningCharacters(str)
```
pbiggar: I'd suggest using Prelude.String.toEgcSeq to make a sequence of EGCs (which are what dark uses for characters) and then reversing and then using String.concat.

pbiggar (commented):
I'm going to merge this so you don't have to do another rebase. Can you follow up on this comment later? https://github.com/darklang/dark/pull/2893/files/56efcf0cef707e4c022bb6c7e6f48ee338991705..6f136839d50b25d90947591b3ceea377380be340#diff-485d62bb2fbc5ac34002d1401a84955dadb69a2bc568be890301c71c93577128

PR #2901, opened 2020-12-24
Libstring fs

Athinanarof:
Last PR comments addressed.

Functions modified and new test cases included..

Test cases included.

Revert changes.


Athinanarof:
Hi @pbiggar, new PR ready. 😁

pbiggar:
Nice work!

PR #2905, opened 2021-01-04
Libstring fs

Athinanarof:
LibString.fs functions implemented. Unit test included.
Functions implemented.

Unit test included for each function. However, I have 3 test cases of slice_v0 to continue working on. WIP.

Revert changes.


Athinanarof:
Hi @pbiggar I have created a new MR.

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -883,8 +883,25 @@ let fns : List<BuiltInFn> =
       fn =
         (function
         | _, [ DStr s; DInt n ] ->
-            let characterCount = int n
-            Value(DStr(s.[characterCount..]))
+            let drop_first_n s (num_egcs : bigint) =
```
pbiggar, line 1000: This looks a bit complex. I think you could have the same effect by doing:

```
s
|> String.toEgcSeq
|> Seq.toList
|> List.skip count
|> String.concat ""

Athinanarof: Hi @pbiggar, I know there were simpler ways to make some functions work but what I did was to translate the current implementation in Ocalm, like in this one uses Unicode_string.drop_first_, I thought I should be doing that 😅
image

pbiggar (commented):
I was going to suggest changing all the behaviour to more idiomatic functional code, but I see your point, so I guess it's better to stick with direct translations.

Can you:
- make sure there are complex emoji tests for everything
- change the code to use camelCase
- use `0I` or `1I` instead of `bigint0` or `bigint`

And I guess we'll ship like that.

Thanks!

Athinanarof:
Hi @pbiggar, changes are done; I just found two functions where the complex emojis do not work; I will continue with the file in the morning, and I will add those in the next MR.

pbiggar:
Great, thanks! Here's a bunch more challenging unicode characters to try: https://stackoverflow.com/a/51539774/104021

PR #2923, opened 2021-01-27
fix regex typo in validateUserName

kclauson:


pbiggar:
Thanks, great catch! Thanks for reading the F# code, really appreciate it. It's a lot of new code, so it's great to know someone is reading it. Thanks again 🙏

Issue #2863, opened 2020-11-09
Dark editor does not allow you to create a list consisting of several values produced by |>


Stvad:
You can construct list of one element, but if you are to add a coma after that element - editor won't understand it: 

![image](https://user-images.githubusercontent.com/81347/98492351-6d056280-21ec-11eb-9262-d04c6683b1e6.png)

If you just press enter - the editor wold extend the pipeline:
![image](https://user-images.githubusercontent.com/81347/98492358-6f67bc80-21ec-11eb-8433-cabb70113646.png)

Slack discussion: https://darkcommunity.slack.com/archives/CNTCWD6MP/p1604881807171700

xtopherbrandt:
@pbiggar looking at this bug, I'd like to know your thinking about how to fix it. The main problem is that when the cursor is at the end of a variable within a list (and nothing is selected), createPipe (in Fluid.ml) is given the id of the list and no other context to know if the cursor is inside the list or outside.  The key handling code in updateKey uses getTopmostSelectionID which finds the list. 

One option would be to make the key handling code for ShiftEnter a bit smarter and set findParent to false if the cursor is inside a list. I think this would be a simpler change but spreads out the logic of finding the right thing to pipe from.

Another option would be to provide more context about the cursor location to createPipe and probably findAppropriateParentToWrap. This would be a bigger riskier change but may centralize the logic.

pbiggar:
Thanks for looking at this @xtopherbrandt. I'm not sure which the correct solution is. This code is very hacky, but is held together by lots of unit tests, so if the tests pass you can consider it a decent solution. (you'll definitely want to add some tests too!)

The editor was designed with the knowledge that we're going to have to support lots of edge cases, though obviously if you find a way to clean it up sincely, that would be great. I'll note that getTopmostSelectionID might not be the best function to use for an "empty" selection, though to my knowledge it works quite well for non-empty selections.

pbiggar:
This was fixed by @xtopherbrandt in #2930.

PR #2930, opened 2021-02-03
Issue #2863

xtopherbrandt:
Elements within a list are not currently pipeable. The main problem is that when the cursor is at the end of a variable within a list (and nothing is selected), createPipe (in Fluid.ml) is given the id of the list and no other context to know if the cursor is inside the list or outside. The logic simply assumes that if nothing is selected it should find the top most element and pipe from there. 

Add a list of pipeable Fluid Tokens to check if the element being piped from is pipeable. So when nothing is selected the token to the left is checked and if it is pipeable a pipe is created from it.

With no token selected and pressing shift-enter:
Current behaviour:
image

New behaviour:
image

Three new tests have been added and two have been updated to check the new behaviour.

Back out the changes to Fluid.ml and fluid_test.ml

No

### DIFF
--- a/client/src/fluid/Fluid.ml
+++ b/client/src/fluid/Fluid.ml
@@ -2021,9 +2021,8 @@ let rec findAppropriateParentToWrap
     (oldExpr : FluidExpression.t) (ast : FluidAST.t) : FluidExpression.t option
     =
   let child = oldExpr in
-  Js.log2 "child" child ;
   let parent = FluidAST.findParent (E.toID oldExpr) ast in
-  Js.log2 "parent" parent ;
+
   match parent with
   | Some parent ->
     ( match parent with
@@ -2073,7 +2072,7 @@ let createPipe ~(findParent : bool) (id : ID.t) (astInfo : ASTInfo.t) :
   let action =
     Printf.sprintf "createPipe(id=%s findParent=%B)" (ID.toString id) findParent
   in
-  Js.log2 "createPipe id" id ;
+
   let astInfo = recordAction action astInfo in
   let exprToReplace =
     FluidAST.find id astInfo.ast
@@ -4169,17 +4168,11 @@ let rec updateKey
             if startPos = endPos
             then
               begin
-                Js.log "No selection" ;
                 let tokenAtLeft = getLeftTokenAt astInfo.state.newPos (ASTInfo.activeTokenInfos astInfo) in
                 match tokenAtLeft with
                 | Some current when T.isPipeable current.token ->
-                  Js.log "Pipeable Token at left " ;
-                  Some (T.tid current.token), false
-                | Some ({token = TPipe _; _} as current) ->
-                  Js.log "Pipe Token at left " ;
                   Some (T.tid current.token), false
                 | _ ->
-                  Js.log "Unmatched Token at left " ;
                   match topmostSelectionID with
                   | Some id -> Some (id), startPos = endPos;
                   | None -> Some (T.fakeid), startPos = endPos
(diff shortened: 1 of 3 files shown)


pbiggar:
This looks great, thank you! Will test it and give it closer review later today but the code looks mergeable as is.

CI isn't passing - it looks like the formatting is off. Can you run the formatter on it? `./scripts/format format client` should do it.

xtopherbrandt:
Sorry I hadn't run `format client` . I have now.
How do I include that commit into this PR?

pbiggar:
The failures are due to CI config, unrelated to this. So nothing to worry about there.

client/src/fluid/Fluid.ml
```diff
@@ -4162,8 +4162,34 @@ let rec updateKey
     | Keypress {key = K.ShiftEnter; _}, left, _ ->
         let doPipeline (astInfo : ASTInfo.t) : ASTInfo.t =
           let startPos, endPos = FluidUtil.getSelectionRange astInfo.state in
-          let topmostID = getTopmostSelectionID startPos endPos astInfo in
-          let findParent = startPos = endPos in
+          let topmostSelectionID =
+            getTopmostSelectionID startPos endPos astInfo
+          in
+          let topmostID, findParent =
```
pbiggar, line 4168: It seems like findParent is always `startPos = endPos`, so that could be simplified.

client/src/fluid/Fluid.ml
```diff
@@ -4162,8 +4162,34 @@ let rec updateKey
...
+              in
+              match tokenAtLeft with
+              | Some current when T.isPipeable current.token ->
+                  (Some (T.tid current.token), false)
+              | _ ->
+                ( match topmostSelectionID with
+                | Some id ->
+                    (Some id, startPos = endPos)
+                | None ->
+                    (Some T.fakeid, startPos = endPos) )
```
pbiggar, line 4184: I'm unclear on why we'd use `Some T.fakeid` instead of `None`? If it is a good idea, can you make topmostID no longer be an option anymore?

pbiggar (changes requested):
This looks nice and works well. I have a few questions, but we should be good to merge shortly.

client/src/fluid/Fluid.ml
xtopherbrandt: There is one case where findParent is explicitly set to false (line 4178), when the token to the left is pipeable.

client/src/fluid/Fluid.ml
xtopherbrandt: We shouldn't be using T.fakeid here. Replaced with None.
Taking the option out of topmostID is a bit beyond my OCaml refactoring ability at the moment.

pbiggar: If we're keeping none in, then no need to refactor :)

client/src/fluid/Fluid.ml
```diff
@@ -4162,8 +4162,34 @@ let rec updateKey
...
+              let tokenAtLeft =
+                getLeftTokenAt
+                  astInfo.state.newPos
+                  (ASTInfo.activeTokenInfos astInfo)
+              in
+              match tokenAtLeft with
+              | Some current when T.isPipeable current.token ->
+                  (Some (T.tid current.token), false)
+              | _ ->
+                ( match topmostSelectionID with
```
pbiggar, line 4180: Can you pull this out into a variable, since it's used on both branches?

client/src/fluid/Fluid.ml
pbiggar: Gotcha, thanks!

pbiggar:
One more change, and then it's mergeable. thanks again!

client/src/fluid/Fluid.ml
xtopherbrandt: Yup, Good catch.

pbiggar:
Wonderful, thanks!

PR #2932, opened 2021-02-04
Libbytes.fs

Athinanarof:
Libbytes.fs functions implementation.

Functions translated 
Functions implemented and unit test included.
Test cases are working as the old Ocaml version does.

Revert changes


Athinanarof:
Hello @pbiggar here is a new pull request 🤞

pbiggar:
Looks good, nice. Am rerunning CI, and should be able to merge then.

PR #2933, opened 2021-02-04
Libstring.fs

Athinanarof:
Libstring.fs functions

Functions implemented

Test cases included and working as the old Ocaml version does.

Revert changes.


Athinanarof:
Hi @pbiggar here is another PR. :)

fsharp-backend/tests/testfiles/string.tests
```diff
@@ -370,6 +368,31 @@ String.padEnd_v0 "﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽" "0" 20 = "
 String.padEnd_v0 "👱👱🏻👱🏼👱🏽👱🏾👱🏿" "0" 10 = "👱👱🏻👱🏼👱🏽👱🏾👱🏿0000"
 String.padEnd_v0 "🧟‍♀️🧟‍♂️" "0" 5 = "🧟‍♀️🧟‍♂️000"
 String.padEnd_v0 "👨‍❤️‍💋‍👨👩‍👩‍👧‍👦🏳️‍⚧️🇵🇷" "0" 10 = "👨‍❤️‍💋‍👨👩‍👩‍👧‍👦🏳️‍⚧️🇵🇷000000"
+String.toUUID_v0 "psp-soslsls==" = (Test.typeError_v0 "`uuid` parameter was not of form XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX")
+String.toUUID_v0 "123456" = (Test.typeError_v0 "`uuid` parameter was not of form XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX")
+String.toUUID_v0 "d388ff30-667f-11eb-ae93-0242ac13000" = (Test.typeError_v0 "`uuid` parameter was not of form XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX")
+String.toUUID_v1 "👱🏼" = Error "`uuid` parameter was not of form XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"
```
Athinanarof, line 374: @pbiggar quick question, for the toUUID, the returned value is a TUUID. How can I compare against this type on a .test file?

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -749,81 +749,81 @@ Don't rely on either the size or the algorithm."
...
+    { name = fn "String" "toUUID" 0
+      parameters = [ Param.make "uuid" TStr "  " ]
+      returnType = TUuid
+      description =
+        "Parse a UUID of form XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX from the input `uuid` string"
+      fn =
+        (function
+        | _, [ DStr s ] ->
+            match s with
+            | Parse (x : System.Guid) -> System.Guid.Parse s |> DUuid |> Value
```
pbiggar, line 761: I dont understand what's happening here? How does a String match with `| Parse`? Is this an active pattern?

fsharp-backend/src/LibExecution/StdLib/LibString.fs
```diff
@@ -749,81 +749,81 @@ Don't rely on either the size or the algorithm."
...
+    { name = fn "String" "toUUID" 1
+      parameters = [ Param.make "uuid" TStr "" ]
+      returnType = TUuid
+      description =
+        "Parse a UUID of form XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX from the input `uuid` string"
+      fn =
+        (function
+        | _, [ DStr s ] ->
+            match s with
+            | Parse (x : System.Guid) ->
```
pbiggar, line 781: Same question as above

fsharp-backend/tests/testfiles/string.tests
pbiggar: I would call toString the newly created UUID and verify it has the same contents.

pbiggar (commented):
A few questions on this.

fsharp-backend/src/LibExecution/StdLib/LibString.fs
gusty: Yes, `Parse` is an old active pattern defined in F#+ which use a generic `tryParse` function internally. The name is a bit unfortunate, as it should be called `Parsed` instead.

So, doing `| Parse (x : System.Guid) -> System.Guid.Parse s |` makes no sense as `x` contains the already "parsed" value, so here we're parsing twice.

Also, since `Parse` is an active patters it could be combined with the DU pattern in one, so:

```
(function
    | _, [DStr (Parse x)] -> x |> DUuid |> Value
    | _ -> ....
```
and as you can see the type annotation is not needed, since there is a `DUuid` which AFAIK it accepts a GUID which makes enough context for type inference.

Hope this helps !
should do the work.

Athinanarof: Thank you @gusty for the great explanation :)

pbiggar: I'd prefer to avoid using f#+ where possible (esp in stdlib). I'm also a little weary of active patterns as they obscure what's happening a bit imo. Would you mind doing the tryParse directly?

Athinanarof: Understood @pbiggar, now I'm aware of that  :) I pushed the changes. Thank you.

pbiggar:
Great, thank you!

PR #2934, opened 2021-02-05
Fix typo in Api.fs

kclauson:


pbiggar:
Nice catch, thank you!

PR #2936, opened 2021-02-08
Libfloat fs

Athinanarof:
Implement LibFloat functions.
Functions translated.

Unit test included for each function, only still pending support nan in tests.

Revert changes


Athinanarof:
Hi @pbiggar here is a new PR :)

fsharp-backend/src/LibExecution/StdLib/LibFloat.fs
```diff
@@ -1,317 +1,317 @@
...
+//   l
+//   |> List.map (fun dv ->
+//          match f dv with Some v -> Result.Ok v | None -> Result.Error (l, dv))
+//   |> Result.all
+
+
+// let ( >>| ) = Result.( >>| )
+
+let fns : List<BuiltInFn> =
+  [ { name = fn "Float" "ceiling" 0
```
pbiggar, line 20: The old function had 2 names: "Float::ceiling" and "Float::roundUp" (its been mangled a bit by some regexes). So you need to add a function that does the same thing called Float::roundup_v0.

fsharp-backend/src/LibExecution/StdLib/LibFloat.fs
```diff
@@ -29,22 +29,19 @@ let fns : List<BuiltInFn> =
...
-
-  //         (function
-  //         | _, [DFloat a] ->
-  //             DInt (Float.round_down a |> Dint.of_float)
-  //         | args ->
-  //             incorrectArgs ())
-  //   ; sqlSpec = NotYetImplementedTODO
-  //     ; previewable = Pure
-  //   ; deprecated = NotDeprecated }
+    { name = fn "Float" "floor" 0
```
pbiggar, line 31: Needs a "Float::roundDown_v0" too

fsharp-backend/tests/testfiles/float.tests
```diff
@@ -6,10 +6,16 @@ Float.ceiling_v0 1.3 = 2
 Float.ceiling_v0 1.0000001 = 2
 Float.ceiling_v0 -2147483647.8 = -2147483647
 Float.ceiling_v0 2147483647.0 = 2147483647
+Float.ceiling_v0 0000000.1 = 1
```
pbiggar, line 6: This seems like the wrong answer.

fsharp-backend/tests/testfiles/float.tests
```diff
@@ -4,27 +4,30 @@ Float.ceiling_v0 1.0000001 = 2
...
 Float.absoluteValue_v0 -5.6 = 5.6
-Float.absoluteValue_v0 -0 = 0
-Float.absoluteValue_v0 -0 = 0
-
+Float.absoluteValue_v0 -0.0 = 0.0
+Float.absoluteValue_v0 -0.0 = 0.0
+// //Float.negate_v0 (0.0 / 0.0) = nan // TODO: support nan in tests
+// //Float.negate_v0 (1.0 / 0.0) = -inf // TODO: support nan in tests
+Float.negate_v0 -5.6 = 5.6
+Float.negate_v0 5.6 = -5.6
```
pbiggar, line 30: Can you add a test for 0 and -0?

pbiggar (changes requested):
Thanks! A few fixes and we should be good to go. Can you add a test for `-0.0` and `0.0` for all functions too?

pbiggar:
Merged in https://github.com/darklang/dark/pull/2945, thanks!

I didn't do the fixes, will do them in my branch.

PR #2955, opened 2021-02-23
Libfloat.fs

Athinanarof:
Libfloat functions ported.

Functions implemented.

Unit test included.

Revert changes


Athinanarof:
Hi @pbiggar, can you take a look at this PR please :).

fsharp-backend/tests/testfiles/float.tests
```diff
@@ -42,16 +42,16 @@ Float.negate_v0 -5.6 = 5.6
...
-// Float.clamp_v0 -2.0 -5.0 5.0 = -2.0
-// Float.clamp_v0 -3.0 -2.0 1.0 = -2.0
-// Float.clamp_v0 -5.0 1.0 1.0 = 1.0
-// Float.clamp_v0 0.5 (Test.negativeInfinity_v0) 1.0 = 0.5
-// Float.clamp_v0 1.0 2.0 1.0 = 1.0
-// Float.clamp_v0 3.0 0.0 2.0 = 2.0
-// Float.clamp_v0 0.5 Test.nan_v0 1.0 = Test.typeError "Internal Float.clamp exception"
-// Float.clamp_v0 0.5 1.0 Test.nan_v0 = Test.typeError "Internal Float.clamp exception"
+//Float.clamp_v0 (0.0 / 0.0) -1.0 1.0 = -Test.nan_v0
+Float.clamp_v0 (1.0 / 0.0) -1.0 0.5 = 0.5
```
pbiggar, line 46: You can use NaN instead of dividing by zero here.

fsharp-backend/src/LibExecution/StdLib/LibFloat.fs
```diff
@@ -272,37 +272,33 @@ let fns : List<BuiltInFn> =
...
+      description =
+         "If `value` is within the range given by `limitA` and `limitB`, returns `value`.
+          If `value` is outside the range, returns `limitA` or `limitB`, whichever is closer to `value`.
+         `limitA` and `limitB` can be provided in any order."
+      fn =
+        (function
+        | _, [DFloat v; DFloat a; DFloat b] ->
+            let output = false in
+            let min, max = if a < b then (a, b) else (b, a) in
+            ( match Single.TryParse(Math.Clamp(v, min, max), output) with
```
pbiggar, line 287: What is this for? Math.Clamp seems to return a double, and parsing it doesn't make sense. I'm also unclear on why we'd use Single instead of Double.

pbiggar (changes requested):
Looks good, thanks!

I'm not sure what's going on with Math.Clamp though.

pbiggar:
Copied into #2958

PR #2984, opened 2021-03-23
Add tests for Dict::get_v0 and Dict::get_v1

levlaz:
Hi - this is my first pull request. I noticed that Dict::get_v0 and Dict::get_v1 didn't have any tests so I added some.


pbiggar:
Awesome. If you had a moment, would love to see tests for the negative cases too!

PR #2985, opened 2021-03-24
Port Twitter::urlencode function

levlaz:
This PR ports over the Twitter::urlencode function from OCaml to F#. 

The main change is removing the use of the previous OCaml unicode library, and use the built in `Uri.EscapeDataString` method from .NET. 

The existing twitter.tests unit test passes, I also tested it out against the [examples on the Twitter documentation](https://developer.twitter.com/en/docs/authentication/oauth-1-0a/percent-encoding-parameters) and the output was correct.


pbiggar:
Wonderful, thank you. Gonna merge as is, but it would be great to have the twitter documentation examples added as tests if possible.

PR #2986, opened 2021-03-24
Implement FSharp version of Int::sum_v0

jwalter:
Hello, this is my first code contribution to Dark. I'm totally new to both F# and OCaml so I 100% expect there is room for improvement in this PR.

### DIFF
--- a/fsharp-backend/src/LibExecution/StdLib/LibInt.fs
+++ b/fsharp-backend/src/LibExecution/StdLib/LibInt.fs
@@ -275,34 +275,27 @@ let fns : List<BuiltInFn> =
       sqlSpec = NotQueryable
       previewable = Pure
       deprecated = NotDeprecated }
-    // ; { name = fn "Int" "sum" 0
-    //   ; parameters = [Param.make "a" TList ""]
-    //   ; returnType = TInt
-    //   ; description = "Returns the sum of all the ints in the list"
-    //   ; fn =
-    //         (function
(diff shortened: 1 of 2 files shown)


fsharp-backend/src/LibExecution/StdLib/LibInt.fs
```diff
@@ -275,34 +275,27 @@ let fns : List<BuiltInFn> =
...
+      fn =
+        (function
+        | _, [DList l] ->
+            let ints =
+              List.map
+                (fun i ->
+                  match i with
+                  | DInt it -> it
+                  | _ ->
+                      Errors.throw "Sum expects you to pass a list of ints")
```
pbiggar, line 291: Can you use `Errors.argumentWasnt "a list of strings" ...` here?

pbiggar (changes requested):
Awesome, thanks! Just one minor change and then we should be able to ship it.

fsharp-backend/src/LibExecution/StdLib/LibInt.fs
jwalter: Absolutely. I assume it should be "a list of ints" though?

pbiggar: oh yeah 🤦

pbiggar:
Wonderful, thank you!

PR #2988, opened 2021-03-27
Implement Float::sum in F#

jwalter:
Float::sum is not implemented in F#

Implement using 'fold'

Added unit tests in float.tests for successful and failed execution


pbiggar:
Fantastic, thank you!

PR #2989, opened 2021-03-27
Implement parts of F# Dict

jwalter:
F# version of Dict is not fully implemented

This PR implements a handful of the less complicated functions in Dict

All implemented functions are covered by unit tests


pbiggar:
Fantastic, thanks so much for the contribution. Looks good as is!

PR #2995, opened 2021-03-30
Add implementation and test for List::reverse

mariajdab:
Implementation and test for List::reverse 
Implementation in F# List::reverse and the tests
Implementation in F# List::reverse and the tests
I ran ./scripts/builder --compile --test --watch and all the test passed
It will be run `git revert <commit hash>`
Yes


pbiggar:
Wonderful, thanks!

PR #2997, opened 2021-03-31
Test cases for Int::clamp

mjeffryes:
Hi - this is my first pull request. I noticed that Int::clamp didn't have any tests so I added one.


pbiggar:
Wonderful, thanks!

CI is failing but it's unrelated. I'm rerunning it and that should be fine.

Would love to see some tests for larger numbers (the F# implementation of int is infinite-precision, so numbers that are larger than 64-bit would be ideal). You might need `// FSHARPONLY` on those to make the tests pass.

I'll merge this as-is once the tests pass. If you had time for the extras, we can do another PR.

PR #2998, opened 2021-04-01
List functions

mariajdab:
## What is the problem/goal being addressed?
Added functions to LibList and the test cases in list.test
## What is the solution to this problem?
Added functions to LibList and the test cases in list.test
## What are the changes here? How do they solve the problem and what other product impacts do they cause?
*Please include before/after screenshots/gifs if appropriate*
Added the functions: List::drop, List::take, List::repeat, List::getAt_v0, List::getAt_v1 and the test cases for each one of them
Changed for the function List::getAt_v0 the line: deprecated = ReplacedBy(fn "" "" 0) to deprecated = NotDeprecated 
## How are you sure this works/how was this tested?
Test cases passed
## What is the reversion plan if this fails after shipping?
It will be run `git revert <commit hash>`
## Has this information been included in the comments?
Yes

PR #2999, opened 2021-04-02
Port More List functions

mjeffryes:
Porting stdlib List functions to F# backend

Ported `List::repeat_v0`, `List::drop_v0` and `List::flatten_v0`

`List::repeat_v0` raised an interesting choice: since Dark ints are now infinite precision, should we allow functions like this to create truly massive lists or should we just expose some sort of limit (the most obvious being F#'s own limit of Int32). 
I chose the latter option because materializing extremely large lists is a very good way to undermine the stability of a shared runtime. 
I noticed other list functions have similar problems (eg. `List::take`, which I ran out of time to finish, has no efficient way to take more than `Int32.MaxValue` items), which suggests to me that a lazy list implementation might be a better pairing with infinite precision integers.

Added test cases and iterated with `scripts/builder --compile --watch --test`

I can't figure out how to get VSCode to format this for me, so I expect that will probably fail in CI

### DIFF
--- a/fsharp-backend/src/LibExecution/StdLib/LibList.fs
+++ b/fsharp-backend/src/LibExecution/StdLib/LibList.fs
@@ -325,12 +325,12 @@ let fns : List<BuiltInFn> =
       fn =
         (function
         | _, [ DList l ] ->
-          let f a b =
-            match (a, b) with
-            | DList a, DList b -> DList (List.append a b)
+          let f acc i =
+            match i with
+            | DList l -> List.append acc l
             | _ -> Errors.throw "Flattening non-lists"
           in
-          Value(List.fold f (DList []) l)
+          List.fold f [] l |> DList |> Value
         | _ -> incorrectArgs ())
       sqlSpec = NotYetImplementedTODO
       previewable = Pure
--- a/fsharp-backend/src/LibExecution/StdLib/LibList.fs
+++ b/fsharp-backend/src/LibExecution/StdLib/LibList.fs
@@ -318,27 +318,23 @@ let fns : List<BuiltInFn> =
       sqlSpec = NotYetImplementedTODO
       previewable = Pure
       deprecated = NotDeprecated }
-    //   ; { name = fn "List" "flatten" 0
-//     ; parameters = [Param.make "list" TList ""]
-//     ; returnType = TList
-//     ; description =
-//         "Returns a single list containing the values of every list directly in `list` (does not recursively flatten nested lists)."
-//     ; fn =
-//           (function
-//           | _, [DList l] ->
-//               let f a b =
-//                 match (a, b) with
-//                 | DList a, DList b ->
-//                     DList (List.append a b)
-//                 | _ ->
-//                     RT.error (DList [a; b]) "Flattening non-lists"
-//               in
-//               List.fold (DList []) ~f l
-//           | _ ->
-//               incorrectArgs ())
-//     ; sqlSpec = NotYetImplementedTODO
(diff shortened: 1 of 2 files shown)


fsharp-backend/tests/testfiles/list.tests
```diff
@@ -33,11 +40,11 @@ List.filter_v0 [-20; 5; 9;] (fun x -> x < 1) = [-20]
...
-//List.flatten_v0 [[1];[2];[3]] = [ 1, 2, 3 ]
-//List.flatten_v0 [[1];[[2;3]]] = [ 1, [ 2, 3 ] ]
-//List.flatten_v0 [[[]]] = [ [] ]
-//List.flatten_v0 [[]] = []
-//List.flatten_v0 [] = []
+List.flatten_v0 [[1];[2];[3]] = [ 1, 2, 3 ]
+List.flatten_v0 [[1];[[2;3]]] = [ 1, [ 2, 3 ] ]
+List.flatten_v0 [[[]]] = [ [] ]
+List.flatten_v0 [[]] = []
+List.flatten_v0 [] = []
```
pbiggar, line 47: Can you add a test for lists of non-lists as well?

pbiggar (changes requested):
Thanks!

Looks like there are some duplicate work with #2998, which is ready to go. Can you drop List.drop and List.repeat, as I think #2998 has clearer implementations?

Great tests though, please leave them in!

mjeffryes:
Sounds good I'll wait until #2998 ships and then update my branch

pbiggar:
@mjeffryes #2998 just shipped

mjeffryes:
Hmm... I tried to rebase and squash on top of the updated main branch (since most of the old commits were irrelevant), but now it appears github is thoroughly confused about the commit list.

mjeffryes:
Ok, just one commit now as intended. I had to comment out the bigint tests for drop and repeat. I could not figure out how to mimic the `System.OverflowException` that they throw now.

fsharp-backend/src/LibExecution/StdLib/LibList.fs
```diff
@@ -318,27 +318,23 @@ let fns : List<BuiltInFn> =
...
+      description = "Returns a single list containing the values of every list directly in `list` (does not recursively flatten nested lists)."
+      fn =
+        (function
+        | _, [ DList l ] ->
+          let f a b =
+            match (a, b) with
+            | DList a, DList b -> DList (List.append a b)
+            | _ -> Errors.throw "Flattening non-lists"
+          in
+          Value(List.fold f (DList []) l)
```
pbiggar, line 333:
```suggestion
          let f list dv =
            match dv with
            | DList l -> List.append list l
            | _ -> Errors.throw "Flattening non-lists"
          in
          Value(DList(List.fold f (DList []) l))
```

pbiggar: Was there a reason to put a DList in the accumulator/state, vs just a list. Comment above has a fix to avoid creating a DList each time; not tested though.

pbiggar (changes requested):
Thanks! One question on the need to allocate a DList each time.

fsharp-backend/src/LibExecution/StdLib/LibList.fs
mjeffryes: No, just an oversight. Fixed in the latest version

pbiggar:
Awesome, thank you!

PR #3000, opened 2021-04-03
Port AWS::urlencode to F#

jwalter:
Porting AWS::urlencode  to F#

Duplicate the behavior of the OCaml implementation, including some potential bugs.

Added aws.tests, including some tests showing that the potential bug in the OCaml implementation is also in the F# version

### DIFF
--- a/fsharp-backend/src/LibExecution/StdLib/LibNoModule.fs
+++ b/fsharp-backend/src/LibExecution/StdLib/LibNoModule.fs
@@ -146,7 +146,17 @@ let fns : List<BuiltInFn> =
       description = "Url encode a string per AWS' requirements"
       fn =
           (function
-          | _, [ DStr s ] -> 
+          | _, [ DStr s ] ->
+              (* Based on the original OCaml implementation which was slightly modified from
+               * https://github.com/mirage/ocaml-cohttp/pull/294/files (to use
+               * Buffer.add_string instead of add_bytes); see also
+               * https://github.com/mirage/ocaml-uri/issues/65. It's pretty much a straight
+               * up port from the Java example at
+               * https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html,
+               * which calls it UriEncode *)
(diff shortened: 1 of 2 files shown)


fsharp-backend/src/LibExecution/StdLib/LibNoModule.fs
```diff
@@ -140,22 +140,37 @@ let fns : List<BuiltInFn> =
...
-// ; previewable = Pure
-//   ; deprecated = NotDeprecated }
+    { name = fn "AWS" "urlencode" 0
+      parameters = [Param.make "str" TStr ""]
+      returnType = TStr
+      description = "Url encode a string per AWS' requirements"
+      fn =
+          (function
+          | _, [ DStr s ] -> 
+              let n = String.length s
```
pbiggar, line 150: Can you adapt whatever comments are relevant from https://github.com/darklang/dark/blob/main/backend/libexecution/stdlib_util.ml#L5 and add them here?

pbiggar (changes requested):
Awesome, looks great, thank you! Just one comment and we should be good to merge.

pbiggar:
Awesome, thank you!

pbiggar:
I've been fuzzing and have found some tests that aren't passing in the new version:

```
AWS.urlencode_v0 "گچپژ" = "%DA%AF%DA%86%D9%BE%DA%98"
AWS.urlencode_v0 "<foo val=”bar“ />" = "%3Cfoo%20val%3D%E2%80%9Dbar%E2%80%9C%20/%3E"
AWS.urlencode_v0 "表ポあA鷗ŒéＢ逍Üßªąñ丂㐀𠀀" = "%E8%A1%A8%E3%83%9D%E3%81%82A%E9%B7%97%C5%92%C3%A9%EF%BC%A2%E9%80%8D%C3%9C%C3%9F%C2%AA%C4%85%C3%B1%E4%B8%82%E3%90%80%F0%A0%80%80"
```

I haven't had time to look into them yet, so they might be the result of the test harness not encoding these properly, but it probably means the fn has a few small bugs.

PR #3001, opened 2021-04-05
Port Char to F#

jwalter:
Porting char to F#

char.tests was added to ensure compatibility with OCaml implementation


fsharp-backend/tests/testfiles/char.tests
```diff
@@ -0,0 +1,11 @@
...
+Char.toASCIICode "a" = Test.typeError_v0 "Char::toASCIICode was removed from Dark" // FSHARPONLY
+
+Char.toASCIIChar 1 = Test.typeError_v0 "This function no longer exists." // OCAMLONLY
+Char.toASCIIChar 1 = Test.typeError_v0 "Char::toASCIIChar was removed from Dark" // FSHARPONLY
+
+Char.toLowercase "A" = Test.typeError_v0 "This function no longer exists." // OCAMLONLY
+Char.toLowercase "A" = Test.typeError_v0 "Char::toLowercase was removed from Dark" // FSHARPONLY
+
+Char.toUppercase "A" = Test.typeError_v0 "This function no longer exists." // OCAMLONLY
+Char.toUppercase "A" = Test.typeError_v0 "Char::toUppercase was removed from Dark" // FSHARPONLY
```
pbiggar, line 11: These tests should have `_v0` in their names.

fsharp-backend/src/LibExecution/StdLib/LibChar.fs
```diff
@@ -19,41 +19,37 @@ let incorrectArgs = LibExecution.Errors.incorrectArgs
...
-  //   ; deprecated = ReplacedBy(fn "" "" 0) } ]
+let fns : List<BuiltInFn> = 
+  [ { name = fn "Char" "toASCIICode" 0
+      parameters = [Param.make "c" TChar ""]
+      returnType = TInt
+      description = "Return `c`'s ASCII code"
+      fn = Errors.removedFunction
+      sqlSpec = NotQueryable
+      previewable = Pure
+      deprecated = DeprecatedBecause("Dark does not support Characters yet") }
```
pbiggar, line 30: Good guess, but not exactly right. Dark does actually support characters (just not character literals, you have to use `String::foreach` to get them I think), but it's that these functions supported an old ascii-based char, and these days chars are EGC-based. So deprecated because they "used an old Character type that no longer exists"

pbiggar (changes requested):
Awesome, thank you! Some small comments inline.

pbiggar:
Great, thank you!

PR #3002, opened 2021-04-05
Port the rest of Crypto

jwalter:
Porting remaining parts of Crypto to F#

Enabled the commented out tests and added a few more


pbiggar (approved):
Fantastic, thank you!

PR #3005, opened 2021-04-08
Port some more Date to F#

jwalter:
Date has not been fully ported to F#

Ported the remaining functions except toHumanReadable which I frankly didn't understand how it was supposed to work based on the tests I tried


pbiggar (commented):
Looks great, thanks!

I'm curious about the ToUniversalTime method call. Was that "just in case", or was something going wrong without it?

jwalter:
I did add ToUniversalTime "just in case". As long as Dark uses UTC for all dates I don't see a reason to keep the ToUniversalTime call. I'll remove it.

pbiggar:
Fantastic, thanks!

PR #3012, opened 2021-04-16
Http functions

mariajdab:
Add functions to `LibHttp.fs` and test to `http.test`
Add functions to `LibHtttp.fs` and test to `http.test`
Test cases passed
Yes

### DIFF
--- a/fsharp-backend/src/LibExecution/StdLib/LibHttp.fs
+++ b/fsharp-backend/src/LibExecution/StdLib/LibHttp.fs
@@ -184,24 +184,22 @@ let fns : List<BuiltInFn> =
       sqlSpec = NotYetImplementedTODO
       previewable = Pure
       deprecated = ReplacedBy(fn "Http" "responseWithJson" 0) }
-// ; { name = fn "Http" "responseWithJson" 0
-//   ; parameters = [Param.make "response" varA ""; Param.make "code" TInt ""]
-//   ; returnType = TResp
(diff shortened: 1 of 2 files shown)


fsharp-backend/src/LibExecution/StdLib/LibHttp.fs
```diff
@@ -48,21 +48,28 @@ let fns : List<BuiltInFn> =
...
+        "Returns a Response that can be returned from an HTTP handler to respond with HTTP status `code`, `response` body, and `headers`."
+      fn =
+        (function
+        | _, [dv; DObj o; DInt code] ->
+          let pairs =
+            Map.toList o
+              |> List.map
+                  (fun (k, v) ->
+                    match k, v with
+                    | k, DStr v  -> k, string v
```
pbiggar, line 64: I don't think you need this `string` call here.

fsharp-backend/src/LibExecution/StdLib/LibHttp.fs
```diff
@@ -70,21 +70,28 @@ let fns : List<BuiltInFn> =
...
```
pbiggar, line 86: don't need `string` here

pbiggar (changes requested):
Looks great! There's two tiny changes and then we can merge.

Thanks!

pbiggar:
Great, thanks!

PR #3044, opened 2021-05-06
Upgrade to dotnet 6

pbiggar:


baronfel:
hey @pbiggar I noticed that the support script to run `dotnet fsi` with your packages set up pins a load script from .net 5 [here](https://github.com/darklang/dark/blob/main/scripts/support/fsi-setup.fsx#L11). I think after this MR that that needs updating, is that correct?

pbiggar:
Thanks for the headsup @baronfel! I added a fix at https://github.com/darklang/dark/pull/3045

PR #3046, opened 2021-05-09
More LibList functions

mariajdab:
Add functions to `LibList.fs` and test to `list.test`
Add functions to `LibList.fs` and test to `list.test`
Test cases passed
Yes

### DIFF
--- a/fsharp-backend/src/LibExecution/StdLib/LibList.fs
+++ b/fsharp-backend/src/LibExecution/StdLib/LibList.fs
@@ -658,47 +658,54 @@ let fns : List<BuiltInFn> =
       sqlSpec = NotYetImplementedTODO
       previewable = Pure
       deprecated = ReplacedBy(fn "List" "filter" 2) }
-//   ; { name = fn "List" "filter" 2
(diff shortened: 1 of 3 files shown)


fsharp-backend/src/LibExecution/StdLib/LibList.fs
```diff
@@ -614,79 +633,101 @@ let fns : List<BuiltInFn> =
...
+                    | v ->
+                        Errors.throw (
+                          Errors.argumentWasnt
+                            "`true` or `false` for every value in `list`"
+                            "f"
+                            v
+                          + $"for the input {DvalRepr.toDeveloperReprV0 dv}"
+                        )
+
+                  return run && cont
```
pbiggar, line 717: Same comment as for v1:

I think you want to change this so something like:

```
if run then ... Interpreter.applyFnVal ... match result ...
else return false
```

mariajdab:
Thank you. I will add the revisions and comments.

pbiggar:
Nice job, thank you!

PR #3057, opened 2021-05-19
Build fixes

sxmanton:
1. I could not access the postgres db using pgAdmin (pgAdmin installed windows side with container running in wsl2)
2. I was encountering an error during run-fsharp-tests saying that one of the databases did not exist yet. Turns out the migrations were not completing from the ocaml server before the fsharp server started writing tests.

1. Manually edited postgreql.conf and saw that worked to allow pgAdmin to connect, so I uncommented the line which was already in the dockerfile, but adjusted the postgres version number to be correct
2. Fix provided by @pbiggar

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -215,8 +215,7 @@ RUN /etc/init.d/postgresql start && \
 # Adjust PostgreSQL configuration so that remote connections to the
 # database are possible.
 RUN echo "host all  all    0.0.0.0/0  md5" >> /etc/postgresql/9.6/main/pg_hba.conf
-
-# RUN echo "listen_addresses='*'" >> /etc/postgresql/10/main/postgresql.conf
+RUN echo "listen_addresses='*'" >> /etc/postgresql/9.6/main/postgresql.conf
 
 USER dark
 # Add VOLUMEs to allow backup of config, logs and databases
--- a/scripts/run-fsharp-tests
+++ b/scripts/run-fsharp-tests
@@ -33,6 +33,9 @@ createdb testdb
 ./scripts/support/runlegacyserver
 ./scripts/support/run-ocaml-test-server
 
+# CLEANUP This is where the migrations are run for testdb at the moment
+DARK_CONFIG_STATIC_HOST=static.darklang.localhost:8001 ./scripts/wait-until-server-ready
+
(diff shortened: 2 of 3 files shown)


sxmanton:
Hmm, seems it is using port 8000 instead of 8001 on circleci, not sure why

pbiggar:
Oh, interesting. Where did you see that?

I had thought the problem was that the files in `wait-until-server-ready` weren't there (I see it loaded appsupport.js but not app.js. appsupport.js is created in circleci.yml). My thought for a patch was

```
diff --git a/scripts/wait-until-server-ready b/scripts/wait-until-server-ready
index faf74e4e2..1c1a416aa 100755
--- a/scripts/wait-until-server-ready
+++ b/scripts/wait-until-server-ready
@@ -14,7 +14,4 @@ function wait_for {
 }

 wait_for analysis.js
-wait_for app.js
-wait_for appsupport.js
-wait_for app.css
```

sxmanton:
`curl: (7) Failed to connect to static.darklang.localhost port 8000: Connection refused` at line 3225

pbiggar:
> `curl: (7) Failed to connect to static.darklang.localhost port 8000: Connection refused` at line 3225

Yeah, that does seem rather weird. It certainly implies this passes after waiting for the port 8000 server, not the port 8001 server, which is not what we want.

pbiggar:
I can't imagine why the env var would be winning over the one we set when calling wait-until-server-ready. I'm going to try SSHing in to figure this out.

pbiggar:
OK, I think I figured out what's happening here.

In CircleCI, DARK_CONFIG variables are not set in the environment. Instead, they are set using `BASH_ENV`, which re-sets them every time a script is started in non-interactive mode.

This means that although we set `DARK_CONFIG_STATIC_HOST` when we call `wait-until-server-ready`, when `wait-until-server-ready` starts it reads from `BASH_ENV` automatically, which overwrites `DARK_CONFIG_STATIC_HOST` to the one set in `config/circleci` (that is to say, the one using port 8000).

Not sure what to do here, but I guess one thought is to not override env vars that are already set.

pbiggar:
Merged these changes in #3067 as I couldn't push here.

Thanks a million for the fix!

PR #3059, opened 2021-05-19
Add Dict and Object functions

mariajdab:
Add functions to `LibDict.fs` and `LibObject.fs` and add test cases
Add functions to `LibDict.fs` and `LibObject.fs` and add test cases
Add `Legacy.fs` to `LibExecution` in order to use `Legacy.PrettyResponseJsonV0.toPrettyResponseJsonV0` on `LibObject`
Test cases passed
Yes


fsharp-backend/src/LibExecution/Legacy.fs
```diff
@@ -0,0 +1,155 @@
...
+         | Error rdv ->
+             w.writeObject
+               (fun () ->
+                 w.WritePropertyName "type"
+                 w.WriteValue("result")
+                 w.WritePropertyName "constructor"
+                 w.WriteValue("Error")
+                 w.WritePropertyName "values"
+                 w.writeArray (fun () -> writeDval rdv)))
+    | DBytes bytes -> wrapStringValue "bytes" (System.Convert.ToBase64String bytes)
```
pbiggar, line 152: In legacy, this was not base64 encoded, AFAICT. So I think this should be:

`bytes |> System.Text.Encoding.ASCII.GetString |> wrapStringValue "bytes"`

pbiggar (approved):
This is excellent, nice job! Should be good to go with some small changes.

mariajdab:
Thank you! I will add the revisions.

PR #3061, opened 2021-05-21
Daniela/lib result changes

danicampagna:
Add functions to LibResult.fs and add test cases

Test cases passed


pbiggar (changes requested):
This looks great, thanks! Two small changes and we're good to go.

pbiggar:
Took care of the review things. Thanks again!

PR #3072, opened 2021-06-01
Add Json functions

mariajdab:
Add functions to `LibJson.fs` and the test cases
Add functions to `LibJson.fs` and the test cases
Test cases passed
Yes


backend/libexecution/libjson.ml
```diff
@@ -35,7 +35,7 @@ let fns : fn list =
               fail args)
     ; preview_safety = Safe
     ; deprecated = true }
-  ; { prefix_names = ["JSON::parse_v0"]
+  ; { prefix_names = ["JSON::parse"]
```
pbiggar, line 38: I think it would be better to not do this rename, as it might change something in the code. Can this be avoided?

fsharp-backend/src/LibExecution/StdLib/LibJson.fs
```diff
@@ -9,81 +9,71 @@ open LibExecution.RuntimeTypes
...
+      fn =
+        (function
+        | _, [ DStr json ] -> json |> DvalRepr.ofUnknownJsonV1 |> Value
+        | _ -> incorrectArgs ())
+      sqlSpec = NotYetImplementedTODO
+      previewable = Pure
+      deprecated = ReplacedBy(fn "JSON" "parse" 1) }
+    { name = fn "JSON" "parse" 1
+      parameters = [ Param.make "json" TStr "" ]
+      returnType = TResult(varA, varErr)
```
pbiggar, line 66: varErr here should be `TStr` (as the error always contains a string)

pbiggar (commented):
Nice job, thanks!

backend/libexecution/libjson.ml
mariajdab: I think it's necessary because the test cases for this Ocaml function will cause problems. Something that I realized working with it and after some debug if I leave `JSON::parse_v0` the test for Ocaml doesn't recognize the name of the function. Also, I took a look at other functions of the Ocaml version and the first version of a function (or version 0) doesn't have `_v0` just the name function. I think maybe it was a typo?

fsharp-backend/src/LibExecution/StdLib/LibJson.fs
mariajdab: Thank you! I will add the revision.

backend/libexecution/libjson.ml
pbiggar: Yeah, that's nasty. I had to deal with this before and I think the same solution will work for you.

I think you should be able to leave the code in backend as it was (with the `_v0` suffix), and just add an if statement like this one:

https://github.com/darklang/dark/blob/main/fsharp-backend/src/ApiServer/Functions.fs#L87

pbiggar:
Thanks!

Will follow up with some fixes in another PR.

PR #3074, opened 2021-06-04
Add HttpClient functions

mariajdab:
Add functions to `LibHttpClient.fs` and the test cases
Add functions to `LibHttpClient.fs` and the test cases
Test cases passed
Yes

### DIFF
--- a/fsharp-backend/src/LibExecution/StdLib/LibHttpClient.fs
+++ b/fsharp-backend/src/LibExecution/StdLib/LibHttpClient.fs
@@ -100,24 +100,17 @@ let fns : List<BuiltInFn> =
         | _ -> incorrectArgs ())
       sqlSpec = NotYetImplementedTODO
       previewable = Pure
-      deprecated = ReplacedBy(fn "HttpClient" "bearerToken" 1) } ]
-// ; { name = fn "HttpClient" "bearerToken" 1
-//   ; parameters = [Param.make "token" TStr ""]
-//   ; returnType = TObj
-//   ; description =
-//       "Returns an object with 'Authorization' set to the passed token"
-//   ; fn =
-//         (function
-//         | _, [DStr token] ->
-//             let auth_string =
-//               Unicode_string.append
-//                 (Unicode_string.of_string_exn "Bearer ")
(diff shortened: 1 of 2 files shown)


fsharp-backend/src/LibExecution/StdLib/LibHttpClient.fs
```diff
@@ -19,109 +19,98 @@ let incorrectArgs = LibExecution.Errors.incorrectArgs
...
+      returnType = TDict varA
+      description = "Returns an object with 'Authorization' set to the passed token"
+      fn =
+        (function
+        | _, [ DStr token ] ->
+            let authString =
+              System.Text.Encoding.UTF8.GetString(
+                Array.append
+                  (System.Text.Encoding.UTF8.GetBytes "Bearer ")
+                  (System.Text.Encoding.UTF8.GetBytes token)
```
pbiggar, line 95: I think you should use string append here instead of converting to bytes.

pbiggar (changes requested):
Looks great. I change noted in text. Can you also change all the signatures to `TDict TStr` as they all return dicts of strings.

fsharp-backend/src/LibExecution/StdLib/LibHttpClient.fs
mariajdab: Thank you! I will add the changes

fsharp-backend/src/LibExecution/StdLib/LibHttpClient.fs
```diff
@@ -83,17 +83,12 @@ let fns : List<BuiltInFn> =
...
       fn =
         (function
         | _, [ DStr token ] ->
-            let authString =
-              System.Text.Encoding.UTF8.GetString(
-                Array.append
-                  (System.Text.Encoding.UTF8.GetBytes "Bearer ")
-                  (System.Text.Encoding.UTF8.GetBytes token)
-              )
+            let authString = [ "Bearer "; token ] |> String.concat ""
```
pbiggar, line 91: You can use `"Bearer " + token`

PR #3091, opened 2021-06-16
Fix container on Windows

danicampagna:
Fix a problem which happened during VS initialize, if code runs under WSL it needs to change folder /home.

### DIFF
--- a/scripts/support/initialize-vscode
+++ b/scripts/support/initialize-vscode
@@ -1,6 +1,4 @@
 #!/usr/bin/env bash
-set -x
-echo $@
 # this is run on the host when starting the vscode devcontainer. It makes sure
 # the host is in the right shape so that the devcontainer can start up. This is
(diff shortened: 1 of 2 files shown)


scripts/support/initialize-vscode
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
-
+set -x
```
pbiggar, line 2:
```suggestion
```

scripts/support/initialize-vscode
```diff
@@ -1,5 +1,6 @@
...
+echo $@
```
pbiggar, line 3:
```suggestion
```

pbiggar (changes requested):
Thanks!

The CI failure is just from the debugging code that was left in - remove that and we should be fine.

PR #3096, opened 2021-06-23
Origin/daniela/liblist migration

danicampagna:
Continue with the migration of the functions written in OCaml to F#, in this case those of the `LibList.fs` class


fsharp-backend/src/LibExecution/StdLib/LibList.fs
```diff
@@ -684,239 +701,258 @@ let fns : List<BuiltInFn> =
...
+            | Some v -> return v
+          }
+        | _ -> incorrectArgs ())
+      sqlSpec = NotYetImplementedTODO
+      previewable = Pure
+      deprecated = NotDeprecated }
+    { name = fn "List" "filterMap" 0
+      parameters =
+        [ Param.make "list" (TList varA) ""
+          Param.makeWithArgs "f" (TFn([ varA ], TBool)) "" [ "val" ] ]
```
pbiggar, line 757: The `TBool` here should be a `TOption varB`

fsharp-backend/src/LibExecution/StdLib/LibList.fs
```diff
@@ -684,239 +701,258 @@ let fns : List<BuiltInFn> =
...
+      returnType = TList varA
```
pbiggar, line 758: This should be a `TList varB`

pbiggar (changes requested):
Nice job!

Apart from the small signature change, can you reformat using the `script/format` script and rebase the changes so that we don't have any formatting changes anymore? Thanks!

danicampagna:
I used the `script/format` script and cherry-picked the commits in PR [#3100](https://github.com/darklang/dark/pull/3100)

PR #3100, opened 2021-06-28
Liblist migrations

danicampagna:
Continue with the migration of the functions written in OCaml to F#, in this case those of the LibList.fs class. In my last PR [#3096,](https://github.com/darklang/dark/pull/3096) I had some formatting issues that I corrected in this single PR. 


pbiggar:
I think the formatting issues are still there - looking in the diffs, the indentations are changed.

PR #3103, opened 2021-06-30
LibDict functions migration to F#

danicampagna:
Continue with the migration of the functions written in OCaml to F#, in this case those of the LibDict.fs class. I added some functions to Prelude too to help


pbiggar:
Look great, nice job!

PR #3104, opened 2021-07-01
LibDB2 functions migration

danicampagna:
Continue with the migration of the functions written in OCaml to F#, in this case were missing only DB2::delete_v1 and DB2::deleteAll_v1 from LibDB2.


pbiggar:
Awesome, thanks!

PR #3108, opened 2021-07-05
LibList functions migration

danicampagna:

Rewrite LibList.fs functions to F#. I also added a new function to Prelude (`List.filter_map`), to help with the code.

Tests passed and I added some more tests to check


pbiggar:
Great, thanks!

PR #3110, opened 2021-07-05
LibNoModule functions migration

danicampagna:

I added some functions left in `LibNoModule.fs` and a file with tests of all functions there

All tests I added passed


pbiggar:
Looks great, thanks!

PR #3111, opened 2021-07-05
LibDate function migration

danicampagna:

There were only one function left to migrate in `LibDate.fs`, `Date::toHumanReadable`

It is a function that has been deprecated because it doesn't work but I added some tests just in case we ever use it again.


pbiggar:
Nice, thanks!

PR #3112, opened 2021-07-06
Prelude functions rewrited

danicampagna:
While working on the functions migration in `LibDict`, I needed to write a `fold_s` function to help with the code. So now I rewrite functions in Prelude to work with this new one.

Wrote `map_s`, `filter_s` and `filter_map` from Map module in `Prelude` using the new function `fold_s` to reuse code

I ran the tests again and they passed


pbiggar:
Nice!

PR #3167, opened 2021-08-18
Update to latest versions of things (.NET 6 RC1), and split into more projects

pbiggar:
Update most things to latest:
- paket
- honeytail
- .NET (to RC1, as there were bugs in preview 7)
- most paket deps

Interesting changes:
- Npgsql.FSharp to 4.0.0 (removes Tasks namespace)
- lots of small changes in FSharp.compiler.services
- start using paket groups to avoid conflicts
- fix bug using opentelemetry
- reimplement follow manually for HttpClient (now works exactly the same as in OCaml)
- use .NET's new System.Random.Shared (which is thread-safe, the value we were using wasn't)
- start using honeycomb library instead of doing it manually, plus refactor
- move packagereferences out of Wasm.fsproj


pbiggar:
OK, this grew quite a bit while waiting to make CI pass. This now also:

- Adds a type parameter to `Prelude.ignore`
- Starts configuring binaries for deployment, including supporting using the new binaries
- Split things into their own projects:
  - LibExecutionStdLib
  - BackendOnlyStdLib
  - Tablecloth
  - LibRealExecution
  - QueueWorker
- No longer use Tablecloth in LibExecution or LibExecutionStdLib
- Disable migrations in ApiServer (again)

fsharp-backend/src/QueueWorker/README.md
```diff
@@ -0,0 +1,5 @@
+# LibBackend
+
+The backend application around Dark. Everything from user management, web
+servers, logging, DBs, etc. Also includes functions which cannot be implemented
+in the client.
```
prekel, line 5: Duplicate

pbiggar: Awesome, thanks!

PR #3172, opened 2021-08-22
Fix converting newlines to spaces when add secrets

### LINKED ISSUE #2861: JWT signing doesn't work with secrets
On secrets, I’m storing secrets for jwt signing / verifying. I noticed that if I use the value from “secrets”. It wouldn’t work:

Screenshot 2020-11-07 at 12 47 38 PM

But if I type the value literally, it would work:

Screenshot 2020-11-07 at 12 50 02 PM

You can reproduce it by generating your ow private key and passing it to the `JWT::signAndEncode` function

prekel:

Fix #2861

Switch from an `input` to a `textarea`, and change some corresponding styles

I have changed the horizontal layout direction to vertical, because textarea is resizeable.

image
image
image
image

🤷‍♂️

No.

### DIFF
--- a/client/src/components/InsertSecret.res
+++ b/client/src/components/InsertSecret.res
@@ -157,7 +157,7 @@ let view = (m: ST.insertModal): Html.html<Types.msg> =>
               },
               list{},
             ),
-            Html.input'(
+            Html.textarea(
               list{
                 Attr.placeholder("secret value"),
                 Attr.name("secret-value"),
--- a/client/src/components/InsertSecret.res
+++ b/client/src/components/InsertSecret.res
@@ -162,7 +162,7 @@ let view = (m: ST.insertModal): Html.html<Types.msg> =>
                 Attr.placeholder("secret value"),
                 Attr.name("secret-value"),
                 Attr.value(m.newSecretValue),
-                Html.classList(list{("modal-form-input", true), ("error", !m.isValueValid)}),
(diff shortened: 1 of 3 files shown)


pbiggar:
Thanks, looks great! I'll take a proper look tomorrow.

pbiggar (changes requested):
This is great.

One issue I found when testing it is that sometimes when you open the secret box now, the textarea will have the value of an old secret in it. This seems bad. here's an example:

image

In this example, I had entered that value for a previous secret, and then opened the secret box again, where I discovered it was still showing that value.

Any idea what causes this? I couldn't reproduce on darklang.com, which makes me think it's caused by the textarea change.

prekel:
> 
> 
> This is great.
> 
> One issue I found when testing it is that sometimes when you open the secret box now, the textarea will have the value of an old secret in it. This seems bad. here's an example:
> 
> ...

Sorry, I forgot that `textarea` does not support `value` attribute. `m.newSecretValue` must be provided through children. I'll do it later

prekel:
Cannot reproduce... (on both Microsoft Edge 92.0.902.78 and Google Chrome Canary 95.0.4620.0 on Windows 11).

But maybe https://github.com/darklang/dark/pull/3172/commits/1a3473ab1c1633385baf3f7bbb398e446a52093c should fix it.

PR #3192, opened 2021-09-11
Glue together more of the CI pipeline

pbiggar:
Fix deploys


pbiggar:
This only changes deploy, which is broken on main, so merging.

PR #3193, opened 2021-09-11
Add BwdServer to the deployment

pbiggar:
Adds the bwdserver to the deployment, making it end-to-end in the same places as the apiserver.

Also refactors nginx - this should keep working the same for ocaml but it's not fully wired up for the F# servers just yet.


pbiggar:
Redid this off main branch

PR #3199, opened 2021-09-14
Fix legacy server in apiserver

pbiggar:


pbiggar:
Should be a safe merge so merging

PR #3248, opened 2021-11-23
Http Stack

### LINKED ISSUE #3125: Finish HTTP pipeline / middleware
- [x] add support for CORS settings
- [x] test what happens for different transfer-encodings in HTTP requests (esp chunked)
- [x] check that there is no error if you set the Content-Length header
- [x] Check vars in bwdpipeline are not in scope
- [x] Fully test CORS
- [x] Fully test requests to BwdServer with toPretty, CORS, and fields of the request obj
- [x] test form upload
- [x] use nginx during tests
- [x] add headers so that tests mimic production
- [x] split out middleware into its own module
- [x] ensure all underscore tests are fixed
- [x] more tests
  - [x] requests setting no content-length
  - [x] different accept-encodings
  - [x] different cache-control
  - [x] ~different accept-language~
  - [x] different accept values
  - [x] sitemap.xml
  - [x] favicon.ico
  - [x] ~builtwithdark.lvh.me~
  - [x] user not found
  - [x] canvas not found
  - [x] x forward for ip address
  - [x] user agent header
  - [x] other x forwarded for
  - [x] unicode (and bad unicode)
- [ ] find test suite of http edge cases
- [x] port non-underscore tests to underscore
- [x] add tests for path segments
  - [x] superfluous slashes
- [x] responses:
  - [x] incoplete
  - [x] derror
  - [x] Result.Ok
  - [x] Result.Error
  - [x] Option.Nothing
  - [x] Option.Just
  - [x] Redirect
  - [x] as above with error rail
- [x] ~track in Heap.io~

pbiggar:
This should finish the Http framework/stack/middleware, except for some small TODOs. Fixes #3125.

- Add lots of tests that cover different parts of HTTP, especially different headers
- Separates the dark http stack from the ASP.NET stuff, as much as possible. In the future I want to be able to change the middleware, and the first step is to ensure there's a line where things can be split. It's not a super great split, but it's a start.
- Splits HttpHeaders up again - using it for both HttpClient and the Http server seemed smart, but it made it challenging to have separate logic for both.

Also, some small new features:
- Allow users to set the "server" header
- Change the default server header to "darklang" (lowercase)
- Make an error match
- Remove bodies for GETs
- separate the nginx logs so we can see them all
- fix parameters to Tablecloth Array.slice
- change tests to use binary strings (allows testing images and invalid unicode)
- port remaining bwd unit tests


fsharp-backend/src/BackendOnlyStdLib/LibHttpClient5.fs
```diff
@@ -39,23 +36,45 @@ let parametersNoBody =
...
+// Middleware, as we want to be able to change them separately.
+
+let hasFormHeader (headers : HttpHeaders.T) : bool =
+  headers
+  |> HttpHeaders.get "content-type"
+  |> Option.map Tablecloth.String.toLowercase = Some
+                                                  "application/x-www-form-urlencoded"
+
+let hasJsonHeader (headers : HttpHeaders.T) : bool =
+  // CLEANUP: don't use contains for this
```
StachuDotNet, line 49: I'm curious what's wrong with Contains here?

pbiggar: It's just that it's not super strict: it could be `my-application/jsonry`

Issue #3379, opened 2022-01-16
Switch to new event table


pbiggar:
There are way too many function_results in the DB, costing us quite a bit of money and making the DB very big. The DB is 97% function_results, mostly from one customer. That cost $3392 in December.

The old traces aren't being cleaned up because the garbage collector process hangs on it. We've stopped collecting traces for this customer, but the old ones are all there.

Even if we deleted the table, Cloud SQL will keep charging us for the many TB of space we use. So we need to delete the data, then switch to a new DB server.

There's also 100GB of stored_events (trace inputs), and 65GB in function_arguments. However, it's not worth blowing this up to save an extra 1.5% of the current total, even though that will be 50% of the new total.

The rough order of operations:

- [x] make a new function_results_v3 table
- [x] read from function_results_v3, and when there aren't enough values, from function_results_v2
- [x] don't run migrations on startup
- [x] deploy
- [x] run migration with exechost
- [x] read from v3 and v2
- [x] deploy
- [x] start writing to only function_results_v3
- [x] deploy
- [x] copy from function_results_v2 for the last 10 traces from stored_events. Copy to function_results v3
   - [x] add darkinternal functions to copy results
   - [x] prototype this locally
   - [x] build this on darklang.com/a/ops
     - [x] use logging!
   - [x] do the copy
- [x] alert users to upcoming downtime in #status
- [x] change garbage collector to run in function_results_v3
- [x] stop reading from function_results_v2
- [x] check that user's canvases still work and that traces are loading
  - [x] functions
  - [x] handlers
- [x] drop/truncate function_results_v2
  - [x] drop or truncate or something else?
  - [x] disable transaction on the migration?
  - [x] validate that there are no queries on it
  - [x] check there are queries locking the table
  - [x] check there foreign keys on it
- [ ] clone DB and do a trial run
  - [x] drop/truncate function_results_v2 in clone
  - [x] validate space savings in clone
  - [ ] go through the google migration service process on the clone
  - [ ] connect something to this and see how it goes
  - [ ] anything to change after we migrate? Size, settings, version?
- [x] test out postgres 14 on the codebase
- [x] investigate if we want to use Google Migration Service, or merely add a new secondary
- [ ] rollover
- [x] switch F# stuff to function_results_v3
- [x] remove all references to function_results_v2

pbiggar:
Notes on using the Google Database Migration Service on the clone (from https://cloud.google.com/database-migration/docs/postgres/setup):

- enabled the API

[...]

PR #3386, opened 2022-01-19
Try out postgres 14

pbiggar:
I'm considering upgrading to postgres 14 as we have to take downtime now anyway for #3379.

Tests all pass.

For local dev, there's a guide to upgrade: https://gorails.com/guides/upgrading-postgresql-version-on-ubuntu-server

Summary:
- `rm -Rf /etc/postgres/14`
- `sudo apt update && sudo apt install postgres-9.6`
- `sudo pg_upgradecluster 9.6 main`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ executors:
       IN_DEV_CONTAINER: true
     docker:
       # DOCKERFILE_REPO: see Dockerfile note about how this is built.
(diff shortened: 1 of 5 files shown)


pbiggar:
The upgrade was too much work on the backend side, so not doing this for now. good to know it works though.

pbiggar:
We've moved to postgres 14 already.

PR #3431, opened 2022-02-01
Redirect some traffic to fsharp

pbiggar:
This adds a path from the main darklang ingress (the load balancer for darklang.com/*) to the apiserver. I've made it go there just for the trydarkfsharp canvas, which will be enough to test it out.

This will be deployed manually so I can quickly revert if it's going badly.


pbiggar:
I tested all this manually and discovered that nginx wasn't configured correctly. This fixes the nginx config for apiserver, and also for bwdserver (which I didn't test, just applied the same changes).

Alas, this doesn't enable access to the new apiserver, as I couldn't get the ingress (the main load balancer for darklang.com) to point to the new service, as it's in a different kubernetes namespace. I tried to use a workaround I found on stackoverflow, but it doesn't work on GKE. I have a request in with google support and I'll see what they say, but we don't currently have a solution for now.

Note that one solution is to put everything in the same namespace. This is strictly worse from a security perspective, especially since the darklang namespace has a default-deny policy that would be error-prone to add to the default namespace.

PR #3432, opened 2022-02-01
Add more volumes to clear-builder-volumes script

StachuDotNet:
per 1:1

Does this look OK as-is @pbiggar?

### DIFF
--- a/scripts/build/clear-builder-volumes
+++ b/scripts/build/clear-builder-volumes
@@ -2,6 +2,10 @@
 
 set -euo pipefail
 
+# Clear all the volumes that we use
+
+# Must be run from outside (stopped) container
+
 docker volume rm \
   dark_build \
   dark_dotesy \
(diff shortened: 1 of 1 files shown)


scripts/build/clear-builder-volumes
```diff
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+# clear all the volumes that we use
```
pbiggar, line 5: The formatting I've been using for these is:

- Capitalize 1-line description
- newline
- other notes

So in this case it would be

```
set -euo pipefail

Clear all the volumes that we use

Must be run from outside (stopped) container

docker volume ...
```

We should probably find somewhere to put coding guidelines so that we stay consistent when the automated formatters don't help.

pbiggar (commented):
Quick formatting request, and then good to go

pbiggar:
Thanks! That reminds me I need to fix CI for 3rd parties.

PR #3466, opened 2022-02-09
Fix Date.parse

pbiggar:


StachuDotNet (approved):
Functionally this looks good to me, and it seems pretty well-tested.

I noticed a little pyramid of List.maps - I'd like to offer an alternative:

a common helper fn for cartesian products (I stole this from SO. I usually do in any projects https://stackoverflow.com/a/50306962)
```
let product lists = 
    let folder list state =
         state |> Seq.allPairs list |> Seq.map List.Cons 
    Seq.singleton List.empty |> List.foldBack folder lists
```

usage
```
  [dateFieldSeparators; dateTimeSeparator; timeSeparator; millisecondFormat; seconds; meridians; tzSuffixes]
  |> product
  |> Seq.map(function
    | [dfs; dts; ts; ss; tt; msf; tzs] ->
      // Dont allow separator and then no seconds
      let ss = if ss = "" then "" else ts + ss
      
      // AM/PM requires 12 hour format
      let hh = if tt = "" then "HH" else "hh"

      // Don't allow milliseconds if no seconds
      let msf = if ss = "" then "" else msf
      
      $"yyyy{dfs}MM{dfs}dd{dts}{hh}{ts}mm{ss}{msf}{tt}{tzs}"
    | _ -> failwith "Unexpected"
  )
 ```

You should be able to get rid of those List.concats, then, too

pbiggar:
> I noticed a little pyramid of List.maps - I'd like to offer an alternative:

Oh, very cool!

> https://gist.github.com/StachuDotNet/9da00ecdb12b09f78a1bb2188f6a924f tested

Question: how do you run files like this? Can I just do `dotnet fsi filename.fs` or something?

StachuDotNet:
Usually I would copy the code into https://fable.io/repl and run it there. That said, it's been down for a day or so.
Otherwise, I typically have a scratch console app open in Rider, and copy/paste stuff there.

If you save the file as an .fsx, you should be able to run in FSI with
`dotnet fsi`
`#load file.fsx`

PR #3469, opened 2022-02-10
For integration tests, fixes BWD URL

StachuDotNet:
When running integration tests locally, the BWD urls of test canvases was broken as-is, yielding several failing tests locally.

Prior to this commit, 
```
function bwdUrl(testInfo: TestInfo, path: string) {
  return (
    "http://test-" + testInfo.title + ".builtwithdark.lvh.me:8000" + path
  );
}
```

Adjusts the URL to
```
function bwdUrl(testInfo: TestInfo, path: string) {
  return (
   "http://test-" + testInfo.title + ".builtwithdark.localhost:8000" + path
  );
}
```
, which works for me

@pbiggar _should_ the lvh.me URL work? if so, any ideas what could be wrong?
Also, what does "lvh" stand for?

### DIFF
--- a/integration-tests/tests.ts
+++ b/integration-tests/tests.ts
@@ -233,7 +233,7 @@ test.describe.parallel("Integration Tests", async () => {
 
   function bwdUrl(testInfo: TestInfo, path: string) {
     return (
-      "http://test-" + testInfo.title + ".builtwithdark.lvh.me:8000" + path
+      "http://test-" + testInfo.title + ".builtwithdark.localhost:8000" + path
     );
   }
 


pbiggar:
> @pbiggar _should_ the lvh.me URL work? if so, any ideas what could be wrong?
> Also, what does "lvh" stand for?

I don't really remember doing this, but I'm guessing this is what I was going for: https://nickjanetakis.com/blog/ngrok-lvhme-nipio-a-trilogy-for-local-development-and-testing#lvh-me

StachuDotNet:
does the lvh form work for you locally? if so, is there anything I need to set up for it to work for me?

PR #3474, opened 2022-02-14
Correct mimetype of WASM files in ApiServer

StachuDotNet:
When loading Blazor assemblies (.dlls and .wasm files), console warnings indicate that they are being loaded with the incorrect MIME type. (application/octet-stream)

Ensures that `ApiServer` returns all statically-hosted `.dll` and `.wasm` files are returned with a MIME type of `application/wasm`

Use `FileExtensionContentTypeProvider` type exposed in `Microsoft.AspNetCore.StaticFiles` to configure the MIME types per file  extension

Tested locally; console warnings that existed previously no longer do.

### DIFF
--- a/fsharp-backend/src/ApiServer/ApiServer.fs
+++ b/fsharp-backend/src/ApiServer/ApiServer.fs
@@ -20,6 +20,7 @@ open Tablecloth
 
 open Http
 open Middleware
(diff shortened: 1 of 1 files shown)


pbiggar:
Thanks!

I don't think this needs CI, so just going to merge as is.

PR #3477, opened 2022-02-14
Add gatsby support for static asset upload

kate-grant:
Currently Dark builds static asset url from prefix (dark magic DARK_STATIC_ASSETS_BASE_URL) without schema. This supports React PUBLIC_URL, but
Gatsby assetPrefix needs schema (otherwise it will be formatted as a relative path), so Dark needs option to build static asset url from prefix option with schema.

Add a way to search and replace for dark magic with schema, so that dark magic with schema can be successfully replaced in Gatsby config assetPrefix the way dark magic without schema is currently replaced in React PUBLIC_URL.

Added logic to webserver.ml block where DARK_STATIC_ASSETS_BASE_URL is replaced with sa.url to first search and replace the replace dark magic with schema with the sa.url, before continuing with existing logic that searches and replaces the dark magic-without-schema with the sa.url.

This was developed in Dark docker env running dark on localhost, with the limitation of commenting out analysis.js in compile file. Dark on localhost functioned as expected with the limitation of error Uncaught Reference Error: darkAnalysis not defined

If this fails, the added logic to the webserver.ml static asset url for https://DARK_STATIC_ASSETS_BASE_URL can be removed

Information about Gatsby support has been added to the comment prefacing this block.


pbiggar:
Great, thanks!

Issue #3254, opened 2021-11-25
Switch to blazor in the client


pbiggar:
- [x] Get blazor working on darklang.com
- [x] switch integration tests over to using blazor
- [x] ensure when blazor fails that we get a rollbar
- [x] add changelog entry explaining how to change back
- [x] Go through user canvases and check Blazor works properly - @pbiggar only
- [ ] switch the client's default to blazor

StachuDotNet:
I've found the `use-blazor` query param, and have found that basic functionality seems OK with that flag on.

Could you please remind me the issues that I can address? @pbiggar 

Many in the TODO list here seem only actionable by you, besides the rollbar one which I'll think/look through. I recall that there were some perf. concerns in particular, but forget _where_.

pbiggar:
I've reordered them and marked the one that I think only I can do.

To switch integration tests, you would add the setting [here](https://github.com/darklang/dark/blob/51dab2e9a1612ebf382e69d56f1216fcddb018e7/integration-tests/tests.ts#L128).

(separately, I just realized integration tests are running on port 8000 by default, we should be safe to switch the default (see `--fsharp` in run.sh)).

The issue that I expect to come up is that blazor is so slow to initialize that the integration tests will be hard to make work. That slowness is a real issue, and potentially blocking. The issue to solve it is https://github.com/darklang/dark/issues/3168

The changelog is in the docs repo, maybe make a PR saying that it will happen and we can update when we switch the default?

Btw, feel free to make new issues for any of these if you prefer.

StachuDotNet:
Small check-in here; working on this in #3485. Basic transition is set up there, and remaining work at this point is:
- [ ] resolve a small group of integration tests
- [ ] punt some others with FSTODOs
- [ ] ensure when blazor fails that we get a rollbar
- [ ] add changelog entry

After that point, I'll have to hand off to Paul to test with various user canvases.

StachuDotNet:
Catching up with some Slack threads, the remaining things to do here are:
- cancel analysis executions when new requests come in, to decrease UX pains of non-AOT Blazor being slow.
  This is my current focus
- reproduce and resolve the loading issue that Paul faced. I haven't been able to repro - @pbiggar if you could share your experience in more detail, that'd be appreciated.
- resolve a (separate?) load-time issue that I faced[1]

---

[...]

PR #3485, opened 2022-02-15
Make blazor default for canvas; switch integration tests to F# backend

StachuDotNet:
#3254

- Integration tests should run against the F# server (on port 9000) by default
- The canvas should use the WASM-compiled (Blazor) F# backend for analysis, rather than the OCaml one.


pbiggar:
If you rebase/merge this on main, CI should now run, I believe.

StachuDotNet:
Some integration tests rely on the package manager (incl. uploading of packages).

Should this PR be blocked by that functionality, in which case I could pivot to such, or pull out into a future PR, with an FSTODO on the tests?

pbiggar:
I would say you can do an FSTODO on it for now and we can reassess later if it's truly necessary.

StachuDotNet:
The `test_admin/stdlib/Test::one` fn exposed [here](https://github.com/darklang/dark/blob/main/integration-tests/clear-db.sh#L43) does not pass any of the patterns that our F# backend accepts [here](https://github.com/darklang/dark/blob/main/fsharp-backend/src/LibExecution/ProgramTypes.fs#L44-L75).

image

```
Bad format in function name [2/22/2022 3:38:48 PM +00:00]
        exception: True
        exception.type: Prelude+InternalException
        exception.stacktrace: 
        exception.message: Bad format in function name
        fnName: test_admin/stdlib/Test::one
```

and so a test that uses said function (see [here](https://github.com/darklang/dark/blob/main/integration-tests/tests.ts#L1288) and [here](https://github.com/darklang/dark/blob/main/backend/test_appdata/test-package_function_references_work.json#L12)) fails when targeting the F# backend.

What should be done here @pbiggar?
- adjust the insertion somehow
- add another allowable pattern to `ProgramTypes.fs`
- adjust how the test calls/references the function
- something else

pbiggar:
> The `test_admin/stdlib/Test::one` fn inserted [here](https://github.com/darklang/dark/blob/main/integration-tests/clear-db.sh#L43) does not pass any of the patterns that our F# backend accepts [here](https://github.com/darklang/dark/blob/main/fsharp-backend/src/LibExecution/ProgramTypes.fs#L44-L75).
> 
> image
> 
> ```
> Bad format in function name [2/22/2022 3:38:48 PM +00:00]
>         exception: True
>         exception.type: Prelude+InternalException
>         exception.stacktrace: 
>         exception.message: Bad format in function name
>         fnName: test_admin/stdlib/Test::one
> ```
> 
> and so a test that uses said function (see [here](https://github.com/darklang/dark/blob/main/integration-tests/tests.ts#L1288) and [here](https://github.com/darklang/dark/blob/main/backend/test_appdata/test-package_function_references_work.json#L12)) fails when targeting the F# backend.
> 
> What should be done here @pbiggar?
> 
> * adjust the insertion somehow
> * add another allowable pattern to `ProgramTypes.fs`
> * adjust how the test calls/references the function
> * something else

I think rename it (so adjust the insertion) so that it's valid.

backend/test_appdata/test-fluid_test_copy_request_as_curl.json
```diff
@@ -3,13 +3,13 @@
   [
     "SetHandler",
     91390945,
-    { "x": 576, "y": 32 },
+    { "x": 76, "y": 32 },
```
StachuDotNet, line 6: with the original coordinates, the relevant UI was hidden under the "first time" prompts. Is this a reasonable fix?

backend/test_appdata/test-package_function_references_work.json
```diff
@@ -9,7 +9,7 @@
       "ast": [
         "EFnCall",
         503886666,
-        "test_admin/stdlib/Test::one_v0",
+        "test_admin/stdlib/Test::one_v1",
```
StachuDotNet, line 12: Maybe something to pull out into a separate issue:
without changing this (and other corresponding locations) to be a v1 rather than a v0, the _v0 portion of the function name was stripped at the call site. Seemingly this is an OCaml interop issue, but couldn't pin down the specific issue.

integration-tests/README.md
```diff
@@ -38,6 +48,12 @@ If you're trying to eliminate a flaky test that is hard to reproduce, use `--rep
 
 `./scripts/run-in-docker ./integration-tests/run.sh`
 
+### Run Against F#
+
+By default, the integration tests are run against the OCaml backend.
```
StachuDotNet, line 53:
```suggestion
### Run Against OCaml

By default, the integration tests are run against the F# backend.
```

integration-tests/README.md
```diff
@@ -38,6 +48,12 @@ If you're trying to eliminate a flaky test that is hard to reproduce, use `--rep
 
 `./scripts/run-in-docker ./integration-tests/run.sh`
 
+### Run Against OCaml
+
+By default, the integration tests are run against the F# backend.
+
+If you'd like to run against the F# backend, use `--ocaml`
```
StachuDotNet, line 55:
```suggestion
If you'd like to run against the OCaml backend, use `--ocaml`
```

integration-tests/README.md
```diff
@@ -133,3 +149,7 @@ button in the browser, which runs the testing function for the test
 - `rundir/integration-tests/`
 
   - videos, traces, and console logs from the test executions
+
+- `backend/test_appdata`
+
+  -
```
StachuDotNet, line 155:
```suggestion
```

integration-tests/run.sh
```diff
@@ -7,12 +7,15 @@ set -euo pipefail
 # This deliberately can be run outside the container - sometimes you want to test on
 # the host
 
+# todo: this script fails if you cd to this directory.
```
StachuDotNet, line 10:
```suggestion
# CLEANUP this script fails if you cd to this directory.
```

integration-tests/tests.ts
```diff
@@ -1277,26 +1281,27 @@ test("feature_flag_in_function", async ({ page }) => {
...
-    await expectContainsText(page, ".return-value", "0");
+    // await page.click(".execution-button");
+    // await expectContainsText(page, ".return-value", "0");
 
-    // check if we can get a result from the bwd endpoint
-    let response = await get(page, bwdUrl(testInfo, url));
-    expect(response).toBe("0");
+    // // check if we can get a result from the bwd endpoint
+    // let response = await get(page, bwdUrl(testInfo, url));
+    // expect(response).toBe("0");
```
StachuDotNet, line 1304: TODO for self: uncomment these lines.

backend/test_appdata/test-fluid_test_copy_request_as_curl.json
pbiggar: The first time prompts aren't supposed to be there. We write some values to local storage to ensure they aren't so it sounds like something is broken there.

backend/test_appdata/test-package_function_references_work.json
pbiggar: Yeah, that sounds right. This fix is fine - it's an awkward thing to fix in both versions

backend/test_appdata/test-fluid_test_copy_request_as_curl.json
StachuDotNet: My issue here must have been related to something else - test is passing now with original position.

StachuDotNet:
two steps resulted in progress on integration tests running in CI against F#:
- the `integration-tests` job was updated to run `run-fsharp-server --published` instead of `run-backend-server`
- `run-fsharp-server` itself was updated to reference the `/publish`ed binaries rather than the ones at `Release/net6.0/linux-x64/publish/`, when the `--publish` flag is present. This is to match what's persisted to workspace in the circle job
  image

Given those two changes, tests are _almost_ running in CI.
At this point, it's held up on `_wait-until-server-ready` waiting for appsupport.js to be available _on port 8000_.
image

My plan now is to update `_wait-until-server-read` so that it accepts a `--use-apiserver` flag, in which case `wait_for {}` can use `DARK_CONFIG_APISERVER_STATIC_HOST` rather than `DARK_CONFIG_STATIC_HOST` - I think that will do the trick.

integration-tests/tests.ts
```diff
@@ -750,7 +751,8 @@ test("feature_flag_in_function", async ({ page }) => {
     await awaitAnalysis(page, timestamp);
 
     await page.waitForSelector(".return-value");
-    const expectedText = "but + only works on Ints.";
+
+    const expectedText = "Try using Float::+, or use Float::truncate to truncate Floats to Ints.";
```
StachuDotNet, line 755: TODO: this will pass in F# but not in OCaml. fix that.

integration-tests/tests.ts
```diff
@@ -1064,6 +1066,8 @@ test("feature_flag_in_function", async ({ page }) => {
     await gotoHash(page, testInfo, "fn=602952746");
     const timestamp = Date.now();
 
+    await page.pause();
```
StachuDotNet, line 1069: TODO: I added it but in hindsight not sure it was needed;  re-evaluate.

integration-tests/tests.ts
```diff
@@ -1277,20 +1283,19 @@ test("feature_flag_in_function", async ({ page }) => {
...
+    // this await confirms that test_admin/stdlib/Test::one_v1 is in fact
     // in the autocomplete
     let before = Date.now();
     await page.type("#active-editor", "test_admin");
     await awaitAnalysis(page, before);
     await expectExactText(
       page,
       ".autocomplete-item.fluid-selected.valid",
-      "test_admin/stdlib/Test::one_v0Any",
+      "test_admin/stdlib/Test::one_v1Any",
```
StachuDotNet, line 1294: Question for @pbiggar: what's the "any" for here? how does this not break?

StachuDotNet:
Finally passing in CI!

I added a few TODOs for myself to resolve after some lunch, but ready to have another set of eyes on for now.

Some notes:
- I doubled the playwright timeout. Doesn't feel great, but acceptable?
- I'm not sure what level of detail you typically receive in your Rollbars, and if these are satisfactory.
- Relevant `docs` PR to merge as/after this merges: darklang/docs/pull#227

integration-tests/tests.ts
pbiggar: The "Any" is the type. The text contents of the entire `div` is jammed together (even though they are in sibling sub-divs).

fsharp-backend/src/LibBackend/Canvas.fs
```diff
@@ -679,15 +679,20 @@ let loadJsonFromDisk (root : Config.Root) (c : Meta) : List<tlid * PT.Oplist> =
...
+        |> List.map (fun (tlid, oplist) ->
+          let tl =
+            let oplist = fromOplist meta [] oplist
+            let tls = toplevels oplist
+            let dtls = deletedToplevels oplist
+            (Map.mergeFavoringLeft tls dtls) |> Map.get tlid |> Option.unwrapUnsafe
+          (tlid, oplist, tl, NotDeleted))
+      with
+      | ex ->
+        debuG "Couldn't find test file - assuming empty oplists" meta
```
pbiggar, line 694: What's happening here?

integration-tests/tests.ts
pbiggar: You could pass it if it includes either the OCaml OR the F# error message.

integration-tests/tests.ts
pbiggar: I don't understand why this would be useful. Doesn't this pause the execution? If it's needed, can we have a comment explaining it?

scripts/devcontainer/_wait-until-server-ready
```diff
@@ -3,8 +3,22 @@
 
 set -euo pipefail
 
+base_url="$DARK_CONFIG_STATIC_HOST"
+
+for cmd in "$@"; do
+  case $cmd in
+    --use-apiserver )
```
pbiggar, line 10: I think I would rather see this split out into a script called `wait-until-apiserver-ready`

pbiggar (changes requested):
Looks great!

integration-tests/tests.ts
StachuDotNet: I removed the line now and there was no change. I must have added it when diagnosing a test locally, and wanted to quickly pause at that point in debug ode.

integration-tests/tests.ts
StachuDotNet: We must have fixed this error message since I did this - the error message is consistent for me now, it seems.

fsharp-backend/src/LibBackend/Canvas.fs
StachuDotNet: At one point, I believed that "tests without a test file" were failing here. It seems they're not, though, so I can remove this try/catch now.

StachuDotNet: https://app.circleci.com/pipelines/github/darklang/dark?branch=pull%2F3485&filter=all seems I'm not crazy, actually. These failing after removing.

pbiggar: Ah, it looks like the old load_json_from_disk_fluid uses `File.maybeReadFile` and uses options. Do we have a preference for Options vs Exceptions? I would imagine we prefer options.

pbiggar (changes requested):
One remaining question, happy to merge with or without.

fsharp-backend/src/LibBackend/Canvas.fs
StachuDotNet: Agreed - just pushed a PR that gets us to a slightly nicer place

PR #3487, opened 2022-02-15
Adjust legacy binary serializer to parse as UserFn before DB

### LINKED ISSUE #3486: Canvas fn loaded in F# backend interpreted as data store
If you create a canvas function, and (re)load the canvas against the F# backend, the function somehow "becomes" a broken data store.
If you load the equivalent URL on :8000, the canvas is loaded correctly, with the function existing as a function. (so, the problem indeed seems F#-specific)

https://user-images.githubusercontent.com/906686/154117934-fe958229-8490-4da6-b313-4253a0b8558f.mp4

This is currently causing some integration tests to fail, so I am working on this as part of #3485. (currently digging into `OCamlInterop.toplevelBin2Json` which may be relevant here.

StachuDotNet:
Resolves #3486

If you
- create a canvas 
- add a user function
- (re)load the canvas with the F# backend

currently the user function is (seemingly consistently) 'parsed' as a DB TopLevel rather than a UserFn.

In the legacy binary serialization server, attempt to parse UserFns before DBs.

I've tested this in the UI, and confirmed no regressions in other tests.

As existing integration tests already covered this bug (when the integration tests target F#, which they shortly will by default), I did not write new tests to cover this.


pbiggar:
Interesting situation. Does this mean that DBs and Functions have the same binary represenation? Can you test whether serialized DBs are read as functions with this change?

pbiggar:
If you rebase/merge this on main, CI should now run, I believe.

pbiggar:
FYI, I've seen this issue locally.

pbiggar:
Maybe something like this: https://github.com/darklang/dark/blob/main/fsharp-backend/tests/Tests/OCamlInterop.Tests.fs#L84

StachuDotNet:
I overcomplicated this for a while, but believe these 2 tests cover us pretty well.

I mentioned (on Slack) writing some more extensive TopLevel fuzzing, but somehow glanced over the existing infra. around that in the existing test files. It might make sense to create a separate Issue around generally fuzzing TopLevels and ensuring OCaml interop?

pbiggar:
Nicely done, thanks.

PR #3489, opened 2022-02-16
Push integration test results only when credentials present

StachuDotNet:
CI fails to post the results of integration tests to honeycomb, for PRs against a fork of Dark.

Example: https://app.circleci.com/pipelines/github/darklang/dark/2149/workflows/1877e29a-765c-49c1-b9c4-3bb46564c7bd/jobs/22397?invite=true#step-112-3

Only attempt push to honeycomb if credential `BUILDEVENT_APIKEY` is recent, in the appropriate shell script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -485,12 +485,6 @@ jobs:
       - darkcheckout
       - setup-app
       - attach_workspace: { at: "." }
-      # This is needed because kubectl --dry-run uses it
-      - auth-with-gcp:
-          background: false
-          filters:
-            branches:
-              ignore: /^(pull\/).*$/
       - build-gcp-containers
       - slack-notify-job-failure
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -486,7 +486,11 @@ jobs:
       - setup-app
       - attach_workspace: { at: "." }
       # This is needed because kubectl --dry-run uses it
-      - auth-with-gcp: { background: false }
+      - auth-with-gcp:
+          background: false
+          filters:
+            branches:
+              ignore: /^(pull\/).*$/
       - build-gcp-containers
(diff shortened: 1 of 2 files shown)


StachuDotNet:
This helped, but now we're hitting another issue:
the `auth-with-gcp` command depends on a variable `GCLOUD_SERVICE_KEY`, which is not present for such builds.

This command is currently used by the `predeployment-checks` and `gcp-containers-test` jobs called for this workflow.

Do you think it's reasonable to wrap the internals of `auth-with-gcp` in the same way, wrapping https://github.com/darklang/dark/blob/dbf2357469f9f9b856ed8971444fe87b3a4bfb30/.circleci/config.yml#L219-L222 with the same sort of `if`? @pbiggar

pbiggar:
> This helped, but now we're hitting another issue: the `auth-with-gcp` command depends on a variable `GCLOUD_SERVICE_KEY`, which is not present for such builds.
> 
> This command is currently used by the `predeployment-checks` and `gcp-containers-test` jobs called for this workflow.

There's quite some value from the `build-gcp-containers` step, so I'd suggest that we try to keep as much of it as possible. My suggestion would be:

- don't do `shipit release push` in `build-gcp-containers`. Instead do `shipit release prepare`, which is a good enough test for that step (the rest is just repeating the test in predeployment-checks). Then it doesn't need auth-with-gcp and we don't need an `if`

- skip the predeployment-checks step on PRs. I'd suggest using the built-in CircleCI feature for this rather than a bash `if`, this explains how: https://circleci.com/docs/2.0/workflows/#branch-level-job-execution. You can match on `branches.ignore: /pull\/.*/` or something like that.

StachuDotNet:
@pbiggar how does this look to you?

pbiggar:
Perfect!

Issue #3262, opened 2021-11-26
Test every function


pbiggar:
We want to make sure there are tests for every function. The list of functions without any tests is below.

Tests can be added in https://github.com/darklang/dark/tree/main/fsharp-backend/tests/testfiles. It's fairly self-explanatory, but the [readme](https://github.com/darklang/dark/blob/main/fsharp-backend/tests/testfiles/README.md) provides all the details.

<details

## How I created this list

To get the expected list of functions:

- log in to darklang.com, open the JS console, and type `copy(complete)`.
- paste this into a file: `pbpaste | jq '.[] | .name' -r | sort > expected`

To get the actually tested list of functions:
- `cd dark/fsharp-backend/tests/testfiles/ && grep -o -e "[A-Z][a-zA-Z0-9_]*\.[a-zA-Z0-9]\+_v[0-9]" * -h | sed 's/_v0//g' | sed 's/\./::/' | grep -v "Test::" | sort | uniq | sort > actual`

To see the list of functions that are not tested:
- ` comm -13 actual expected`

With some additional tweaking (remove Json::parse_v0, DarkInternal functions, and HttpClient functions that are tested elsewhere), we get the list of untested functions.

Note that the tested functions doesn't include any that don't have `::` in their names, so things like `assoc` or `||` won't be included. I manually removed those from the list.

</details>

- [x] DB::add_v0
- [x] DB::count
- [x] DB::delete
- [x] DB::deleteAll
- [x] DB::fetchAll
- [x] DB::fetchBy
- [x] DB::fetchByMany
- [x] DB::fetchOneBy
- [x] DB::fetchOneByMany
- [x] DB::get_v2
- [x] DB::insert
- [x] DB::keys
- [x] DB::queryExactFields
- [x] DB::queryExactFieldsWithKey
- [x] DB::queryOneWithExactFieldsWithKey
- [x] DB::queryOne_v2
- [x] DB::queryOne_v3
- [x] DB::schema
- [x] DB::schemaFields_v1
- [x] DB::schema_v1
- [x] DB::update
- [x] Date::<
- [x] Date::<=
- [x] Date::>
- [x] Date::>=
- [x] Date::greaterThanOrEqualTo
- [x] Date::lessThanOrEqualTo
- [x] Date::sub
- [x] Float::roundTowardsZero
- [ ] Http::redirectTo
- [ ] Http::respondWithHeaders
- [ ] StaticAssets::baseUrlFor
[...]

PR #3493, opened 2022-02-16
Add missing tests around Date module

StachuDotNet:
Per #3262, several functions in the Date module were un- or under-tested.

Write additional tests covering these functions: `subtract`, `add`, `<`, `>`, `<=`, `>=`, `lessThan`, `greaterThan`, `lessThanOrEqualTo`, `greaterThanOrEqualTo`

Ran unit tests locally


pbiggar:
Looks good, thanks!

PR #3497, opened 2022-02-18
Switch to NodaTime

pbiggar:
Switch from using System.DateTime to using NodaTime. NodaTime uses a much larger set of types which allows much more static safety.

I went with using `LocalDateTime` for `DDate` and `Instant` for everything else. `DDate` is implicitly in UTC, and has properties for manipulating days, months, minutes, etc, which was useful for LibDate. `Instant` is basically a timestamp, which is intended for things marking points in time.

When saving to the DB, there are different Npgsql functions to use, based on the table definition: some are tables use `timestamp with timezone` and some use `timestamp without timezone`. It's pretty inconsistent. I added `Sql.instantWithoutTimezone` for instants that needed to be stored as `timestamp without timezone`.

I also refactored and improved test error reporting, the fuzzing server's logging, and Ast traversals, as they were all needed for testing.


fsharp-backend/tests/testfiles/date.tests
```diff
@@ -161,19 +161,26 @@ Date.greaterThanOrEqualTo (d "2020-11-26T04:37:46Z") (d "2020-11-26T04:37:46Z")
...
 Date.second_v0 (d "2019-07-28T22:42:36Z") = 36
 Date.second_v0 (d "1905-12-13T10:12:14Z") = -46
 Date.second_v0 (d "1965-04-24T09:54:28Z") = -32
-
 Date.second_v0  (d "1969-07-05T12:34:00Z") = 0
 Date.minute_v0  (d "1955-05-16T18:08:00Z") = -52
 Date.minute_v0  (d "1949-06-27T12:26:00Z") = -34
 Date.minute_v0  (d "1908-05-12T10:18:00Z") = -42
 
+Date.weekday_v0 (d "2019-07-22T22:42:36Z") = 1
```
StachuDotNet, line 175:
```suggestion
Date.weekday_v0 (d "2019-07-22T22:42:36Z") = 1 // Monday
```
gentle reminder in a case where it may not be obvious

fsharp-backend/src/LibBackend/Cron.fs
```diff
@@ -38,27 +38,26 @@ let lastRanAt (cron : CronScheduleData) : Task<Option<System.DateTime>> =
...
-  // Must use the 5-value constructor as the 3-value constructor doesn't support days
+let convertInterval (interval : PT.Handler.CronInterval) : NodaTime.Period =
   match interval with
-  | PT.Handler.EveryDay -> System.TimeSpan(1, 0, 0, 0, 0)
-  | PT.Handler.EveryWeek -> System.TimeSpan(7, 0, 0, 0, 0)
-  | PT.Handler.EveryFortnight -> System.TimeSpan(14, 0, 0, 0, 0)
-  | PT.Handler.EveryHour -> System.TimeSpan(0, 1, 0, 0, 0)
-  | PT.Handler.Every12Hours -> System.TimeSpan(0, 12, 0, 0, 0)
-  | PT.Handler.EveryMinute -> System.TimeSpan(0, 0, 1, 0, 0)
+  | PT.Handler.EveryDay -> NodaTime.Period.FromDays 1
```
StachuDotNet, line 46: Nice helpers, NodaTime!

fsharp-backend/src/LibBackend/Db.fs
```diff
@@ -122,12 +123,43 @@ module Sql =
       LibExecution.DvalReprInternal.toInternalRoundtrippableV0 (RT.DObj dvalmap)
     Sql.parameter param
 
+  // We sometimes erroneously store these as timestamps that are not Utc. But we mean them to be Utc.
+  let instantWithoutTimeZone (i : NodaTime.Instant) : SqlValue =
+    // We can't use a Utc System.DateTime here as
```
StachuDotNet, line 128: as...?

StachuDotNet (approved):

PR #3498, opened 2022-02-18
Reorg (no change) fuzzer tests

StachuDotNet:
The fuzzer tests are all in one file right now, and it can be confusing to understand them
(and therefore discouraging to contribute).

Break the file up into several sub-modules.

### DIFF
--- a/fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
+++ b/fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
@@ -1,6 +1,7 @@
 module FuzzTests.ExecutePureFunctions
 
 open System.Threading.Tasks
+open FSharp.Control.Tasks
 open System.Text.RegularExpressions
 
 open Expecto


StachuDotNet:
I was tempted to rename `ExecutePureFunctions` to `PureFunctionExecution`. Both don't sound great to me. any ideas? Otherwise I can leave it as-is.

StachuDotNet:
Not sure what's up with that build error
`/home/dark/app/fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs(560,5): error FS3511: This state machine is not statically compilable. A resumable code invocation at '(572,6--573,74)' could not be reduced. An alternative dynamic implementation will be used, which may be slower. Consider adjusting your code to ensure this state machine is statically compilable, or else suppress this warning. [/home/dark/app/fsharp-backend/tests/FuzzTests/FuzzTests.fsproj]`
but I'll take a peek

StachuDotNet:
Ready now @pbiggar

PR #3501, opened 2022-02-21
Add various in-code documentation to ApiServer

StachuDotNet:
Increase in-code documentation of ApiServer, for future developers' understanding.

- adjust existing commentary from //-style to ///-style, so IDEs show the the descriptions embedded in their UIs.
  image
- add additional commentary where relevant

### DIFF
--- a/fsharp-backend/src/ApiServer/Api/APIOps.fs
+++ b/fsharp-backend/src/ApiServer/Api/APIOps.fs
@@ -1,11 +1,10 @@
+/// API endpoint to receive and handle an Operation
 module ApiServer.AddOps
-
-// Functions and API endpoints for the API
-
-open Microsoft.AspNetCore.Http
 
 open System.Threading.Tasks
 open FSharp.Control.Tasks
+open Microsoft.AspNetCore.Http
+
 open Prelude
 open Tablecloth
 open Http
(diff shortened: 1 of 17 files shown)


StachuDotNet:
I started with ApiServer as it's relatively small. I hope to follow up with other projects, with more back-and-forth questions to discover what comments are "helpful" along the way.

fsharp-backend/src/ApiServer/Api/APIOps.fs
```diff
@@ -43,6 +42,10 @@ let empty : Op.AddOpResult =
 
 let causesAnyChanges (ops : PT.Oplist) : bool = List.any Op.hasEffect ops
 
+/// API endpoint to add a set of Operations in a Canvas
```
pbiggar, line 45: These are properly called Ops, not Operations. Might be worth defining in Op.fs that an Op is an Operation, but I wouldn't use the full word here when we only use an abbreviation throughout the code.

fsharp-backend/src/ApiServer/Ui.fs
```diff
@@ -1,3 +1,7 @@
+/// Supports loading of static HTML pages
```
pbiggar, line 1: They're not totally static, as we replace per-user things.

pbiggar, line 1:
```suggestion
/// Supports loading of HTML pages
```

pbiggar (changes requested):
This is great! Very much in support of this.

I've made a few notes where appropriate, ready to merge once they're resolved.

pbiggar:
> In future similar PRs, expect many /// comments of /// WHATISTHIS. I think that'll be a good way for me to ask many questions all at once, and document the responses in-code.

Nice idea

PR #3516, opened 2022-02-26
Fix production configs

pbiggar:
Allow the bwdserver to join the load balancer:
- Fix the configuration settings to get the healthcheck to the bwdserver. The missing configuration was the nodeport. 
- Copy the bwdserver settings to apiserver.
- Change the host in the healthchecks to gce-ingress-healthcheck
- add note about logging
- turn on logging in apiserver
- Fix infinite redirects (it's the header that indicates https, not actually https on the server, as TLS is terminated at the load balancer)
- support the rest of the tryfsharpdark paths on bwdserver
- reduce bwdserver replicas to 3 (from 4)

### DIFF
--- a/fsharp-backend/src/BwdServer/Server.fs
+++ b/fsharp-backend/src/BwdServer/Server.fs
@@ -215,8 +215,14 @@ let unmatchedRouteResponse
 // ---------------
 // HttpsRedirect
 // ---------------
+
+let isHttps (ctx : HttpContext) =
+  // requests aren't actually https in production. The load balancer terminates https
+  // and forwards the details using headers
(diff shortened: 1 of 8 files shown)


StachuDotNet:
Are these changes what effect the predeployment ci check job?

pbiggar:
> Are these changes what effect the predeployment ci check job? Could be a totally off-base question :)

Yes!

Any changes to services (typically yaml files in the services/ directory) need to be manually applied before deploying. Sadly they break other branches which don't have that config yet.

They can typically be ignored (as I'm tracking the changes), but open to other solutions.

PR #3517, opened 2022-02-27
Improve circleci message

pbiggar:
Send a better circleci message to the server


pbiggar:
managed to do this via the github api without changing anything here

PR #3519, opened 2022-02-28
Backfill tests around DB module

StachuDotNet:
Per #3262, several functions in the DB module were un- or under-tested.

Write additional tests covering the relevant fns.

### DIFF
--- a/fsharp-backend/tests/testfiles/db.tests
+++ b/fsharp-backend/tests/testfiles/db.tests
@@ -73,9 +73,9 @@ DB.schema_v1 SortedX = { x = "Str"; sortBy = "Int" }
 
 // DB.set
 [test.db.set does upsert] with DB X
-(let old = DB.set_v1 { x = "foo" } "hello" X in
- let newval = DB.set_v1 { x = "bar" } "hello" X in
- DB.getAllWithKeys_v1 X) = [["hello"; { x = "bar"}]]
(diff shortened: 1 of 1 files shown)


StachuDotNet:
If/when this looks good to you @pbiggar, I'd appreciate an update to #3262, removing the DB and Date fns

fsharp-backend/tests/testfiles/db.tests
```diff
@@ -38,6 +43,46 @@
...
+
+
+
+// ------------
+// Setting values
+// ------------
+
+// DB.set
+[test.db.set does upsert] with DB X
+(let old = DB.set_v1 { x = "foo" } "hello" X in
```
pbiggar, line 76: I've been trying hard to avoid "foo", "bar" and other things in the codebase (though there are definitely some in there now). My understanding is that it makes the code less approachable, see for example [this post](https://dev.to/kiarathedev/why-i-don-t-use-foo-bar-baz-1kf7), which while it does not exactly match our situation, gives a general idea of why to avoid them.

pbiggar (changes requested):
Looks great, thanks!

One minor change and we should be good to go.

fsharp-backend/tests/testfiles/db.tests
StachuDotNet: Understood and agreed - I was admittedly being lazy/noncreative. Pushed a fix.

PR #3535, opened 2022-03-02
Default in-canvas analysis to use Blazor by default

StachuDotNet:
In-canvas "analysis" is currently powered by the OCaml backend compiled via `js_of_ocaml`.
We'd like to migrate to instead use our F# backend compiled via Blazor.

We attempted a merge of darklang/dark#3485, which resulted in notable production slowness of analysis.
So, that PR was broken down into several others, and this is the "final" PR to address.

Rather than default to the compiled OCaml WASM, with a truthy `use-blazor` query param to use the F# one, default to the Blazor-compiled backend, availing a `use-ocaml-analysis` query param to use the OCaml version

Note: this change implicitly changes which WASM-built backend is used during integration tests.


StachuDotNet:
Quick note:
during yesterday's testing, I found that production analysis felt significantly slower than locally.
I need to verify that with more testing, but this is curious.

StachuDotNet:
Catching up with some Slack threads, the remaining things to do here are:
- cancel analysis executions when new requests come in, to decrease UX pains of non-AOT Blazor being slow
- reproduce and resolve the loading issue that Paul faced. I haven't been able to repro - @pbiggar if you could share your experience in more detail, that'd be appreciated.
- resolve a (separate?) load-time issue that I faced[1]

---

[1] I managed to hit issues loading blazor.boot.js when using F# ApiServer in prod (using a-testing-fsharp). I believe the problem here is involved with some of:
- the blazor folder is copied with hashes (i.e. not in this 'unhashed' branch of logic https://github.com/darklang/dark/blob/main/scripts/deployment/_push-assets-to-cdn#L35-L36)
- F# API server excluding backend/static/blazor from `HASH_REPLACEMENTS` (https://github.com/darklang/dark/blob/main/fsharp-backend/src/ApiServer/Ui.fs#L63)

Relevantly, the OCaml equivalent not doesn't exclude the blazor folder from things it expects to be hashed: (https://github.com/darklang/dark/blob/main/backend/libbackend/webserver.ml#L1501)

I'm guessing the thing to do here will be to either change `_push-assets-to-cdn` to _not hash_ those files, or change the F# solution to not expect them to be hashed. Thoughts, @pbiggar?

StachuDotNet:
Since this PR was last left, we've improved Blazor performance (by switching to System.Text.Json for I/O between Wasm and JS)

pbiggar:
If this goes awry, we'll get rollbars, right?

StachuDotNet:
Yes, we should. BlazorWorker catches load errors and forwards them to appsupport's window.onerror which calls window.Rollbar.error

StachuDotNet:
I wouldn't merge this quite yet - I'm seeing some integration tests failing in my AOT branch and I'm not yet convinced they're specific to the work there.

pbiggar (approved):

StachuDotNet:
Currently #4007 is the only issue I'm aware of that _blocks_ this merge.

After that's done, my thoughts are to sanity test a bunch more, do some light perf checks, and call it OK.

Beyond sanity testing and some light perf checks, do we need anything else before flipping the switch, @pbiggar? Is the load time within reason? Personally I'm a bit tired of Blazor, so looking to merge this sooner than later.

(relevantly, not feeling super energetic about a _deep_ perf comparison between this and js_of_ocaml, and writing some post.)

pbiggar:
> Beyond sanity testing and some light perf checks, do we need anything else before flipping the switch, @pbiggar? Is the load time within reason? Personally I'm a bit tired of Blazor, so looking to merge this sooner than later.

No, I think we're good after that. We can call the whole rewrite done.

StachuDotNet:
- [ ] remove the `use-blazor=true` flag from `integration-tests` as part of this

PR #3538, opened 2022-03-03
Add tests around `emit` StdLib fn

StachuDotNet:
`emit_v0` and `emit_v1` are untested.

Add a `events.tests` file with tests - and generally allow for testing of `emit` within testfiles.

- in `EventQueue.fs`, add fn `testingGetQueue` for test purposes, to fetch all events on a queue
- add a Test-only lib fn `getQueue` that uses `testingGetQueue`
- adjust `LibExecution.Test.fs`
  - to allow for the parsing of `[test.name] with Worker MyWorker` tests
  - to add the handlers to test programs when relevant, using `Canvas.addOps` (if there's a better way to do this, please let me know)
- update the relevant README.md
- add tests around `emit`

### DIFF
--- a/backend/libbackend/event_queue.ml
+++ b/backend/libbackend/event_queue.ml
@@ -254,6 +254,27 @@ let dequeue (parent : Span.t) (transaction : Int63.t) : t option =
       result)
 
 
+(* Fetches (only) the values stored in a queue,
+   only to be used for writing tests *)
+let testing_get_queue canvas_id (event_name : string) =
+  Db.fetch
+    ~name:"get_queue"
+    "SELECT e.value
+      FROM events AS e
+      JOIN canvases AS c ON e.canvas_id = c.id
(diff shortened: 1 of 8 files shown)


StachuDotNet:
Took a while, but this was a fun one

StachuDotNet:
I still need to figure out why these fail for ocaml

pbiggar (commented):
Looks good!

To support OCaml, you'd need to add `Test.getQueue` to OCaml's libtest. I'm fine if you don't add this though - let me know which way you want to go.

StachuDotNet:
I took it as a personal challenge to get it working for OCaml - took me _way_ too long to figure out why the new fn wasn't exposed in `event_queue` (because I didn't add it to .mli file).

pbiggar (changes requested):
Looks good! One tiny change and we can merge.

PR #3541, opened 2022-03-06
Backfill string tests

StachuDotNet:
Several String lib functions are un-tested or undertested.

Add tests.

### DIFF
--- a/fsharp-backend/tests/testfiles/string.tests
+++ b/fsharp-backend/tests/testfiles/string.tests
@@ -88,7 +88,12 @@ String.foreach_v1 "🧟‍♀️🧟‍♂️" (fun x -> x) = "🧟‍♀️🧟
 String.foreach_v1  "👨‍❤️‍💋‍👨👩‍👩‍👧‍👦🏳️‍⚧️‍️🇵🇷" (fun x -> x) = "👨‍❤️‍💋‍👨👩‍👩‍👧‍👦🏳️‍⚧️‍️🇵🇷"
 String.foreach_v1 "abcdef" (fun x -> blank) = blank
 
-//String.fromChar_v1 'a' = "a"
+// fromChar
+String.fromChar_v0 'a' = Test.typeError_v0 "String::fromChar was removed from Dark" // FSHARPONLY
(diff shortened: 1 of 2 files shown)


fsharp-backend/tests/testfiles/string.tests
```diff
@@ -88,7 +88,12 @@ String.foreach_v1 "🧟‍♀️🧟‍♂️" (fun x -> x) = "🧟‍♀️🧟
 String.foreach_v1  "👨‍❤️‍💋‍👨👩‍👩‍👧‍👦🏳️‍⚧️‍️🇵🇷" (fun x -> x) = "👨‍❤️‍💋‍👨👩‍👩‍👧‍👦🏳️‍⚧️‍️🇵🇷"
 String.foreach_v1 "abcdef" (fun x -> blank) = blank
 
-//String.fromChar_v1 'a' = "a"
+// fromChar
+String.fromChar_v0 'a' = Test.typeError_v0 "String::fromChar was removed from Dark" // FSHARPONLY
+String.fromChar_v0 ("1" |> String.toChar_v1) = Test.typeError "This function no longer exists." // OCAMLONLY
+String.fromChar_v1 'a' = "a" // FSHARPONLY
+String.fromChar_v1 ("1" |> String.toChar_v1) = Test.typeError "String::fromChar_v1 was called with the wrong type to parameter: c" // OCAMLONLY
```
StachuDotNet, line 95: Is this the best we can do?

pbiggar (changes requested):
Looks good.

Can you also do some unicode tests here? Given that Dark is supposed to support characters are EGCs (extended grapheme pairs), this would be a key place to test that.

pbiggar:
You might need `open Types.RuntimeT`?

StachuDotNet:
> You might need `open Types.RuntimeT`?

Great, that unblocked that error, thanks.

StachuDotNet:
Somehow struggling to get `String.fromList_v1 [Test.toChar_v0 "a"]` working, I suspect because of the Option returned. My attempts to deconstruct in a way that both OCaml and F# like are failing.

Separately, I don't have good intuition to what "unicode tests" should be implemented here. Going to step away from this as I feel somehow stuck. Feel free to finish this up, or maybe we can pair for a few minutes, otherwise I'll circle back.

PR #3549, opened 2022-03-09
Tidy some .unported files

StachuDotNet:
Several .unported.ml files exist within the F# project. These should be reviewed, and deleted where relevant.

Review each function/module, checking:
- where was it used in the OCaml backend?
- have we already ported _that_ code?
  - If so, how? Likely able to delete the code in .unported
  - If not, either implement it or leave it for later


StachuDotNet:
There are still .unported files to go, but this is a start.

There's one remaining fn in `db.ml.unported`: `iter_with_cursor`. Is that still relevant, or can I kill it? @pbiggar

pbiggar:
> There are still .unported files to go, but this is a start.
> 
> There's one remaining fn in `db.ml.unported`: `iter_with_cursor`. Is that still relevant, or can I kill it? @pbiggar

kill it

Issue #3540, opened 2022-03-05
Canvas loads are too slow


pbiggar:
Loading medium-to-large canvasas, such as ops-adduser, are much too slow. An initial-load takes over 30seconds.

In each case, the data is fetched quickly enough from the DB. However, running it though the legacy server is slow. We use a legacy server to deserialize the request, and sending the data, decoding it, serializing it to JSON, and send it back, are all rather quick. However, decoding the JSON on the F# end is extremely slow.

In addition, because all functions and DBs (not just the required ones) are fetched for every http call, that can be a lot of times these errors happen.

pbiggar:
BwdServer loads aren't too slow at the moment. Calls to http://ops-adduser.builtwithdark.localhost:8000/isavailable/paul take 21ms in OCaml and ~300ms in F# (when warm). This is mostly calls to the legacy server - executeDarkHandler took only 9ms vs loadFrom taking 303ms out of a total of 321ms.

So still too slow, but not quite at the level of ApiServer.initialLoad. I think this is because initialLoad also loads deleted handlers, which are less likely to be cached as well.

pbiggar:
The plan:

- [ ] identify a stable (probably binary) serializer that deserializes fast.
- [ ] In ApiServer, save oplists and cached oplists with the new serializer, in the toplevel_oplists table, in addition to the old serializer
- [ ] in parallel the the old OCaml code will still be running. We should identify a way of knowing when a toplevel has been saved that uses the old code. Ideas:
  - update the ocaml code to delete those fields. Thus any time the field exists it must be valid
  - update the ocaml code to add a "sync token". The f# code would also add a sync token (but in both columns). If the sync tokens are that same then the new data is valid (both are the same). Otherwise only  
  - save a hash of the ocaml-encoded binary data as part of the fsharp-encoded data, and validate they match. If they don't the old code is valid.
- [ ] have the apiserver load from either a new field if it exists or via the legacy server if it doesnt. If it loads it from legacy server, also save it. Watch out for race conditions.
- [ ] switch over all canvases to the new apiserver

pbiggar:
Looks like MessagePack.CSharp is plenty fast and needed [few annotations](https://github.com/pocketberserker/MessagePack.FSharpExtensions). Found via https://aloiskraus.wordpress.com/2019/09/29/net-serialization-benchmark-2019-roundup/.

I ran every toplevel through it and verified it was able to fully serialize and deserialize it.

pbiggar:
[...]

PR #3566, opened 2022-03-11
Save and read from new serialization where possible

pbiggar:
Does most of #3540.

Since serialization with the legacy server is so slow, both BwdServer and ApiServer give bad experiences. Using the editor is a very bad experience for even medium-sized canvases, as the API call to `initial_load` makes one call to the legacy_server for each toplevel (including deleted toplevels). Calling the BwdServer is also bad, as it needs to load every function and database regardless of whether the handler uses them.

The slow part is not actually calling the server or deserializing the binary, but rather decoding the JSON that the server returns. Encoding the JSON isn't super great either but decoding it is extremely bad.

Instead, this PR lets us save handlers using MessagePack.CSharp binary serialization format. This is not as fast as the OCaml binary serializer, but it's much much much much better than using the legacy_server.

This works by adding new columns on the toplevel_oplists table (which is where the data is currently stored). In addition to the two rows where the ocaml-serialized data is stored (data and rendered_oplist_cache), we add two more for the fsharp-serialized data (oplist and oplist_cache).

When saving data in F#, we save all 4 fields. When saving data in OCaml, we save the 2 ocaml-serialized fields, and set the other two to NULL.

When reading data in F#, we read all 4 fields and pick appropriately: if the new ones are NULL we'll deserialize via the legacy server). When reading in OCaml, we only use the two old columns.

In this way, F# writes are still readable from OCaml. This is important because the bwdservers and queueservers are still ocaml.

OCaml writes are also still readable from F#, because they'll delete (overwrite with null) old fsharp-serialized values, and F# will read the new values (deserializing using the legacy server).

I also want to automatically upgrade values on read. However, I'm going to leave that to a future PR as that is a little more sketch and I'd like to get this merged without increasing scope.


fsharp-backend/src/LibBackend/BinarySerialization.fs
```diff
@@ -0,0 +1,53 @@
...
+open Npgsql
+open LibBackend.Db
+
+open Prelude
+open Tablecloth
+open Prelude.Tablecloth
+
+module PT = LibExecution.ProgramTypes
+
+// TODO inline this file into canvas
```
StachuDotNet, line 18: Does this mean "make this a Dark program someday"?

StachuDotNet, line 18: Does this mean "move this to LibBackend/Canvas.fs"?

StachuDotNet:
Maybe it'll be clearer to me after more reading, but does this prepare us for a case where the shape of `Spec` or `Op` changes? Or will the binary serialization fail then? (one nice thing about serializing and then deserializing JSON is that as long as your changes are _additive_, it works out okay. 

or maybe we don't anticipate those types to change? if we're relying on that somehow, let's add a note at least.

fsharp-backend/src/Prelude/Prelude.fs
```diff
@@ -1526,6 +1526,29 @@ module Task =
...
+    }
+
+  let mapWithConcurrency
+    (concurrencyCount : int)
+    (f : 'a -> Task<'b>)
+    (list : List<'a>)
+    : Task<List<'b>> =
+    let semaphore = new System.Threading.SemaphoreSlim(concurrencyCount)
+    let f = execWithSemaphore semaphore f
+    List.map f list |> flatten
```
StachuDotNet, line 1550: neat little alg.!

fsharp-backend/src/LibBackend/Serialize.fs
```diff
@@ -64,6 +64,62 @@ let isLatestOpRequest
...
+  let deserializePos (pos : Option<string>) : pos =
+    try
+      pos
+      |> Option.map Json.OCamlCompatible.deserialize<pos>
+      |> Option.unwrap { x = 0; y = 0 }
+    with
+    | e -> { x = 0; y = 0 }
+  match typ with
+  | "db" ->
+    task {
```
StachuDotNet, line 82: Does this compile/work if you put the `match` inside the `task {}`, rather than have a `task` in each branch? If so, that'd read a bit nicer I think

fsharp-backend/src/LibBackend/BinarySerialization.fs
pbiggar: Yes, but it was an error - I copied it from another file by accident.

fsharp-backend/src/LibBackend/Serialize.fs
```diff
@@ -76,106 +132,62 @@ let loadOnlyRenderedTLIDs
...
-                  |> PT.TLType
-                | _ ->
-                  Exception.raiseInternal
-                    "Invalid tipe for toplevel"
-                    [ "type", typ; "tlid", tlid; "canvas_id", canvasID ]
-            })
-          |> Task.flatten
-      }))
-
+    // Instead we just limit it to 2 threads at a time does the
```
StachuDotNet, line 176: without "does the," maybe?

fsharp-backend/src/LibBackend/Serialize.fs
pbiggar: It didn't work when I did that initially, even though I expected it would. Since this code isn't going to survive too long, i didn't spend much time investigating.

fsharp-backend/src/Prelude/Prelude.fs
pbiggar: Thanks! I was surprised how easy this came out!

fsharp-backend/src/LibBackend/Serialize.fs
pbiggar: Thanks.

pbiggar:
> Maybe it'll be clearer to me after more reading, but does this prepare us for a case where the shape of `Spec` or `Op` changes? Or will the binary serialization fail then? (one nice thing about serializing and then deserializing JSON is that as long as your changes are _additive_, it works out okay.
> 
> or maybe we don't anticipate those types to change? if we're relying on that somehow, let's add a note at least.

Yeah, I should add some notes and some tests around this.

Currently I have no idea what happens if we change the Op or Toplevel or Expr types. In the OCaml serializers, we had to be very careful, and were basically only able to add extra variants to existing sumtypes. And so we had a few checks:
- a file was generated that represented the serialization format. If a new serialization format was created, we would know we had changed the format
- a digest representing the format was stored in the DB (toplevel_oplists.digest)
- we had a big set of warnings on the affected files and types

I think I can get a similar level of confidence with the following:

- generate a file, which we commit, representing a serialized value (shows if things have changed)
- a test that we can read that file and get the right answer
- a test that we can serialize it and get the same value
- warnings around ProgramTypes.fs
- refactor ProgramTypes.fs to just represent the program types
- some quick manual tests as to what happens if we make some changes to the files

StachuDotNet (approved):
Looks good, apart from the comment I made earlier:
Let's say we want to add some property to `ProgramTypes.Op.CreateDB` in a few months. I think right now with JSON, we wouldn't have any issue deserializing old data into the new type. Does that hold true with MessagePack? I did some quick searches of binary serializers and msgpack specifically, (just a minute) and didn't see an obvious conclusion.

StachuDotNet:
That sounds like a good plan

pbiggar:
> Looks good, apart from the comment I made earlier: Let's say we want to add some property to `ProgramTypes.Op.CreateDB` in a few months. I think right now with JSON, we wouldn't have any issue deserializing old data into the new type. Does that hold true with MessagePack? I did some quick searches of binary serializers and msgpack specifically, (just a minute) and didn't see an obvious conclusion.

Right now we don't use JSON for serialization - we use the bin_io serializers. We also can't use JSON as it's much too slow.

The existing bin_io serializers that we use also have this problem. The solution we use is to never add properties to `Op.CreateDB`, instead we create `Op.CreateDBButWithAnotherProperty` (good example actually since we already have `Op.CreateDBWithBlankOr`, which was a sitation exactly like this).

I don't know if that holds true for MessagePack, but I'd be very surprised if it doesn't. Certainly I will be testing.

This makes me realize I should also add exception handling during the new binary serialization, and add a fallback to use the old serializers.

The current version of the C# serializer supports a string key serializer, which is a bit more resilient to versioning changes. It's the default right now, but I think that we would be better with very explicit changes and tests rather than relying on this behaviour. It's also 2x slower with this and generates data 2x bigger, so I'd prefer to change away from this now.

pbiggar:
Remaining TODO list:
- [ ] run migration before merge
- [ ] write test case for the new binary serialization. Cover all possible cases of Toplevel and Op, including Expr
- [ ] add unit tests that the test cases can roundtrip
- [ ] generate file from the unit test data when apiserver loads. Commit to repo.
- [ ] add test that we can read the generated files
- [ ] add test that the generated bytes for the test cases match the saved files
- [ ] switch to int keys
- [ ] refactor ProgramTypes 
- [ ] experiment with adding new fields and enums and see what happens. Document in comments.
- [ ] fix 2 erroneous comments

pbiggar:
Changes since last review (starts at https://github.com/darklang/dark/pull/3566/commits/6bebf89c6cb8bcc053d76a1434dda8be1f15a6a0):

- I change the format to use ints instead of strings for record member names. This dramatically dropped the size of the generated data, to roughly match the old sizes (vs approximately 2x before).

- enable compression for oplist serialization. Since there's a lot of duplicate here (eg, if you press two keys you'll get two nearly identical SetHandler ops), this reduces stored size by about 15x.

- made the options part of the serialization function, not something global that needs to be initialized.

- add a big test case hitting every form of every type in ProgramTypes
- made a set of test cases to ensure the format doesn't change. I won't describe it here as it should be described enough in the comments (and if not I need to imprve the comments).
- add a lot of discussion of what changes can safely be made
- remove old DB migration Ops since they're not used anywhere

pbiggar:
@StachuDotNet Ready for another review.

fsharp-backend/tests/Tests/SqlCompiler.Tests.fs
```diff
@@ -11,7 +11,7 @@ module PT = LibExecution.ProgramTypes
 open LibExecution.RuntimeTypes
 
 module C = LibBackend.SqlCompiler
-module S = LibExecution.Shortcuts
+module S = TestUtils.RTShortcuts
```
StachuDotNet, line 14: much better place, great!

fsharp-backend/src/LibBackend/BinarySerialization.fs
```diff
@@ -0,0 +1,55 @@
+module LibBackend.BinarySerialization
+
+// Conversion to/from Dark values to binary formats
```
StachuDotNet, line 3:
```suggestion
/// Conversion to/from Dark values to binary formats
module LibBackend.BinarySerialization

```

fsharp-backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -975,25 +349,27 @@ type Op =
   | SetExpr of tlid * id * Expr
   | TLSavepoint of tlid
   | DeleteFunction of tlid
-  | CreateDBMigration of tlid * id * id * (string * id * string * id) list
-  | AddDBColToDBMigration of tlid * id * id
-  | SetDBColNameInDBMigration of tlid * id * string
-  | SetDBColTypeInDBMigration of tlid * id * string
-  | AbandonDBMigration of tlid
-  | DeleteColInDBMigration of tlid * id
```
StachuDotNet, line 983: note to self: I'm not immediately sure why these could be removed - will check history and follow up here.

fsharp-backend/tests/Tests/BinarySerialization.Tests.fs
```diff
@@ -0,0 +1,462 @@
+module Tests.BinarySerialization
```
StachuDotNet, line 1: this file reads really nicely!

fsharp-backend/tests/Tests/Canvas.Tests.fs
```diff
@@ -79,17 +91,29 @@ let testHttpLoadIgnoresDeletedFns =
     do!
       Canvas.saveTLIDs
         meta
-        [ (handler.tlid, [ hop handler ], PT.TLHandler handler, Canvas.NotDeleted)
-          (f.tlid, [ PT.SetFunction f ], PT.TLFunction f, Canvas.NotDeleted) ]
+        [ (handler.tlid,
+           [ hop handler ],
+           PT.Toplevel.TLHandler handler,
+           Canvas.NotDeleted)
+          (f.tlid, [ PT.SetFunction f ], PT.Toplevel.TLFunction f, Canvas.NotDeleted) ]
```
StachuDotNet, line 98: I find myself asking "... what's 'f' here?". Really easy to lose context in a PR review. Could we rename?

fsharp-backend/src/LibBackend/PackageManager.fs
```diff
@@ -331,14 +333,20 @@ let allFunctions () : Task<List<PT.Package.Fn>> =
...
               version = version }
           use _span =
-            LibService.Telemetry.child "decode package binary" [ "fn", string name ]
+            LibService.Telemetry.child
+              "decode package binary"
+              [ "fn",
+                name
+                |> PT2RT.FQFnName.PackageFnName.toRT
+                |> RT.FQFnName.PackageFnName.toString
+                :> obj ]
```
StachuDotNet, line 342: I'm curious why this needs to be `:> obj`'d.

fsharp-backend/src/LibExecution/ProgramTypesParser.fs
```diff
@@ -0,0 +1,272 @@
+/// Provides toString and fromString functions for program types
+///
+/// We are moving to not needing parsers, but for now they are still necessary. They
+/// mostly convert to a format used in the database, or in the OCAml compatible
```
StachuDotNet, line 4:
```suggestion
/// mostly convert to a format used in the database, or in the OCaml compatible
```

fsharp-backend/src/LibExecution/ProgramTypesParser.fs
```diff
@@ -0,0 +1,272 @@
...
+/// representation (spoken by the client and used in the API for example)
+module LibExecution.ProgramTypesParser
+
+// CLEANUP: almost all users of functions in this file could be replaced by having
+// better types and/or propagating the type definitions end-to-end (such as to the
+// frontend). It would likely also increase the reliability and reduce the
```
StachuDotNet, line 10:
```suggestion
// client). It would likely also increase the reliability and reduce the
```
Just trying to be consistent with the first line [here](https://docs.darklang.com/contributing/tour-of-editor)

StachuDotNet: editor also works

StachuDotNet (changes requested):
This is looking really great! Code is reading much nicer now, generally.
I like the way stability of the binary serialization is ensured. I wonder if there's a nice way to ensure that the test in `BinarySerialization.test.fs` contains all available Ops (i.e. walk the tree and error if any ops aren't included) or if that's worth exploring.

> enable compression for oplist serialization. Since there's a lot of duplicate here (eg, if you press two keys you'll get two nearly identical SetHandler ops), this reduces stored size by about 15x.

awesome.

---

Just a few pending comments/questions, but this looks good to me!

fsharp-backend/tests/Tests/SqlCompiler.Tests.fs
pbiggar: Yeah, I like this a lot better too.

fsharp-backend/tests/Tests/BinarySerialization.Tests.fs
```diff
@@ -0,0 +1,462 @@
...
+
+module File = LibBackend.File
+module Config = LibBackend.Config
+module PT = LibExecution.ProgramTypes
+module BinarySerialization = LibBackend.BinarySerialization
+
+/// The test values below are used to check the exact output of test file. So we need
+/// the test inputs to be consistent, which is why we never use `gid ()` below, or
+/// FSharpToExpr functions.
+let testExpr =
```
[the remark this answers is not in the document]
pbiggar: I aimed to write one that hit every type of the variant.

fsharp-backend/tests/Tests/Canvas.Tests.fs
pbiggar: I don't disagree, but I just hit this from a module rename and didn't touch it otherwise. Would rather not increase scope on this.

fsharp-backend/src/LibBackend/PackageManager.fs
pbiggar: `child` takes a `List<string * obj>` (which eventually makes its way into OpenTelemetry which uses objs.

PR #3571, opened 2022-03-14
Upgrade .NET SDK to 6.0.201

StachuDotNet:
- .NET, FSharp.Core and FSharp.Compiler.Service are out of date

Update them


StachuDotNet:
UI failing for me locally; investigating

StachuDotNet:
Actually, the UI is fine for me if I go incognito or clear browser stuff. Is it expected that upgrades should work automatically locally? (I imagine on prod, our CDN takes care of these sorts of things, but not sure about locally)

StachuDotNet:
The build is failing in CI due to an apparent difference in expected (6.0.100) vs actual (6.0.201) .NET versions.

Perhaps `DOTNET_SDK_VERSION` is set somewhere in CircleCI config?
Or maybe the old version is hanging around from an earlier build.

pbiggar:
CI uses the SDK from the dockerfile. However, the docker image it uses is specified [here](https://github.com/darklang/dark/blob/53270926c6c493444dadf607f2ffb6578a210442/.circleci/config.yml#L20).

See this note: https://github.com/darklang/dark/blob/5e29b1c423e9888e45a3d67f9f73eb0064f15948/Dockerfile#L4

StachuDotNet:
Outstanding issue in FSharpToExpr to be handled - likely related to FCS update. Looking at that.

pbiggar:
> Updated both references of the sha - the 'normal' one and the rust one. Not sure if the Rust one should have stayed, but I suppose consistency is best.

Yeah, they should stay in sync.

pbiggar:
> Outstanding issue in FSharpToExpr to be handled - likely related to FCS update. Looking at that.

They occasionally change the AST format, should be straightforward to accomodate.

PR #3578, opened 2022-03-16
Update ApiServer to expect static blazor files to be hashed

StachuDotNet:
Blazor assets that are stored in `backend/static/blazor` are hashed in prod, and the OCaml server expects them to be such, but the F# server doesn't.

This updates `ApiServer` to expect hashed `blazor` folder files.


pbiggar (approved):
Looks good. Will merge this once #3566 goes out - I'm slightly nervous about it.

PR #3579, opened 2022-03-16
Include metadata in F# Wasm exceptions

### LINKED ISSUE #3577: Report exception metadata in blazor
In #3576, the metadata field "expr" from the exception was not reported to rollbar or to the console. We should report it as that's where we put the important context-sensitive info to debug the exception.

StachuDotNet:
Aimed to resolve #3577.

Tested locally by updating `src/LibExecution/OCamlTypes.fs` locally, commenting out the handling of `EIntegers` from `rt2ocamlExpr`, and running an analysis on something that returned an Int.
image

If I try to change the shape of the Error returned (e.g. by returning an anonymous record of `{ message: string, metadata: ...}` , I get issues in decoding the error on the client. So, I shoe-horned both into the string - happy to take recommendations there.

### DIFF
--- a/fsharp-backend/src/Wasm/Wasm.fs
+++ b/fsharp-backend/src/Wasm/Wasm.fs
@@ -313,7 +313,9 @@ type EvalWorker =
           let metadata = Exception.toMetadata e
           System.Console.WriteLine("Error parsing analysis in Blazor")
           System.Console.WriteLine($"called with message: {message}")
-          System.Console.WriteLine($"caught exception: \"{e.Message}\" \"{metadata}\"")
+          System.Console.WriteLine(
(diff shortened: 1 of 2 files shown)


fsharp-backend/src/Wasm/Wasm.fs
```diff
@@ -310,10 +310,13 @@ type EvalWorker =
           )
         with
         | e ->
```
StachuDotNet, line 312: The 'top' case here handles parsing, and the below one handles exceptions running analysis.

pbiggar:
> If I try to change the shape of the Error returned (e.g. by returning an anonymous record of { message: string, metadata: ...} , I get issues in decoding the error on the client. So, I shoe-horned both into the string - happy to take recommendations there.

We can do better on the client another day!

PR #3597, opened 2022-03-18
Don't page when the canvas is one that we know is broken

pbiggar:


fsharp-backend/src/LibBackend/Canvas.fs
```diff
@@ -358,6 +358,19 @@ let empty (meta : Meta) : T =
...
+  [ "danbowles"
+    "danwetherald"
+    "ellen-dbproblem18"
+    "ellen-dbtests"
+    "ellen-preview"
+    "ellen-stltrialrun"
+    "ellen-trinity"
+    "jaeren_sl"
+    "jaeren_sl-crud"
+    "sydney" ]
```
StachuDotNet, line 371: Are the canvases in the list ones that you've found broke when you loaded all `initial_load`s yesterday, or previously-known canvases with issues?

pbiggar: Some of them were known, some of them were new (though I discovered them doing the tests on prodclone a few days ago). They were all manually investigated. Most of them failed because of a known issue with having two DBs of the same name, except for a few that were due changing a user's username.

StachuDotNet: OK, thanks

Issue #3554, opened 2022-03-09
Add package errors to startup readiness check


pbiggar:

pbiggar:
Done (the checklist was in #3607)

PR #3607, opened 2022-03-20
Avoid risk of bad packages at startup

pbiggar:
This adds a path to use a binary serialization for packages. This makes it faster to load the first request on a server, but also makes it less error prone - we won't end up with a lazy exception.

The plan from here is:
- [x] run the migration
- [x] merge #3608
- [x] manually test to make ensure this still works
- [x] figure out why ExecHost isn't working in production
- [ ] merge this PR
- [ ] manually test to make ensure everything still works
- [ ] Run `ExecHost convert-packages` from production
- [ ] manually test to make sure this works

### DIFF
--- a/fsharp-backend/src/LibBackend/Init.fs
+++ b/fsharp-backend/src/LibBackend/Init.fs
@@ -70,6 +70,8 @@ let init (serviceName : string) (runSideEffects : bool) : Task<unit> =
       EventQueue.WorkerStates.STJJsonConverter.WorkerStateConverter()
     )
 
+    do! PackageManager.init ()
+
     if runSideEffects then do! Account.init serviceName
 
     print $" Inited LibBackend in {serviceName}"
--- a/fsharp-backend/src/LibBackend/Init.fs
+++ b/fsharp-backend/src/LibBackend/Init.fs
@@ -59,7 +59,7 @@ type WaitForDB =
   | WaitForDB
   | DontWaitForDB
 
-/// Initialize LibBackend.
+/// <summary>Initialize LibBackend.</summary>
 ///
 /// <remarks> This function does not do any behaviour which accesses DB tables and
 /// data, as the DB might not be migrated to it's correct form at this point (eg in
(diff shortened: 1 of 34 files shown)


StachuDotNet:
What does body2 mean here?

pbiggar:
> What does body2 mean here?

New version of the body column.

StachuDotNet (approved):

pbiggar:
I continued with this PR to solve #3554. The plan was to ensure that the exception happened at startup, not during later execution. This led to a medium-sized refactor of initialization code. As a result, I pulled the initial commits into their own PR: #3608.

fsharp-backend/src/LibBackend/Init.fs
```diff
@@ -55,9 +55,18 @@ let waitForDB () : Task<unit> =
...
+  | WaitForDB
+  | DontWaitForDB
+
+/// Initialize LibBackend.
+///
+/// <remarks> This function does not do any behaviour which accesses DB tables and
+/// data, as the DB might not be migrated to it's correct form at this point (eg in
+/// the test of dev environment). This is called by ExecHost, which does the
+/// migration. You cannot expect the DB to be ready when LibBackend is initialized -
+/// call `waitForDB` if you need that.</remarks>
```
StachuDotNet, line 68:
```suggestion
/// <summary>Initialize LibBackend.</summary>
///
/// <remarks> This function does not do any behaviour which accesses DB tables and
/// data, as the DB might not be migrated to it's correct form at this point (eg in
/// the test of dev environment). This is called by ExecHost, which does the
/// migration. You cannot expect the DB to be ready when LibBackend is initialized -
/// call `waitForDB` if you need that.</remarks>
```

In my experience, a bare `///` is great if you just want "one summary section".

Otherwise, if you want something like `<remarks>`, it's best to wrap each 'section' explicitly (i.e. wrap the beginning here with `<summary>`...`</summary>`. Otherwise, the tool tip you get won't be ideal.

This tends to be true across F# IDEs. Here's the same issue in Rider:
image
image

It's minor and silly, but something to keep in mind.

pbiggar: Thanks!

Issue #3524, opened 2022-02-28
Fix "Prelude+DeveloperException: expecting str"


pbiggar:
https://rollbar.com/darkops/darklang/items/3024/

pbiggar:
Unclear where this is happening so adding further instrumentation in https://github.com/darklang/dark/pull/3522.

There is probably enough info to give this a try though:

- it happens when calling toStringPairs from LibHttpClientv5.

Next step
- replicate
- make a test case

pbiggar:
This code triggers the exception:

```
HttpClient::getv5 "https://google.com" {} "test"
```

The goal here is for this exception not to end up in Rollbar.

pbiggar:
Fixed by #3610

PR #3610, opened 2022-03-21
Adjust Interpreter.fs to handle DevExceptions

StachuDotNet:
Within the `HttpClient` stdlib fns, passing in a `string` where a `DObj<string>` is expected for `headers` results in a DeveloperException, both in OCaml and F# backends. Issue #3524.

## Context

This error comes from [Dval.to_string_pairs_exn request_headers](https://github.com/darklang/dark/blob/4de2da65b007579ec0e36ff27539007620101120/backend/libbackend/libhttpclient.ml#L104) and [DvalReprExternal.toStringPairs reqHeaders](https://github.com/darklang/dark/blob/4de2da65b007579ec0e36ff27539007620101120/fsharp-backend/src/BackendOnlyStdLib/HttpClient.fs#L439), respectively.

In the OCaml backend, these sort of expected runtime exceptions seem to be handled by the `InProcess` wrappers that surround many functions.
For example, `HttpClient::get_v5` is wrapped by `InProcess` [here](https://github.com/darklang/dark/blob/4de2da65b007579ec0e36ff27539007620101120/backend/libbackend/libhttpclient.ml#L223).
The actual `InProcess` impl. is [here](https://github.com/darklang/dark/blob/4de2da65b007579ec0e36ff27539007620101120/backend/libexecution/ast.ml#L802-L833), and the `try`/`catch` within works well to handle this.

In the F# backend, `Interpreter.fs`.`execFn` is the closeset equivalent to `InProcess` that I've found. This has a `try`/`catch`, but the `catch` doesn't handle DeveloperExceptions, so the [final wildcard branch](https://github.com/darklang/dark/blob/4de2da65b007579ec0e36ff27539007620101120/fsharp-backend/src/LibExecution/Interpreter.fs#L777) ends up getting triggered. This branch yields a rollbar and no useful UI that tells the developer what to do.

--- 

## WIP

Here I've attempted to resolve the issue by adding a `| :? DeveloperException` branch, which does get used, but it seems the error message still isn't visible in the UI. I suspect I need to change the `state.notify` usage.

### DIFF
--- a/fsharp-backend/src/LibRealExecution/RealExecution.fs
+++ b/fsharp-backend/src/LibRealExecution/RealExecution.fs
@@ -60,6 +60,7 @@ let createState
 
     // Any real execution needs to track the touched TLIDs in order to send traces to pusher
     let touchedTLIDs, traceTLIDFn = Exe.traceTLIDs ()
+    HashSet.add tlid touchedTLIDs
 
     let tracing =
       { Exe.noTracing RT.Real with
--- a/fsharp-backend/src/LibRealExecution/RealExecution.fs
+++ b/fsharp-backend/src/LibRealExecution/RealExecution.fs
@@ -60,7 +60,6 @@ let createState
 
     // Any real execution needs to track the touched TLIDs in order to send traces to pusher
     let touchedTLIDs, traceTLIDFn = Exe.traceTLIDs ()
-    HashSet.add tlid touchedTLIDs
 
     let tracing =
       { Exe.noTracing RT.Real with
--- a/fsharp-backend/src/LibExecution/Interpreter.fs
+++ b/fsharp-backend/src/LibExecution/Interpreter.fs
@@ -776,11 +776,9 @@ and execFn
           state.notify state "fakedval found" [ "dval", dv ]
           return dv
         | :? DeveloperException as e ->
-          printfn "DevExn"
-          state.notify state "fakedval found" [ "dval", DStr "test" ]
+          state.notify state "Developer error found" [ "error", e.Message ]
           return Dval.errSStr sourceID e.Message
         | e ->
-          printfn "Unhandled thing %A" e
           // CLEANUP could we show the user the execution id here?
           state.reportException
             state
--- a/fsharp-backend/src/LibExecution/Interpreter.fs
+++ b/fsharp-backend/src/LibExecution/Interpreter.fs
@@ -7,6 +7,7 @@ open FSharp.Control.Tasks.Affine.Unsafe
 
 open Prelude
 open RuntimeTypes
+open Prelude
 
 let globalsFor (state : ExecutionState) : Symtable =
   let secrets =
@@ -605,7 +606,7 @@ and execFn
        && not state.onExecutionPath
        && Set.contains fnDesc state.callstack then
       // Don't recurse (including transitively!) when previewing unexecuted paths
-      // in the editor. If we do, we'll recurse forever and blow the stack. *)
+      // in the editor. If we do, we'll recurse forever and blow the stack.
       return DIncomplete(SourceID(state.tlid, id))
     else
       // CLEANUP: optimization opp
@@ -638,10 +639,10 @@ and execFn
           arglist
 
       match badArg, isInPipe with
-      | Some (DIncomplete src), InPipe _ ->
+      | Some (DIncomplete _src), InPipe _ ->
         // That is, unless it's an incomplete in a pipe. In a pipe, we treat
         // the entire expression as a blank, and skip it, returning the input
-        // (first) value to be piped into the next statement instead. *)
+        // (first) value to be piped into the next statement instead.
         return List.head arglist
       | Some (DIncomplete src), _ -> return DIncomplete src
       | Some (DError (src, _) as err), _ ->
@@ -664,7 +665,7 @@ and execFn
                 do! state.tracing.storeFnResult fnRecord arglist result
 
               return result
-          | PackageFunction (tlid, body) ->
+          | PackageFunction (_tlid, body) ->
             // This is similar to InProcess but also has elements of UserCreated.
             match TypeChecker.checkFunctionCall Map.empty fn args with
             | Ok () ->
@@ -697,7 +698,7 @@ and execFn
                       |> Dval.unwrapFromErrorRail
                       |> typeErrorOrValue Map.empty
                   }
-              // there's no point storing data we'll never ask for *)
+              // there's no point storing data we'll never ask for
               if fn.previewable <> Pure then
                 do! state.tracing.storeFnResult fnRecord arglist result
 
@@ -774,7 +775,12 @@ and execFn
         | Errors.StdlibException (Errors.FakeDvalFound dv) ->
           state.notify state "fakedval found" [ "dval", dv ]
           return dv
+        | :? DeveloperException as e ->
+          printfn "DevExn"
+          state.notify state "fakedval found" [ "dval", DStr "test" ]
+          return Dval.errSStr sourceID e.Message
         | e ->
+          printfn "Unhandled thing %A" e
           // CLEANUP could we show the user the execution id here?
           state.reportException
             state
--- a/fsharp-backend/src/LibExecution/Interpreter.fs
+++ b/fsharp-backend/src/LibExecution/Interpreter.fs
@@ -778,7 +778,7 @@ and execFn
         | :? DeveloperException as e ->
           state.notify
             state
-            $"Developer error found against {fnDesc}"
+            $"DevException against {fnDesc}"
             [ "context", "An exception was caught in fncall"
               "fn", fnDesc
               "args", args
--- a/fsharp-backend/src/LibExecution/Interpreter.fs
+++ b/fsharp-backend/src/LibExecution/Interpreter.fs
@@ -776,7 +776,15 @@ and execFn
           state.notify state "fakedval found" [ "dval", dv ]
           return dv
         | :? DeveloperException as e ->
-          state.notify state "Developer error found" [ "error", e.Message ]
+          state.notify
+            state
+            $"Developer error found against {fnDesc}"
+            [ "context", "An exception was caught in fncall"
+              "fn", fnDesc
+              "args", args
+              "callerID", id
+              "isInPipe", isInPipe
+              "error", e.Message ]
           return Dval.errSStr sourceID e.Message
         | e ->
           // CLEANUP could we show the user the execution id here?
--- a/fsharp-backend/src/LibExecution/Interpreter.fs
+++ b/fsharp-backend/src/LibExecution/Interpreter.fs
@@ -785,7 +785,7 @@ and execFn
               "callerID", id
               "isInPipe", isInPipe
               "error", e.Message ]
-          return Dval.errSStr sourceID e.Message
+          return Dval.errSStr sourceID (Exception.toDeveloperMessage e)
         | e ->
           // CLEANUP could we show the user the execution id here?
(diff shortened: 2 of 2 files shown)


StachuDotNet:
@pbiggar I'll circle back to this later in the day, but if you have any quick suggestions based on the above, please let me know. Thanks!

StachuDotNet:
To study this, I ran this test REPL

image


against both backends, and watched network traffic for the `execute_function` request.

OCaml's response:
```
{
    "result": [
        "DError",
        [
            [
                "SourceId",
                1993998850,
                2125571192
            ],
            "expecting str"
        ]
    ],
    "hash": "vzxSEFFnUP4Vs-lVZ_KwlzmWIBeiuGnOdFmwaZTUXQGLke9nYJvAMOS-cfWDcqZa",
    "hashVersion": 1,
    "touched_tlids": [
        1993998850
    ],
    "unlocked_dbs": [
        820101588,
        1315303241,
        1541193719,
        1852027420,
        2124904283
    ]
}
```

F#'s response:
```
{
    "result": [
        "DError",
        [
            [
                "SourceId",
                1993998850,
                2125571192
            ],
            "{\n  \"short\": \"expecting str 2\",\n  \"long\": null,\n  \"tipe\": \"code\",\n  \"actual\": \"\\\"teaaazaaazaast\\\"\",\n  \"actual_tipe\": null,\n  \"expected\": null,\n  \"result\": null,\n  \"result_tipe\": null,\n  \"info\": {},\n  \"workarounds\": []\n}"
        ]
    ],
    "hash": "QHrJV7T6HIu3_MAH_9yhZgYizYRRcUAvlGE2Vq1IUwFwbeTqlF7oETVu9VZK69CS",
    "hashVersion": 1,
    "touched_tlids": [],
    "unlocked_dbs": [
        820101588,
        1315303241,
        1541193719,
        1852027420,
        2124904283
    ]
}
```

StachuDotNet:
FWIW, adding the following test to `httpclient.tests` seems to not be helpful, passing against both backends:

`HttpClient.get_v5 "https://google.com" {} "test" = Test.typeError_v0 "expecting str"`

So, the issue is likely 'above' Interpreter.fs somewhere. :bulb:

pbiggar:
Thanks for the excellent write-up!

The intent here was that we know there might be exceptions in our implementation of stdlib functions. As a result, we want to be notified when they happen. Once they happen we want to triage them, possibly fixing it or creating a new version of the function. We would also want to stop the error from alerting in the future, as it has now been triaged (we often leave the old code intact and just deprecate it, so it's not useful to keep alerting us).

The old mechanism to stop alerting was [here](https://github.com/darklang/dark/blob/4de2da65b007579ec0e36ff27539007620101120/backend/libexecution/ast.ml#L816). We would catch exceptions at their throw sites and throw Code exceptions instead, which allows them to be turned into error messages at this point.

So one issue at least is that we're throwing a Developer exception instead of Library exceptions (the new name for Code exceptions). Discussing this now makes me realize that there is a lot of confusion around the exception types and that I personally don't understand them that well. I think it would be beneficial to talk through the purposes of the Dark built-in exceptions and then audit them.

So yes, I think your approach is solid except it should be a Library exception. Also, we would like to call `state.notify` about it, I think, so that its existence is recorded.

I notice as well that in the old version we would just reraise the exception if it wasn't a code exception. In the new version we rollbar it but still catch it and proceed with a DError. Do we think that's a behaviour change from the old version? Are we able to trigger it?

pbiggar:
> the encoded OCaml part tidied up:
> 
> ```
> {
>     "short": "expecting str",
>     "long": null,
>     "tipe": "code",
>     "actual": "\"fs TesT\"",
>     "actual_tipe": null,
>     "expected": null,
>     "result": null,
>     "result_tipe": null,
>     "info": {},
>     "workarounds": []
> }
> ```

We made the format for the new version just a string. However, the client should deal with either one just fine. See uses of `Decoders.exception_`. It may be there are some places where the client does not expect a plain string in a DError to hunt down.

StachuDotNet:
> we're throwing a Developer exception instead of Library exceptions 
Isn't it the Developer's "fault" that they are passing in a `string` when the fn calls for a `Dict`?
That's why `DeveloperException` made sense to me, anyway.

(certainly it'd be best if we had a type checker to prevent that, but that's beyond status quo for now)

If we _did_ migrate [this]() to be a library exception, my guess is that we'd simply `throw` it, resulting in a `Errors.StringError msg` which is already handled in `execFn`. Is that right?

> we would like to call state.notify about it
I tried this! But kept hitting runtime exceptions.
Specifically, I tried `state.notify state "Developer error found" [ ]` to keep it simple, and get these in the logs:
image

Maybe rollbar is just failing for me locally, though. Not sure, but this doesn't seem to work/help.

StachuDotNet:
Something else I noticed:
When I perform this test in the F# canvas, watching the network requests, I see two requests: `execute_function` and `get_trace_data`. The F# canvas does not yield any `get_trace_data` request. Why? I'm guessing that this is stemming from `.notify` not working?

pbiggar:
> > we're throwing a Developer exception instead of Library exceptions
> Isn't it the Developer's "fault" that they are passing in a `string` when the fn calls for a `Dict`?
> That's why `DeveloperException` made sense to me, anyway.

That makes sense to me. Though it could also be argued that it's a problem that happened when calling a function (really a type error). Makes me wonder why this isn't a type error caught when calling the function?


> (certainly it'd be best if we had a type checker to prevent that, but that's beyond status quo for now)

Yes 😢


> If we _did_ migrate this to be a library exception, my guess is that we'd simply `throw` it, resulting in a `Errors.StringError msg` which is already handled in `execFn`. Is that right?

`throw` I believe is used when there's no context about the error, unlike here. 


> > we would like to call state.notify about it
> > I tried this! But kept hitting runtime exceptions.
> > Specifically, I tried `state.notify state "Developer error found" [ ]` to keep it simple, and get these in the logs:
> > image

I could be wrong but is this working as intended? That is to say, notify is supposed to log this sort of info in the telemetry, including a stack trace - is that what you're seeing? Or is that another level of error?

pbiggar:
> Something else I noticed: When I perform this test in the F# canvas, watching the network requests, I see two requests: `execute_function` and `get_trace_data`. The F# canvas does not yield any `get_trace_data` request. Why? I'm guessing that this is stemming from `.notify` not working?

Does `.notify` cause `execute_function` to return a 500? If so then this would probably cause that.

Otherwise is might be related to the different values in `touched_tlids`?

StachuDotNet:
I think you're right about the log being 'as intended.' I was just seeing :warning: red text :warning: and reacted abruptly to it.

---

`touched_tlids` is definitely a component here, it seems!
Here's something that the F# impl. does that the OCaml impl. apparently doesn't:
It ensures that, when executing a handler, the TLID of the handler _itself_ is included in `touched_tlids`. It does this [here](https://github.com/darklang/dark/blob/537402756559a251a5153242f7f90b0a952b6bfb/fsharp-backend/src/LibRealExecution/RealExecution.fs#L63).

There are a few instances of `params.tlid :: touched_tlids` in the OCaml code, but none apparently within this code path. (I've tested this both for success and failure cases of REPLs)

If I comment out that line (the one that ensures the incoming tlid is included within `touched_tlids`), the UI stays like this:
image

I believe what happened before was:
- we got the response, including the non-empty TLIDs
- the above UI was shown
- the recent trace was fetched
- the response from ^ came back _quickly_ and "overwrote" the UI above, seeming like it was never there.

While this yields some sort of improvement, the UI still pales in comparison to the equivalent OCaml response, which provides more context:
image

pbiggar:
> While this yields some sort of improvement, the UI still pales in comparison to the equivalent OCaml response, which provides more context:

That's OK - it was a deliberate choice I made to move away from that style as it didn't really fulfil its function (it had more context, but the connection was lost). We can work on doing better here in the future, don't worry about it for now.

pbiggar:
> the response from ^ came back quickly and "overwrote" the UI above, seeming like it was never there.

I wonder if that's why it wasn't included in the OCaml response. Maybe we want to go back to that?

StachuDotNet:
> > the response from ^ came back quickly and "overwrote" the UI above, seeming like it was never there.
> 
> I wonder if that's why it wasn't included in the OCaml response. Maybe we want to go back to that?

Yeah I'm thinking that may be best.
[The commit that did this](https://github.com/darklang/dark/commit/2281a940c23bb9dfb5fab26a54e769edb72f7b4b#diff-8fa14372e7288683e53d24d630b4cbbb3b47895d10f065abd04d348735dcf905R57) doesn't have many other hints to go off of. pushing that change and seeing how tests like it. Manual testing, seems everything's the same or improved.

fsharp-backend/src/LibRealExecution/RealExecution.fs
```diff
@@ -60,7 +60,6 @@ let createState
 
     // Any real execution needs to track the touched TLIDs in order to send traces to pusher
     let touchedTLIDs, traceTLIDFn = Exe.traceTLIDs ()
-    HashSet.add tlid touchedTLIDs
```
pbiggar, line 63: Can you add a hefty comment?

StachuDotNet:
Removing `HashSet.add` like that breaks something else, shown by the failing `execute_function_works` integration test: when you go to execute a handler (via the repeat button), it doesn't return anything in `touched_tlids`, then the client doesn't fetch the trace, and so the new trace's results aren't displayed.

I experimented for a while trying to adjust where we `HashSet.add` the TLID (thinking that maybe if we only include in success cases, we'd fair well), but couldn't quickly find a way to make this work.

I'm spending a bit longer attempting to trace diffs between OCaml and F#, in hopes of finding _just_ the right places to includ the tlid, but it may be worth considering handling this in the client somehow?

pbiggar:
Yeah, sounds like a bug in the client all right.

StachuDotNet:
Given this discussion, let's split into two issues?
- (this one) Update `Interpreter.fs` to catch and properly handle `DeveloperException`s, exposing them to the client rather than yielding rollbars
- Update `client` to handle error results of `execute_function`, ensuring error text displays as expected

StachuDotNet:
Added more context to the Rollbar.

StachuDotNet:
Extracting separate client issue shortly; just gathering context

pbiggar:
Is this ready to review/merge?

StachuDotNet:
I believe so - an Exception Rollbar is no longer created in these cases. (tested against my personal Rollbar)
A Rollbar _is_ still created, but it's an `Info` Rollbar rather than an exception one.

---

I tried to write a test around this in `httpclient.tests`:
`HttpClient.get_v5 "https://google.com" {} "notadict" = Test.typeError_v0 "expecting str"`

But this passes even without the changes in `Interpreter.fs`, since the ["or anything else" branch](https://github.com/darklang/dark/blob/0a17707f45d5b5d4fac83957d1f72c24b1a3ab56/fsharp-backend/src/LibExecution/Interpreter.fs#L777) _also_ returns a `return Dval.errSStr sourceID (Exception.toDeveloperMessage e)`. Is there a reasonable way to test that an "exception Rollbar" isn't generated? If so, the test would be worth including.

pbiggar:
state.notify and state.reportException are both tracked in tests, [here](https://github.com/darklang/dark/blob/0a17707f45d5b5d4fac83957d1f72c24b1a3ab56/fsharp-backend/tests/TestUtils/TestUtils.fs#L287). I think in this case they are ignored, but we could do something different.

StachuDotNet:
Continuing to research this for now. Added some notes in #3620 but it's starting to feel like something deeper is going on, and hopefully getting closer.

PR #3613, opened 2022-03-22
Add Bytes::base64Decode

mjibson:


mjibson:
This is an F#-only implementation, and has failing and insufficient tests. They are currently producing things like:

```
expected: DResult (Error (DStr "Not a valid base64 string"))
  actual: DIncomplete SourceNone
```

and I thought it more prudent to open a PR with readable code (instead of Slack) to request help for this. I am unable to search or understand why this `DIncomplete` type is being produced.

pbiggar:
Tests by default are tested against both OCaml and F# implementations, and run against OCaml first. So the DIncomplete here is in the OCaml implementation, because there isn't an OCaml version of the function.

The solution is to add the comment ` // FSHARPONLY` at the end of the line, and that will ensure it is not run in the OCaml version.

mjibson:
Got a bit further. I'm not able to get the bytes literal syntax to work. I tried a bunch, all in the current revision. Example failure:

```
[04:33:12 ERR] tests.LibExecution.All files.Tests from bytes.tests..FSHARPONLY (line 16) (Bytes.base64Decode_v0 "Kw" = (DBytes [|43uy|])) errored in 00:00:00.0140000 <Expecto>
Expecto.AssertException: .
expected:
("Prelude+InternalException: Unsupported expression
```

pbiggar:
Apologies, I led you astray. When referring to `[| 45uy |]`, that's actually F# syntax, not Dark.

I can't think of a good way to create bytes that exists right now, except from UTF8. However, it would probably be easy to create one. You could make a function called `Test::asBytes_v0` in LibTest.fs, and call it like so `Test.asBytes_v0 [45;6;72]`.

mjibson:
RFAL

pbiggar:
Thanks! 👀

pbiggar:
It occurred to me to check that there were tests here for both types of base64, as this supports both. I think some of the tests touch on it but we should make those explicit.

Still planning to merge this once we've stopped using the OCaml execution engine entirely. Any day now...

StachuDotNet:
@pbiggar I think we can do this?

pbiggar:
Yes!

StachuDotNet:
I made a mistake in the conflict-resolution; resolving.

fsharp-backend/tests/TestUtils/LibTest.fs
```diff
@@ -217,6 +217,25 @@ let fns : List<BuiltInFn> =
       sqlSpec = NotQueryable
       previewable = Impure
       deprecated = NotDeprecated }
+    { name = fn "Test" "asBytes" 0
```
pbiggar, line 220: There should be two lines between separate function definitions

fsharp-backend/src/LibExecutionStdLib/LibBytes.fs
```diff
@@ -2,14 +2,52 @@ module LibExecutionStdLib.LibBytes
...
+            |> Convert.FromBase64String
+            |> DBytes
+            |> Ok
+            |> DResult
+            |> Ply
+        | _ -> incorrectArgs ())
+      sqlSpec = NotYetImplementedTODO
+      previewable = Pure
+      deprecated = NotDeprecated }
+
```
pbiggar, line 49: There should be two lines between separate function definitions

fsharp-backend/tests/testfiles/bytes.tests
```diff
@@ -3,6 +3,40 @@
 
 Bytes.length_v0 (String.toBytes_v0 "abcdef") = 6
 
+Bytes.base64Decode_v0 "random string" = Error "Not a valid base64 string" // FSHARPONLY
```
pbiggar, line 6: We can remove the FSHARPONLY annotations now

fsharp-backend/tests/TestUtils/LibTest.fs
```diff
@@ -217,6 +217,25 @@ let fns : List<BuiltInFn> =
...
+            match x with
+            | DInt x -> byte x
+            | _ -> incorrectArgs ())
+          |> Array.ofList
+          |> DBytes
+          |> Ply
+        | _ -> incorrectArgs ())
+      sqlSpec = NotYetImplementedTODO
+      previewable = Pure
+      deprecated = NotDeprecated }
```
pbiggar, line 238: There should be two lines between separate function definitions

pbiggar (changes requested):
A few small changes left and we can merge

pbiggar:
Great, thanks for pushing this over the line @StachuDotNet.

Thanks for the contribution @mjibson!

mjibson:
Yay, thanks for the followup.

Issue #3589, opened 2022-03-17
Investigate and resolve blazor load error


pbiggar:
When loading blazor, many of the files it needs will not load in the browser. The browser stops being able to make a connection to the server. This only happens in production. It is most commonly reproduced by loading and then doing another load from a clean cache.

I managed to reproduce the error:

```
Uncaught exception Blazor worker failure TypeError: Failed to fetch
    at loadResource (blob:https://darklang.com/dacea6f8-981c-4a6a-9522-f78faeed6aeb:738:24)
    at blob:https://darklang.com/dacea6f8-981c-4a6a-9522-f78faeed6aeb:732:11
    at Array.map (<anonymous>)
    at Object.loadResources (blob:https://darklang.com/dacea6f8-981c-4a6a-9522-f78faeed6aeb:731:39)
    at createEmscriptenModuleInstance (blob:https://darklang.com/dacea6f8-981c-4a6a-9522-f78faeed6aeb:220:52)
    at blob:https://darklang.com/dacea6f8-981c-4a6a-9522-f78faeed6aeb:772:25 undefined undefined undefined
e.<computed> @ appsupport-0aab008024f907c251bfe9927856eeebed075655f5139be469f7af1ceaf66459.js:179373
```

![image](https://user-images.githubusercontent.com/181762/158612539-68bd0423-d2e5-4a95-b32f-362e4d7dddaa.png)

And it seems to happen when part of the load time's out:

![image](https://user-images.githubusercontent.com/181762/158618815-e711f9e7-6526-4225-a697-f6b0f6646726.png)

Unclear why these loads time out, but this happens in about 1/4 of the times I refresh.

_Originally posted by @pbiggar in https://github.com/darklang/dark/issues/3254#issuecomment-1069216438_

pbiggar:
Some investigation:
- the items being requested are uploaded to the bucket and configured properly
- no observable differences between things that succeed and things that fail, at the bucket level
- taking the failing url and fetching it via curl succeeds
- the load balancer does not have any logs for these requests
- the error in Chrome is ERR_CONNECTION_CLOSED
- all requests (success and failure) to the static assets load balancer are "h3" (http3)

ERR_CONNECTION_CLOSED means it wasn't able to make a connection:
- chrome has a connection limit?
- ddos protection at google?
- my os/router/isp/other doesn't like that many connections of that size/content-type/etc?

pbiggar:
I don't have any clues here so I'll see who I can ask.

[...]

PR #3618, opened 2022-03-22
Fix content types for blazor

pbiggar:
I'm investigating [why fetching blazor assets](
https://github.com/darklang/dark/issues/3589) from the CDN causes Chrome to not connect anymore. I have no idea why this would be, but one idea I had is that perhaps the content-types are weird and triggering some sort of built-in DDOS protection. Not a great guess, but it does make all the content-type headers right instead of weird, so it's a net improvement anyway.

This was tested by pointing at a different bucket and comparing against the darklang-static-assets bucket.


pbiggar:
I'm going to run this locally without the `-n` to overwrite existing files and their mimetypes, after this deploys.

StachuDotNet (approved):

PR #3625, opened 2022-03-23
Load testing

pbiggar:
I've started load testing the new F# implementation, using for now:

`./wrk -t 8 -c 5000 -d 5s https://trydarkfsharp.builtwithdark.com/fizzbuzz --timeout 4s`

I'm seeing a lot of 500 and 502 errors, along with an alert in the nginx logs that the container doesn't have enough workers. Hopefully this should fix this and we'll see where we are then.

Also, it seems that after switching to an unprivileged version of nginx, we didn't use the default config from that container. So this switches to that, and then makes the appropriate changes.


pbiggar:
I have a stray file committed here, removing before I commit. Just broke my dev environment and I'll do this after.

Issue #3630, opened 2022-03-24
try removing the nginx reverse-proxy for BwdServer


pbiggar:
We have no real use for the nginx proxy for bwdserver, and it just provides another thing to configure and maintain. We need nginx for the apiserver and we used to have the apiserver and the bwdserver in the same process.

Let's try without it.

pbiggar:
If we stick with this we'll need:
- DDOS protection (try cloud armor) as we used nginx to allow load-shedding
- a redirect for one customer

pbiggar:
Need a test with what happens with incorrect content-length:
- [ ] too long
- [ ] too short

pbiggar:
- [ ] test the xforwarded for

pbiggar:
There was a bug I couldn't figure out when developing this. It turns out Kestrel really does not like you to have the wrong content-length. It manifested as a full 400 error being returned by Kestrel AFTER a response had already happened (just jammed right on the end of the HTTP output stream, and in some cases in the middle of it).

I speculate that nginx really didn't care that much which is why we didn't notice the tests were wrong.

pbiggar:
- [ ] Remove "Kestrel" header

pbiggar:
- [ ] check assets passed through via serveLatest work

pbiggar:
> * [x]  test "host" header in production - does it have the right port

Yes, no port exists. (in the url field either)

pbiggar:
> * [ ]  check assets passed through via serveLatest work

Being handled in #3658

pbiggar:
Closing this - the remaining work in covered in #3664.

PR #3635, opened 2022-03-25
Remove nginx bwdserver

pbiggar:
Part 1 of https://github.com/darklang/dark/issues/3630

This removes nginx from the BwdServer. We needed nginx for a couple of reasons in the past that are no longer necessary or can be handled in other (better) ways:

- we needed some form of way to protect the app from expensive requests without processing them or recompiling and redeploying the whole thing. nginx gave us a quick way to update the config and restart the containers - instead we can use cloud armor, which is better at it

- we needed to support requests with missing content-lengths, as users like to make POST requests from curl sometimes and getting the right content-length is annoying. Now kestrel supports omitting the content-length header

- we needed to proxy requests to other servers for darklang.com, but now we've split the ApiServer from the BwdServer so it's not as important

- we needed to do gzip on large returned files. Admittedly nginx is better at that than kestrel but it can still be done in kestrel

- ultimately, i want to remove it as it's "yet another thing to understand" and I'd rather remove a complex step from our infra

This removes nginx from bwdserver only, including removing the deployment/k8s config files.

One major problem was that the test suite had a lot of files with the wrong content-length, which nginx papered over. So a lot of this PR is making sure the content-length is right in all the test files, and adding support to the test suite to make sure of that.


pbiggar:
The predeployment checks CI job failed - that's fine, there's a different config and we'll just have bwdserver down for a few minutes. Since it's prerelease it doesn't matter.

StachuDotNet (approved):
Great!

pbiggar:
- [ ] do shipit manual apply just before merging

PR #3636, opened 2022-03-28
Fuzztests: fixes and reorg

### LINKED ISSUE #3500: Resolve NodaTime-related fuzz test issues
Fuzztests using NodaTime types are failing to run locally due to the NodaTime migration:

```
FuzzTests.known good.jwtJson.roundtrip jwt v1 errored in 00:00:00.0170000 <Expecto>
System.Exception: The type NodaTime.LocalDateTime is not handled automatically by FsCheck. Consider using another type or writing and registering a generator for it.
```

I've found this template for a .Instant generator:
- http://www.fssnip.net/nK/title/A-quick-NodaTimeInstant-generator-for-FsCheck

but have not yet tested this

StachuDotNet:
WIP.

- errant `|> debug` was causing tests to fail
- `NodaTime` usage was causing tests to fail. Copying these to every generator got tiresome quickly, so started to create a reusable set of Generators
- it seems we need a `LibBackend.init` to run properly now
- opinion: it's always best to provide an explicit generator, even if it's the default one
- some tests still are eventually failing on odd characters; likely that `Generators.ocamlSafeString` needs adjustment
- idea: expose the ability to change `maxTests` via CLI so we don't have to recomplie every time you change it in the Utils file
- idea: run some of the tests in CI, likely with a lower maxTest #
- idea: benchmark both our generators and our property tests
- finding: password tests are _very_ slow. (8tests/second).
- finding: HttpClient.tests fail eventually, on "&=&". Check this out further.
- Others vary in duration. Once we have the above-mentioned benchmarks
- idea: consider the "complexity" of a ;type that we're testing against. Given the type's complexity, what's a "reasonable" number of tests to run?
- I will remove the "Playground.fs" file prior to merge

will resolve #3500

### DIFF
--- a/fsharp-backend/src/LibExecution/OCamlTypes.fs
+++ b/fsharp-backend/src/LibExecution/OCamlTypes.fs
@@ -795,7 +795,7 @@ module Convert =
 
       let pipeTarget = ORT.EPipeTarget(gid ())
       let fnCall = ORT.EFnCall(id, name, pipeTarget :: actualArgs, rt2ocamlSter rail)
-      ORT.EPipe(pipeID, pipeStart @ [ fnCall ]) |> debug "converted to pipe"
+      ORT.EPipe(pipeID, pipeStart @ [ fnCall ])
     | RT.EFQFnValue _
     | RT.EApply (_, _, _, _, _) ->
       // these shouldn't happen in practice at the moment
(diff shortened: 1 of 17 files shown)


pbiggar:
> idea: run some of the tests in CI, likely with a lower maxTest #

I'm not a fan of this approach to fuzzing. The purpose of CI is to get confidence that ones PR works. It's not helpful to learn of new fuzzing errors to unrelated code in ones PR.

We could run fuzzing full-time on another machine thoug.

pbiggar:
Is this no longer needed?

StachuDotNet:
> Is this no longer needed?

Correct - created the other as a replacement to this, then forgot. thanks

Issue #3617, opened 2022-03-22
Address exception mess


pbiggar:
Exceptions have a few problems:

- I split the old Code exception into LibraryException and DeveloperException. This is confusing - it confused me a lot. DeveloperException should be renamed to CodeException. LibraryException should be renamed to KnownIssueException. Both should be caught in a callFn.

- Errors.fs has a number of different specialized exceptions that are caught in Interpreter.fs, such as FakeValFoundInQuery and StdlibException. These are because the function name wasn't available for the exception at the time I wrote it, but it is now available in state.executingFnName

- Errors.throw throws strings all the time. Why is this different from Code? Why are there specialized exceptions in other places?

- Errors.fs has a bunch of functions to get consistent error messages for classes of functions. this is a good pttern and one we should double down on.

Issue #3620, opened 2022-03-22
F# server doesn't persist failed fn/handler executions


StachuDotNet:
(WIP - still adjusting the wording of this PR, and researching)
Title likely needs some adjustment.

This applies only to the F# backend - the issue stems from a difference between the F# and OCaml backends, but we believe that the F# backend is doing the right thing, and that the client should handle this better. This issue has been extracted from a conversation in #3610

--- 
Issue:
Create a handler that will result in an error, and play it.
We'd expect the new error to pop up now, but instead we see `This trace returns <Incomplete>` and `Click Play to execute function`

https://user-images.githubusercontent.com/906686/159572072-d41d6399-c822-4db4-ae6f-9ffd894d48ff.mp4

--- 
Context:
When I execute this function, an `execute_function` API call is made. The response includes both the result, as well as any `touched_tlids`:
```
{
    "result": [
        "DError",
        [
            [ "SourceId", 2077505382, 714461266 ],
            "expecting str"
        ]
    ],
    "hash": "7MvOlhz7MMr-Ce95_UJdRlst14TX8WHcijuk5Hr5O5Pr4zx1mAF1HRkEaLx4t9h6",
    "hashVersion": 1,
    "touched_tlids": [
        2077505382
    ],
    "unlocked_dbs": [ 820101588, 1315303241, 1541193719, 1852027420, 2124904283 ]
}
```

The F# server makes a point of always including the relevant TLID in the `touched_tlids` bag, while the OCaml server in some cases doesn't include it.

If the `touched_tlids` contains anything, subsequent requests are made to the API, to route `get_trace_data`.

Once that data is loaded, the _previous_ execution seems to paint over the quickly-shown error message, with the messaging we display in the video.

StachuDotNet:
New discovery: if I execute the REPL even once against the OCaml server, I can then see the error message in the F#-loaded client. (though, it's finnicky, and I cannot produce _new_ errors - only load the OCaml-produced ones)

StachuDotNet:
Small update here, though I think I'm close to finishing the picture:

When `execute_function` results in a failure,
- the OCaml version successfully stores the results in the DB (`function_results_v3`, `function_arguments`)
- the F# version does not store those results in the DB

So, this is a backend issue, I believe. Currently tracing where the F# server is failing to persist these. Once the backend is properly storing this, the load of traces `get_trace_data` would result in the right error being displayed, and we shouldn't need to adjust anything in the client code.

[...]

PR #3638, opened 2022-03-29
Ensure traces are stored if there are errors

pbiggar:
I'm not done with this yet but would like a directional review.

It's hard to see from the diff, but there's almost no change here:
- the try catch is moved from around all function calls to just the StdLib fn call
- the catch logic is moved into a function
- added type checking for the results of functions

I'm going to continue looking at this to add tests and some defense in depth:
- [ ] tests for the feature flag exception (not caught as is)
- [ ] tests for bad types returning from:
  - [ ] package manager functions
  - [ ] user functions
- [ ] tests for error rails returning from:
  - [ ] package manager functions
  - [ ] user functions
- [ ] reduce the size of the exception handler by folding in #3617 

My main question is when to merge. I'm tempted to merge here or after adding tests. wdyt?

### DIFF
--- a/fsharp-backend/src/LibExecution/Interpreter.fs
+++ b/fsharp-backend/src/LibExecution/Interpreter.fs
@@ -188,6 +188,7 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
 
       let! conditionResult =
         // under no circumstances should this cause code to fail
+        // FSTODO this can fail
         try
           eval state st cond
         with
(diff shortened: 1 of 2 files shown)


StachuDotNet (approved):
Great, this makes sense and is a pretty small incision - I'd say this is OK to merge as-is, especially with the [context of the old ast.ml](https://github.com/darklang/dark/issues/3620#issuecomment-1081968420).

pbiggar:
Ok, I'll add everything else to an issue and then merge as is.

PR #3640, opened 2022-03-30
FuzzTest fixes and cleanup

### LINKED ISSUE #3500: Resolve NodaTime-related fuzz test issues

StachuDotNet:
It's been several weeks since FuzzTests have been run; this PR offers several fixes, dev tools, and code clarity improvements.

- the `--fscheck-max-tests` flag built into expecto/fscheck does not work, which can make testing a bit cumbersome, having to wait for compilations when you want to change how many tests are run. While a bit noisy, this PR offers a solution to enable this feature for now.
- will resolve #3500; many FuzzTests fail quickly due to NodaTime `Instant` and `LocalDateTime` types not having default generators
- several StdLib fns have `previewable` status that may be adjusted (later, as CLEANUP)
- `Previewable` descriptions were inexact
- many fuzztests have regressed since last checked, and have been now moved to the `knownBad` bucket of tests


fsharp-backend/src/LibExecutionStdLib/LibDate.fs
```diff
@@ -222,6 +222,7 @@ let fns : List<BuiltInFn> =
           Ply(DDate(DDateTime.T(now.Year, now.Month, now.Day, 0, 0, 0)))
         | _ -> incorrectArgs ())
       sqlSpec = NotQueryable
+      // CLEANUP mark as impure
```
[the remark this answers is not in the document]
StachuDotNet: Fix casing of comment

fsharp-backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -744,13 +744,22 @@ module Package =
...
   | Pure
 
-  /// Save their results. We can preview these safely
+  /// Can either:
+  /// - only be run on the server but is actually pure (e.g. `Crypto::md5`)
+  ///   Generally, these are functions that aren't compilable to WASM.
+  /// - be safely run without side effects but is not pure (e.g. `Date::now`)
+  ///
+  /// We should save the results.
   | ImpurePreviewable
```
pbiggar, line 757: Seems I steered you wrong here as my understanding was off.

There are 5 functions currently marked ImpurePreviewable - those are all crypto functions, which are technically pure and can be moved over to Pure once the ocaml is turned off.

Then there are functions which can safely be run but not pure, but none of them are marked as ImpurePreviewable.

To be usable in the editor/wasm, a function must be both Pure and shipped as part of the Wasm build (eg excludes BackendOnlyStdLib).

fsharp-backend/src/LibExecutionStdLib/LibDate.fs
pbiggar: In a cleanup, this can be marked ImpurePreviewable

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
```diff
@@ -23,7 +23,7 @@ module OCamlInterop = LibBackend.OCamlInterop
 module G = Generators
 
 /// Ensure we only work with OCaml-friendly floats
-let filterFloat (f : float) : bool =
+let isValidOCamlFloat (f : float) : bool =
```
pbiggar, line 26: I suspect this can be removed. I think it was about passing them to ocaml via json, and I'm pretty sure that's supported now.

pbiggar: Fyi these are all "valid" floats (floats are floats whereever they are)

fsharp-backend/tests/FuzzTests/FQFnName.FuzzTests.fs
```diff
@@ -32,7 +32,7 @@ let modName : Gen<string> = nameGenerator [ 'A' .. 'Z' ] G.alphaNumericString
 let fnName : Gen<string> = nameGenerator [ 'a' .. 'z' ] G.alphaNumericString
 
 type Generator =
-  static member SafeString() : Arbitrary<string> =
+  static member String() : Arbitrary<string> =
```
pbiggar, line 35: Could these be included via inheritence as well, like the NodaTime ones (or is there only single inheritence, I forget)?

fsharp-backend/tests/FuzzTests/OCamlInterop.FuzzTests.fs
```diff
@@ -155,7 +161,7 @@ module Roundtrippable =
     static member Dval() : Arbitrary<RT.Dval> =
       Arb.Default.Derive() |> Arb.filter (DvalReprInternal.isRoundtrippableDval true)
 
-  let roundtrip (dv : RT.Dval) : bool =
+  let canRoundtrip (dv : RT.Dval) : bool =
```
pbiggar, line 164: I'm not sure "canRoundtrip" is a good name here. It implies some properties of the dval that would allow it to be roundtripped, while these functions actually do the roundtrip and report success.

fsharp-backend/tests/FuzzTests/Passwords.FuzzTests.fs
```diff
@@ -21,7 +21,7 @@ module RT = LibExecution.RuntimeTypes
 module G = Generators
 
 type Generator =
-  static member SafeString() : Arbitrary<string> =
+  static member String() : Arbitrary<string> =
```
pbiggar, line 24: Why did you rename these from SafeString?

fsharp-backend/tests/FuzzTests/HttpClient.FuzzTests.fs
```diff
@@ -74,16 +73,23 @@ let queryToEncodedString (q : List<string * List<string>>) : bool =
   DvalReprExternal.queryToEncodedString q
   .=. (OCamlInterop.paramsToQueryString q).Result
 
-let tests =
-  let test name fn = testProperty typeof<Generator> name fn
-  testList
-    "HttpClient"
-    [ test "dvalToUrlStringExn" dvalToUrlStringExn // FSTODO: unicode
```
pbiggar, line 82: Was this FSTODO handled?

pbiggar (changes requested):
Nice fixes!

fsharp-backend/tests/FuzzTests/FQFnName.FuzzTests.fs
StachuDotNet: Yes - will do

fsharp-backend/tests/FuzzTests/Passwords.FuzzTests.fs
StachuDotNet: I did it to prevent [this problem](https://github.com/darklang/dark/blob/4237ff6b227abe87c2580d0072377630633d368c/fsharp-backend/tests/FuzzTests/HttpClient.FuzzTests.fs#L35-L42). Which of these two generators will be used?

It's really easy to write conflicting generators unless we decide to consistently name the members to be of the type that we're overriding a generator for. Happy to just "keep this in mind" always if you think that's best, but this seemed like a decent way to prevent the problem.

fsharp-backend/tests/FuzzTests/HttpClient.FuzzTests.fs
StachuDotNet: No, my mistake - placing back in.

fsharp-backend/tests/FuzzTests/OCamlInterop.FuzzTests.fs
StachuDotNet: Good point - trying `roundtripsSuccessfully` - ok?

fsharp-backend/src/BackendOnlyStdLib/LibStaticAssets.fs
```diff
@@ -117,6 +117,7 @@ let fns : List<BuiltInFn> =
           SA.url state.program.canvasName deployHash SA.Short |> DStr |> Ply
         | _ -> incorrectArgs ())
       sqlSpec = NotQueryable
+      // CLEANUP may be marked as ImpurePreviewable
```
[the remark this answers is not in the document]
pbiggar: I guess we can technically consider the canvas to be an input, even if the input doesn't come from the developer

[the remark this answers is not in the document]
StachuDotNet: OK agreed; will leave as-is, then.

fsharp-backend/src/LibExecution/RuntimeTypes.fs
StachuDotNet: My current impression is that `ImpurePreviewable` might fall off entirely? In any case, going back-and-forth on this so will vastly simplify comment here to not spend too much time on it

fsharp-backend/tests/FuzzTests/Passwords.FuzzTests.fs
pbiggar: So what's the solution to the problem? Can you add a comment describing the problem and solution?

fsharp-backend/src/LibExecution/RuntimeTypes.fs
StachuDotNet: "Can be safely previewable, and we should save the results." feels pretty safe.

StachuDotNet: OK I gave the wording on these another go; feels better to me,  anyway :)

fsharp-backend/tests/FuzzTests/Passwords.FuzzTests.fs
StachuDotNet: The solution is consistent naming to prevent conflicting `static member`s.
One solution would be to update the coding guides - I've just pushed an adjustment there.

I haven't yet thought of a better solution - open to ideas

fsharp-backend/src/BackendOnlyStdLib/LibStaticAssets.fs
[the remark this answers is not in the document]
pbiggar: I meant that to mean it is pure.

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
pbiggar: Especially since `NaN <> NaN`

fsharp-backend/tests/FuzzTests/Passwords.FuzzTests.fs
pbiggar: Oh so the idea is name them all String so we don't accidentally have two types of string?

StachuDotNet: yes, though I suppose that strategy fails a bit when we define generators for tuples - what's the naming convention then? One name followed by the other? (e.g. in ExecutePureFunctinos)

Issue #3639, opened 2022-03-29
Followups to recent Interpreter.fs work


pbiggar:
Combines remaining issues from #3638, #3620

There were differences between ast.ml and Interpreter.fs that we did not have tests for.

- [x] tests for the feature flag exception (not caught as is)
- [x] are there tests for exceptions/errors in calls to package manager functions?
  - [x] simple wrong number of arguments
  - [x] with/without unwrapFromErrorRails
- [x] are there tests for exceptions/errors in calls to user functions functions?
  - [x] simple wrong number of arguments
  - [x] with/without unwrapFromErrorRails
- [ ] trim Interpreter.handleException by looking at #3617

pbiggar:
Solved. in #3644

PR #3644, opened 2022-03-31
Interpreter function call followups

pbiggar:
Solves the function call parts of #3639. I decided to call a stop here and to do the exception work from #3617 in another PR.

This is a followup to #3638, where I noticed that calling functions was largely untested. This adds tests for functions and package manager functions.

It found one actual problem, which is that function names were improperly printed in errors.

Also lots of testing fixes:
- does a better job of printing exception metadata on errors
- catch exceptions which could potentially happen in feature flags (couldn't find a way to trigger this though)
- throw exceptions during tests and add a way to check if exceptions happened
- add types to test function definitions
- add package function defintions

Some broader change:
- I thought I had banned Option.unwrapUnsafe and friends, but turns out I didn't. Now I do.
- change TStr to be printed as `Str` instead of `String`. That matches OCaml and lowers the risk of breaking something obscure.


fsharp-backend/tests/testfiles/language.tests
```diff
@@ -64,16 +64,19 @@ toString 5 = "5"
...
 flag "test" null "old" "new" = "old"
 flag "test" 5 "old" "new" = "old"
 flag "test" blank "old" "new" = "old"
 flag "test" (Test.typeError_v0 "test") "old" "new" = "old"
 flag "test" (List.head_v1_ster []) "old" "new" = "old"
 flag "test" {x = true} "old" "new" = "old"
 flag "test" [true] "old" "new" = "old"
+flag "test" (Test.errorRailValue "error") "old" "new" = "old"
+flag "test" (Test.negativeInfinity "extra arg") "old" "new" = "old"
+flag "test" (Test.raiseException "message") "old" "new" = "old" // EXPECTED_EXCEPTION_COUNT: 1
```
StachuDotNet, line 79: handy!

StachuDotNet:
The [readme](https://github.com/darklang/dark/blob/paul/interpreter-followups/fsharp-backend/tests/testfiles/README.md) could use an update for package and user fns, as well as typing constraint updates

StachuDotNet (approved):
Looks good - just readme that should be updated

PR #3645, opened 2022-04-01
Clean up exception mess

pbiggar:

We have a lot of exceptions that all attempted to do some portion of the job of exceptions in stdlib:
1. DeveloperException
2. LibraryException
3. Errors.StdlibException.IncorrectArgs
4. Errors.StdlibException.StringError
5. Errors.StdlibException.FakeDvalFound
6. Errors.DBQueryException
7. Errors.FakeValFoundInQuery

Ostensibly they all had a different reason for existing, but the reasons were confused and didn't hold a lot of weight. For example, LibraryExceptions were for "known exceptions", but any exception we throw ourselves is by definition known, so how does this differ from the other exceptions we created for the purpose.

I tried combining them into a single exception, but there are actually 2 different use cases:
- raising a string exception (1,2,4,6)
- exiting from the current function execution. This has the subcategories of
  - return to the callsite to create a good error message (3)
  - we have found a fakeDval and wish to return it instead of continuing (5,7)

For the second use case, it was impossible to craft a good `.Message`. This broke a lot of tests in ways that were very difficult to fix, which was what indicated that these did in fact need to be separate. 

I added `CodeException`, and put all string exceptions into it.

I broke out `FakeDvalFoundException` (combining it with `FakeValFoundInQuery`)

I also broke out IncorrectArgs into its own exception

### DIFF
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
@@ -569,12 +569,10 @@ that's already taken, returns an error."
                 | Some Canvas.AllOrigins -> "*" |> DStr |> Some |> DOption
                 | Some (Canvas.Origins os) ->
                   os |> List.map DStr |> DList |> Some |> DOption
-              try
-                let! c = Canvas.getMeta (CanvasName.create host)
-                let! cors = Canvas.fetchCORSSetting c.id
-                return corsSettingToDval cors
-              with
-              | e -> return DError(SourceNone, Exception.toDeveloperMessage e)
+
+              let! c = Canvas.getMeta (CanvasName.create host)
+              let! cors = Canvas.fetchCORSSetting c.id
+              return corsSettingToDval cors
             }
           | _ -> incorrectArgs ())
       sqlSpec = NotQueryable
@@ -606,7 +604,7 @@ that's already taken, returns an error."
                     os
                     |> List.map (function
                       | DStr v -> v
-                      | _ -> Exception.raiseGrandUser "Invalid origin string")
+                      | _ -> Errors.throw "Invalid origin string")
                     |> Canvas.Origins
                     |> Some
                     |> Ok
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
@@ -604,7 +604,7 @@ that's already taken, returns an error."
                     os
                     |> List.map (function
                       | DStr v -> v
-                      | _ -> Errors.throw "Invalid origin string")
+                      | _ -> Exception.raiseCode "Invalid origin string")
                     |> Canvas.Origins
                     |> Some
                     |> Ok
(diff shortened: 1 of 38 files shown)


fsharp-backend/src/LibBackend/SqlCompiler.fs
```diff
@@ -19,7 +19,10 @@ module DvalReprExternal = LibExecution.DvalReprExternal
 module RuntimeTypesAst = LibExecution.RuntimeTypesAst
 module Errors = LibExecution.Errors
 
-let error (str : string) : 'a = raise (Errors.DBQueryException str)
+let errorTemplate =
+  "You're using our new experimental Datastore query compiler. It compiles your lambdas into optimized (and partially indexed) Datastore queries, which should be reasonably faster.\n\nUnfortunately, we hit a snag while compiling your lambda. We only support a subset of Dark's functionality, but will be expanding it in the future.\n\nSome Dark code is not supported in DB::query lambdas for now, and some of it won't be supported because it's an odd thing to do in a datastore query. If you think your operation should be supported, let us know in #general.\n\nError: "
```
StachuDotNet, line 23:
```suggestion
  "You're using our new experimental Datastore query compiler. It compiles your lambdas into optimized (and partially indexed) Datastore queries, which should be reasonably faster.\n\nUnfortunately, we hit a snag while compiling your lambda. We only support a subset of Dark's functionality, but will be expanding it in the future.\n\nSome Dark code is not supported in DB::query lambdas for now, and some of it won't be supported because it's an odd thing to do in a datastore query. If you think your operation should be supported, let us know on Slack in #general.\n\nError: "
```

pbiggar:
I've added more changes:

- inlined handleException
- no longer show an error message for InternalException
- switch some InternalExceptions to CodeException (failing internal logic is still InternalException, but failing input returns CodeException)
- continue to show an error message for GrandUserException
- only notify for FakeDvals exceptions in NotDeprecated functions
- fix a try that wasn't dereferencing its task

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
```diff
@@ -569,12 +569,10 @@ that's already taken, returns an error."
                 | Some Canvas.AllOrigins -> "*" |> DStr |> Some |> DOption
                 | Some (Canvas.Origins os) ->
                   os |> List.map DStr |> DList |> Some |> DOption
-              try
-                let! c = Canvas.getMeta (CanvasName.create host)
-                let! cors = Canvas.fetchCORSSetting c.id
-                return corsSettingToDval cors
-              with
-              | e -> return DError(SourceNone, Exception.toDeveloperMessage e)
```
StachuDotNet, line 577: Why can we remove the try/with here?

StachuDotNet (approved):
Great - no concerns here beyond the question I have about a try/with being removed. I couldn't figure out why that's OK.

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
pbiggar: Turns out I did in a later commit!

PR #3650, opened 2022-04-05
Add types to shipit

pbiggar:
I started adding a very small feature to shipit and did not feel very confident doing it, so I added types to it. This makes no change to functionality, but pretty aggressively types everything.

My plan is to merge this then do a refactor to remove dups and use real types instead of dicts everywhere.

PR #3652, opened 2022-04-05
Move namespace to shipit top-level

pbiggar:
Again, no change to functionality, just moving namespace to top-level and confirming that the namespace matches any other namespace listed.

This is based on #3650, the diff will be cleaner once that merges.


services/honeycomb-agent/shipit.yaml
```diff
@@ -1,13 +1,16 @@
-# honeycomb-agent is deployed via helm. I don't know a great deal about it, but it seems we can update the deployment with
+# honeycomb-agent is deployed via helm. I don't know a great deal about it, but it
+# seems we can update the deployment with
+#
```
StachuDotNet, line 3: Slightly confusing with the newline - I assume the 'with' is supposed to continue with the `help upgrade ...` bit?

StachuDotNet (approved):
Looks good - feels like many of these changes of these changes will flatten with #3653 being merged, so may review again after that merges if relevant.

services/honeycomb-agent/shipit.yaml
pbiggar: Will handle in followup.

PR #3653, opened 2022-04-06
Actually make a new bwdserver deployment

pbiggar:
This makes a new BwdServer service, ingress, and deployment, all in the darklang namespace.

The purpose is twofold:
- safely add a cloud armor config (without risking doing something to the existing builtwithdark.com ingress)
- provide something where users can test against without converting their app over (here they can test with a new IP address and specifying the host)

This does not yet add cloud armor, which I'm looking into and plan to do later in another PR.

As of now the IP address is 34.120.181.153


services/bwdserver-deployment-darklang/bwdserver-deployment.template.yaml
```diff
@@ -2,13 +2,13 @@ apiVersion: apps/v1
...
   name: bwdserver-deployment
-  namespace: default
+  namespace: darklang
   annotations:
     kubernetes.io/change-cause: "{ARG:CHANGE_CAUSE}"
 
 spec:
   revisionHistoryLimit: 10
-  replicas: 3
+  replicas: 1 # FSTODO
```
StachuDotNet, line 11: Is this to "see how we do with only 1" and update if necessary?

scripts/deployment/shipit
```diff
@@ -872,96 +858,147 @@ properties:
...
-      bad(f"Error in {dir}/shipit.yaml:\n  {e}")
+  k8s = config['k8s']
 
-    k8s = config['k8s']
+  namespace = k8s.get("namespace")
+  if namespace is None:
+    bad(f"missing namespace field in {dir}/shipit.yaml")
+  else:
+    if not re.match(r"^([-a-z0-9]+)$", namespace):
+      bad(f"namespace \"{namespace}\" not in valid")
```
StachuDotNet, line 928:
```suggestion
      bad(f"namespace \"{namespace}\" not valid")
```

services/bwdserver-deployment-darklang/bwdserver-deployment.template.yaml
```diff
@@ -0,0 +1,362 @@
...
+          # ---------------------
+          imagePullPolicy: Always
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - NET_RAW # https://docs.bridgecrew.io/docs/bc_k8s_27
+            privileged: false
+            # this is expected to run as root. We're getting rid of the container
+            # soon so not worth fixing.
```
StachuDotNet, line 180: Can you elaborate here?
Is it _currently_ running as root and we want to not run as root? and the comment here is that it's not worth fixing the existing/old container?

services/bwdserver-deployment-darklang/bwdserver-deployment.template.yaml
```diff
@@ -0,0 +1,362 @@
...
+          # ---------------------
+          # Environment
+          # ---------------------
+          envFrom:
+            - configMapRef:
+                name: "{VERSIONED-CONFIGMAP:app-config}"
+          env:
+            - name: DARK_CONFIG_RUNNING_IN_GKE
+              value: "true"
+            # database - we don't use it, but we need to connect to it
```
StachuDotNet, line 235: Why do we need to connect to it?

services/bwdserver-deployment-darklang/bwdserver-service-backendconfig.yaml
```diff
@@ -0,0 +1,22 @@
...
+    port: 30212 # the nodeport from bwdserver-service
+    # Host is not allowed here, but it has been manually set on this healthcheck so
+    # that the server can recognize this is a healthcheck and not a granduser request
+    # host: gce-ingress-healthcheck
+    healthyThreshold: 1
+    unhealthyThreshold: 1
+  logging:
+    # Changing this setting didn't seem to do anything - I needed to Edit it in the
+    # GCP web console (or presumably using the gcloud command line utilily)
+    # Note that only a change is status is logged.
```
StachuDotNet, line 21:
```suggestion
    # Note that only a change in status is logged.
```
?

StachuDotNet (approved):
Reviewed to the best of my current ability, though certainly a bit kubernetes-lost.

services/bwdserver-deployment-darklang/bwdserver-deployment.template.yaml
pbiggar: the legacyserver connection to the DB is deep deep in Libbackend, and cutting it out is very hard. But why bother when we're going to delete it.

services/bwdserver-deployment-darklang/bwdserver-deployment.template.yaml
pbiggar: This is copied from the old config. I did a pass to make it as secure as possible a while back, including getting as many containers as possible not running as root. However, legacyserver expects to run as root because we run everything in ocaml using `sudo`.

Issue #3556, opened 2022-03-09
Fuzz all pure functions


pbiggar:
Mark LibJwt, x509 and crypto functions as ImpurePreviewable to run in the fuzzer

StachuDotNet:
Just to make sure I understand correctly, would this be an appropriate rephrase?
"Fuzztest LibJwt, x509 and crypto functions. You'll need to mark them as ImpurePreviewable to run in the fuzzer"?

StachuDotNet:
[Here](https://dynalist.io/d/xYMggOIEkXlxNh01DVXlNG4_#collapse&theme=dark)'s an analysis of the Dark fns, noting which ones can/can't be fuzztested.

I gathered the list by writing a [a variant of `DarkInternal::allFunctions`](https://gist.github.com/StachuDotNet/3b0ef98e09d71fac3e228dd8ab58592e) locally which includes `Previewable` status, and manipulating the resultant JSON into the Dynalist doc, organizing fns where they seemed appropriate.

Some notes:
- I assume `toForm` should be ignored, as it's marked as for demo only
- `Date::today` is currently marked as `Pure` - I'm planning as marking this as `Impure`
- while _most_ of `StaticAssets` can't be fuzztested, several fns (`StaticAssets::baseUrlFor`, `StaticAssets::fetch`, `StaticAssets::urlFor`) should be fuzz-testable w/r/t ocaml interop, at least. Would those be appropriate to be marked as `ImpurePreviewable`?
- I haven't yet tested `X509::pemCertificatePublicKey`
- I'm still a bit confused why some functions are `ImpurePreviewable` rather than `Pure` (e.g. `Crypto::md5`); does `ImpurePreviewable` really mean "pure, but let's save it because it's a bit costly?" or am I misinterpreting?

If there are any fns in that document out of place, please let me know as I'm working my way through them.

pbiggar:
> I'm still a bit confused why some functions are ImpurePreviewable rather than Pure (e.g. Crypto::md5); does ImpurePreviewable really mean "pure, but let's save it because it's a bit costly?" or am I misinterpreting?

"Pure" doesn't actually mean pure, it means "this can be run in the editor via js_of_ocaml". Some things were impure in the OCaml version as they were not runnable in the editor, mostly because of their use of C/native bindings.

"ImpurePreviewable" means "this can only run on the server but is actually pure" (eg `Crypto::md5`) or "you can safely run this without side-effects, but it is not pure" (eg `Date::now`).

[...]

PR #3654, opened 2022-04-06
Fuzztest and fstodo questions

StachuDotNet:
This PR contains some WIP, but isn't intended to be merged.
Rather, it's intended as a space to ask some questions around several FSTODOs, in a way that may be more convenient than an Issue. Right after creating the PR, I'll follow up with in-line questions.

Currently feeling a bit unproductive on [this task](https://github.com/darklang/dark/issues/3556) and hoping these questions will help steer in the right direction.


fsharp-backend/src/BackendOnlyStdLib/HttpClient.fs
```diff
@@ -319,7 +319,7 @@ let rec httpCall
 
           // CLEANUP no reason to do this
           // Match curls default redirect behaviour: if it's a POST with content, redirect to GET
-          // FSTODO: are some headers involved
+          // FSTODO: are some headers involved EXTRACT
```
StachuDotNet, line 322: I'm going to extract this FSTODO and the one in `LegacyBaseHttpClient` into their own joined issue.

fsharp-backend/tests/FuzzTests/DvalRepr.FuzzTests.fs
```diff
@@ -36,7 +36,6 @@ module DeveloperRepr =
       |> Arb.filter (function
         | RT.DFnVal _ -> false
         | RT.DFloat 0.0 -> false
-        | RT.DFloat infinity -> false // TODO I don't think this is doing what we want it to!!
```
StachuDotNet, line 39: FYI this was excluding all DFloats from this DvalRepr test. I've implemented the intention in `Generators.fs` (along with a NaN exclusion)

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
```diff
@@ -55,11 +55,11 @@ module Generators =
           let! v = Arb.generate<int64>
           return RT.EInteger(gid (), v)
         | RT.TStr ->
-          let! v = Generators.ocamlSafeString
+          let! v = G.ocamlSafeString
```
StachuDotNet, line 58: For consistency, used the `G` module shorthand for the _common_ generators

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
```diff
@@ -128,17 +128,20 @@ module Generators =
               (fun i (v : RT.DType) -> (id i, $"{v.toOldString().ToLower()}_{i}"))
               paramTypes
 
-          // FSTODO: occasionally use the wrong return type
-          // FSTODO: can we use the argument to get this type?
+          let! returnType =
+            Gen.frequency [ (98, Gen.constant returnType)
+                            (2, G.RuntimeTypes.dType) ]
```
StachuDotNet, line 133: I believe this matches your expectation for the FSTODO? In any case, I've tested that this works OK.

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
```diff
@@ -128,17 +128,20 @@ module Generators =
...
+
+          // FSTODO: can we use the argument to get this type? WHATISTHIS
           let! body = genExpr' returnType size
           return RT.ELambda(gid (), parameters, body)
         | RT.TBytes ->
-          // FSTODO: this doesn't really do anything useful
+          // FSTODO: this doesn't really do anything useful WHATISTHIS
           let! bytes = Arb.generate<byte []>
           let v = RT.EString(gid (), Base64.defaultEncodeToString bytes)
           return callFn "String" "toBytes" 0 [ v ]
```
StachuDotNet, line 142: I'm not sure what the expectation is here.
I recall there being an issue with bytes interoping with OCaml backend well, which is why there's a `containsBytes` filter elsewhere. Was the intention behind this to get around that?
In any case, what should be done?

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
```diff
@@ -155,9 +158,12 @@ module Generators =
         | RT.TError ->
           let! msg = genExpr' RT.TStr size
           return callFn "Test" "typeError" 0 [ msg ]
+        | RT.TIncomplete ->
+          // WHATISTHIS does this seem reasonable, given the context?
+          let! returnType = G.RuntimeTypes.dType
+          return! genExpr' returnType size
```
StachuDotNet, line 164: This function `exprFromType` is only called from `dvalFromType`, which is in turn used only by the `FnAndArgs` generator, to supply arguments for StdLib fns.

Given that context, does this seem a reasonable implementation for the rare case where a [random value](https://github.com/darklang/dark/blob/63057e05434fb26134ace13693443c69111a0e76/fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs#L365) is to be generated for a fn param?

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
```diff
@@ -179,7 +185,8 @@ module Generators =
       if isKnownDifference then
         false
       elif allowedErrors.functionToTest = None then
-        // FSTODO: Add JWT and X509 functions here
+        // FSTODO: Add JWT functions here
```
StachuDotNet, line 188: the JWT fns are:
- JWT::signAndEncode_v0, v1
- JWT::signAndEncodeWithHeaders_v0, v1
- JWT::verifyAndExtract_v0, v1

signAndEncode have already been marked as ImpurePreviewable, and I've tested that these work ok against the (static) cert generator.

`verifyAndExtract` is a bit tricky - I believe the fns in Json.FuzzTests.fs are sufficient here?

I recall having some issues fuzztesting `signAndEncodeWithHeaders`, but forget the details now - will follow up with those tomorrow.

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
```diff
@@ -513,7 +517,8 @@ let equalsOCaml ((fn, args) : FnAndArgs) : bool =
         if not allowed2 then
           debugFn ()
 
-          print $"Got different error msgs:\n\"{aMsg}\"\n\nvs\n\"{eMsg}\"\n\n"
+          print
+            $"Got different error msgs:\n\"{aMsg}\"\n\n(F#) vs (OCaml)\n\"{eMsg}\"\n\n"
```
StachuDotNet, line 521: it was confusing to me which error message was which - this helps, esp. as it matches the order in `allowedFuzzerErrors.jsonc`

fsharp-backend/tests/FuzzTests/Generators.fs
```diff
@@ -25,7 +25,7 @@ module DvalReprInternal = LibExecution.DvalReprInternal
 let isSafeOCamlString (s : string) : bool = s <> null && not (s.Contains('\u0000'))
 
 /// We disallow `\u0000` in OCaml because Postgres doesn't like it; see `of_utf8_encoded_string.ml`
-/// FSTODO: add in unicode
+/// FSTODO: add in unicode WHATISTHIS
```
StachuDotNet, line 28: What's the expectation here?

fsharp-backend/tests/FuzzTests/HttpClient.FuzzTests.fs
```diff
@@ -24,7 +24,7 @@ type Generator =
   static member Instant() : Arbitrary<NodaTime.Instant> = Generators.NodaTime.Instant
 
   static member String() : Arbitrary<string> =
-    // FSTODO: add in unicode
+    // FSTODO: add in unicode WHATISTHIS
```
StachuDotNet, line 27: What's the same expectation here? is it the same expectation as in `Generators.fs`?

fsharp-backend/tests/FuzzTests/HttpClient.FuzzTests.fs
```diff
@@ -82,7 +82,7 @@ module Tests =
     let test name fn = testProperty config typeof<Generator> name fn
     testList
       "HttpClient, known good"
-      [ test "dvalToUrlStringExn" dvalToUrlStringExn // FSTODO: unicode
+      [ test "dvalToUrlStringExn" dvalToUrlStringExn // FSTODO: unicode WHATISTHIS
```
StachuDotNet, line 85: What's the same expectation here? is it the same expectation as in Generators.fs?

fsharp-backend/tests/TestUtils/TestUtils.fs
```diff
@@ -298,7 +298,7 @@ let executionStateFor
                 // This should be greater than zero, but there's confusion between
                 // different kinds of exceptions. Once we untangle handleException
                 // this can be re-enabled.
-                // FSTODO tc.exceptionReports.Length > 0
+                // FSTODO tc.exceptionReports.Length > 0 WHATISTHIS
```
StachuDotNet, line 301: What's the same expectation here?
Has `handleException` been "untangled"?

fsharp-backend/src/BackendOnlyStdLib/HttpClient.fs
pbiggar: Good idea

fsharp-backend/tests/FuzzTests/DvalRepr.FuzzTests.fs
pbiggar: Oh lol!

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
pbiggar: Yup, perfect!

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
pbiggar: This is here because there isn't a literal byte representation in Dark. I don't remember why that comment is there, but the intention is to expand our byte level testing to include non-stringable bytes (to test edges of conversions to strings, and just be more robust in general). If we're able to remove `containsBytes` or reduce its scope, all the better.

fsharp-backend/tests/FuzzTests/Generators.fs
pbiggar: To generate all unicode (except \u0000). If we already do that here then it's done.

fsharp-backend/tests/FuzzTests/HttpClient.FuzzTests.fs
pbiggar: Yes, I think so

fsharp-backend/tests/TestUtils/TestUtils.fs
pbiggar: Yes it has actually! See here for the new logic https://github.com/darklang/dark/blob/63057e05434fb26134ace13693443c69111a0e76/fsharp-backend/src/LibExecution/Interpreter.fs#L680


I _think_ but am not sure that you could remove the `if isDError` here.

Any exception except CodeException and GrandUserException should indicate that the code is misbehaving. We may have to triage those misbehaviours individually, but the general idea is that other exceptions should be caught and handled closer to the callsite (and if they still want to throw exceptions, they should use a CodeException).

pbiggar (commented):
This is looking good.

Overall, the goal here is to gain confidence that users' code isn't going to break. To accomplish that, we want to get as many functions fuzzed as possible, with as few classes of values filtered out as possible.

In particular, anywhere we have a "we don't test this type because X" we would want to see if that can be removed or tickled differently (eg, tests with infinite perhaps didn't work because we couldn't serialize infinity, but now we can serialize infinity so we can test it again).

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
StachuDotNet: That's exactly why it's been confusing for me :)

Generally in "unit testing" literature (at least what I've read), `expected` comes first, then `actual`.

pbiggar: Oh funny. Expecto puts `actual` first so that's what I've gotten used to now

StachuDotNet:
Other PRs are replacing this one; I have two more bits to extract out of this, then I'll close it in favor of the others.

fsharp-backend/tests/FuzzTests/ExecutePureFunctions.FuzzTests.fs
StachuDotNet: In this context, we're returning `RT.Expr`s; there is no `EIncomplete`.

Since there is no `EIncomplete`, my best idea was to return `ENull`, `EBlank`, or something random. I figured "something random" would test the most.

Thoughts?

pbiggar: Ah, I see. An `EBlank` is the Expr equivalent of `TIncomplete`

StachuDotNet:
Closing this stale PR.
Most of this was done (after some tidying) in #3656.
The two threads I didn't follow up on have been just now linked in #3556.

PR #3656, opened 2022-04-06
Resolve several FSTODOs; refine fuzztests

StachuDotNet:


pbiggar:
It seems like the `NaN` found a real error in `Float::clamp` that should be fixed - am I understanding that right?

StachuDotNet:
> It seems like the `NaN` found a real error in `Float::clamp` that should be fixed - am I understanding that right?

For some reason I thought that we didn't support `NaN` - if we do, then yes!
How do you achieve NaN in Dark?

pbiggar:
`Test.nan_v0`. I think `0.0/0.0` will also get you there.

I'd like to not support NaN but given it's supported by the underlying float implementation, it might sneak in there. There's a lot of tests around it.

StachuDotNet:
Seems I was slightly misinterpreting; now I'm thinking that the difference is in types?

```
fn ({ module_ = "Float"
  function_ = "clamp"
  version = 0 })
DEBUG: args ([DFloat 1.718281828; DFloat nan; DFloat 9.223372037e+18])
Got different error msgs:
"clamp requires arguments to be valid numbers"

vs
"{
  "short":
    "Internal Float.clamp exception: (\"clamp requires [min <= max]\" (min 9.2233720368547758E+18) (max NAN))",
  "long": null,
  "tipe": "code",
  "actual": null,
  "actual_tipe": null,
  "expected": null,
  "result": null,
  "result_tipe": null,
  "info": { "a": "nan", "b": "9.2233720368547758e+18" },
  "workarounds": []
}"
```

and because of the type difference, a rule of 
```
    [
      "Float::clamp",
      "clamp requires arguments to be valid numbers",
      "Internal Float.clamp exception: (\"clamp requires [min <= max]\" (min .*) (max .*))"
    ],
```
didn't catch the case.

Bit surprising - working through this, but guessing OCaml didn't 'catch' as well here?

StachuDotNet:
Ah, right, seems the [F# impl](https://github.com/darklang/dark/blob/b4ec433bb24be0f917d51b6d5f2219caf0d77de5/fsharp-backend/src/LibExecutionStdLib/LibFloat.fs#L317-L321) avoids exceptions, while [OCaml impl](https://github.com/darklang/dark/blob/b4ec433bb24be0f917d51b6d5f2219caf0d77de5/backend/libexecution/libfloat.ml#L278-L293) yields exception.code:

fsharp-backend/tests/allowedFuzzerErrors.jsonc
```diff
@@ -157,6 +157,11 @@
       "Expected the argument `.*` to be a string representation of an IEEE float, but it was `.*`",
       "Expected a string representation of an IEEE float"
     ],
+    [
+      "Float::clamp",
+      "clamp requires arguments to be valid numbers",
+      ".*Internal Float.clamp exception: \\(\\\"clamp requires \\[min <= max\\]\\\" \\(min .*\\) \\(max .*\\)\\).*"
```
StachuDotNet, line 163: An example error this matches:
```
{
  "short":
    "Internal Float.clamp exception: (\"clamp requires [min <= max]\" (min NAN) (max -1))",
  "long": null,
  "tipe": "code",
  "actual": null,
  "actual_tipe": null,
  "expected": null,
  "result": null,
  "result_tipe": null,
  "info": { "a": "-1.", "b": "nan" },
  "workarounds": []
}
```

PR #3657, opened 2022-04-06
Tidy test utility around expected exception count checking

StachuDotNet:
Following the advice provided [here](https://github.com/darklang/dark/pull/3654/files#r844069178), three tests failed locally as a result.

I've added `EXPECTED_EXCEPTION_COUNT: 1` to those tests, and they're now passing. Is that the expected change?


pbiggar:
Yeah, actually this is good as is. The last two changes are perfect, as the tests are "what happens if we throw some rando exception in the middle of a function". The first change we could possibly wrap it, but since it's a darkinternal function I think this is probably fine.

I'm good to merge once CI passes.

Issue #3658, opened 2022-04-06
investigate LibStaticAssets issue with darklang.com


pbiggar:
I added ops-corpsite.builtwithdark.com to the new backend, and it caused the page source to display instead of being interpreted by the browser:

![image](https://user-images.githubusercontent.com/181762/162080220-79b866b6-8ca1-488c-93b5-c8e473dae45c.png)

The code is:

```
StaticAssets::serveLatestv1 "index.html"
```

This needs to be fixed before we can switch more canvases over to BwdServer.

pbiggar:
Differences between the responses:

old response uses:
```
content-type: text/html
server: nginx/1.16.1
cache-control: public, max-age=604800, immutable
expires: Sat, 09 Apr 2022 17:42:07 GMT
last-modified: Mon, 12 Oct 2020 19:13:11 GMT
```

while the BwdServer uses:
```
content-type: text/plain; charset=utf-8
server: UploadServer
```
and doesn't have the other servers.

So most likely the issue is the `content-type`.

pbiggar:
- [ ] validate fix for serveLatest_v1
- [ ] find way to test fix
- [ ] find way to test class of errors
- [ ] adapt fix/tests for all variations of fetch, serve, latest, bytes, etc.

pbiggar:
Functions to test:
- [ ] StaticAssets::baseUrlFor
- [ ] StaticAssets::baseUrlForLatest
- [ ] StaticAssets::fetch
- [ ] StaticAssets::fetchBytes
- [ ] StaticAssets::fetchLatest
- [ ] StaticAssets::fetchLatestBytes
- [ ] StaticAssets::fetchLatest_v1
- [ ] StaticAssets::fetch_v1
- [ ] StaticAssets::serve
- [ ] StaticAssets::serveLatest
- [ ] StaticAssets::serveLatest_v1
- [ ] StaticAssets::serve_v1
- [ ] StaticAssets::urlFor
- [ ] StaticAssets::urlForLatest

pbiggar:
- [ ] Test serve in the bwdserver tests too and check everything makes it through as expected.

pbiggar:
- [ ] don't change UploadServer as nginx doesnt

pbiggar:
Merged and confirmed this works on darklang.com

PR #3660, opened 2022-04-07
Also get the content headers when making StaticAssets calls

pbiggar:
This fixes the StaticAssets issue.

I want to add tests before merging though.


pbiggar:
The fix to static assets was to pass through the content-headers as well as the other headers. We had code to do this for LibHttpClient but forgot it here, so I refactored to make it harder to forget.

I added tests and refactored the test suite to make it easier to do so. Because static assets test implicitly have the canvasName in them (that is where the first hash is generated from), it was necessary to stop using random canvas names for these particular tests, so I added `ExactCanvasName`. I also made it so we could have multiple annotations following a `with` on a test or test group, which was needed for this test. Finally, I added a way to add a static deploy to the DB.

I uploaded the 3 files in static-assets-tests to the cloud storage bucket pointed to by the `test-static-assets-deploy` canvas. This is at `https://test-static-assets-deploy.darksa.com/u0n38z-mv-t0tkvnwiqyvankw6w/abcdef1234/index.html`. I noticed that index.html would 404 sometimes, hopefully that's just a brief hiccup and doesn't stick around.

As well as adding tests for the remaining functions (listed [here](https://github.com/darklang/dark/issues/3658#issuecomment-1091041907)), the tests currently have a major race condition. Since each test uses the same canvas, when each test is cleared between test runs, they clean the DB. If another test is running using the same canvas, the tests will fail. I thought about creating a canvas for each test, but it would also need a bucket for each test, which would be frustrating. but maybe that's the only way.

To get this over the line we need to:
- make tests work consistently (eliminate the race condition)
- add tests for the other functions

scripts/formatting/format
```diff
@@ -125,6 +125,8 @@ all_prettier_files() {
     -o -path "./fsharp-backend/.paket" \
     -o -path "fsharp-backend/Build" \
     -o -path "./fsharp-backend/Build" \
+    -o -path "fsharp-backend/tests/staticassets-tests" \
+    -o -path "./fsharp-backend/tests/staticassets-tests" \
```
StachuDotNet, line 129: is this to deal with OS inconsistencies?

pbiggar: You mean the fact that it's there twice? I don't remember. I'm cargo culting it a bit here.

StachuDotNet: yeah that's what I meant; OK, np

fsharp-backend/tests/Tests/LibExecution.Tests.fs
```diff
@@ -51,15 +54,40 @@ let setUpWorkers meta workers =
     do! Canvas.saveTLIDs meta oplists
   }
 
+let setupStaticAssets
+  (meta : Canvas.Meta)
+  (deployHashes : List<string>)
```
StachuDotNet, line 59: nitpick that we can move elsewhere if better:

built-in types like `list` are generally (at least, by docs I've read, incl. built-in types like `list` are generally (at least, by docs I've read) used in the suffix notation (`x list` rather than `List<x>`)
I'm totally fine getting away from that, but we should mark that in the coding guide.) used in the suffix notation (`x list` rather than `List<x>`), standardize to that, and (as we see it) adjust to use that generally.

There's a LinkedList/Generic.List type that also exists in .NET

upon seeing List<x>, I think many devs brains would think of that mutable array list thing first, esp. if any C# experience

pbiggar:
Am continuing with this. I just remembered that we can use `testSequencedGroup` to avoid races.

pbiggar:
Summary of whole PR:

The fix to static assets was to pass through the content-headers as well as the other headers. We had code to do this for LibHttpClient but forgot it here, so I refactored to make it harder to forget.

I added tests and refactored the test suite to make it easier to do so. Because static assets test implicitly have the canvasName in them (that is where the first hash is generated from), it was necessary to stop using random canvas names for these particular tests, so I added ExactCanvasName. I also made it so we could have multiple annotations following a with on a test or test group, which was needed for this test. Finally, I added a way to add a static deploy to the DB.

I uploaded the 3 files in static-assets-tests to the cloud storage bucket pointed to by the test-static-assets-deploy canvas. This is at https://test-static-assets-deploy.darksa.com/u0n38z-mv-t0tkvnwiqyvankw6w/abcdef1234/index.html. (Note that this would occasionally fail, not consistently, probably an eventual consistency thing. Anyway, it has stopped).

Since each test uses the same canvas, I had to run all the tests in serial. This slows down the test suite so we may need to split this file up later to speed it up, but fine for now.


Also:

- created a specific config var (STATIC_ASSETS_SALT_SUFFIX) instead of using ENV_DISPLAY_NAME (this was the make tests work in CI, but it's good practice anyway)

- add `Test.regexReplace` (I ended up not using it in the end, but nice to have)

- Add `Test.httpResponseHeaders`, `Test.httpResponseStatusCode`, and  `Test.httpResponseBody`. There isn't another way to get values out of a httpresponse

- Add `Test.intArrayToBytes` - allow creating a DBytes value in a test

- Fix error message when not a 200 code (we would throw the exception then catch it immediately)

- better error message when there's no deploy hash

- Change which headers are removed in the F# version:
  - remove Content-length cause asp.net likes to set it
  - remove Server as nginx overrode that before, and we don't want to set the Google server header as our server header (it's `UploadServer`)
  - allow Cache-Control (we tried to disable it before but it didn't work, so this keeps things the same)

backend/liblegacyfuzzing/fuzzing.ml
```diff
@@ -472,6 +472,83 @@ let fns : Types.RuntimeT.fn list =
...
+    ; return_type = TList
+    ; description = "Get status code from a HttpResponse"
+    ; func =
+        InProcess
+          (function
+          | state, [DResp response] ->
+            ( match response with
+            | Redirect _, _ ->
+                Dval.dint 302
+            | Response (code, headers), _ ->
```
StachuDotNet, line 533:
```suggestion
            | Response (code, _), _ ->
```
nitpick

fsharp-backend/tests/Tests/LibExecution.Tests.fs
```diff
@@ -441,38 +504,48 @@ let fileTests () : Test =
...
 
         currentGroup <- { currentGroup with tests = currentGroup.tests @ [ test ] }
 
       | _ -> raise (System.Exception $"can't parse line {i}: {line}"))
 
     finish ()
-    testList $"Tests from {filename}" allTests)
+    let tests = testList $"Tests from {filename}" fileTests
+    // Staticassets.tests use a combination of ExactCanvasName and
+    // StaticAssetsDeployHash, which can be racy. See comment in staticassets.tests.
```
StachuDotNet, line 542: replace 'racy' with less ambiguous term like "Race conditions can cause problems with these tests" here and in `staticassets.tests`?

Wasn't sure if 'racy' meant 'involves race conditions' or 'annoying/dicy' or something else, at first read

fsharp-backend/tests/httptestfiles/serve-latest.test
```diff
@@ -0,0 +1,50 @@
+[http-handler GET /]
+StaticAssets.serve_v1_ster "abcdef1234" "index.html"
+
+[canvas-name test-static-assets-deploy]
```
StachuDotNet, line 4: Not a big deal, but I'd expect this at the top of the file for single-file tests

fsharp-backend/tests/testfiles/staticassets.tests
```diff
@@ -0,0 +1,547 @@
...
+  198; 156; 204; 102; 179; 51; 40; 83; 70; 49; 138; 35; 161; 2; 185; 92; 110; 50; 176; 3;
+  120; 111; 173; 253; 200; 31; 202; 0; 35; 136; 200; 0; 142; 66; 42; 144; 36; 201; 114;
+  99; 76; 190; 183; 183; 247; 51; 35; 164; 210; 225; 158; 56; 182; 223; 85; 147; 89; 34;
+  65; 123; 75; 62; 247; 144; 255; 197; 47; 29; 238; 219; 222; 92; 167; 151; 81; 0; 0; 0;
+  0; 73; 69; 78; 68; 174; 66; 96; 130 ]
+
+
+// baseUrlFor
+[test.baseUrlFor] with ExactCanvasName test-baseurlfor-0
+StaticAssets.baseUrlFor_v0 "abcdefghi" = "https://test-baseurlfor-0.darksa.com/gd57quywk2db6zncs3h2hxgfdtc/abcdefghi"
```
StachuDotNet, line 58: where does `gd57quywk2db6zncs3h2hxgfdtc` come from?

StachuDotNet:
Tests to consider adding to `staticassets.tests`:
- (easy?) for `serve` and similar fns, tests where we _have_ a deploy has, but reference a different one
- (more complicated) for `serveLatest`, deploy _twice_, and ensure the latest is the one served (unless I missed this)?

StachuDotNet:
Done another round of review @pbiggar

fsharp-backend/tests/testfiles/staticassets.tests
pbiggar: These are all hashes made from the canvas name

https://github.com/darklang/dark/blob/d88146e9d404ce8aa59e82096f15b044f130a8cf/fsharp-backend/src/LibBackend/StaticAssets.fs#L67

fsharp-backend/tests/testfiles/staticassets.tests
```diff
@@ -0,0 +1,547 @@
...
+[test.fetchLatestBytes_v0 nohash FSHARPONLY] with ExactCanvasName test-static-assets-deploy
+StaticAssets.fetchLatestBytes_v0 "index.html" = Test.typeError "No deploy hash found"
+
+
+
+// ----------------------------
+// Serve
+// ----------------------------
+
+[fn.ocaml200HtmlHeaders i:int]
```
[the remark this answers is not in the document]
pbiggar: Yeah, exactly. As far as I recall, we can't call functions without arguments in these tests.

I'll rename the parameters to `_`

pbiggar:
> Tests to consider adding to `staticassets.tests`:
> 
> * (easy?) for `serve` and similar fns, tests where we _have_ a deploy has, but reference a different one


> * (more complicated) for `serveLatest`, deploy _twice_, and ensure the latest is the one served (unless I missed this)?

Yes, I can add two deploy hashes easily enough.

PR #3694, opened 2022-04-13
Fix DeleteTLForever

pbiggar:
This fixes the [Rollbar](https://rollbar.com/darkops/darklang/items/3075/) "couldn't find the TL we supposedly just looked up", which was triggered in [ApiOps](https://github.com/darklang/dark/blob/47e7f861d61786815cd1516d89c10430389e2677/fsharp-backend/src/ApiServer/Api/APIOps.fs#L116).

This exception was triggered when users tried to permanently delete their handlers/DBs/etc, because then the handler was no longer in the list of toplevels (since it was deleted). It was also possible to trigger by Undo-ing past the start of a handler, although users hadn't done that in practice.

The issue occurred because we changed in rewriting this part of the code, I changed the interface making it hard to handle these situations.

In the OCaml version, when you got a DeleteForever Ops (there were three, one for handler/DB, one for types, and one for functions), we would process it as an Op. Then when trying to save ops, we would recognize the op was missing, and save it in one of two special deleted formats:
- for normal deleted (still there but "in the trash"): we would save it in the DB with the row set `deleted=true`, and delete the `name`, `space`, and `module` columns (typically used to find a Http handler when a request is made).
- for "Forever" deleted handlers, we would set the `deleted=null`, and also put a near-empty handler in the cached handler column (keeping the oplist however).

In the first F# version, we accidentally made these changes:
- for normal deletes, we correctly set `deleted=true` but did not set the `name`, `space` and `module` handlers to blank. This meant that they could still be loaded and found in some cases when they're shouldn't have been.
- for "forever" deletes, we raised an exception

To solve the normal delete issue, we now set the fields correctly. This also adds `deleted` checks to all the places that should have had them (they didn't strictly need this in the past as the searches used the `name` (etc) fields to find them, so since we accidentally didn't clear those fields, they could be found).

To solve the "forever" deletes, I made a new API and told the client to use it instead. Using Ops for Forever deletes made no sense and was likely a hack at some point. We also shouldn't use ops for regular deletes, but we'll maybe deal with that at a later point.

I didn't remove the ops yet as they're almost certainly in the DB, but hopefully only in permanently deleted handlers. I'll make a plan for this and add an issue for followups.

One further issue is that this exception is also triggered if a user does Undo too many times. When users did Undo too far, we changed their handlers into `HTTP GET /` handlers. This wasn't good either, so now we just ignore the final undo, which is better. Also, that's what appears to happen in the UI, so it was a surprise that the DB was storing something different).

More to do:
- [x] filter out the ops on the backend (leave them in the DB)
- [x] after deleting a handler, test it can't be loaded in live situation
- [x] after deleting a handler, check its fields are empty
- [ ] FOLLOWUP: find deleted handlers in the DB and clear the name/module/space fields (there should be 113 of these)
- [ ] FOLLOWUP: find handlers with null in their deleted field (indicating they should be deleted forever) in the DB and delete them (there should be 37685 of these)


fsharp-backend/tests/Tests/ApiServer.Tests.fs
```diff
@@ -577,6 +577,12 @@ let testInsertDeleteSecrets (client : C) (canvasName : CanvasName.T) =
     Expect.equal fResponse oResponse "compare responses"
   }
 
+let canonicalize404s (fofs : List<TI.F404>) : List<TI.F404> =
+  fofs
+
+  |> List.map (fun ((space, name, modifier, datetime, traceID) : TI.F404) ->
+    (space, name, modifier, datetime.truncate (), traceID))
```
StachuDotNet, line 584:
```suggestion
let canonicalize404s (fofs : List<TI.F404>) : List<TI.F404> =
  fofs
  |> List.map (fun ((space, name, modifier, datetime, traceID) : TI.F404) ->
    (space, name, modifier, datetime.truncate (), traceID))
```
errant newline

StachuDotNet (approved):
Wish I had more to add - looks good!

StachuDotNet:
Actually - are there tests we can write around this?

pbiggar:
> Actually - are there more tests we can write around this? I'm not coming up with examples, but feels like significant change for little test change

Yeah, I agree. Working on it shortly.

pbiggar:
Review from "[Remove the Delete*Forever Ops from ProgramTypes](https://github.com/darklang/dark/pull/3694/commits/6409b9a3e0e4f4e081cf35da6e3572afefc09717)" would be welcome

pbiggar:
I still have tests to add.

pbiggar:
> after deleting a handler, test it can't be loaded in live situation

This is actually handled by Canvas.Tests.testLoadAllDBs and Canvas.Tests.testHttpLoadIgnoresDeletedFns.

StachuDotNet (approved):

StachuDotNet:
Looks good - will read through tests too once they're out

pbiggar:
Tests added

StachuDotNet (approved):

pbiggar:
Yay! This turned into quite the slog 😭

StachuDotNet:
> Yay! This turned into quite the slog sob

Happens! Looks great.

PR #3695, opened 2022-04-14
Add LibBinarySerialization

pbiggar:
I was going to do this in #3694 but it's cleaner as a standalone PR.

For execution, we separate RuntimeTypes from ProgramTypes, so that the way users write code does not have to be the exact same format that we use to execute it. However, on the backend, the programTypes that we use to process programs is the same as the Serialization format.

This is inflexible, as I discovered when I wanted to delete some Ops in #3694. I can't delete the ops because they're an essential part of the serialization format, and things will break if they go away. However, that also means we need to handle them in every place they're used, which is ugly.

So this adds an extra type that is currently identical. After this we'll be able to change the ProgramTypes to match more how we think about programs, without breaking the serialization format.


pbiggar:
The tests here caught a couple of mistakes I made - glad we added them.

fsharp-backend/src/LibBinarySerialization/SerializedTypes.fs
```diff
@@ -0,0 +1,354 @@
...
+// field in the record should be annotated with `[<MessagePack.Key 0>]` (the zero
+// should be replaced with a unique sequential index):
+//
+// [<MessagePack.MessagePackObject>] type X = { [<MessagePack.Key 0>] x : int }
+//
+// If you forget to annotate all parts of a type (or a type referred to by that type)
+// the serializer will raise an exception. (It seems to be OK to not annotate some
+// variants but not others; since it's unclear we annotate them all)
+//
+// All "code" is Dark is serialized using these types and stored in the DB, and we
```
StachuDotNet, line 24:
```suggestion
// All "code" in Dark is serialized using these types and stored in the DB, and we
```

fsharp-backend/src/LibBinarySerialization/SerializedTypes.fs
```diff
@@ -0,0 +1,354 @@
...
+  | TResult of DType * DType
+  // A named variable, eg `a` in `List<a>`, matches anything
+  | TVariable of string // replaces TAny
+  | TFn of List<DType> * DType // replaces TLambda
+  | TRecord of List<string * DType>
+  | TDbList of DType // TODO: cleanup and remove
+// This allows you to build up a record to eventually be the right shape.
+// | TRecordWithFields of List<string * DType>
+// | TRecordPlusField of string (* polymorphic type name, like TVariable *)  * string (* record field name *)  * DType
+// | TRecordMinusField of string (* polymorphic type name, like TVariable *)  * string (* record field name *)  * DType
```
StachuDotNet, line 176: Are these hanging around for future use?

StachuDotNet (approved):
Great. I actually like this more - keeps the ProgramTypes tidier

fsharp-backend/src/LibBinarySerialization/SerializedTypes.fs
pbiggar: Yes

fsharp-backend/src/LibBinarySerialization/SerializedTypes.fs
pbiggar: Will fix in the next PR

Issue #3699, opened 2022-04-14
Fix "operator does not exist: integer = text


pbiggar:
https://rollbar.com/darkops/darklang/items/3100

pbiggar:
What's happening here is that there are type errors in the SQL and instead of being caught as an error and reported, the SQL is getting called, where it fails. The tests had a "Fn called with error as an argument" thing which caused the problem not to be noticed.

pbiggar:
There's a few issues here:
1. we're not catching exceptions thrown by the DB driver (eg if there's bad sql)
2. we're papering over the errors in tests
3. there's an actual error being thrown

The actual error is the query `\p -> p.intField = "string"`. The `=` function has two arguments of type `Variable "a"`, but at no point do we check they have the same concrete type (this is because in the old version they were both type `TAny`.

This failed in the old version so I'm going to handle 1. and 2., but not 3.

pbiggar:
This is fixed

PR #3702, opened 2022-04-15
Handle sql errors better

pbiggar:
Partial fix for https://github.com/darklang/dark/issues/3699

There's an exception in rollbar that comes from the following code:

```
DB::query TableWithIntField \p -> p.intField = "string"
```

This is because for the `=` function we check the two arguments separately to see if they are the `TAny` type, which they are. In fact, we should check that both arguments are the **same** type.

However, this was also a bug in the old version, so I'm not fixing that. Instead, I'm fixing two related problems:
- we didn't catch the error in tests
- we raised an internal exception instead of showing a semi-nice error to users

This catches the exception and tidies up how we do tests a little in db.tests to be able to see the error.

We don't need to do the rest of #3699 for the migration to be complete (in fact, it's bad to do so, as the two versions will have slightly different behaviour).

While I was here, I also marked a few QueryFunctions that were not previous marked.

### DIFF
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDB2.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDB2.fs
@@ -8,6 +8,8 @@ module UserDB = LibBackend.UserDB
 module Errors = LibExecution.Errors
 module Db = LibBackend.Db
 
+open LibService.Exception
+
 let fn = FQFnName.stdlibFnName
 
 let incorrectArgs = LibExecution.Errors.incorrectArgs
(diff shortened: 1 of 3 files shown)


fsharp-backend/src/BackendOnlyStdLib/LibDB2.fs
```diff
@@ -24,6 +26,21 @@ let tableParam = Param.make "table" dbType ""
 let ocamlCompatibleSpecParam = Param.make "spec" ocamlTObj ""
 let queryParam = Param.makeWithArgs "filter" (TFn([ varA ], TBool)) "" [ "value" ]
 
+let handleUnexpectedExceptionDuringQuery
```
StachuDotNet, line 29: best binding name

StachuDotNet (approved):

PR #3705, opened 2022-04-15
We have two canvases we can switch over

pbiggar:
Applied, will merge once it's clear they work

### DIFF
--- a/services/bwd-deployment/bwd-ingress.yaml
+++ b/services/bwd-deployment/bwd-ingress.yaml
@@ -315,6 +315,5 @@ spec:
                   number: 80
 
 
-
   tls:
     - secretName: builtwithdark.com-tls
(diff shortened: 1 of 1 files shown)


services/bwd-deployment/bwd-ingress.yaml
```diff
@@ -293,6 +293,27 @@ spec:
                 name: bwdserver-service
                 port:
                   number: 80
+    - host: ops-pinger.builtwithdark.com
```
StachuDotNet, line 296: What does this canvas do?

pbiggar: pages me if it doesn't get a request for 8 minutes. This tests that cron is still working.

Funnily enough, switching it over paged me!

Issue #3706, opened 2022-04-15
Some FuzzTests aren't clearing their canvases


StachuDotNet:
Discovery found [here](https://github.com/darklang/dark/issues/3689#issuecomment-1100260392_). Known to be an issue at least for `executepurefunctions` and `bytes-to-string` fuzztests, but I see others as well.

I've been running FuzzTests a lot recently. Apparently I locally have 100k canvases of the form `test-executepurefunctions-[random]` right now.

We can resolve this by either clearing the canvases via code, or extract/re-use logic in the `integration-tests/clear-db.sh` script.

If going the latter route:
- [ ] extract test-canvas-deletion logic from `clear-db.sh` into a new script (`clear-test-canvases.sh`)
- [ ] edit that^ script to either:
  - (ideally) handle _many_ tests (right now it fails for more than 143 test canvases)
  - (good enough for now) provide commentary of how to hackily deal with such
- [ ] clear test canvases as part of fuzztest execution
  calling `clear-test-canvases` as part of `run-fsharp-fuzzer` should be fine for now

StachuDotNet:
Misc. idea: when a round of tests is complete, run one final test to ensure there aren't any `test-` canvases hanging around, to ensure tests have cleaned up after themselves properly.

pbiggar:
> idea: when a round of tests is complete, run one final test to ensure there aren't any `test-` canvases hanging around, to ensure tests have cleaned up after themselves properly.

Leaving canvases around is desirable as it allows us to inspect them afterwards. Though test canvases are expected to be in the testdb database.

StachuDotNet:
Aha, that's fair! So, #3504 is the focus here, and this can be closed?

pbiggar:
Sort of.

Some context:
- integration tests are supposed to run on the `devdb` DB, so you can look at the canvas with your browser if it fails. That's why clear-db only clears `test-` canvases
- non-integration tests are supposed to be run on the `testdb` DB. That's intended to be cleared on each new run of tests, and can be cleared with a simple `dropdb`
- ApiServer.Tests also operates on `devdb`. This is by-design, as we want to test do we get the same values from the ocaml server as we do on ApiServer. However, once we've removed the ocaml server, all the ApiServer tests will need to change in some form.

[...]

Issue #3689, opened 2022-04-13
ApiServer tests failing for me locally


StachuDotNet:
Apologies for this being bare for now - I'll add some context in the morning or after done current task.

All I know is that I get this when they're included:
```
[00:18:04 INF] EXPECTO? Running tests... <Expecto>
[00:18:06 ERR] tests.ApiServer.check apiserver permissions.check apiserver permissions[1]: (Value is not created., test-something) -> (200, 200) errored in 00:00:01.6060000 <Expecto>
System.Threading.Tasks.TaskCanceledException: A task was canceled.
[00:18:06 ERR] tests.ApiServer.check apiserver permissions.check apiserver permissions[3]: (Value is not created., sample-something) -> (200, 200) errored in 00:00:01.6060000 <Expecto>
System.Threading.Tasks.TaskCanceledException: A task was canceled.
[00:18:06 ERR] tests.ApiServer.check apiserver permissions.check apiserver permissions[2]: (Value is not created., sample) -> (200, 200) errored in 00:00:01.6080000 <Expecto>
System.Threading.Tasks.TaskCanceledException: A task was canceled.
[00:18:06 ERR] tests.ApiServer.check apiserver permissions.check apiserver permissions[0]: (Value is not created., test) -> (200, 200) errored in 00:00:01.6060000 <Expecto>
System.Threading.Tasks.TaskCanceledException: A task was canceled.
[00:18:06 ERR] tests.ApiServer.local.initial load errored in 00:00:01.0050000 <Expecto>
System.Threading.Tasks.TaskCanceledException: A task was canceled.
Stack overflow.
Repeat 52329 times:
```

StachuDotNet:
I'm looking into this now as it's partially blocking my local testing of #3665.

So far, this seems related to the permission issues I was seeing a few weeks back when working on integration tests - local credentials set up seem to not be OK for me. Diagnosing.

StachuDotNet:
Somehow, I have... quite a few test canvases, which is yielding issues when attempting to deserialize the initial load.
![image](https://user-images.githubusercontent.com/906686/163603194-b590dbeb-d4cb-451e-ac1a-49126a1f2b56.png)

I'm not sure how it got to this point. @pbiggar I'm planning to locally resolve this locally by purging local canvases and trying again. But for reference, [here](https://gist.github.com/StachuDotNet/9593a63baad83fb170f3b1dc86193b37) is a dump of what I got locally.

StachuDotNet:
[...]

PR #3707, opened 2022-04-15
Add notes to clear many test canvases

StachuDotNet:
I've been encountering situations where I somehow accumulate _many_ test canvases locally; most recently FuzzTests haven't been cleaning up their canvases, which prevented ApiServer tests from running. See #3689 and #3706.

To resolve this, I've been locally hacking `integration-tests/clear-db.sh`, running it many times until test canvases are cleared.

This isn't an ideal solution, but these notes are an improvement over nothing.

### DIFF
--- a/integration-tests/clear-db.sh
+++ b/integration-tests/clear-db.sh
@@ -13,7 +13,9 @@ function run_sql { psql -d "$DB" -c "$@" >> "$DBLOG" ; }
 
 function fetch_sql { psql -d "$DB" -t -c "$@"; }
 
-# If you're looking to purge a _lot_ of data, here's some advice:
+# If you need to delete many test canvases at once, this script is currently
+# insufficient, as it breaks when more than 143 test canvases are present.
+# To get around this limitation, follow these steps:
 #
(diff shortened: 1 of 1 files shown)


integration-tests/clear-db.sh
```diff
@@ -13,6 +13,20 @@ function run_sql { psql -d "$DB" -c "$@" >> "$DBLOG" ; }
 
 function fetch_sql { psql -d "$DB" -t -c "$@"; }
 
+# If you're looking to purge a _lot_ of data, here's some advice:
```
StachuDotNet, line 16:
```suggestion
# If you need to delete many test canvases at once, this script is currently
# insufficient, as it breaks when more than 143 test canvases are present.
# To get around this limitation, follow these steps:
```

pbiggar:
Feels like we could write to a file (`mktemp`) and then run the commands with `psql -f` instead

StachuDotNet:
That sounds great, and a better solution - I just wanted to get this thing out of my head. I'd like to keep PR open to follow up on it next week, then close + replace with something like that.

Issue #3710, opened 2022-04-16
getting 500 on big user, unclear why


pbiggar:
Debugging this issue now.

pbiggar:
So far I've found two issues:

## Issue 1:
calling a function like this:
```
value
|> myFunction dErrorValue dErrorValue
```

in OCaml returns `dErrorValue` but in F# returns a `DIncomplete`.

No obvious cause so far.

## Issue 2:

F# `DB::keys` is returning `[]` while OCaml `DB::keys` is returning `["some value"]` (OCaml seems to be correct here).

The DB contents are set in the background every minute, by getting the response, adding a new value, then deleting the old values one-by-one (note this is using OCaml code in the queueworker).

I don't have any obvious leads.

pbiggar:
Not having any success reproducing this locally. My plan is to deploy an instrumented version of the runtime to make just this request and see what I can learn.

pbiggar:
OK, found issue 1. The user named their function `myfunction_v1`, and we parsed that as a StdLib and not as a UserFunction.

1) needs tests
2) probably if they don't have a module we should check the stdlib before we decide it's a stdlib function
3) these are probably saved in the DB like this so some sort of migration or fix is required

StachuDotNet:
relates to #3641, it seems

pbiggar:
Fixed and followup done.

PR #3721, opened 2022-04-20
Fix 500 due to bad StdLib function names

pbiggar:
This solves #3641 and the second part of #3710.

This root cause of the issue is that we incorrectly parsed user functions ending in `_v6` (or any version number) as a StdLib function. Then we tried to call the StdLib function and couldn't find it.

So functions like `myFunction_v2` would return `<Incomplete>`.

However, the problem is a bit deeper because we have serialized these in the F# SerializedTypes that are stored in the DB.

This works around the issue for now. I still need to think about how to migrate the data.

The workaround is that if there is a Stdlib function found with no module, which is not a built-in infix function (eg `+`), and is not one of the built-in one word functions (eg `emit_v0`), then it must be one of these functions that is incorrectly parsed. As such, it should be converted to a `User` function instead.

To help ensure the bug is gone, I also changes the ProgramTypes format of function names used in EBinOps, adding checks and assertions during the translation process. Unfortunately, I had to step it back a little bit upon realizing that `Date::>` is an infix function, which we should remove as that was a bad choice we made once upon a time.

There's a checklist below which I intend to do before merging.

Also does:
- tidies up gcp-proxy-db script
- significant imporvements to gcp-psql script
- add some instrumentation
- add scripts to run exec-host locally (with the option to use prodclone)
- adds a selection of new assertion functions
- catch errors better in exechost

Options for migrating the data:
- read it again from ocaml (where it is OK as it uses string names).Advantages are that this is clean. It can't be out of sync as this is still what nearly all users are using. But the parsing code has changed so this might trigger more bugs.
- read and then write the serialized types. This should go through the conversion that cleans things up. This might trigger bugs in the change code.

TODO:
- [x] read known-failing canvas and see if this fixes it
- [x] read all toplevel_oplists from the DB and confirm we can process them
- [x] make a plan for how to migrate the data
- [x] add bug to remove Date infix functions and change the ProgramTypes representation

### DIFF
--- a/fsharp-backend/src/LibBinarySerialization/SerializedTypesToProgramTypes.fs
+++ b/fsharp-backend/src/LibBinarySerialization/SerializedTypesToProgramTypes.fs
@@ -32,7 +32,7 @@ module FQFnName =
     | ST.FQFnName.Stdlib fn when
       fn.module_ = ""
       && not (Set.contains fn.function_ PTParser.FQFnName.oneWordFunctions)
-      && not (Set.contains fn.function_ PTParser.FQFnName.infixFunctions)
+      && not (LibExecutionStdLib.StdLib.isInfixName fn.module_ fn.function_)
       ->
       // It must have had a "_v0" or similar to get here, so always print the
       // version, even if it's 0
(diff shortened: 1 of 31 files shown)


fsharp-backend/src/LibBinarySerialization/SerializedTypesToProgramTypes.fs
```diff
@@ -65,8 +78,27 @@ module Expr =
       PT.EFieldAccess(id, toPT obj, fieldname)
     | ST.EFnCall (id, name, args, ster) ->
       PT.EFnCall(id, FQFnName.toPT name, List.map toPT args, SendToRail.toPT ster)
-    | ST.EBinOp (id, name, arg1, arg2, ster) ->
-      PT.EBinOp(id, FQFnName.toPT name, toPT arg1, toPT arg2, SendToRail.toPT ster)
+    | ST.EBinOp (id, ST.FQFnName.Stdlib fn, arg1, arg2, ster) ->
+      assertIn
+        "serialized binop should have blank/Date module"
+        [ ""; "Date" ]
```
StachuDotNet, line 84: should we ad a CLEANUP to this specific line? (I think we're looking to make Date fns not allow this form?)

StachuDotNet (approved):
No concerns from me - just a few places where we may want CLEANUPs around Date fns.
Will review again once the related TODOs are resolved

pbiggar:
> read all toplevel_oplists from the DB and confirm we can process them

I loaded all the known canvases, and created #3725 for an issue that came up that isn't a blocker.

pbiggar:
Plan:

TODO: might still be in oplists

for each tl:
- oplists, tl
- do roundtrip on both, see if different
- manually check the difference
- record that it differs

for each recorded differing tl:
- get new tl/oplist
- update oplist with saveOplistToFSharpCache
- update cached with saveToplevelToFSharpCache

fsharp-backend/src/LibExecution/OCamlTypes.fs
```diff
@@ -344,6 +344,32 @@ module Convert =
...
+      ">="
+      "<="
+      "<"
+      "^"
+      "%"
+      "/"
+      "Date::<"
+      "Date::>"
+      "Date::<="
+      "Date::>="
```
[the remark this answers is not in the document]
pbiggar: I left it out because we're just going to remove the entire OCamlTypes "soon".

fsharp-backend/src/LibBinarySerialization/SerializedTypesToProgramTypes.fs
pbiggar: Added.

StachuDotNet (approved):

PR #3730, opened 2022-04-22
Upgrade to .net 7 preview 3

StachuDotNet:
Upgrading to .NET 7 Preview 3, mostly in hopes of our Blazor usage improving in a few ways:
- [x] using the publish build rather than the debug build
- [ ] ? AOT builds succeeding
- [ ] AOT builds yielding better performance

Additionally, using .NET 7 Preview 3 allows us to try out [dotnet-wasi-sdk](https://github.com/SteveSandersonMS/dotnet-wasi-sdk)


StachuDotNet:
[this](https://github.com/darklang/dark/pull/3730/commits/c5f55426cec1938b27e98af3efafe6098ab6d6e6) took a while to hunt down, and I'm still not sure which party is at fault. The key change here is that the `function` form fails, but the `match` form passes. It may be an issue with F#, or Ply?

Prior to this change, [several `dropWith` LibList tests fail](https://app.circleci.com/pipelines/github/darklang/dark/2787/workflows/041df8b3-87ed-453b-9130-e535ba408145/jobs/29787/parallel-runs/0/steps/0-111)

StachuDotNet:
@pbiggar given discussion here and within the dotnet-wasi-sdk attempt PR, how are you feeling about this .NET 7 preview 3 upgrade by itself? The only issue I've run into is the external regression mentioned around the `function` usage, but there's certainly some degree of risk still in committing.

pbiggar:
If it doesn't come with Blazor/AOT stuff, then the question is when is this a good risk to take on.

On the one hand, if we do it now, then whatever behaviour is in .NET 7 is the default, and won't be a change when .NET7 comes out filly. On the other hand, we're right at the end of the BwdServer migration and a significant change could set up back - and not only that, it won't be super solvable except by rolling back, which will be a risk again because of new code on top.

I guess let's not do it now then.

PR #3735, opened 2022-04-24
Fix gid with NEVER_REBUILD_DOCKER env var

benwaffle:

`gid` is needed in the last `docker run` command, but it `gid` is not defined if NEVER_REBUILD_DOCKER is set

Define `gid` unconditionally.


pbiggar:
Wonderful, thank you!

Issue #3745, opened 2022-04-26
HSTS headers are wrong in ApiServer


pbiggar:
In "local" tests, I get an error:

```
[19:19:28 ERR] tests.ApiServer.local.hsts errored in 00:00:00.2420000 <Expecto>
Expecto.AssertException: Strict-Transport-Security header either missing or incorrect.
expected: ["max-age=31536000; includeSubDomains; preload,max-age=31536000; includeSubDomains; preload"]
  actual: ["max-age=31536000; includeSubDomains; preload"]
```

On darklang.com I see that we have 2 HSTS headers:

<img width="475" alt="image" src="https://user-images.githubusercontent.com/181762/165376739-e017f551-1cb9-46ca-974e-eb2b68f06cdd.png">

This is served by proxying from BWD and then going through nginx, so we might have too many steps along the way.

StachuDotNet:
I took a stab at this, but am currently blocked by ignorance of nginx + proxying that's going on.

Some notes:
- I was able to reproduce locally
- I wasn't seeing this issue before, so either I wasn't caught up with latest when I was testing before, or something external has since changed
- https://github.com/darklang/dark/blob/ba1617a21903ec50a64d7c977946b4097a0dd92a/services/apiserver-deployment-darklang/nginx.conf#L84 I tried removing this line to see what would happen - somehow this yields a situation in which _no_ HSTS headers are present
  ```
  expected: ["max-age=31536000; includeSubDomains; preload"]
  actual: []
```

Do you have any advice to investigate deeper?

pbiggar:
I actually wouldn't bother with this.

Looking at the nginx file, we actually would like to remove it in the future, and almost everything could be removed now except the static assets and compression. There doesn't seem to be a downside to having two identical hsts headers either, except the untidyness.

StachuDotNet:
Perplexing: I remove UseHsts + AddHsts from `ApiServer.fs` and still get the original error. (2 HSTS headers are present).

Just saw note - sounds good.

pbiggar:
Closing as it doesn't apply to darklang-next

PR #3746, opened 2022-04-27
Correct arg. order of Expect.equal in HSTS test

StachuDotNet:
Inspired by being confused by this error report: https://github.com/darklang/dark/issues/3745, as the expectation/actual are reversed.


pbiggar:
Merged. It won't go out cause the predeployment checks are off (my fault, unrelated), but this is small and in tests so that doesn't matter.

PR #3747, opened 2022-04-27
Fix query string behaviour

pbiggar:
This fixes the behaviour for http requests to Dark with the query string `?p=a%2Cb` which should be parsed into `{ p: "a, b" }`

I pulled the query encoding stuff into a new file (`HttpQueryEncoding`), it was confusing to me what was used in DvalReprExternal.

I also rewrote the parsing algorithm. The old one used `System.Web.HttpUtilist.ParseQueryString` and then tried to shimmy the result into the right shape. The new one is very basic, it just splits the query string using `&` and then `=`.

I realized that we had two implementations of the logic, one being the ParseQueryString I mentioned above, the other being in BwdServer.Server.getQuery, which also took a pre-parsed querystring and tried to turn it into the right shape. Isntead I passed the string through and used the same logic.

I also realized that `String.split` had bad behaviour that we weren't checking for anywhere. `String.split anything ""` resulted in `[""]`, not `[]` as you might more reasonably expect. So I fixed that and checked all the call sites for potential fallout.

A remaining problem here is that the HttpQueryEncoding is shared between the two HttpClients and the Http framework. Unclear what to do about that.

Questions:
- is HttpQueryEncoding a good name for the file?
- is BackendOnlyStdlib a good place for it (remembering that it currently is used for the http framework too) 
- should the logic be duplicated into each place that uses it? (currently the http client, legacy http client, and http server)


StachuDotNet:
> I also realized that String.split had bad behaviour that we weren't checking for anywhere. String.split anything "" resulted in [""], not [] as you might more reasonably expect.

:exploding_head:

fsharp-backend/src/BackendOnlyStdLib/HttpQueryEncoding.fs
```diff
@@ -0,0 +1,125 @@
...
+  | RT.DUuid uuid -> string uuid
+  | RT.DHttpResponse (RT.Redirect _) -> "null"
+  | RT.DHttpResponse (RT.Response (_, _, hdv)) -> r hdv
+  | RT.DList l -> "[ " + String.concat ", " (List.map r l) + " ]"
+  | RT.DObj o ->
+    let strs = Map.fold [] (fun l key value -> (key + ": " + r value) :: l) o
+    "{ " + (String.concat ", " strs) + " }"
+  | RT.DOption None -> "none"
+  | RT.DOption (Some v) -> r v
+  | RT.DResult (Error v) -> "error=" + r v
```
StachuDotNet, line 47: A bit tangential - why do these differ?

```
| RT.DError _ -> "error="
| ...
| RT.DResult (Error v) -> "error=" + r v
```

fsharp-backend/src/BackendOnlyStdLib/HttpQueryEncoding.fs
```diff
@@ -0,0 +1,125 @@
...
+  | RT.DFloat f -> LibExecution.DvalReprExternal.ocamlStringOfFloat f
+  | RT.DChar c -> c
+  | RT.DNull -> "null"
+  | RT.DDate d -> RT.DDateTime.toIsoString d
+  | RT.DDB dbname -> dbname
+  | RT.DErrorRail d -> r d
+  | RT.DError _ -> "error="
+  | RT.DUuid uuid -> string uuid
+  | RT.DHttpResponse (RT.Redirect _) -> "null"
+  | RT.DHttpResponse (RT.Response (_, _, hdv)) -> r hdv
```
StachuDotNet, line 40: can we rename this? I don't know what hdv is supposed to be

StachuDotNet:
I'm really happy to see this extracted out of `DvalRepr` and `DvalReprExternal` - that confused me. (was unexpected)

> Questions:
> 
> * is HttpQueryEncoding a good name for the file?
> * should the logic be duplicated into each place that uses it? (currently the http client, legacy http client, and http server)

I think `HttpQueryEncoding` is a good name, and appropriate for this.

I also _like_ that there's shared logic between Http framework and HttpClient for this - I can't see why our logic would need to differ. Similarly, I'd expect the same code to be used to read http headers in both cases, etc. It sort of reminds me of [Giraffe](https://github.com/giraffe-fsharp/Giraffe)'s cousin [Oryx](https://github.com/cognitedata/oryx).

If we come across a need to have separate logic between the two, so be it, but this seems good for now.

> * is BackendOnlyStdlib a good place for it (remembering that it currently is used for the http framework too)

I have a general feeling of "If something is pure and doesn't _need_ to be in the backend, it shouldn't be in the Backend-specific project." So, I'd search to put it in `LibExecution` or `LibExecutionStdLib`. Unfortunately, it feels out of place in both of those projects - LibExecution is more "core" stuff, and each file currently in `LibExecutionStdLib` is a bag of fns. :/

I'd be a bit tempted to throw it in `LibExecutionStdLib` and break the pattern, but also fine with `BackendOnlyStdlib`.

StachuDotNet (approved):
Approved with some minor thoughts about where to place HttpQueryEncoding.fs

fsharp-backend/src/BackendOnlyStdLib/HttpQueryEncoding.fs
pbiggar: I'd rather not touch that code. `hdv` is the http response dval - the dval contained in the http response (as opposed to the dval that contains it).

pbiggar:
I tried to clean it up a little bit, but that opened some new cans of worms and I think it's best to stop here. Added a comment instead.

PR #3748, opened 2022-04-27
Replace `function` with `match` throughout F# code

StachuDotNet:
The use of `function` can easily hide details of a match/function, similar to excluding type annotations in ambiguous situations.

This replaces the usages of this keyword with more simple `match`es.

As a side-benefit, this improved the formatting of code in several places - `function` seems to be formatted 2 spaces to the left of where one would expect. (at least, personally)

---

This does not replace the many usages within the stdlib fns. Worth a separate discussion (maybe some active pattern would be appropriate - I think it's fine as-is)


fsharp-backend/tests/Tests/DvalRepr.Tests.fs
```diff
@@ -373,9 +374,10 @@ module Password =
 
         let wrappedSerialize dval =
           dval
-          |> (function
-          | RT.DObj dvalMap -> dvalMap
-          | _ -> Exception.raiseInternal "dobj only here" [])
```
StachuDotNet, line 378: Here's an example of what I'd consider to be a formatting improvement.

pbiggar:
Code looks good.

Overall, I think I have a slight preference for the new code. I always found `function` a little impenetrable to read. wdyt?

StachuDotNet:
Agreed - `function` also comes with a downside of being slightly more difficult to decipher the difference when you `match` against something different. (it's a subtle thing, like including `,` at the end of lines in lists in some langs

PR #3756, opened 2022-04-30
Add F# code coverage analysis

### LINKED ISSUE #3688: Run F# backend against coverage tool
either as a one-off, or long-term through CI.

Some options to consider:
- [ ] altcover [1](https://www.hanselman.com/blog/altcover-and-reportgenerator-give-amazing-code-coverage-on-net-core) [2](https://github.com/SteveGilham/altcover/wiki/The-Visualizer#user-guide)
- [ ] [dotnet-coverage](https://docs.microsoft.com/en-us/dotnet/core/additional-tools/dotnet-coverage)
- [ ] [jetbrains dotCover](https://www.jetbrains.com/help/dotcover/Running_Coverage_Analysis_from_the_Command_LIne.html)
- [ ] [coverlet](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-code-coverage)
- [ ] [codecov](https://about.codecov.io/product/integrations)

StachuDotNet:
This is a work in progress aimed to add code coverage to the F# backend. It will resolve #3688.

[AltCover](https://github.com/SteveGilham/altcover) was chosen as the coverage tool.

Seemingly, all .NET code coverage tools assume that `dotnet test` will run your tests.
Expecto is not build to expose tests to `dotnet test`, either plainly or via `[<Tests>]` attributes common in .NET, so we needed to tackle this first. The NuGet packages `YoloDev.Expecto.TestSdk` and `Microsoft.NET.Test.Sdk` were added for this effect. After adding those, we also needed 1+ `[<Tests>]` attributes for the test-runner to discover - I opted to add this to the root `Tests` module containing all tests, rather than drop the attribute in many files. `[<Tests>]` seems to only be respected when attributing a `Test` binding, so `Tests.fs` had to adjust a bit, in order to expose such in a new `tests` binding.

With these changes, we can now `dotnet test` within the `fsharp-backend` directory, and it runs many of the tests well.

With `dotnet test` now discovering and running those tests successfully, the next step was to involve a code coverage tool - AltCover. Altcover is installed as a global tool via the `dotnet tool` CLI, installed in the dockerfile.

With altcover installed, we can run `dotnet test /p:AltCover=true`.
It seems to work at least at a basic level - here are some results:
```
A total of 773,417,102 visits recorded
Coverage statistics flushing took 6.862 seconds
Visited Classes 1807 of 15372 (11.76)
Visited Methods 3221 of 38990 (8.26)
Visited Points 10094 of 157756 (6.4)
Visited Branches 4714 of 134704 (3.5)
Maximum CRAP score 1176140

==== Alternative Results (includes all methods including those without corresponding source) ====
Alternative Visited Classes 3528 of 22131 (15.94)
Alternative Visited Methods 5981 of 59821 (10)
Alternative maximum CRAP score 1176140
```

Some outstanding tasks:
- [ ] Adjust altcover settings to only cover our source code.
  Currently, altcover is likely ensuring that we're testing things like TestUtils, and potentially some external dependencies
- [ ] Ensure _all_ of our tests in `Tests` are included in the new `tests` binding
  Several test modules are commented out - some of them fail during `dotnet test` _discovery_, and others fail during `dotnet test` _execution_. The ones that fail due in discovery, I forget the details of now; will have to follow up. The ones that fail during execution are likely failing due to the _sequenced_ aspects of those tests not being respected, though there may be additional reasons.
- [ ] Determine where we want to store the coverage analysis and reporting (in the directory), and set up to do such
  currently, the coverage ends up at `fsharp-backend/tests/Tests/coverage`. I imagine we'd want to create a _top-level_ `coverage` folder, with a sub-folder for pretty .html output
- [ ] Review the actual report! Tidy things until it looks reasonable (testing the right things, with the right results)
- [ ] Determine if we need the `altcover` NuGet package in addition to the global tool (i.e. can we remove it?)
  I added this when borrowing code from [MiniScaffold](https://github.com/TheAngryByrd/MiniScaffold/), and need to follow up
- [ ] Generate a nice (.html?) report from the coverage analysis
  I took an initial attempt with [ReportGenerator](https://github.com/danielpalme/ReportGenerator), but sticking with AltCover's [Visualizer](https://github.com/SteveGilham/altcover/wiki/The-Visualizer) may be our best move here. I haven't yet spent much time here. It may make sense to actually install the visualizer tool on the dev's host machine (outside of container) and then inspect the resultant analysis file, and use the traditional ReportGenerator for CI or other uses.
- [ ] Determine if/how we want to hook coverage analysis into CI
- [ ] Consider including tests from other test projects (fuzz? integration?)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -357,6 +357,7 @@ RUN curl -SL --output dotnet.tar.gz https://dotnetcli.azureedge.net/dotnet/Sdk/$
 
 RUN sudo dotnet workload install wasm-tools
 RUN dotnet tool install -g dotnet-sos
+
 # TODO: is this the right directory?
 RUN echo "plugin load /home/dark/.dotnet/tools/.store/dotnet-sos/5.0.160202/dotnet-sos/5.0.160202/tools/netcoreapp2.1/any/linux-x64/libsosplugin.so" > ~/.lldbinit
 
@@ -364,6 +365,10 @@ RUN echo "plugin load /home/dark/.dotnet/tools/.store/dotnet-sos/5.0.160202/dotn
 RUN dotnet tool install fantomas-tool --version 4.6.3 -g
 RUN curl https://raw.githubusercontent.com/darklang/build-files/main/ocamlformat --output ~/bin/ocamlformat && chmod +x ~/bin/ocamlformat
 ENV PATH "$PATH:/home/dark/bin:/home/dark/.dotnet/tools"
+
+# code coverage
+RUN dotnet tool install -g altcover.global
+RUN dotnet tool install -g dotnet-reportgenerator-globaltool
 
 #############
 # tunnel user
--- a/fsharp-backend/paket.dependencies
+++ b/fsharp-backend/paket.dependencies
@@ -28,6 +28,11 @@ nuget FSharp.Compiler.Service = 41.0.3
 nuget NReco.Logging.File = 1.1.3
 nuget SimpleBase = 3.1.0
 
+// Code Coverage
+nuget YoloDev.Expecto.TestSdk 0.12.12
+nuget Microsoft.NET.Test.Sdk 17.1.0-preview-20211109-03
+nuget altcover ~> 8.2.825
+
 // Services
 nuget Lib.AspNetCore.ServerTiming = 4.3.0
 nuget PusherServer = 4.6.1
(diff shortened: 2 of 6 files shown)


StachuDotNet:
This is stale, and I'm thinking to move to something other than AltCover if/when I revisit backend code coverage. (something more mainstream?). See #3688 for further discussion

Issue #3753, opened 2022-04-28
Make a new ingress+service for darkcustomdomain in the darklang namespace


pbiggar:
- [x] read https://cert-manager.io/docs/installation/kubernetes/
- [x] add an nginx-ingress-controller deployment
  - [x] copy over nginx-ingress-controller.yaml
  - [x] change the name and the part-of
  - [x] change namespace
  - [x] update nginx container version
  - [x] change the `--publish-service` config?
  - [x] make sure it only deals from the right ingress
  - [x] extract cert-manager config and move into another file
- [x] add a new service in darklang namespace
  - [x] add darkcustomdomain-ip-svc.yaml
  - [x] get new ip address
  - [x] set the namespace
  - [x] what is `app.kubernetes.io/part-of: ingress-nginx`? Do I need to change this?
- [x] add a new ingress
  - [x] copy and start darkcustomdomain-ingress.yaml.original
  - [x] copy over all `.spec.tls` and `.spec.rules`
- [x] adapt all the darkcustomdomain scripts
  - [x] configure cert-manager to use new ingress
  - [x] change the solver
  - [x] test them
- [ ] testing
  - [x] test access
  - [ ] add honeycomb triggers to continually check this works
- [x] add networkpolicy
  - [x] get it working
  - [x] trim down networkpolicy appropriately
- [x] adapt for aspnet
  - [x] pass through the server header
  - [x] can we do http3?

May need to restart cert-manager at some point:
`kubectl rollout restart deployment cert-manager -n cert-manager`

cert-manager is running in the cert-manager namespace, and doesn't seem to need to be changed. But it might need to be reconfigured

pbiggar:
I did a pass at this, and it went ok but it's not done yet.

Unfortunately, the new controller tried to update the old controller's config. It still seems to work but we're now in a precarious position. I tried to restart the old controller to fix it, but that didn't work.

It seems like rolling forward might be the best option. All custom domain users are already using the new BwdServer, so we're in a good position to do that.

The new controller tries to use the settings from the old controller and update the ip address and ingress info, so we have to specify in the settings which ingress we're talking about.

pbiggar:
- [ ] what about the health or healthz route - is that still accessible?

- [ ] add the nginx headers for aspnet.

pbiggar:
- [ ] switch customers to the new ingress via DNS
- [ ] delete the old ingress and nginx-controller
- [ ] carefully trim back the cert-manager stuff
[...]

PR #3767, opened 2022-05-01
Create a new custom-domain deployment in the k8s darklang namespace

pbiggar:
This adds a new nginx-ingress-controller in the darklang namespace, and includes a few updates to improve it. The idea is we'll have a new load balancer (with a new IP) that lives in the darklang k8s namespace and we'll switch the DNS over once it's up and running, which it seems to be.

For now I just want to get this merged and then I can think about next steps. This does most of #3753 but it doesn't actually confirm that we can add new certs just yet, and the custom-domain tooling has not been ported over.

Some of the changes here were to make the ingress behave a little more like the gce-ingress, which is to say, only adding three headers (via, x-forwarded-for, x-forwarded-proto) instead of the default 7. Those headers weren't available in gce and made it so that the same http requests to the same canvas via two different routes would give two different inputs, which is weird.

Other changes:
- update nginx-ingress-controller to 1.2.0 from 0.30.0
- since there are two load nginx load balancers, make sure the new ingress only talks to the one in the darklang namespace
- rename the various components to match our naming scheme for other k8s components
- break compound yaml files into multiple files
- add a network policy


pbiggar:
Since all these components are already applied, this is safe to merge.

Issue #3642, opened 2022-03-31
Use dotnet-wasi-sdk once it supports AOT compilation


StachuDotNet:
It'd be great to ship one big .wasm file rather than many .dlls and other files.

[dotnet-wasi-sdk](https://github.com/stevesandersonms/dotnet-wasi-sdk) allows us to do this.

We've tested (after upgrading to .net 7 in another branch) that Dark works with `dotnet-wasi-sdk`, but the project does not currently support AOT builds. Once it does, or once this project is assumedly merged into an official feature of .NET, we should reconsider this.

pbiggar:
Closing as it doesn't apply to darklang-next

PR #3788, opened 2022-05-09
Try Blazor AOT after replacing Blazor Worker

StachuDotNet:
## Phase 1: make blazor-loading simpler
- [x] Build an un-minified copy of `blazor.webassembly.js` from the the [aspnetcore repo](https://github.com/dotnet/aspnetcore)
  this took a `./restore.sh` and `./build.sh` in the `src/Components` directory, after adjusting `Components/Web.JS/src/webpack.config.js` such that `minimize: false` under the `optimization` option
- [x] Copy that into Dark (fsharp-backend/src/Wasm/static` as `blazor.webassembly.js`)
- [x] Prepare a custom resource-loader to respect Dark having files at `{static host}/blazor/*` rather than `.../_framework/*`
  I tossed this into `appsupport.js` for now
- [x] Reference the new `blazor.webassembly.js` in `ui.html`
  `autostart="false"` is needed so `appsupport` has a chance to avail some relevant pieces
- [ ] With that in place, load Blazor resources fully, ready for interop
  I'm loading this with `Blazor.start({ loadBootResource: loadBootResource })`
  Most resources (boot.js and all the .dlls) are loading again well.
  The next failure type (not currently blocked - just pausing for a break now) is on `dotnet.wasm` - for some reason it's trying to locate at `http://darklang.localhost:9000/a/dotnet.wasm` rather than using the normal resource-loader. Maybe we need to pass something else into `Blazor.start`

## Phase 2: Find the minimal bug relevant to AOT builds
Once the simpler blazor-loading implementation is in place, the goal is to get to _some_ working build using AOT compiles, and a minimal step required to _break_ that

- [ ] Gut as much of the Wasm-build Dark as possible (e.g. most fns)
- [ ] Get basic calls working
- [ ] Add as much back in as we can, excluding anything that breaks
- [ ] Collect a report on the minimal breakage
- [ ] Figure out what to do from there

Misc. notes and TODOs:

- For now I'm not trying to load both the old-style BlazorWorker and the more 'raw' version - too much distracting network traffic
-  blazor.webassembly.js has changed a bit since we last copied it, so I opted to grab a fresher copy, more relevant for conversations with aspnetcore folks
- [ ] Review #3642 changes; there may be something here worth salvaging.
- [ ] Find/create an issue in `aspnetcore` of `Blazor.start` not being documented
- [ ] Publish npm package for `blazor.webassembly.js`, and document? create bindings (.ts, .fs, rescript)

### DIFF
--- a/fsharp-backend/src/ApiServer/ApiServer.fs
+++ b/fsharp-backend/src/ApiServer/ApiServer.fs
@@ -68,6 +68,19 @@ let addRoutes
   let std = standardMiddleware
   let html = htmlMiddleware
 
+  // dotnet.js loads dotnet.wasm without respecting our custom resource-loader
+  // this is a _rough_ hack to deal with such. TODO: remove this hack
+  let loadDotnetWasm : HttpHandler =
+    (fun (ctx : HttpContext) ->
+      task {
+        return
+          ctx.Response.Redirect(
+            "http://static.darklang.localhost:9000//blazor/dotnet.wasm",
+            false
+          )
+      })
+  addRoute "GET" "/a/dotnet.wasm" std None loadDotnetWasm
+
   let api name perm f =
     let handler = jsonHandler f
     let route = $"/api/{{canvasName}}/{name}"
@@ -93,6 +106,7 @@ let addRoutes
     let userInfo = loadUserInfo ctx
     Exception.raiseInternal "triggered test exception" [ "user", userInfo.username ]
   addRoute "GET" "/a/{canvasName}/trigger-exception" std R exceptionFn
+
 
   api "add_op" RW AddOps.addOp
   api "clean_ops" RW AddOps.clean
--- a/backend/templates/ui.html
+++ b/backend/templates/ui.html
@@ -57,7 +57,8 @@
       type="text/javascript"
       src="//{{STATIC}}/unsupportedBrowser.js"
     ></script>
-    <script type="text/javascript" src="//{{STATIC}}/BlazorWorker.js"></script>
+    <!-- <script type="text/javascript" src="//{{STATIC}}/BlazorWorker.js"></script> -->
+    <script type="text/javascript" src="//{{STATIC}}/blazor.webassembly.js" autostart="false"></script>
     <script type="text/javascript" src="//{{STATIC}}/appsupport.js"></script>
   </head>
   <body onKeyDown="stopKeys(event)" onKeyUp="stopKeys(event)"></body>
--- a/client/src/appsupport.js
+++ b/client/src/appsupport.js
@@ -6,6 +6,27 @@ const mousewheel = function (callback) {
   });
 };
 
+// WIP...
+let appRoot = `${window.location.protocol}//${staticUrl}`;
+function toDarklangSetupUri(file) {
+  file = file.replace(/_framework\//, "");
+  file = "/blazor/" + file;
+  return `${appRoot}/${file}`;
+}
+window.loadBootResource = (resourceType, file, incomingUrl, hash) => {
+    const params = {ass: resourceType, file, incomingUrl, hash};
+    console.log('Loading...', params);
+
+    const url = toDarklangSetupUri(file);
+    if (resourceType == 'dotnetjs'){
+      return url;
+    } else {
+      return fetch(url, { method: 'GET', cache: 'no-cache', })
+    }
+};
+
+
+
 // ---------------------------
 // Check unsupported browser
 // ---------------------------
@@ -323,7 +344,8 @@ window.Dark = {
   analysis: {
     useBlazor: (function () {
       const urlParams = new URLSearchParams(window.location.search);
-      return urlParams.get("use-blazor");
+      const v = urlParams.get("use-blazor");
+      return v == "true";
     })(),
     requestAnalysis: function (params) {
       if (window.Dark.analysis.useBlazor) {
--- a/fsharp-backend/src/Wasm/static/blazor.webassembly.js
+++ b/fsharp-backend/src/Wasm/static/blazor.webassembly.js
@@ -1,0 +1,4355 @@
+/******/ (() => { // webpackBootstrap
+/******/ 	"use strict";
+var __webpack_exports__ = {};
+
+;// CONCATENATED MODULE: ../../../JSInterop/Microsoft.JSInterop.JS/src/dist/Microsoft.JSInterop.js
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// This is a single-file self-contained module to avoid the need for a Webpack build
+var DotNet;
+(function (DotNet) {
+    window.DotNet = DotNet; // Ensure reachable from anywhere
+    const jsonRevivers = [];
+    const byteArraysToBeRevived = new Map();
+    const pendingDotNetToJSStreams = new Map();
+    const jsObjectIdKey = "__jsObjectId";
+    const dotNetObjectRefKey = "__dotNetObject";
+    const byteArrayRefKey = "__byte[]";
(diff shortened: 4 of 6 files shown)


StachuDotNet:
Still haven't figured out why `dotnet.wasm` is failing to download.
image

The file is requested multiple times:
- once at the correct URL (`http://static.darklang.localhost:9000//blazor/dotnet.wasm`), which 200s
- twice at `http://darklang.localhost:9000/a/dotnet.wasm`, which 500s
- each of those 500s is somehow paired with this console error message:
  `Unhandled promise rejection unhandledrejection RuntimeError: abort(RuntimeError: abort(both async and sync fetching of the wasm failed). Build with -s ASSERTIONS=1 for more info.). Build with -s ASSERTIONS=1 for more info.`
- searching for that text, I find it (only) in 3 files (one relevant):
  ```
  dark@dark-dev:~/app$ grep -rnw '.' -e 'Build with -s ASSERTIONS=1'
  ./backend/static/blazor/dotnet.6.0.4.dze57nx010.js:1:...
  ./fsharp-backend/Build/out/Wasm/Debug/net6.0/dotnet.js:1:...
  ./fsharp-backend/Build/out/Wasm/Debug/net6.0/wwwroot/_framework/dotnet.6.0.4.dze57nx010.js:1:...
  ```

Something is loading dotnet.wasm, outside of blazor.webassembly.js and it seems to break the loading.

Some questions to pursue:
- [ ] review the above screen-shot some more; I haven't looked too deeply yet
- [ ] what else is loading this file?
- [ ] why is it doing so _twice_?
- [ ] how/can I get it to load via our custom resource-loader?
- [ ] is the problem in `dotnet.js`, or is it something in blazor.webassembly.js?
- [ ] maybe it's worth reading through blazor.webassembly.js more to figure out how/where it I/O's with blazor.js

pbiggar:
> The next failure type (not currently blocked - just pausing for a break now) is on dotnet.wasm - for some reason it's trying to locate at http://darklang.localhost:9000/a/dotnet.wasm rather than using the normal resource-loader. Maybe we need to pass something else into Blazor.start

This script is being run at `darklang.localhost/a/canvasname`, while the old script ran in a web worker, so presumably it had to http path to be relative to. There's a function somewhere where we translate the blazor path into a url, you can probably use that to solve this.

StachuDotNet:
If you're referring to `toDarklangSetupUri`, that's already been ported, used in the code here, and respected in most of the requests. The issue is that there are _extra_ calls which don't seem to be flowing through the `loadBootResource` param passed into `Blazor.start`. (our argument of `loadBootResource` includes flowing through `toDarklangSetupUri`)

StachuDotNet:
- [ ] actually follow `Build with -s ASSERTIONS=1 for more info.`
  (I'm assuming this is a flag to set in the wasm.fsproj - haven't looked yet)

StachuDotNet:
I just added a _rough_ hack to proceed - added a new route in apiserver to redirect the attempted /a/dotnet.wasm requests to the static host. It seems to get me past that step. Something to follow up on.

fsharp-backend/src/ApiServer/ApiServer.fs
```diff
@@ -68,6 +68,17 @@ let addRoutes
   let std = standardMiddleware
   let html = htmlMiddleware
 
+  // this is a bad hack since I can't seem to adjust dotnet.js to load dotnet.wasm anywhere else...
+  // todo: load the real dotnet.wasm here... using httpclient? gross.
+  // todo: a 301 would be a bit less hacky - but would it work?
```
StachuDotNet, line 73: it does.

StachuDotNet:
Blazor.start() is now executing without error, but after throwing in many console.logs, I've seen that we're getting stuck in an `createEmscriptenModuleInstance` fn - trying to diagnose why.

StachuDotNet:
I spent a while trying to get this (using blazor.webassembly.js rather than our BlazorWorker) working with .NET 6.
It kept getting stuck, with no console errors.

Then, I tried the same approach (using blazor.webassembly.js rather than BlazorWorker) against the .NET 7 branch, and I'm able to get basic evals working. (without AOT - about to try such)

I'm guessing that the issue here is that I was using a more 'modern' blazor.webassembly.js, incompatible with .net 6 for some reason. I'm going to continue working on the .net7 branch tonight, but plan to follow up and generate a raw .net 6-friendly blazor.webassembly.js tomorrow if possible. (by checking out `aspnetcore` at an earlier point).

Happy to finally get evals working with blazor.webassembly.js rather than BlazorWorker, now that it's finally loading fully rather than getting stuck.

Now, I'm going to try AOT builds with this set-up, and go from there.

StachuDotNet:
When attempting to run against AOT builds (using the simplified blazor-loading in .net 7 branch), I'm coming across this error:
image

`dotnet.js` is looking for `Microsoft.AspNetCore.Components.WebAssembly`, but
- it doesn't exist in the `publish`ed build, so doesn't end up in `backend/static/blazor`
- because of this, it also isn't listed in `blazor.boot.js`, which `dotnet.js` uses to try to find the appropriate `Microsoft.AspNetCore.Components.WebAssembly` dll

Why do some .dlls simply not exist in the `publish` directory? This is a fundamental question that I'm not getting. I can't figure out why `Microsoft.AspNetCore.Components.WebAssembly` wouldn't be published with the other .dlls

StachuDotNet:
Sneaking in `Microsoft.AspNetCore.Components.WebAssembly` (copying the .dll from non-publish artifacts, and adding the relevant line to blazor.boot.js) got a bit further, but failed shortly after.
image.

Not sure.

StachuDotNet:
https://github.com/fsprojects/Paket/issues/3999 This may be relevant to our issue - trying the workaround mentioned in comments.

StachuDotNet:
Made some great Blazor discoveries yesterday. Following up on them today.

I need to expand upon these notes more, and push my local code[1], but here are some take-aways:
- After varied adjustments, I'm able to load and run the AOT build with the `blazor.webassembly.js` script, rather than our hack
  Note: This testing has been _without_ a web worker. Will need to follow up to not block the UI thread
- My to-go test case is simple (`List::repeat 10000 1 |> List::length`), but it shows clear differences between AOT and non-AOT results, so I'm happy with it for now
- The root cause of perf, for my testing, is: serialization/deserialiation.
- Specifically, my test's results include some intermediate results for the `List::repeat` - that long list of `1`s is taking ages to serialize
- The serialization of these 10k elements takes ~13s with a non-AOT build, and ~9s with an AOT build. While this isn't a huge improvement, it's great that:
  1. the AOT build is working
  2. we see actual improvement with the AOT build
  3. seemingly, we've centralized the problem to serialization (again, at least for my cases)

[1] my local code is in a bit of a mess now; for now I've pushed the mess to [another branch](https://github.com/StachuDotNet/dark/commit/b911f4c1a2ab1f2ae790f532c2325217c31b0885) - currently tidying and including here.

StachuDotNet:
Some additional notes:
- I had to do things to our Wasm.fsproj to get the AOT build to work, because of some bug in Paket, I believe. This might be fine for now, but we're including a lot of stuff that doesn't really matter right now - I think we can trim a lot more with a bit more research. (and maybe a contribution to Paket)
- the mono console warnings that I saw before in trying to use AOT builds seemingly relate to tasks - I had to update some things in Wasm.fs to get around this, though I'm hoping to simplify and have more answers here

StachuDotNet:
@pbiggar you mentioned hitting perf issues on various canvases before while just poking around. My guess is that those perf issues were _also_ due to the serialization bottleneck - if you happen to recall if those handlers would involve large[1] inputs, outputs, or intermediate results, that'd be promising context to have.

[1] "large" here means "involves many elements" such that serialization may take a while.

StachuDotNet:
Closing in favor of tidier PRs

PR #3801, opened 2022-05-13
Use System.Text.Json for Blazor IO Serialization

StachuDotNet:
for massive performance improvements


StachuDotNet:
Doing a quick test of integration tests before I request review

pbiggar:
I suspect that the json format will no longer be compatible between the client and Wasm, but I guess that's a next step, and since this doesn't change the default that seems fine.

PR #3819, opened 2022-05-17
Queueworker pass 11 - prevent stackoverflow

pbiggar:
When moving traces out of the main execution, I added them to a dictionary. However, it seems there is a very-bad performance impact from this: each item we add to a dictionary has a HashCode calculated, and hashcodes to Lists can be very slow.

This PR makes these more performant. We calculate our own hash for FunctionResults, so that we don't duplicate saving the data (still using Dictionarys). For FunctionArguments, don't use the arguments to hash a dictionary, instead just save the whole list (I'm not sure we saved anything by having a dictionary).

### DIFF
--- a/scripts/build/compile
+++ b/scripts/build/compile
@@ -234,6 +234,7 @@ def fsharp_backend_test():
 def fsharp_tool_restore():
   start = time.time()
   build = "./scripts/build/_dotnet-wrapper tool restore"
+
   return run_backend(start, build)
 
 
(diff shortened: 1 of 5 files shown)


StachuDotNet (approved):

scripts/build/compile
```diff
@@ -284,6 +285,7 @@ def fsharp_backend_quick_build():
     command = "build"
 
   build = f"./scripts/build/_dotnet-wrapper {command} \
+      /clp:ForceConsoleColor \
```
StachuDotNet, line 288: great

PR #3820, opened 2022-05-17
Support reading more events from the queue

pbiggar:
Also improve comments a lot

The reason for this is that pulling events from the queue is quite slow right now - we do a full loop iteration for each event.

For context, here are events all queued at roughly the same time via cron:

image

Even though we have the 2 workers set to take 8 events each, it takes over two seconds to get all 27 values out of the queue (most complete in about 250ms). Really we have the capacity to start them all at once, so let's not be coy about this.


StachuDotNet:
Should we add an additional test against the case of allowing more than 1 message to be dequeued?

StachuDotNet:
Otherwise, looks great!

StachuDotNet (approved):

Issue #3811, opened 2022-05-16
`use-blazor=true` flag currently broken in production


StachuDotNet:
I just noticed that ?use-blazor=true in the Editor currently is broken - many console errors around invalid integrity attributes.

AFAIK this was working Friday afternoon - not yet sure what may be wrong here.

StachuDotNet:
![image](https://user-images.githubusercontent.com/906686/168647234-e1dc8d30-c7cc-4cf9-8271-b61dbc22afe6.png)

StachuDotNet:
I haven't encountered this error before @pbiggar if you have any advice, I can look into this.

StachuDotNet:
It seems some of the blazor assets (dlls) are 200-ing OK when being loaded, but others are 404-ing.

pbiggar:
Relevent, don't know if this is it https://stackoverflow.com/a/71171026/104021

pbiggar:
So what I think this is all referring to is the hashes loaded via blazor.boot.json

<img width="1010" alt="image" src="https://user-images.githubusercontent.com/181762/168648158-6c0ca220-cab8-4931-86b6-46b49be3423d.png">

pbiggar:
I don't know why some of them would be bad, but perhaps as the Stackoverflow link suggests, we are perhaps loading them from cache incorrectly?

pbiggar:
> It seems some of the blazor assets (dlls) are 200-ing OK when being loaded, but others are 404-ing.

Hmm, I was thinking it wasn't really 404-ing, just that was how the SRI check failure appears in the chrome UI. But it could be the other way: they're 404-ing and we're getting an SRI *because* they're 404-ing.

StachuDotNet:
What cache are you referring to?

pbiggar:
Stackoverflow said "Manually deleted all bin and obj folders in the solution and then redeployed. After this everything worked."

pbiggar:
Cleared the CI cache in this build: https://app.circleci.com/pipelines/github/darklang/dark/2949/workflows/fdf26f5c-e511-4c96-865f-1cd0f76fcdc1

StachuDotNet:
Unfortunately that didn't seem to help

StachuDotNet:
Resolved recently - this was due to bad CDN uploads in CI, because of the more+larger Blazor static assets resulting in more big concurrent calls to gsutil cp

PR #3831, opened 2022-05-19
F# Analysis: Ahead-of-Time Compile, use published build, and use in integration tests

### LINKED ISSUE #3168: Optimize / trim Blazor
You should be able to compile the Blazor files using RunAOTCompilation. However, it failed with the following error:

```
/usr/share/dotnet/sdk/6.0.100-rc.1.21417.9/Sdks/Microsoft.NET.Sdk.BlazorWebAssembly/targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets(446,5): error : System.ArgumentException: An item with the same key has already been added. Key: /home/dark/.nuget/packages/fsharp.core/5.0.3-beta.21352.5/contentFiles/any/netstandard2.0/FSharp.Core.xml [/home/dark/app/fsharp-backend/src/Wasm/Wasm.fsproj]
/usr/share/dotnet/sdk/6.0.100-rc.1.21417.9/Sdks/Microsoft.NET.Sdk.BlazorWebAssembly/targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets(446,5): error :    at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior) [/home/dark/app/fsharp-backend/src/Wasm/Wasm.fsproj]
/usr/share/dotnet/sdk/6.0.100-rc.1.21417.9/Sdks/Microsoft.NET.Sdk.BlazorWebAssembly/targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets(446,5): error :    at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value) [/home/dark/app/fsharp-backend/src/Wasm/Wasm.fsproj]
/usr/share/dotnet/sdk/6.0.100-rc.1.21417.9/Sdks/Microsoft.NET.Sdk.BlazorWebAssembly/targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets(446,5): error :    at Microsoft.NET.Sdk.BlazorWebAssembly.ComputeBlazorPublishAssets.GroupResolvedFilesToPublish(Dictionary`2 resolvedFilesToPublishToRemove, Dictionary`2 resolvedAssemblyToPublish, Dictionary`2 satelliteAssemblyToPublish, Dictionary`2 resolvedSymbolsToPublish, Dictionary`2 resolvedNativeAssetToPublish) [/home/dark/app/fsharp-backend/src/Wasm/Wasm.fsproj]
/usr/share/dotnet/sdk/6.0.100-rc.1.21417.9/Sdks/Microsoft.NET.Sdk.BlazorWebAssembly/targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets(446,5): error :    at Microsoft.NET.Sdk.BlazorWebAssembly.ComputeBlazorPublishAssets.Execute() [/home/dark/app/fsharp-backend/src/Wasm/Wasm.fsproj]
```

Nothing I could do would make this work.

I suspect that once we're not on bleeding edge F# & .NET (we're on a nightly build to make something else work), this might work.

I doubt trimming will work, as we use reflection in the JSON serializers. It doesn't look like [source generators](https://www.stevejgordon.co.uk/playing-with-system-text-json-source-generators) are going to be in F# anytime soon, but it might be possible to write the source generator in C#.

StachuDotNet:
- the Wasm project is now AOT-compiled, and trimmed
  - increased the `resource_class` of the corresponding step in CI
    without such, we get a SIGKILL due to high RAM usage
- integration tests now use Blazor as the analysis engine
  Blazor takes longer to load than the OCaml wasm did, so:
  - increased the `resource_class` of the corresponding step in CI
  - increased the playwright timeout
- BlazorWorker is now "warmed up" with a `2+3` expr before declaring that it's ready for eval requests
  it was found that the first eval takes some time to run
- adjusted our `compile` script to use published build, and remove Wasm-specific hacks
- tidies Wasm.fs (no-op)

Will resolve #3168


StachuDotNet:
integration tests are still being flaky due to blazor load time. I'm nearly complete a commit that will only load Blazor once per machine/worker, which should allow us to decrease the playwright timeout and `resource_class` of that CI step.

StachuDotNet:
https://app.circleci.com/pipelines/github/darklang/dark/3006/workflows/5520aad8-306c-4ffb-8868-5335ac97d955/jobs/32334

While I still want to tidy a few things, this would have passed if the `Assert the worktree is clean` error hadn't occurred. ~5mins (compared to ~4min before), and using only 1 machine, instead of the 4 that `main` currently demands. 

- [ ] lower resource class to large
- [ ] increase concurrency from 3 to 4 or more
- [ ] get the `Assert the worktree is clean` error resolved
  (Actually, maybe this is OK - is this the thing that resolves only after it's been merged to `main`?)
- [ ] uncomment one test currently commented (should be at 70)
- [ ] break down the commit into more reasonable chunks
- [ ] generally tidy code

StachuDotNet:
I haven't been able to fully figure out the problem where integration tests hang. It seems it only occurs when we retry a text (more specifically, maybe, when we re-create a context?). Regardless, getting closer, and integration tests are imo easier to work with now.

StachuDotNet:
I believe we're now at a merge-able state. The build and tests pass*.

There's still some flakiness with the build, but that's not relevant to the PR - just F# tests failing, documented in a separate issue.

*Technically, the integration test step is _failing_, but on the `Assert the worktree is clean` substep. I'm not sure what to do with that. [Here](https://app.circleci.com/pipelines/github/darklang/dark/3027/workflows/38d728b7-c05d-45ff-b507-43672acb5373/jobs/32559/parallel-runs/1/steps/1-112) is a sample failure. Please advise.

I took advantage of that sub-step failing by using it as a pivot point for "run-run since last failure," ensuring integration tests passed consistently. (I got 3x in a row, happy with that)

---

I spent time attempting to adjusts tests such that we don't load Blazor more than once per worker, but ran into a number of snags. I think there's valuable work to salvage there, but I'll leave it for a separate PR.

Once this is merged, we'll see if this will or won't conveniently help the current issue of #3811. If it doesn't help, I'll pivot to that, otherwise I'll focus on testing so we can merge #3535.

pbiggar:
> *Technically, the integration test step is failing, but on the Assert the worktree is clean substep. I'm not sure what to do with that. [Here](https://app.circleci.com/pipelines/github/darklang/dark/3027/workflows/38d728b7-c05d-45ff-b507-43672acb5373/jobs/32559/parallel-runs/1/steps/1-112) is a sample failure. Please advise.

So the assertion tests that CI does not change any check-ed in files. In this case, integration-tests/package-lock.json is changed. I think you deleted it? Unclear why. We should certainly have a lock file checked in so maybe replace it with the latest version?

StachuDotNet:
> > *Technically, the integration test step is failing, but on the Assert the worktree is clean substep. I'm not sure what to do with that. [Here](https://app.circleci.com/pipelines/github/darklang/dark/3027/workflows/38d728b7-c05d-45ff-b507-43672acb5373/jobs/32559/parallel-runs/1/steps/1-112) is a sample failure. Please advise.
> 
> So the assertion tests that CI does not change any check-ed in files. In this case, integration-tests/package-lock.json is changed. I think you deleted it? Unclear why. We should certainly have a lock file checked in so maybe replace it with the latest version?

I didn't delete it - it's changed, as I've upgraded playwright and included the node types package.

pbiggar:
> I didn't delete it - it's changed, as I've upgraded playwright and included the node types package.

It's been deleted in this PR:

image

StachuDotNet:
Wild. I see that same line count difference, but this is my view:
image

pbiggar:
No, this is my mistake. You're right, it's not deleted.

So then what's happening is that `npm update` is changing the lock file because - I assume - we're not using the latest packages? So presumably you can solve the CI problem by updating the lock file to the latest versions?

fsharp-backend/src/Wasm/Wasm.fs
```diff
@@ -260,51 +264,50 @@ module Eval =
...
-  static member getPostMessageDelegate() : InvokeDelegate =
-    let assemblyName = "System.Private.Runtime.InteropServices.JavaScript"
-    let typeName = "System.Runtime.InteropServices.JavaScript.Runtime"
-
-    // Get the `self` object from the webworker.
-    let sourceAssembly : Assembly = Assembly.Load(assemblyName)
-    let typ = sourceAssembly.GetType(typeName)
+  static member GetGlobalObject(_globalObjectName : string) : unit = ()
+
+  static member selfDelegate =
```
pbiggar, line 283: Are `static member x()` and `static member x` equivalent? If not, why the change?

StachuDotNet:
Just pushed a commit to try using one larger machine. (though xlarge is the largest I have access to, I believe).

Regarding the package-lock.json, will investigate shortly. Some theories:
- [ ] one of the packages is outdated
- [ ] my updated package-lock.json is of "version 2," which may be incompaible with the npm version in CI. I forget if I installed the packages from my host machine in that directory, or from inside the container.

pbiggar (approved):
This looks good but I'm really confused. This PR doesn't seem to do all that much, so why does it work now when it didn't before?

fsharp-backend/src/Wasm/Wasm.fs
StachuDotNet: They're not the same - the first of these requires the unit param.
In this case, the unit param seemed useless, and I have a habit of removing useless params (if something can be just a simple value, why not?)

In this specific case, by the end of executing this file `postMessageDelegate` was already resolved (no hanging () argument) anyway, so it feels the same to me. `getPostMessageDelegate()` had a `()` param, but was only used to create `postMessageDelegate`. This all felt like more noise than useful, so I simplified to a plain `selfMessageDelegate`

StachuDotNet:
> This PR doesn't seem to do all that much, so why does it work now when it didn't before?

Really good question. I have two theories:
- `wasm-tools` improved in some way (we [don't lock](https://github.com/darklang/dark/blob/bd1bd7408be9b34099bbf79273f3d716c8ebd57b/Dockerfile#L362) the version currently in our dockerfile)
- serialization was not only the problem with run-time, but also AOT builds - switching to STJ resolved our issue

---

Integration tests just took 9mins in CI with one xlarge machine. Is this OK? I think it can be improved still dramatically once I/we figure out how to safely share Blazor loads across tests.

pbiggar:
> Is this OK for now? I think it can be improved still dramatically once I/we figure out how to safely share Blazor loads across tests.

Speeding up CI is a high priority, but not high enough to stop this from merging IMO. I would do it right after though (if you can stomach it :))

pbiggar:
> wasm-tools improved in some way (we [don't lock](https://github.com/darklang/dark/blob/bd1bd7408be9b34099bbf79273f3d716c8ebd57b/Dockerfile#L362) the version currently in our dockerfile)

Feels like we should

fsharp-backend/src/Wasm/Wasm.fs
pbiggar: I meant given it's called the same way before and after, did it work before or does it work now or if both then why?

Not a big deal, if it works it works, I'm just confused.

StachuDotNet:
> > [we don't lock dotnet tool versions]
> 
> Feels like we should

I added this card: https://github.com/darklang/dark/projects/1#card-82295385

>> [CI takes 9mins now, but is only on one machine]
>
> Speeding up CI is a high priority, but not high enough to stop this from merging IMO. I would do it right after though

Deal! CI time is def. top of mind for me too.

StachuDotNet:
Somehow, an `npm update` within the container vastly simplified the `package-lock.json`, relative to `main`.
Seemed odd to me, but CI is happy now. OK?

pbiggar:
Yay!

I think this doesn't change the default right? Is the plan to test manually and then followup with something to change the default?

StachuDotNet:
This doesn't change the default - yes, hoping to merge the other PR (#3535) if all's well.

StachuDotNet:
We'll see if this PR magically resolves #3811, otherwise we still have some work to do there. So far, I'm ignorant of next steps there, but haven't thought about it in a while.

StachuDotNet:
After this deploy, seeing a `Blazor worker failure SyntaxError: Unexpected end of JSON input` when `use-blazor=true` - investigating. (Note, this _is_ an improvement, and 3811 is semi-resolved. now we just have a different issue.)

StachuDotNet:
blazor.boot.json is 404-ing - probably easily fixable, looking.

PR #3835, opened 2022-05-19
Fix httpclient

pbiggar:
This solves a number of problems with the new httpclient, where the behaviour of the new client(s) differed from the old one(s).

This was a regression in the transition, though it didn't affect many people. To be affected, a user would have to make a http call which redirected from a non-json content-type to a json one. I feel we should make this change as:

- existing users who have been transitioned did not seem to be affected
- if existing users who have been transitioned were affected, we would want to revent and fix to the old behaviour so that they don't have to write code
- known users of the queues are affected by this

There are three problems that I observed when looking through this:
1) when there are multiple headers with the same name (due to redirects), the last one (the one from the redirect destination) should win (this is bad behaviour, but matches the old response)
2) the new http client added extra query params on redirect
3) setting a content-length header crashed the .NET httpclient 

Problem 1 was that we observed that a user made a HTTP request to a server, and the response was consistently "text/html" in the OCaml httpclient, and consistently "application/json" in the F# client. But, the OCaml client was parsing the JSON while the F# client was not!

What was happening was that there was a http redirect happening. The first response had a content-type of "text/html" while the second response had a content-type of "application/json".

In OCaml, we concatenated the responses, and then checked to see if there was a content-type header. We used the first content-type header, which was "application/json", to determine whether to parse the string as JSON. However, we then joined the headers into a map, and so the last content-type header, "text/html", actually was the header returned to the client.

We now replicate this behaviour in F#. This also solves a number of other OCAMLONLY/FSHARPONLY situations.

This also fixes problems 2 and 3, which were found along the way, whose causes and fixes are more straightforward.

This is the only thing blocking remaining queueworker changes.


fsharp-backend/tests/httpclienttestfiles/v0/request-header-override-content-length-post-bad-fsharp.test
```diff
@@ -1,32 +1,33 @@
...
 
-Redirect destination reached
+"Hello back"
 
 [test]
 // FSHARPONLY
-// This is here to be a destination for valid redirect tests
-(let response = HttpClient.get_v4_ster "http://URL" {} {} in
+(let reqHeaders = { ``Content-length`` = "6" }
+ let response = HttpClient.post_v5_ster "http://URL" "morethansix" {} reqHeaders in
```
StachuDotNet, line 20: post_v5 in a v4 test?

fsharp-backend/tests/httpclienttestfiles/v0/request-header-override-content-length-post-bad-ocaml.test
```diff
@@ -0,0 +1,35 @@
...
+Date: xxx, xx xxx xxxx xx:xx:xx xxx
+Content-type: text/plain; charset=utf-8
+Content-Length: LENGTH
+
+"Hello back"
+
+[test]
+// OCAMLONLY
+(let reqHeaders = { ``Content-length`` = "nonsense"; }
+ let response = HttpClient.post_v5 "http://URL" "morethansix" {} reqHeaders in
```
StachuDotNet, line 20: v5 in a v0 test

fsharp-backend/tests/httpclienttestfiles/v0/request-header-override-content-length-post.test
```diff
@@ -0,0 +1,32 @@
...
+HTTP/1.1 200 OK
+Date: xxx, xx xxx xxxx xx:xx:xx xxx
+Content-type: text/plain; charset=utf-8
+Content-Length: LENGTH
+
+"Hello back"
+
+[test]
+(let reqHeaders = { ``Content-length`` = "6" }
+ let response = HttpClient.post_v5_ster "http://URL" "morethansix" {} reqHeaders in
```
StachuDotNet, line 19: v5 in a v0 test

StachuDotNet:
There are a number of v5 fn calls in non-v5 tests; I'll revisit once those are resolved?

fsharp-backend/tests/httpclienttestfiles/v0/request-header-override-content-length-post.test
pbiggar: Nice catch, will fix them all.

PR #3869, opened 2022-05-23
Add start of incident doc

pbiggar:
I figure this is a decent start. A good next step would be to annotate everything with full links and commands.

### DIFF
--- a/docs/production/what-to-do-if-something-goes-wrong.md
+++ b/docs/production/what-to-do-if-something-goes-wrong.md
@@ -25,7 +25,7 @@ The most likely candidates are:
 - bad/broken k8s configuration was set
 - a user, canvas, or handler that's going wild or hitting a slow part of our code
 - a Denial-of-service attack
-  - an attack is much less likely than an accidental one by a user)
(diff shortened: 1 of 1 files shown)


docs/production/what-to-do-if-something-goes-wrong.md
```diff
@@ -0,0 +1,187 @@
...
+## What could be wrong
+
+Typically, you have some evidence of something going wrong. You need to turn that into a cause, and then find a way to undo that.
+
+The most likely candidates are:
+- bad/broken code got deployed
+- bad/broken k8s configuration was set
+- a user, canvas, or handler that's going wild or hitting a slow part of our code
+- a Denial-of-service attack
+  - an attack is much less likely than an accidental one by a user)
```
StachuDotNet, line 28:
```suggestion
  - an attack is much less likely than an accidental one by a user
```
or ( at start

StachuDotNet (approved):
Separately, I'm going to re-read this later and see if I have follow-up questions, but this looks good!

PR #3871, opened 2022-05-23
Add new pusher credentials for a different pusher app

pbiggar:
This one is in Oregon. I'm hoping that it will prevent the timeouts we're seeing (our k8s cluster is in oregon).

I'm a little sketch on this change though - will the old stuff continue to work? Will the new stuff? I've added the credentials to k8s already.


StachuDotNet:
"I'm hoping that it will"?

pbiggar:
> "I'm hoping that it will"?

whoops, fixed

pbiggar:
This broke when I deployed it, because i had used `password` as the key in the kubernetes configuration, and `secret` in the code. `secret` was the right one to use to I fixed the config in kubernetes and they restarted fine.

I had rolled cronchecker back so we didn't miss any crons, will roll it forward soon.

Issue #3872, opened 2022-05-23
blazor.boot.json is 404-ing after AOT deploy


StachuDotNet:
When I go to `use-blazor=true` after #3831 being merged, I get a 404 on blazor.boot.json (specifically, `https://static.darklang.com/blazor/blazor.boot-4cd2df25689084dcc6338095e0f86de9df2593021bc39d6df3884666319f787e.json`)

Either the hash there is incorrect (a blazor.boot.json exists, but not at that location), or no blazor.boot.json is there for some reason. Currently investigating.

StachuDotNet:
I've looked through `.circleci/config.yml`, and I'm guessing this is some sort of problem with our `_push-assets-to-cdn` script?
How can I see the static assets in production? @pbiggar or maybe that's not the right question.

StachuDotNet:
When I check my local `etags.json`, I see that `BlazorWorker.js` has `2770c79104479e17e0b4f140b571406e9cc93a495e0da0926a0832ee6d7cf949` as the value.
That file loads fine in prod currently.

`blazor/blazor.boot.json`'s value is `92e1133126671a4cf32aca593d3d156f53055dca2e92983ffd2977e312a33701`.

For some reason, `ui.html` is attempting to load `https://static.darklang.com/blazor/blazor.boot-4cd2df25689084dcc6338095e0f86de9df2593021bc39d6df3884666319f787e.json` rather than `https://static.darklang.com/blazor/blazor.boot-92e1133126671a4cf32aca593d3d156f53055dca2e92983ffd2977e312a33701.json` which surprises me.

Note: `https://static.darklang.com/blazor/blazor.boot-92e1133126671a4cf32aca593d3d156f53055dca2e92983ffd2977e312a33701.json` also 404s.

pbiggar:
Here's where they're uploaded: https://app.circleci.com/pipelines/github/darklang/dark/3038/workflows/75a662c9-edf4-4331-940c-c5f1d887a4bb/jobs/32702

I note - and have previously seen in other builds - the `[Killed]` output in there. Possibly related?

StachuDotNet:
the `Killed`s definitely could be related. :thinking:

pbiggar:
Good idea. They might be there from a cache - possibly bump the cache number?

pbiggar:
https://github.com/darklang/dark/blob/main/.circleci/config.yml#L318
https://github.com/darklang/dark/blob/main/.circleci/config.yml#L344

[...]

PR #3875, opened 2022-05-23
When deploying, don't try to upload .br files

StachuDotNet:
Relates to #3811 and #3872.

It seems that the CI step of uploading static assets is failing on .br files. This should fix that.


pbiggar:
Crap, I just realized you hadn't asked for a review. Shouldn't have merged this, my bad!

StachuDotNet:
Np worries, it's worth a shot regardless

PR #3877, opened 2022-05-23
Increase CI cache key of build-fsharp-backend step

StachuDotNet:
in hopes of fixing #3872 and in turn #3811

I'm not sure if this will help, but it shouldn't hurt.


StachuDotNet:
Still seeing `Killed` after this, FYI.

PR #3878, opened 2022-05-23
Remove some slow json

pbiggar:
This is the start of removing Json.OCamlCompatible from the ApiServer's responses. This is still delivering the same data (still on the OCamlTypes format too), just using System.Text.Json-based Json.Vanilla.serialize.

The only real difference here should be the behaviour of NaN and Infinity. However, when I tested existing NaN/Infinity behaviour, everything broke pretty badly. It seems to be the same with this change applied.

Also this re-adds traces in dev environment.


config/dev
```diff
@@ -88,15 +88,18 @@ DARK_CONFIG_ROLLBAR_POST_CLIENT_ITEM=c7af77e991aa4edd80cf6a576c1e42f5
 DARK_CONFIG_ROLLBAR_POST_SERVER_ITEM=e2777a52d9a24c00a7e8e8aedaa61017
 
 # Honeycomb
-DARK_CONFIG_TELEMETRY_EXPORTER=console
-DARK_CONFIG_HONEYCOMB_API_KEY=none
+DARK_CONFIG_TELEMETRY_EXPORTER=honeycomb
+DARK_CONFIG_HONEYCOMB_API_KEY=76f291cb893c40ed06f8f91620b5ee3c
```
StachuDotNet, line 92: This is safe to push? If this is some non-prod public key, do we need to worry about over-usage at all?

fsharp-backend/src/ApiServer/ApiServer.fs
```diff
@@ -68,13 +68,25 @@ let addRoutes
   let std = standardMiddleware
   let html = htmlMiddleware
 
-  let api name perm f =
-    let handler = jsonHandler f
+  // Cleanup: switch everything over to vanilla and get rid of ocamlCompatible
```
StachuDotNet, line 71:
```suggestion
  // CLEANUP: switch everything over to vanilla and get rid of ocamlCompatible
```
consistency nitpick

config/dev
pbiggar: Sigh. No, this was a mistake. I've disabled the key

StachuDotNet (approved):

PR #3981, opened 2022-05-25
Exceptions pass

pbiggar:
An attempt to fix some exceptions I saw.

The main thing here is to remove expensive computation (that is, parsing json) out of the DB loop. That keeps the connection open longer, and those aren't an infinite resource.

May be related to some exceptions, might not.


pbiggar:
Hoping that this might be a fix for #3983 but I'm not holding my breathe.

PR #3984, opened 2022-05-25
Rewrite _push-assets-to-cdn in Python

StachuDotNet:
Currently, this script is failing, seemingly due to high memory usage. See commentary in #3872.

The intention here is to reduce the # of gsutil calls, uploading in batches per file extension or content-type.
The needs of the script got complex enough where it made sense to port to Python.

I'm at a point now where this is ported to match the original.

I'm not well-versed in Python or Bash, so I could be doing some plain dumb things here, stylistically.

### DIFF
--- a/scripts/deployment/_push-assets-to-cdn
+++ b/scripts/deployment/_push-assets-to-cdn
@@ -12,114 +12,114 @@ ROOT_OF_FILES="backend/static/"
 BUCKET="gs://darklang-static-assets"
 
 # Gather list of files in known dir
-allStaticFiles = [f for f in glob.glob(ROOT_OF_FILES + '**', recursive=True) if os.path.isfile(f)]
+all_static_files = [f for f in glob.glob(f"{ROOT_OF_FILES}**", recursive=True) if os.path.isfile(f)]
 
 
 # Filter out files we don't care about
-def shouldCopyFile(filename):
-  filesToSkip = [ROOT_OF_FILES + ".gitkeep", ROOT_OF_FILES + "etags.json"]
-  if filename in filesToSkip:
+def should_copy_file(file_name):
+  files_to_skip = [f"{ROOT_OF_FILES}.gitkeep", f"{ROOT_OF_FILES}etags.json"]
+  if file_name in files_to_skip:
     return False
 
-  if filename.endswith(".gz") or filename.endswith(".br"):
+  if file_name.endswith(".gz") or file_name.endswith(".br"):
     return False
 
   return True
-staticFilesToCopy = [f for f in allStaticFiles if shouldCopyFile(f)]
+static_files_to_copy = [f for f in all_static_files if should_copy_file(f)]
 
 
 # We want to call `gsutil cp` in bulk by Content-Type shortly, but it doesn't
 # support a way to copy multiple files in one call while also renaming them.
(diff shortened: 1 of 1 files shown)


StachuDotNet:
Before finishing the refactor, I assumed that `gsutil cp` would have the ability to pass in a list of "input path"+"output path" pairs, but it only seems to support a list of one or the other through stdin, with the target being a simple directory. See notes about `-m` [here](https://cloud.google.com/storage/docs/gsutil/addlhelp/GlobalCommandLineOptions) and `gsutil cp` [here](https://cloud.google.com/storage/docs/gsutil/commands/cp).

Since we're transforming many of the file names for the copy, such that they have the hashes from `etags.json` embedded, it seems we need some additional refactoring here.

My best idea so far is to adjust this script a bit:
- create a temp directory
- copy all of the files in `staticFilesToCopy` to that new directory
- rename the files in-place, with the hashes
- refactor a bit to process the files in chunks by content-type - once the above is done, we can `find` the files we care about, and copy those with the syntax that `gsutil cp` 
- (purge the temp directory)

Does that seem reasonable, @pbiggar? Am I missing something?

pbiggar:
Sounds good

StachuDotNet:
Any preference in temp dir location? I haven't seen prior art, and would throw in `/tmp` or $RUNDIR subfolder without advice.

pbiggar:
> Any preference in temp dir location? I haven't seen prior art, and would throw in /tmp or $RUNDIR subfolder without advice.

I think python has builtins for getting temp dirs.

But rundir isn't a bad option if it doesn't.

StachuDotNet:
While I haven't actually run this against GCP, I've `print`ed a fair amount locally to see that this looks OK

The `gsutil cp` calls are failing for me when executing locally, but for one of two expected reasons:
- the file is already at the target with the 'new' name (i.e. it's had a successful upload recently)
- (when the above isn't true) on me not having GCP creds on my local dev environment

StachuDotNet:
@pbiggar ready for a round of review - please tear my Python apart :)

pbiggar (changes requested):
This looks great.

There's a couple of places where maybe a format-string would be better than concatenation or other options, but I don't care that much except the one I highlighted.

Python var and method names should be snake_case (which we do in our other python scripts).

I think this would be improved by adding mypy type annotations. We use them in shipit and it works great, but shipit also has dependency management built-in, and I'm not sure it's worth adding dependency management for a few type signatures here.

pbiggar (approved):

pbiggar:
Feel free to merge this when you're ready - no need to wait for CI to be clean (except syntax checks)

StachuDotNet:
https://app.circleci.com/pipelines/github/darklang/dark/3081/workflows/aa0946c0-eb11-445d-a711-b8e486d03bee/jobs/33237/parallel-runs/0/steps/0-117 log seems happy! Now checking Blazor...

StachuDotNet:
Ah, this is slightly off, but very close.

We're pushing all files successfully, but files beyond the root directory (e.g. `blazor/LibExecution.dll`) are being pushed to just `{BUCKET}/LibExecution-HASH.dll`, not `{BUCKET}/blazor/LibExeuction-HASH.dll`.

pbiggar:
You should add `-m` also, to make uploads faster

PR #4010, opened 2022-05-27
Simplify CDN push in CI: iter by mimetypes, not dirs

StachuDotNet:
per recommendation yesterday

### DIFF
--- a/scripts/deployment/_push-assets-to-cdn
+++ b/scripts/deployment/_push-assets-to-cdn
@@ -39,13 +39,9 @@ BUCKET="gs://darklang-static-assets"
 # - call `gsutil cp -r` per Content-Type -specific directory
 
 
-# We can't make folders named like "application/json", so these two utility fns
-# are used to deal with this.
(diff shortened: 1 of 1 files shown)


pbiggar:
What level of reading would you like here? Should I inspect deeply or just glance at it?

StachuDotNet:
just glance is fine.
I've tested locally (proves that it works) and this is per your recommendation (so unlikely to be opposed).

StachuDotNet:
In case the intention doesn't ring a bell, the change here is that we used to iterate over the folders in our temp directory when uploading via gsutil, and decoded the foldername (with the hacky #slash# thing) along the way. Now, we don't ever decode - we killed that fn, and instead iterate over the found mimetypes, encoding to find the 'correct' dir instead of iterating over them.

pbiggar:
Yeah, it looked grand. Just wanted to be sure you weren't relying on a deep review for certainty before I rubber stamp it.

PR #4026, opened 2022-05-29
Remove code, containers and k8s config for old queueworker

pbiggar:
Both the qw and the scheduler.

Plan is to wait a while before merging, in case something comes up.


pbiggar:
Note that though there's a change to the Dockerfile, it's small and insignificant enough that we don't need t make a new one for this change.

PR #4032, opened 2022-05-30
Remove writeback to ocaml opcodes

pbiggar:
This PR is to prevent us from needing to make calls to legacyserver when saving ops. For users who had large handlers, these could slow down saving ops significantly and even timeout.

This data was saved in toplevel_oplists columns `data` and `rendered_oplist_cache`. These are non-null, so we still have to write back data, now we write back an empty byte array.

Then I removed all calls to OCamlInterop from anything in src/, and moved it to TestUtils.

This required removing calling ocaml from BwdServer tests as these relied on the shared DB. LibExecution and HttpClient tests did not, so I didn't remove them. I also removed OCAMLONLY and FSHARPONLY, and some hacks that they required. This involved carefully editing the test files.

Finally, removed the legacycontainer and all remnants.

For tests, I had to disable ApiServer tests. They need to be rewritten anyway. I also had to update the integration tests with new data.

Also:
- rename any use of "rendered" to "cached" when discussing opcodes
- removed Benchmark project


fsharp-backend/src/ApiServer/ApiServer.fs
```diff
@@ -107,7 +107,6 @@ let addRoutes
   addRoute "GET" "/a/{canvasName}/trigger-exception" std R exceptionFn
 
   ocamlCompatibleApi "add_op" RW AddOps.addOp
-  ocamlCompatibleApi "clean_ops" RW AddOps.clean
```
StachuDotNet, line 110: What was calling this?

pbiggar: Nothing. I used it to covert the opcodes then it didnt have a use.

fsharp-backend/tests/Tests/ApiServer.Tests.fs
```diff
@@ -988,4 +988,4 @@ let cookies =
        (302, None, Some "/login?error=Invalid+username+or+password"))
       (local, None, (302, None, Some "/login?error=Invalid+username+or+password")) ]
 
-let tests = testList "ApiServer" [ localOnlyTests; permissions; cookies ]
```
StachuDotNet, line 991: Should we delete `localOnlyTests` while we're at it, or are you thinking to update those such that we run them generally? If so, let's make a separate Issue for that.

pbiggar: Yeah, we should test that these work. Issue added

PR #4043, opened 2022-05-30
Blazor analysis result serialization now results in Rollbars

StachuDotNet:
Prior to this, serialization failures led to an infinite loading spinner in the client.

### DIFF
--- a/fsharp-backend/src/Wasm/Wasm.fs
+++ b/fsharp-backend/src/Wasm/Wasm.fs
@@ -340,10 +340,8 @@ type EvalWorker =
         }
 
       let serialized =
-        let serialize = Json.Vanilla.serialize
-
         try
-          serialize result
+          Json.Vanilla.serialize result
         with
         | e ->
           let metadata = Exception.nestedMetadata e
@@ -352,7 +350,9 @@ type EvalWorker =
           System.Console.WriteLine(
             $"caught exception: \"{e.Message}\" \"{metadata}\""
           )
-          serialize (Error($"exception: {e.Message}, metadata: {metadata}"))
+          Json.Vanilla.serialize (
+            $"exception: {e.Message}, metadata: {metadata}"
+          )
 
       EvalWorker.postMessage serialized
     }
--- a/fsharp-backend/src/Wasm/Wasm.fs
+++ b/fsharp-backend/src/Wasm/Wasm.fs
@@ -320,22 +320,39 @@ type EvalWorker =
           )
           Error($"exception: {e.Message}, metdata: {metadata}")
 
-      match args with
-      | Error e -> return Error e
-      | Ok args ->
+      let! result =
+        task {
+          match args with
+          | Error e -> return Error e
+          | Ok args ->
+            try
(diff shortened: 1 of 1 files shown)


pbiggar:
Why did it break before? I'm not seeing the difference.

StachuDotNet:
See here: https://github.com/darklang/dark/blob/main/fsharp-backend/src/Wasm/Wasm.fs#L339-L340

Previously there was no try/with to wrap the call to Vanilla.serialize

fsharp-backend/src/Wasm/Wasm.fs
```diff
@@ -320,22 +320,39 @@ type EvalWorker =
...
+              System.Console.WriteLine("Error running analysis in Blazor")
+              System.Console.WriteLine($"called with message: {message}")
+              System.Console.WriteLine(
+                $"caught exception: \"{e.Message}\" \"{metadata}\""
+              )
+              return Error($"exception: {e.Message}, metadata: {metadata}")
+        }
+
+      let serialized =
+        let serialize = Json.Vanilla.serialize
```
pbiggar, line 343: Is this separated out for any reason?

pbiggar (approved):
One question, feel free to merge once resolved.

fsharp-backend/src/Wasm/Wasm.fs
StachuDotNet: If you're referring to the whole `serialized` binding, I agree that there's some redundancy / noisy code here.

pbiggar: Specifically:

```
let serialize = Json.Vanilla.Serialize
try
  serialize result
```

instead of

```
try
  Json.Vanilla.Serialize result
```

StachuDotNet: Yeah it's " quick attempt to ensure the Error serialization didn't somehow differ from the Ok serialization"
Dumb protection from someone updating the serialization of `result` but not updating the serialization of the `Error` a few lines later.

I can simplify if you'd like.

fsharp-backend/src/Wasm/Wasm.fs
```diff
@@ -320,22 +320,39 @@ type EvalWorker =
...
         | e ->
           let metadata = Exception.nestedMetadata e
-          System.Console.WriteLine("Error running analysis in Blazor")
+          System.Console.WriteLine("Error serializing results of Blazor analysis")
           System.Console.WriteLine($"called with message: {message}")
           System.Console.WriteLine(
             $"caught exception: \"{e.Message}\" \"{metadata}\""
           )
-          return Error($"exception: {e.Message}, metadata: {metadata}")
+          serialize (Error($"exception: {e.Message}, metadata: {metadata}"))
```
StachuDotNet, line 355: this is the other call to `serialize` I'm referring to.

pbiggar: Ah, I missed this. Might be clearer by making both explicit, but could go either way.

PR #4045, opened 2022-05-30
It should be the entire canvasname that is <= 64

pbiggar:
The entire canvasname must fit in a varchar64. (Since that's the length of subdomains
allowed by DNS, this won't have any user-facing effects)


pbiggar:
Just realized more stuff

StachuDotNet (approved):

PR #4048, opened 2022-05-30
Don't error when the push is too big

pbiggar:
Large pushes are happening because the user is renaming something (probably a DB), that causes a ton of other TLs to be pushed.

For now, it's much much easier to drop these. I'd rather put the effort into making renames not change every toplevel, and this happens rarely and canvases are fairly rarely used by two people at once anyway.

### DIFF
--- a/fsharp-backend/src/LibBackend/Pusher.fs
+++ b/fsharp-backend/src/LibBackend/Pusher.fs
@@ -26,12 +26,18 @@ let pusherClient : Lazy<PusherServer.Pusher> =
        options
      ))
 
+type EventTooBigEvent = { eventName : string }
+
 
 /// <summary>Send an event to Pusher.com.</summary>
 ///
 /// <remarks>
 /// This is fired in the background, and does not take any time from the current thread.
 /// You cannot wait for it, by design.
+///
+/// Do not send requests over 10240 bytes. Each caller should check their payload,
+/// and send a different push if appropriate (eg, instead of sending
+/// `TraceData hugePayload`, send `TraceDataTooBig traceID`
 /// </remarks>
 let push
   (canvasID : CanvasID)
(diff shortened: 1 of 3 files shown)


fsharp-backend/src/LibBackend/Pusher.fs
```diff
@@ -69,10 +72,19 @@ let pushNew404 (canvasID : CanvasID) (f404 : TraceInputs.F404) =
...
+type AddOpEventTooBigPayload = { tlids : List<tlid> }
 
 // For exposure as a DarkInternal function
 let pushAddOpEvent (canvasID : CanvasID) (event : Op.AddOpEvent) =
-  push canvasID "add_op" (Json.OCamlCompatible.serialize event)
+  let payload = Json.OCamlCompatible.serialize event
+  if String.length payload > 10240 then
+    let tlids = List.map LibExecution.OCamlTypes.tlidOf event.``params``.ops
+    let tooBigPayload = { tlids = tlids } |> Json.Vanilla.serialize
+    // push canvasID "addOpTooBig" tooBigPayload
```
StachuDotNet, line 83: do you need to uncomment this?

pbiggar: No, that was the point of the comment that immediately follows.

I guess I should make that clearer?

StachuDotNet: My interpretation of the line is "push a notification that there was some add_op involving so many canvases that the serialized payload would be huge." That seems good+safe to have even with no receiver? (i.e. so we can use that data for some analysis, if pushes are persisted?)

pbiggar: I think it'll make a BIG RED ERROR BAR with no receiver, which the user can't do anything about.

I'll fix the comment.

StachuDotNet: Ah, ok. Maybe we should (separately) _not_ raise a big red error bar when there's no receiver?
That sounds like a reasonable way to tee up something to be handled later, recording along the way for potential analysis.

StachuDotNet: PR looks good with comment update.

StachuDotNet (approved):

fsharp-backend/src/LibBackend/Pusher.fs
pbiggar: Thanks

PR #4051, opened 2022-05-31
Fewer calls to load canvas in BwdServer

pbiggar:
Loading canvases for BwdServer and custom domains too a bunch of separate DB calls. This consolidates requests for canvas metadata into a single call. This converts 3 calls into 1 for every BwdServer request, and converts 4 calls into 1 for custom domain requests.

I've confirmed that both use an index for the canvas table (custom domains doesn't have an index, so this is no different than before).


StachuDotNet (approved):

pbiggar:
Yes, it should be converts 4 into 1.

Issue #4059, opened 2022-06-01
Blazor worker "memory access out of bounds" bug


StachuDotNet:
![image](https://user-images.githubusercontent.com/906686/171303014-055015a1-df07-407d-a870-5f8c6d023dc6.png)

seems to happen only with very large requests.

- https://stackoverflow.com/questions/57938081/blazor-preview-9-mono-wasm-memory-access-out-of-bounds-max-string-size-for-dotn
- https://www.reddit.com/r/WebAssembly/comments/g44ldd/what_are_some_possible_causes_of_runtimeerror/
- https://github.com/dotnet/aspnetcore/issues/5617

StachuDotNet:
I've now gotten the analysis request payload that causes the issue.
It's indeed quite large (70814 chars of JSON).

FWIW this request seems to include a lot of stuff that it doesn't need to perform the analysis.

I'm seeing three paths forward:
- somehow increase the memory limit
- send less stuff in the request (i.e. filter out stuff not needed for analysis)
- send the request in batches, e.g. per https://stackoverflow.com/a/57973792

Currently experimenting to see what the char threshold of failure is.

StachuDotNet:
Note: I can replicate this in Release AOT builds, but not Debug builds.
Now checking to see if I can repro in a non-AOT release build.

StachuDotNet:
Seems I can't - i.e. the bug seems AOT-specific.

pbiggar:
Do you know what function this is happening in? Perhaps we're able to separate reading the json from doing the analysis in some way. Maybe we parse the json in JS code and then call wasm using `JSInterop` or whatever it's called.

pbiggar:
Yeah, either of those sound good. I like the ArrayBuffer/file approach but does that also have size limits?

Are we sure this is the problem though? 76k isn't small but it's also not huge. The Stackoverflow thing says the limit is over 50MB now.

StachuDotNet:
I'm now a bit less confident.

I added this to Wasm.fs:

```
  static member CanWeGetHere(message : string) =
    message |> String.length |> printfn "Yes %i"
```

and posted the giant json to _that_ handler rather than the OnMessage one. It works?

StachuDotNet:
I was wrong before! We _are_ getting to within OnMessage. We're even deserializing OK. The failure seems to be happening within the actual analysis.

StachuDotNet:
Even more interesting:

with my code like this:

```
 let! result =
        task {
          match args with
[...]

PR #4065, opened 2022-06-01
Add ?debug-analysis flag to client

StachuDotNet:
If set to true, console.log()s analysis requests before they're made.

I'm not very hopeful, but it's worth a shot.

### DIFF
--- a/client/src/appsupport.js
+++ b/client/src/appsupport.js
@@ -247,6 +247,10 @@ window.Dark = {
     busy: false,
     /* Records the last time a result returned. So Integration tests will know has analysis finished running since a given timestamp */
     utils: require("../../lib/js/client/workers/FSharpAnalysisWrapper.bs.js"),
+    debug: (function () {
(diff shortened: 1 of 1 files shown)


client/src/appsupport.js
```diff
@@ -247,6 +247,10 @@ window.Dark = {
     busy: false,
     /* Records the last time a result returned. So Integration tests will know has analysis finished running since a given timestamp */
     utils: require("../../lib/js/client/workers/FSharpAnalysisWrapper.bs.js"),
+    debug: (function () {
```
StachuDotNet, line 250: note: this is within the fsharpAnalysis. Not bothering to make this work with the ocaml stuff since we'll be deleting that soon anyway.

StachuDotNet:
May not be needed - https://github.com/darklang/dark/issues/4059#issuecomment-1143617825

pbiggar (approved):

Issue #4066, opened 2022-06-01
Fix multiple deploys at once


pbiggar:
Right now, I've been staggering deploys so that they don't overwrite each other, as it is possible for a later deploy to override an earlier one.

If we fix it so that code is strictly deployed in the order it was merged, I would feel safe just hitting the "merge" button even when there were existing merges going.

What should happen:
- I hit merge for deploy a
- I hit merge for deploy b
- a deploys, claiming then releasing to lock
- b deploys, claiming the lock (possibly waiting for a to finish), then releasing the lock

What can happen:
- I hit merge for deploy a
- I hit merge for deploy b
- deploy a runs slowly
- b deploys (claiming then releasing the lock)
- a deploys (no lock to prevent it)

An important part of this is that our locks rely on the build number, but a build number which is higher or lower doesn't actually tell is anything, since a "build" refers to a "job", not a "workflow".

How can we solve this?

There is existing code to prevent an earlier deploy from overwriting a later one, but it works with build numbers which are not strictly ordered. So we'd want to use something that is:

- sha1 (we can get the ordering from git)
- commit date (easy in that we don't need git)

But also, we should start it at the start of a build. We would like every build to deploy in the order they are merged, rather than using the fact that a deploy has started.

Algorithm:
- at start of build, say "I exist"
- when it is time to deploy, check no "previous build" exists
  - manual deploys always win
  - if there is a previous deploy, wait for it, failing the build after 20 minutes or so. Ensure "retry from failed" just does the deploy

pbiggar:
Solved

PR #4067, opened 2022-06-01
Better CI locking

pbiggar:
As detailed in https://github.com/darklang/dark/issues/4066, I'm afraid things will go badly if I press "merge" on many PRs at once. This addresses that by ensuring deploys go out in the order they were merged, and will wait for other jobs to go out if they're not next.

The details have all been added in the README - let me know if anything is unclear.


pbiggar:
- [ ] test before merging this

pbiggar:
With `deploy-lock-one-remove`, the intent was to have prefixes: `deploy-lock-all-X` and `deploy-lock-one-X`, where the `one-` suffix implied it affected one build. But it's not a super great name: `remove-one` is much clearer than `one-remove`.

Any thoughts on better names?

StachuDotNet:
I think it's fine as-is; just add a #comment right before their definitions with that explanation?

pbiggar:
Good idea.

PR #4068, opened 2022-06-01
Small build speedups

pbiggar:
- Turn up the parallelization on integration tests
- Don't build OCaml that we're not using, so that it doesn't get added to the workspace and slow things down


.circleci/config.yml
```diff
@@ -329,33 +329,15 @@ jobs:
...
           name: Run integration tests
           command: |
             # Run the server first to set up the DB correctly for the prep script
             scripts/run-fsharp-server --published
             # Wait for users to be added so prep.sh works
             ./scripts/devcontainer/_wait-until-apiserver-ready
-            integration-tests/run.sh --concurrency=3 --retry --pattern="`cat test-pattern`" --published
-            rm test-pattern
-
+            integration-tests/run.sh --concurrency=6 --retry --published
```
StachuDotNet, line 340: I believe this will have no effect. Integration tests are running [in serial](https://github.com/darklang/dark/blob/02f2783f1104fd9ad751b5f8dee982b671bd012c/integration-tests/tests.ts#L100) currently

pbiggar: Thank you, that confused me.

pbiggar: I'll test it - I thought it was running at 3x and we might get 6x, but I'm less optimistic of any speedup now.

StachuDotNet: Neat that it worked for you! (removing the .serial)
For me, it kept yielding flaky tests, but looks good for you :)

pbiggar: So it actually didn't work - I can't get more than 1 worker (at least on my local machine).

pbiggar: > Running 70 tests using 1 worker

Same in CI

pbiggar: > 70 passed (1m)

Just had this locally with --concurrency=16

StachuDotNet: ? hilarious.

StachuDotNet: Really curious why I was seeing _significant_ flakiness before, seemingly due to ApiServer strain, and you're not now.
But, happy just seeing the result.

pbiggar: Maybe you weren't using the `--retry` flag, which is enabled in CI?

StachuDotNet: it was failing for me in CI specifically.

pbiggar: Curious. If it starts flaking in CI again, we can rollback.

PR #4070, opened 2022-06-01
Prevent failures of large analysis requests to Blazor

StachuDotNet:
As discovered/discussed in #4059, large requests to the Blazor-powered WASM analysis sometimes resulted in this nasty runtime error, _only_ when built in Release mode with AOT compilation flag on in `Wasm.fsproj`:
image

(Note: #4059 includes a fair amount of discovery around this issue. If you're reading this in the future, start there)

This PR resolves that. 

It's unclear _exactly_ why this works, but two things have changed:
- the old code had a nested `task{}` and `let!` pair when computing the `result`, before serializing and posting back the result.
- we added a `|> Task.map (fun _ -> ())` at the end of the last `task{}`. Without this, we still get the error

The underlying problem seems to be related to some of { GC, binding scope, `tasks`, AOT/release builds }, but I'm not sure. Personally I'm satisfied with this for now, and separately creating a minimal repro to hand to .net team.

### DIFF
--- a/fsharp-backend/src/Wasm/Wasm.fs
+++ b/fsharp-backend/src/Wasm/Wasm.fs
@@ -354,6 +354,7 @@ type EvalWorker =
       EvalWorker.postMessage serialized)
 
     // Magic.
-    // Without this, AOT builds somehow reach beyond their memory limits
+    // Without this, a "memory access out of bounds" error is somehow raised,
+    // only for AOT builds
     // See https://github.com/darklang/dark/issues/4059
     |> Task.map (fun _ -> ())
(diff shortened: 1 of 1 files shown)


fsharp-backend/src/Wasm/Wasm.fs
```diff
@@ -352,5 +351,9 @@ type EvalWorker =
           )
           Json.Vanilla.serialize ($"exception: {e.Message}, metadata: {metadata}")
 
-      EvalWorker.postMessage serialized
-    }
+      EvalWorker.postMessage serialized)
+
+    // Magic.
+    // Without this, AOT builds somehow reach beyond their memory limits
```
StachuDotNet, line 357:
```suggestion
    // Without this, a "memory access out of bounds" error is somehow raised,
    // only for AOT builds
```

pbiggar (approved):
Looks good.

One suggestion would be to try using `.Result` instead of `Task.map`.

PR #4072, opened 2022-06-02
Use the right executor for locking

pbiggar:


pbiggar:
Merging as main is broken.

PR #4075, opened 2022-06-02
Use the correct name

pbiggar:
I called the wrong script, so calling the right one now.


pbiggar:
Merging as this won't do anything until the deploy is done.

PR #4083, opened 2022-06-03
Split F# build in two

pbiggar:
This tries to reduce our very slow F# builds by separating building the binaries from building the wasm. We put the tests in with the binaries.

Required fixing the weird situation where we tested against the apiserver from the dev environment.


pbiggar:
I ended up adding a bunch of stuff to this PR:
- fix for #4084 
- significantly reduce the size of the .net caches by only caching Build/obj (skipping Build/out)
- reducing the amount of data being copied between "workspaces" (this is how CircleCI shares data between build jobs)
  - not generating .pdb files for .NET
  - reducing the number of publishable projects
  - disabling the `SelfContained` and `PublishSingleFile` options, which we didn't use, for .NET projects
  - these changes also
    - reduced the build time by a minute
    - reduced the file size of the generated binaries by 70%
- Also disabled PublishTrimmed, which didn't do much but did cost build time
- removed some old scripts/run-ocaml-* scripts
- remove old DARK_CONFIG_TEST_* configuration
- turn integration test parallelism up to 16 (made thing slower cause too many failures)

.circleci/config.yml
```diff
@@ -282,6 +287,7 @@ jobs:
       - slack-notify-job-failure
       - deploy-lock-remove-on-fail
 
+  # Build server binaries and run tests (but not blazor, which is separated out for build speed)
   build-fsharp-backend:
     executor: in-container
     resource_class: xlarge
```
StachuDotNet, line 293: can/should this be lowered? (Not sure if xlarge costs us more per CI run. maybe large is enough now since blazor compile has been extracted)

StachuDotNet (approved):
Looks great - just one minor question, posted.

.circleci/config.yml
pbiggar: It does cost us 2x. 
I thought about this. However, builds still aren't exactly fast, so I decided not to turn it down. It does seem like there's a lot of CPU being used, and I was recently looking into whether I could use even larger machines.

I'm continuing the optimization of builds in a new branch shortly. If I see the CPU drop I'll switch to smaller machine.

PR #4087, opened 2022-06-04
Don't upload one directory per file

pbiggar:
Deploying assets took an unreasonably large amount of time, when in fact it should have been pretty quick.

What was happening is that for each file, we were uploading the directory. That is, given N files split into K buckets, we intended to upload K buckets, but in fact uploaded N buckets (and disproportionately the bigger ones of course).

This fixes it to upload K buckets by using a set instead of an array.

### DIFF
--- a/scripts/deployment/_push-assets-to-cdn
+++ b/scripts/deployment/_push-assets-to-cdn
@@ -7,9 +7,10 @@ import json
 import subprocess
 import tempfile
 import shutil
+import time
 
-ROOT_OF_FILES="backend/static/"
-BUCKET="gs://darklang-static-assets"
+ROOT_OF_FILES = "backend/static/"
+BUCKET = "gs://darklang-static-assets"
 
 # This script's goal is to gather all of our static assets in `backend/static/`
 # and upload them to GCP via `gsutil cp`, with the correct Content-Type. Some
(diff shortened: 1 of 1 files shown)


pbiggar:
It was a worst-case $O(N^2)$ operation, and since 85% of the files were in one directory, it behaved close to the worst case.

scripts/deployment/_push-assets-to-cdn
```diff
@@ -118,10 +135,11 @@ def get_mimetype_temp_dir(mimetype):
   mimetype_folder = encode_dirname_with_slash(mimetype)
   return f'{temp_dir}/{mimetype_folder}'
 
-mimetypes_to_process = []
+
+mimetypes_to_process = set()
 for f in static_files_to_copy:
   mimetype = mime_type_for(f)
-  mimetypes_to_process.append(mimetype)
+  mimetypes_to_process.add(mimetype)
```
StachuDotNet, line 142: LOL

StachuDotNet (approved):

scripts/deployment/_push-assets-to-cdn
pbiggar: The change to `set()` was the important change

PR #4088, opened 2022-06-04
More CI speedups

pbiggar:
A few more fixes to make CI faster.

Before: 9m22sm

image

After: 8m7s

image

There's two changes:
- use a 5% smaller docker container (not a great win)
- figure out (and fix) why dotnet was rebuilding so much (which is that the git checkout mtime was recent, while the CircleCI cached build objects' mtime was not recent).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,8 +64,6 @@ RUN curl -sSL https://baltocdn.com/helm/signing.asc | apt-key add -
 # We want postgres 9.6, but it is not in ubuntu 20.04
 RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ `lsb_release -cs`-pgdg main" >> /etc/apt/sources.list.d/pgdg.list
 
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -115,8 +111,9 @@ RUN DEBIAN_FRONTEND=noninteractive \
       postgresql-9.6 \
       postgresql-client-9.6 \
       postgresql-contrib-9.6 \
-      google-chrome-stable \
+      git-restore-mtime \
       nodejs \
+      libgbm1 \
```
StachuDotNet, line 116: So, we previously downloaded chrome because Playwright needed it, but we can get away with only libgbm1 somehow?

pbiggar: No, we previously downloaded chrome because testcafe needed it, but now playwright bundles it itself, requiring just the dependencies on the system.

PR #4095, opened 2022-06-06
Stop hiding error

pbiggar:
When a function is called with an error as the argument, we should propagate the error - instead we created a different error.

I've wanted this for months.

There's no change to semantics because we're allowed change the error messages whenever we want.


StachuDotNet (approved):
Yes, exciting!

PR #4101, opened 2022-06-07
Include inner exception details in Blazor rollbars

StachuDotNet:
- remove some redundancy in Wasm.fs
- use Exception.getMessages to roll inner exception into returned error message for Blazor errors

### DIFF
--- a/client/src/appsupport.js
+++ b/client/src/appsupport.js
@@ -137,9 +137,9 @@ function displayError(msg) {
 }
 
 window.onerror = function (msg, url, lineno, colno, error) {
+  console.error("Uncaught exception", msg, url, lineno, colno, error);
   window.Rollbar.error(msg, error);
   window.lastError = error;
-  console.error("Uncaught exception", msg, url, lineno, colno, error);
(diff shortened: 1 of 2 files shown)


client/src/appsupport.js
```diff
@@ -137,9 +137,9 @@ function displayError(msg) {
 }
 
 window.onerror = function (msg, url, lineno, colno, error) {
+  console.error("Uncaught exception", msg, url, lineno, colno, error);
   window.Rollbar.error(msg, error);
   window.lastError = error;
-  console.error("Uncaught exception", msg, url, lineno, colno, error);
```
StachuDotNet, line 142: Is this dumb? My theory is that Rollbar.error might leave some parts of `erro` out - maybe the console.error in the logs may be visible in the "telemetry" section of rollbar, making some things easier to diagnose.

pbiggar: I think you gotta feel it out for a bit before knowing for sure.

pbiggar (approved):

PR #4104, opened 2022-06-07
Fix missing dataset issue on honeycomb during deploy

pbiggar:
Deploys are currently failing because we're trying to add markers to honeycomb datasets that don't exist.

The actual deploy is fine, just the build fails right after.

This fixes it by getting the datasets from the API instead of listing them out.


pbiggar:
Tested locally.

StachuDotNet (approved):

PR #4118, opened 2022-06-09
ubuntu 2204

pbiggar:
This switches the dev and production containers to Ubuntu 22:04.

This was held up before because of trying to update the ocaml containers, which is now not necessary.

It includes an attempt to make the devcontainer build natively on M1/arm chips - however a number of things prevented that (need to wait til dotnet 8 and rescript 10), so I kept the code that abstracts it, but still made M1 Macs use amd64 emulation.

I used a new heredoc feature - hoping that's supported everywhere.

I added two scripts, inlined in the dockerfile, to install scripts. I inlined them so I wouldn't be copying extra files between the dockerfile repo and the dark repo (copying one is bad enough). The scripts make it easier to install files from the internet with hashes, which we were doing repeatedly and inconsistently.

Includes a hack from #4609 to make postgres work.

Upgrades to python 3.10, and from watchgod to watchfiles (I don't remember why this was necessary).


pbiggar:
Given the next big problem is compiling OCaml, I'm going to take a break from this, until we either remove ocaml or take it off the container in some way.

StachuDotNet (approved):

pbiggar:
The final two steps needed to get this across the line were:
- set ulimit properly and configure nginx appropriately (see comments)
- wait til we can download all files before starting integration tests

I'm not sure sure what changed in this version of ubuntu that made integration tests fail. It used to be that if we waited for app.css to load, then integration tests would work fine. In this PR, I discovered that now, if app.css can load, then we are not sure that blazor dlls will load. The result was that many tests couldn't load blazor files, causing many warnings and some fails.

These can't happen in production because we don't use the dotnet server for blazor assets. We also don't use nginx in our devcontainer (we continue using the same one after this PR as before it, albeit with a slightly different config).

Issue #4127, opened 2022-06-11
Analysis not handling Secrets well


StachuDotNet:
An Analysis request of a REPL with a simple Secret returns `Incomplete` currently

StachuDotNet:
Some potential failure points to check:
- [ ] at deserialization
- [ ] during conversion to OCaml types
- [ ] during eval/Interpeter
- [ ] during serialization (Vanilla2 handles secrets poorly?)

StachuDotNet:
The issue seems to be within Serialization - I see the SECRET trickle through until then.

StachuDotNet:
[`secrets = []` :)](https://github.com/darklang/dark/blob/00ad9db1597a0fe9bf31140c4144ec0435fdac8f/fsharp-backend/src/Wasm/Wasm.fs#L197)

StachuDotNet:
Should have a bug fix shortly

StachuDotNet:
This was resolved by #4128.

PR #4128, opened 2022-06-11
Pass secrets through in F# analysis

StachuDotNet:
An empty list `[]` of Secrets is currently being passed through for F# analysis. This fixes that.

### DIFF
--- a/fsharp-backend/src/LibExecution/OCamlTypes.fs
+++ b/fsharp-backend/src/LibExecution/OCamlTypes.fs
@@ -673,7 +673,8 @@ module Convert =
     | BSTypes.UserType ut -> PT.Toplevel.TLType(ocamlUserType2PT ut)
     | BSTypes.UserFn uf -> PT.Toplevel.TLFunction(ocamlUserFunction2PT uf)
 
-
+  let ocamlSecret2RT (secret : secret) : RT.Secret.T =
+    { name = secret.secret_name; value = secret.secret_value }
 
   // ----------------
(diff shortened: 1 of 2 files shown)


fsharp-backend/src/LibExecution/OCamlTypes.fs
```diff
@@ -673,7 +673,8 @@ module Convert =
     | BSTypes.UserType ut -> PT.Toplevel.TLType(ocamlUserType2PT ut)
     | BSTypes.UserFn uf -> PT.Toplevel.TLFunction(ocamlUserFunction2PT uf)
 
-
+  let ocamlSecret2RT (secret : secret) : RT.Secret.T =
+    { name = secret.secret_name; value = secret.secret_value }
```
StachuDotNet, line 677: I was slightly surprised to see a RT value here rather than a PT value, but that's what's expected by analysis.

StachuDotNet:
will fix #4127

StachuDotNet:
This returns Secrets in plain text to the user - is that OK, or do we only want the secrets to be _usable_ within eval?
I'll check the js_of_ocaml analysis to see what that did - if I can get a plain-text secret back from analysis _there_, then this PR seems good.

StachuDotNet:
My confidence here is high, and it's a regression from the old analysis, so merging this.

StachuDotNet:
- [ ] add an integration tests ensuring Secrets are usable in Analysis requests

PR #4130, opened 2022-06-11
Remove special handling of pre-1970 dates

StachuDotNet:
This logic only exists to temporarily match OCaml backend


StachuDotNet:
Actually, the right thing here is to create new fns.

PR #4132, opened 2022-06-11
Clarify existing CLEANUP items; delete a few resolved ones

StachuDotNet:


StachuDotNet:
100% comments that we went through together; merging.

PR #4135, opened 2022-06-11
Don't strip request bodies in GET requests for BwdServer

StachuDotNet:
per CLEANUP item.

I have not yet tested this, manually or via tests. I will follow up with both.


StachuDotNet:
Needs to be done as part of a larger "create new http stack" PR.

I think we can forego a special `CLEANUP_NEW_HTTP_STACK` for now.

PR #4137, opened 2022-06-11
Handle conversion of F# RT.EApply (ELambda...) to OCamlRT

### LINKED ISSUE #4136: F# Analysis result couldn't be converted to OCaml type client expects
An error in converting F# runtime types to OCaml RuntimeTypes 

https://github.com/darklang/dark/blob/f12d9bd829aaee4ad561a9af2e59173ad36d9c02/fsharp-backend/src/LibExecution/OCamlTypes.fs#L875

was found

So, that conversion fn needs some update

StachuDotNet:
Will fix #4136 and https://rollbar.com/darkops/darklang/items/3377/


StachuDotNet:
My confidence is _somewhat_ high here but not fully.

StachuDotNet:
> The alternative here is to remove the old format from the client and server!

yes! It's still not clear to me how to do this, though.

> Is there a repro/unit test for it?

There's a unit test for it now, which is a reduced case of [the rollbar that inspired this Issue](https://rollbar.com/darkops/darklang/items/3377/).

I've been unable to repro this within the Editor itself - I'm not sure how to get my AST in such a state.

pbiggar:
What's the plan here?

pbiggar:
I don't feel super confident here. This feels like we should run all the saved code through this or some fuzzing, but I think I'd rather put that energy into removing the OCamlTypes instead. WDYT?

StachuDotNet:
I agree with the uncertainty and support closing this in favor of focusing on removing OCamlTypes!

PR #4142, opened 2022-06-12
Adjust query param parsing so empty vals result in DStrs

StachuDotNet:
per a CLEANUP item.

They were previously parsed as DNulls to match the old OCaml backend.


pbiggar (changes requested):
Can't do this one without changing existing user code behaviour. This needs to be done as part of adding a new http stack.

Perhaps we should mark this and similar cleanups as CLEANUP_NEW_HTTP_STACK or similar.

StachuDotNet:
Needs to be done as part of a larger "create new http stack" PR.

I think we can forego a special `CLEANUP_NEW_HTTP_STACK` for now.

PR #4146, opened 2022-06-13
Move several LibCrypto fns to LibExecutionStdLib

StachuDotNet:
Not all of LibCrypto can be moved, as System.Security.Cryptography is not fully available in the WebAssembly-compiled bits.

https://docs.microsoft.com/en-us/dotnet/core/compatibility/cryptography/5.0/cryptography-apis-not-supported-on-blazor-webassembly

The functions that could be marked Pure (moved to LibExecutionStdLib), have been.
The others, I've removed the CLEANUP note. If you'd like, I can replace them with a TODO item (though, we'd then have to find an alternative or roll our own fns)

I've tested the ported functions to ensure that they execute appropriate within Analysis, and saw failures when trying to port the remaining fns.


StachuDotNet:
A bit disappointing, but it's progress.

StachuDotNet:
There's an integration test against one of the now-pure fns, expecting to click a play button. I'll need to follow up here a bit.

pbiggar (approved):

StachuDotNet:
I had to change one integration test here since a fn is no longer impure (and doesn't require a manual execution) - does that look OK to you?

PR #4154, opened 2022-06-13
LibString CLEANUPs

StachuDotNet:


fsharp-backend/tests/testfiles/string.tests
```diff
@@ -412,6 +412,22 @@ String.toInt_v1 "" = Error "Expected to parse string with only numbers, instead
 String.toInt_v1 "0" = Ok 0
 String.toInt_v1 "1" = Ok 1
 
+String.toInt_v2 "" = Error "Expected to parse string with only numbers, instead got \"\""
+String.toInt_v2 "-00001" = Ok -1
+String.toInt_v2 "-10001" = Ok -10001
+String.toInt_v2 "-4611686018427387903" = Ok -4611686018427387903I
```
pbiggar, line 418: These should probably use the `L` suffix, not `I`, as a Dark integer is int64

pbiggar (approved):

fsharp-backend/src/LibExecutionStdLib/LibString.fs
```diff
@@ -23,6 +23,12 @@ let err (str : string) = Ply(Dval.errStr str)
 
 let incorrectArgs = LibExecution.Errors.incorrectArgs
 
+let parseInt64 (str : string) : int64 =
+  try
+    assertRe "int64" @"-?\d+" str
+    System.Convert.ToInt64 str
+  with
+  | e -> Exception.raiseInternal $"parseInt64 failed" [ "str", str; "inner", e ]
```
StachuDotNet, line 31: Is this what you meant by inlining - just copying the fn over?

pbiggar: I meant directly into the functions that use it.

As an example of the problems, I seem to have added more abilities to assertRe since those were added, which probably changed the behavior.

StachuDotNet: OK. In this case would you port the whole parseInt64 as-is into each of the fns, or just System.Convert.ToInt64 usage?

StachuDotNet: I'm going to just copy `System.Convert.ToInt64` into each - porting the whole thing is silly if assertRe has/will change(d). We're just moving the problem from one place to another then.

pbiggar: Looks like assertRe would throw an exception, which would be caught and cause a InternalException, which would be caught and cause an error 🤣

Can you check if there's tests for this? If so, LGTM.

StachuDotNet: I'm not sure where I'd look for tests around this - in string.tests?

Upon failure, `Assertion regex failure` is the message that `assertRe` tries to raise, and the definition of `assertRe` is the only place where I find that text in the code.

Since it's caught everywhere, not sure what's appropriate here - maybe a unit test in Prelude.Tests.fs?

pbiggar: I just mean is the error condition (where the assertion fails) tested for these functions?

StachuDotNet: Yes - for example, `String::toInt ""` is tested.

PR #4160, opened 2022-06-14
Ensure analysis performed in right timezone (UTC)

### LINKED ISSUE #4114: Dates get modified when passed in a function call
It seems dates get added 2h when passed by parameter to functions.

As a minimal example consider a function that takes a time and returns it:
Screenshot 2022-06-08 at 14 33 32

Then here's a REPL that creates a time, passes it to the function above, and compares them for equality:
Screenshot 2022-06-08 at 14 33 21

Notice the trace output shows the time is `"02:00:00"` instead of the expected `"00:00:00"`.

Somehow the bug does not manifest if I "live modify" the timestamp parsed in line 1 of the REPL and the output is correct. But if I click either the "reload" icon on the top right, or the green reload icon for the `sameTime` invocation, the result updates to the wrong one as shown in the screenshots.

StachuDotNet:
Will resolve #4114


StachuDotNet:
that test wasn't supposed to pass... wonder if CI runs in UTC or something?

StachuDotNet:
Ah maybe I need to take the absolute value of the diff

StachuDotNet:
`analysisDate 2022-06-14T21:52:20.000Z bwdDate 2022-06-14T21:52:21.000Z`
Seems CI is being run in UTC? or the bug isn't repro-ing there for some other reason. Anyway, I can still use the local failure.

StachuDotNet:
definitely an issue with serialization. Fwiw, we face the issue if we swap to OCamlCompatible

StachuDotNet:
It's not just an issue with Analysis, fwiw - see test just added.

StachuDotNet:
I've created a fix that works, but it has me a bit concerned.

`LocalDateTime` serialization did not roundtrip well.
I added both an integration test and a unit test around this, both of which failed locally.

Since then, I've adjusted `Prelude` such that both OCamlCompatible and Vanilla serializers roundtrip as expected.

The concern is this: have we been messing up any data in the meantime?

i.e. is there risk of us having previously:
- deserialize date from DB
- serialize date, and insert back into DB, offsetting by X hours
?

I haven't thought of a good way to test this, or assess the risk. Going to walk away from this for a bit to think.

cc @pbiggar 


Also, not sure if the 'consolidation' here is favorable - open to opinions.

pbiggar:
We don't have a list of serializations we do - we really should. Having a comprehensive list would help answer questions like this. Maybe start one as part of this PR? Perhaps we could mark each deserialization with some token (such as `DARK_SERIALIZATION`) and make some categories based on how safe it is to change - not sure.

For now, you'll have to check all uses of OCamlCompatible and Vanilla serializations/deserializations.

StachuDotNet:
Done research for Vanilla serializer/deserializer - there are 2 usages currently where I see some level of risk.

Where should we go from here?
Some potential TODOs:
- insert an `event` with a date before code changes, then load them with the changes, and see if dates have been affected from a user's point of view
- continue researching the Analytics.fs usage.

It's concerning how concerning such an action is. We have so many damn serializers.

pbiggar:
It would seem to me that each of our serializers (by which I mean Vanilla and OCamlCompatable) should have unit tests (record -> string) and roundtripping tests (record -> string -> record), where the record has one of more instance of each type we use. That would give good confidence that it does what we expect.

That would solve the "does it change" question.

The other question is "do these specific serializer use cases work if we change the format". In that case, we should have a test that encompasses the functionality of the component (in this case the queue; this test exists), and it should use a value that has all the properties we need (perhaps `TestUtil.sampleDvals` could be combined into a big massive dval using `DList` or `DObj`) for the event queue case.

pbiggar:
Now that the Vanilla and OCamlCompatible testing is in place, we should be able to assess the risk here. I think just running the fsharp tests will tell us if it's safe.

StachuDotNet:
>  I think just running the fsharp tests will tell us if it's safe (which we expect it to be)
Under this premise, I merged `main` into this branch, so the new tests would be run. We'll see what CI says :)

StachuDotNet:
CI failed, but local all passes. I think this is a matter of build cache or something. I could increase the cache # to make CI happy, but suspect simply merging will fine.

I'm running CI from start again to see if it self-resolves

pbiggar:
Shouldn't we expect changes in the serialization tests saved files?

StachuDotNet:
> > Shouldn't we expect changes in the serialization tests saved files?
> 
> That depends on the timezone of CI. IIRC, CI may be running in UTC+0. Will have to verify.

Yeah I believe that's the case. I had found, when writing the integration test that's a part of this PR, that such integration test failed locally prior to the fix, but not in CI. After some logging (and later killing that logging), I determined that CI was running in UTC, which wouldn't catch things like this.

StachuDotNet:
Should we explicitly set the timezone of where we want to run the build? I'm not sure the best place to do that.

StachuDotNet:
https://circleci.com/docs/faq#how-can-i-set-the-timezone-in-docker-images looking into this

StachuDotNet:
I set the timezone in ci config script as an experiment; may end up deleting that commit.

StachuDotNet:
So, that did result in a failure! https://app.circleci.com/pipelines/github/darklang/dark/3644/workflows/413243c0-6921-40a9-b685-0e9af7062246/jobs/38605

I'm not sure what to do with this. Slightly confused by expectations. Maybe worth a short zoom to discuss?

pbiggar:
> So, that did result in a failure! https://app.circleci.com/pipelines/github/darklang/dark/3644/workflows/413243c0-6921-40a9-b685-0e9af7062246/jobs/38605
> 
> I'm not sure what to do with this. Slightly confused by expectations. Maybe worth a short zoom to discuss?

For sure, I'm pretty free tomorrow.

PR #4164, opened 2022-06-16
Update port in README.md

vicero:

While setting up dark locally, I found the port is now 9000 instead of 8000 as stated in the README.md

Update documentation

Update documentation

Found while setting up locally

N/A

N/A


pbiggar:
Wonderful, thank you!

Issue #3883, opened 2022-05-24
In a pipe, add an `=` shortcut to turn it into an assignment


pbiggar:
Given a pipe:

```
value
|> f1 other
|> f2 another
|> ___
```

You should be able to type `= myVar` to get

```
let myVar = value
            |> f1 other
            |> f2 another
```

pbiggar:
Closing as it doesn't apply to darklang-next

Issue #3810, opened 2022-05-16
Can't run integration tests outside of devcontainer


StachuDotNet:
I've been recently unable to run integration tests outside of the devcontainer.
Usually this isn't an issue, but `--debug` doesn't work from within the devcontainer.

The error that I'm getting is `./integration-tests/run.sh: line 18: DARK_CONFIG_APISERVER_HOST: unbound variable`
Previously, I don't believe I've had to do anything special for this to be available outside of the container - any ideas of how to resolve?

StachuDotNet:
`DARK_CONFIG_APISERVER_HOST=darklang.localhost:9000 DARK_CONFIG_BWDSERVER_HOST=builtwithdark.localhost:11001 ./integration-tests/run.sh`

yields this endlessly:
```
curl: (22) The requested URL returned error: 502 Bad Gateway
waiting for analysis.js at http://static.darklang.localhost:9000/analysis.js
```

pbiggar:
I can't find where we set the DNS for darklang.localhost, but we should add static.darklang.localhost to it

pbiggar:
> I can't find where we set the DNS for darklang.localhost, but we should add static.darklang.localhost to it

Actually, I don't think that's it. When I do: `ping x.x.localhost` it knows it's 127.0.0.1, and when I go `curl -v http://static.darklang.localhost` I get a 404 from the webserver.

I'm thinking perhaps analysis.js isn't available? That's the ocaml version of the analysis - perhaps it wasn't built properly? Can you reach it from inside the container? can you reach it in your browser?

StachuDotNet:
I'm able to run tests outside of the container if I remove [this line](https://github.com/darklang/dark/blob/1308ceb566b36b92518b5422eded83ddb03bae73/integration-tests/run.sh#L90) from `integration-tests/run.sh`, which is `./scripts/run-fsharp-server "${PUBLISHED}"`.

I'm not yet sure why.

pbiggar:
I think what might be happening is that the servers started up by `run-fsharp-server` die after the script ends if they're run outside the container. So the `run-fsharp-server` script would first kill the running servers, then start up its own servers, then those would die after that script ends.

So by making the change, run.sh no longer calls `run-fsharp-server` and so no longer kills the servers.

pbiggar:
> How should we address those hacks, so someone can more easily run the tests outside of the container (as our docs suggest)?

One option here is to stop suggesting that. I'm fully using the devcontainer now, personally.

 
[...]

PR #4167, opened 2022-06-17
Integration test improvements

StachuDotNet:
Replaces #4009. Was originally aimed to help with #3883, but delaying that portion until a further PR.

I'd recommend reviewing commit-by-commit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -434,7 +434,7 @@ jobs:
             scripts/run-fsharp-server --published
             # Wait for users to be added so prep.sh works
             ./scripts/devcontainer/_wait-until-apiserver-ready
-            integration-tests/run.sh --concurrency=8 --retry --published
+            integration-tests/run.sh --concurrency=4 --retry --published
       - run: integration-tests/_integration-test-results-to-honeycomb.sh
       - assert-clean-worktree
       - store_artifacts: { path: rundir }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -434,7 +434,7 @@ jobs:
             scripts/run-fsharp-server --published
             # Wait for users to be added so prep.sh works
             ./scripts/devcontainer/_wait-until-apiserver-ready
-            integration-tests/run.sh --concurrency=4 --retry --published
+            integration-tests/run.sh --concurrency=8 --retry --published
       - run: integration-tests/_integration-test-results-to-honeycomb.sh
       - assert-clean-worktree
       - store_artifacts: { path: rundir }
--- a/integration-tests/run.sh
+++ b/integration-tests/run.sh
@@ -20,7 +20,7 @@ REPEAT=1 # repeat allows us to repeat individual tests many times to check for e
 BASE_URL="http://${DARK_CONFIG_APISERVER_HOST}"
 BWD_BASE_URL=".${DARK_CONFIG_BWDSERVER_HOST}"
 BROWSER="chromium"
-#PUBLISHED=""
+PUBLISHED=""
 
 for i in "$@"
 do
@@ -42,7 +42,7 @@ do
     shift
     ;;
     --published)
-    #PUBLISHED="--published"
+    PUBLISHED="--published"
     shift
     ;;
     --debug)
@@ -89,7 +89,7 @@ fi
 
 # We need to restart the server after adding new packages. Integration tests test
 # against the dev environment, not the test one.
-#./scripts/run-fsharp-server "${PUBLISHED}"
+./scripts/run-fsharp-server "${PUBLISHED}"
 ./scripts/devcontainer/_wait-until-apiserver-ready
 
 ######################
--- a/integration-tests/run.sh
+++ b/integration-tests/run.sh
@@ -5,7 +5,7 @@
 set -euo pipefail
 
 # This deliberately can be run outside the container - sometimes you want to test on
-# the host. CLEANUP this is currently failing, unless you specify some env vars
+# the host. TODO this is currently failing, unless you specify some env vars
 # explicitly. Here's a sample command to get it working for now:
 # DARK_CONFIG_APISERVER_HOST=darklang.localhost:9000 DARK_CONFIG_BWDSERVER_HOST=builtwithdark.localhost:11001 ./integration-tests/run.sh
 
--- a/integration-tests/run.sh
+++ b/integration-tests/run.sh
@@ -97,10 +97,7 @@ fi
 ######################
 echo "Starting playwright"
 integration-tests/node_modules/.bin/playwright --version
-BASE_URL="$BASE_URL" \
-  BWD_BASE_URL="$BWD_BASE_URL" \
-  integration-tests/node_modules/.bin/playwright \
-  \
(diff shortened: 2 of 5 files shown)


integration-tests/run.sh
```diff
@@ -95,7 +97,10 @@ fi
 ######################
 echo "Starting playwright"
 integration-tests/node_modules/.bin/playwright --version
-BASE_URL="$BASE_URL" BWD_BASE_URL="$BWD_BASE_URL" integration-tests/node_modules/.bin/playwright \
+BASE_URL="$BASE_URL" \
+  BWD_BASE_URL="$BWD_BASE_URL" \
+  integration-tests/node_modules/.bin/playwright \
+  \
```
StachuDotNet, line 103: I can undo this

.circleci/config.yml
```diff
@@ -434,7 +434,7 @@ jobs:
             scripts/run-fsharp-server --published
             # Wait for users to be added so prep.sh works
             ./scripts/devcontainer/_wait-until-apiserver-ready
-            integration-tests/run.sh --concurrency=8 --retry --published
+            integration-tests/run.sh --concurrency=4 --retry --published
```
pbiggar, line 437: I think this is a good idea. I noticed in CI that this makes each test take much longer (from like 7s to like 15s), and more likely to time out. I hadn't changed it back cause it seemed stable where it was.

StachuDotNet:
Everything passes now, but not much to say in terms of a perf. boost :/

StachuDotNet:
- [ ] try again using multiple machines but with lower resources
?

StachuDotNet:
Ah, I believe I removed the bits that handled splitting tests up across parallel machines, in some earlier PR. Will have to go back and revert that for a proper test.

.circleci/config.yml
```diff
@@ -414,7 +414,8 @@ jobs:
 
   integration-tests:
     executor: in-container
-    resource_class: xlarge
+    resource_class: large
+    parallelism: 2
```
pbiggar, line 418: This won't work as tests are currently split in the integration tests config. ./circleci/config.yml (integration-tests step) needs to be modifier - the appropriate code is likely in the git history.

StachuDotNet:
Unfortunate: https://github.com/microsoft/playwright/issues/11644#issuecomment-1025796541

I'm going to sleep on this PR in favor of some other work. Not being able to record videos may make the sharing of the browser context unfavorable.

pbiggar:
There's some nice refactoring/improvements here, so even if we have to revert the shared context, we should try to merge those.

StachuDotNet:
OK. I'll tidy this a bit, reserving the shared context for now, mark it as ready for review, and leave shared context for a separate PR.

integration-tests/run.sh
```diff
@@ -5,7 +5,9 @@
 set -euo pipefail
 
 # This deliberately can be run outside the container - sometimes you want to test on
-# the host
+# the host. TODO this is currently failing, unless you specify some env vars
+# explicitly. Here's a sample command to get it working for now:
+# DARK_CONFIG_APISERVER_HOST=darklang.localhost:9000 DARK_CONFIG_BWDSERVER_HOST=builtwithdark.localhost:11001 ./integration-tests/run.sh
```
StachuDotNet, line 10: better than nothing, until #3810 is resolved.

StachuDotNet: I know I keep looking this up in Slack history, anyway

integration-tests/tests.ts
```diff
@@ -1,30 +1,60 @@
...
-} from "@playwright/test";
+import { test, expect, ConsoleMessage, Page, TestInfo } from "@playwright/test";
 import fs from "fs";
 
-const BASE_URL = process.env.BASE_URL;
-const BWD_BASE_URL = process.env.BWD_BASE_URL;
-const options = {
-  baseURL: BASE_URL,
-  bwdBaseURL: BWD_BASE_URL,
-};
```
StachuDotNet, line 16: I previously added `bwdBaseUrl` to this `options` thinking it was something general that _we_ defined, rather than something Playwright specific. Rather, baseURL is something specific to Playwright that we're able to override. This gets rid of that, rather inlining the object into the `.use()` below.

integration-tests/tests.ts
```diff
@@ -1,30 +1,60 @@
...
+  getStyleProperty,
+  gotoAST,
+  gotoHash,
+  Locators,
+  get,
+  post,
+  pressShortcut,
+  selectAll,
+  waitForEmptyEntryBox,
+} from "./utils";
```
StachuDotNet, line 32: I'm not a big fan of so many imports from one file.
I thought about breaking down `utils` but spent too long on _naming things_ so gave up. Open to suggestions.

integration-tests/tests.ts
```diff
@@ -1,30 +1,60 @@
...
-  let setLocalStorage = async (key: string, value: any) => {
+  async function setLocalStorage(key: string, value: any) {
     await page.evaluate(
       ([k, v]) => {
         localStorage.setItem(k, v);
       },
       [key, JSON.stringify(value)],
     );
-  };
+  }
```
StachuDotNet, line 57: I don't remember doing this - I think the formatter did it for me.

integration-tests/utils.ts
```diff
@@ -0,0 +1,232 @@
...
+      return true;
+    }
+    return false;
+  };
+  await page.waitForFunction(analysisFunction, lastTimestamp);
+}
+
+export async function awaitAnalysisLoaded(page: Page) {
+  await page.waitForFunction(() => window.Dark.fsharpAnalysis.initialized);
+}
```
StachuDotNet, line 204: Greatly simplified - happy with this!

StachuDotNet:
Assuming CI passes, I think this is in a merge-able state.

I haven't given up hope on improving CI perf, but have spent enough time on this for now.

.circleci/config.yml
StachuDotNet: Yeah I noticed that too - really wonder why that is.

StachuDotNet:
Ha - the tests got stuck. Not sure if it's related to the concurrency drop from 8 to 4, or if Circle is just toying with me.

StachuDotNet:
Prior to merge:

- [ ] compare artifacts to normal build and ensure they're the same

StachuDotNet:
OK, it's getting stuck for some other reason; looking into it.

StachuDotNet:
These are the issue tests:
- `abridged_sidebar_content_visible_on_hover`
- `create_from_404`
- `record_consent_saved_across_canvases`

StachuDotNet:
I was able to repro, and found that when I reverted my changes to the `beforeEach` hook, the issue went away. Seems best to revert.

integration-tests/run.sh
pbiggar: Yeah, this is a good idea.

integration-tests/tests.ts
pbiggar: Fine by me

integration-tests/tests.ts
pbiggar: Better to be consistent anyway.

pbiggar (approved):
Nice!

PR #4183, opened 2022-06-21
Refactor LibJwt

StachuDotNet:
I found LibJwt to be pretty confusing, while researching the serializers that we use.

This:
- consolidates the existing serialization
- adds needed commentary
- removes Yojson intermediate serialization type
- removes needless circular fn definitions
- adds SERIALIZER_DEF and SERIALIZER_USAGE comments
- places verifyAndExtract functions to above Serialization module, to indicate there's no serialization occuring there

---

I'm researching our serializers and their usages. I got pretty confused by this module's code, so spent some time to better understand it and consolidate the bits we'd like to get away from.

Now that the serialization is consolidated to just 2 `serialize` and `deserialize` fns, I think it'd be pretty reasonable to write a property-based test to ensure a new set of these functions exactly matches the current set. Maybe I'm under-afraid about being _exactly_ the same?

### DIFF
--- a/fsharp-backend/src/BackendOnlyStdLib/LibJwt.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibJwt.fs
@@ -329,13 +329,6 @@ module Serialization =
       let msg = if str = "" then "JSON string was empty" else e.Message
       Error msg
 
-module STJSerialization =
-  // TODO add property tests to check comparisons of these serializers
-  // TODO make these match the ones under Serialization
-  // TODO use these, make them the default one, and purge the old one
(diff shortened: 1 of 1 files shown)


StachuDotNet:
I realize this is generally a one-big-commit PR as-is; I took great care to ensure the serialization/deserialization is unaffected by the refactoring, and tests show this.

pbiggar:
I'd recommend as next steps here to start from scratch with only the comment changes.

StachuDotNet:
Understandable - I was super confused by the existing code, and couldn't resist the compulsion to make it legible. Probably not worth the risk. Closing in favor of #4187

PR #4186, opened 2022-06-21
Fail CI deployment if CDN push fails

### LINKED ISSUE #3880: Fail CI if CDN uploads fail
Currently, if the upload of a static asset to our CDN fails during CI, e.g. with a message such as `Killed` seen in #3872, it silently fails.

Instead, we should block the deploy to -prod.

StachuDotNet:
Will resolve #3880.

Tested locally:
- success case: push to a local folder, see "Done" at end of script, verify assets get tehre
- failed case: change `gsutil` to `gsutilbroken`, no 'Done' at end of script
- failed case: push to a non-existant local folder, no 'Done' at end of script


scripts/deployment/_push-assets-to-cdn
```diff
@@ -167,7 +167,14 @@ for mimetype in mimetypes_to_process:
     cp -Z -n -r . "{BUCKET}"
     '''
 
-  subprocess.run(gsutil_command, shell=True, cwd=mimetype_subfolder)
+  result = subprocess.run(gsutil_command, shell=True, cwd=mimetype_subfolder, capture_output=True)
+  if result.stderr and result.returncode != 0:
+    raise subprocess.CalledProcessError(
```
StachuDotNet, line 172: Maybe `raise` is the wrong thing here - I don't know python.

pbiggar: why non-empty stderr and non-zero returncode?

I would print an error and then use `system.exit`

StachuDotNet: We could suppress the non-error messages getting to stderr with `-q`, but those messages are useful for local debugging https://github.com/GoogleCloudPlatform/gsutil/issues/116#issuecomment-13929972 so I think this is best. Adding a comment.

pbiggar (approved):

PR #4188, opened 2022-06-22
Deprecate `String::toInt_v1` in favor of new `Int::parse_v0`

StachuDotNet:
will resolve #4168 and resolve #4170

PR #4191, opened 2022-06-22
Implement Int::parse in favor of String::toInt

### LINKED ISSUE #4168: Rename `String::toInt` as `Int::parse`

### LINKED ISSUE #4170: Investigate + resolve String::toInt_v2 flakiness
per #4169

### LINKED ISSUE #4196: Int::parse should use a result or an option

StachuDotNet:
Replaces #4188, of which I absolutely botched the Git history.

Will resolve #4168 and resolve #4170 and resolve #4196.


StachuDotNet:
From last review:

> Can you add some tests with whitespace around the number? (front, back, between - and the number).
> What happens with + in the number, can we test that too?
> Maybe a test for other representations that might be accidentally parsed without our knowledge, to confirm they aren't supported (0x, 0b, 5E-5)?
> What about a number in octal format (078)?

These are done.

pbiggar (approved):
Wonderful!

If you had a moment, would be great to add these (great!) new tests to the old version too. I don't want to expand the scope so feel free not to, but I would like us to get into the habit of doing that going forward.

StachuDotNet:
Resolves #4196 shortly (fixing now)

StachuDotNet:
Ah - also, need to use the v1 error message from String::toInt, not the v0 error message. (v1 is better). One moment.

pbiggar (approved):

Issue #4157, opened 2022-06-13
Remove OCaml analysis


StachuDotNet:
- [ ] backend/jsanalysis
- [ ] package.json: `build:analysiswwrapper` and a reference in the `build` script
- [ ] client/src/analysis
- [ ] analysiswrapper.js and analysis.js
- [ ] js_of_ocaml
- [ ] ocamlAnalysis in appsupport.js
- [ ] any `useBlazor` usages
- [ ] ? backend/libtarget.js (I'm not sure what this is)

PR #4192, opened 2022-06-22
Reference window.Dark.analysis over window.Dark.fsharpAnalysis

StachuDotNet:
`fsharpAnalysis` is no longer a thing, per #4157


pbiggar (commented):
Post-commit approval: 👍

PR #4197, opened 2022-06-23
WIP client review

StachuDotNet:
Ignore for now - just noticed some things when perusing client code yesterday. I plan to go through again to study and do reviews similar to my previous F# ones.


StachuDotNet:
I'll do a client review at some point, but this is already stale, so I might as well start from scratch here later.

Issue #2972, opened 2021-03-12
Remove weird behaviour in String.split_V0


pbiggar:
Make String.split_v1 which behaves the same, but does not return an empty strings in the split list.

PR #4202, opened 2022-06-24
Implement a better String::split

StachuDotNet:


fsharp-backend/src/LibExecutionStdLib/LibString.fs
```diff
@@ -533,20 +533,17 @@ let fns : List<BuiltInFn> =
...
           if sep = "" then
             s |> String.toEgcSeq |> Seq.toList |> List.map DStr |> DList |> Ply
           else
-            // CLEANUP: we need a new version of this fn.
-            // This behaviour is the worst. This mimics what OCaml did: There
-            // should be (n-1) empty strings returned for each sequence of n
-            // strings matching the separator (eg: split "aaaa" "a" = ["",
-            // "", ""]). However, the .NET string split puts in n-1 empty
-            // strings correctly everywhere except at the start and end,
-            // where there are n empty strings instead.
```
StachuDotNet, line 549: I wasn't sure what this bit meant - are you happy with .NET's String.split, or no?

pbiggar: This is one of those things where there isn't necessarily a right answer. Different languages and libraries take different approaches, and there's good reasons for different choices. Our old version didn't have good reasons for its choices, and was really quite bad.

So let me turn this around: what should be the behaviour? Are you happy with the test output here? If so, why? If not, what do you think it should do differently?

pbiggar:
I think we should have tests for splitting when using composition. For example (quote from [here](https://www.w3.org/International/articles/definitions-characters/):

> the sequence 👨‍👩‍👧‍👧 [U+1F468 U+200D U+1F469 U+200D U+1F467 U+200D U+1F466] results in a composed emoji glyph for a "family: man, woman, girl, boy" on systems that support this kind of composition.

What happens if we split on U+1F467 here?

fsharp-backend/src/LibExecutionStdLib/LibString.fs
StachuDotNet: The new implementation here fits what most languages seem to do, which is the best option without having some Dark-specific reason to do things differently, so I'm happy with the test output here for that reason.

Beyond that, I don't have any particular investment here - even the current impl. seems reasonable to me.

StachuDotNet:
I got a test to pass around String.split with a composed family emoji. The results look OK to me.

StachuDotNet:
Really not sure why all of the v1 tests pass locally but fail in CI. They're all failing on the `actual` being `Incomplete` somehow.
Also, why are they _flaky_ and not just _failing_? Does the v1 function just ... not exist the first run-through somehow? :thinking: 

Feels like I'm not progressing here right now so will abandon and follow up. Open to ideas as to what's going on.

StachuDotNet:
Check out #2972 along with this

fsharp-backend/tests/testfiles/string.tests
```diff
@@ -385,6 +385,50 @@ String.split_v0 "x22y" "2" = ["x"; ""; "y"]
...
+String.split_v1 "x1" "1" = ["x"; ""]
+String.split_v1 "x666666y" "6" = ["x"; ""; ""; ""; ""; ""; "y"]
+String.split_v1 "x55555y" "5" = ["x"; ""; ""; ""; ""; "y"]
+String.split_v1 "x4444y" "4" = ["x"; ""; ""; ""; "y"]
+String.split_v1 "x333y" "3" = ["x"; ""; ""; "y"]
+String.split_v1 "x22y" "2" = ["x"; ""; "y"]
+String.split_v1 "x1y" "1" = ["x"; "y"]
+String.split_v1 "6a6aa6aaa6aaaa" "a" = ["6"; "6"; ""; "6"; ""; ""; "6"; ""; ""; ""; ""]
+String.split_v1 "👨‍❤️‍💋‍👨👩‍👩‍👧‍👦🏳️‍⚧️‍️🇵🇷" "" = ["👨‍❤️‍💋‍👨"; "👩‍👩‍👧‍👦"; "🏳️‍⚧️‍️"; "🇵🇷"]
+String.split_v1 "👨‍👩‍👧‍👦" "👩" = ["👨‍"; "‍👧‍👦"]
```
pbiggar, line 431: I don't think it should do this.

pbiggar (changes requested):

fsharp-backend/tests/testfiles/string.tests
pbiggar: > Check out #2972 along with this

Good point. I think our investigation here into what other langs do, and the various edge cases, means I'm happy with what we chose to do here and to ignore #2972.

pbiggar: To be clearer, we shouldn't split below the "character" (egc) size.

StachuDotNet:
I wrote a ... rough implementation that doesn't break up ECGs: https://github.com/darklang/dark/pull/4202/commits/b47e2b8322e9704b2cf3df7e595eb7880459d314#diff-36b5caefa314eb883a5942a56a16d01b5f1601a3cce90fd6ab98a48fc4bce9b5R577-R600

Am I missing some super easier way, or is the best path forward simply to tidy this sort of algorithm?

StachuDotNet:
https://stackoverflow.com/questions/72917431/net-string-split-on-egcs-text-elements figured someone else would look for this eventually, so posted on SO

StachuDotNet:
Second-guessing whether `sep` really needs to be turned into TextElements. :thinking:

StachuDotNet:
This simpler solution doesn't turn the separator into ECGs, and seems to work: https://stackoverflow.com/a/72917897/833921
Worth trying that, though it's not immediately clear to me why it works.

pbiggar (approved):
Nice!

Issue #3987, opened 2022-05-25
Support negative floats


pbiggar:
Same as #3250 but for floats

pbiggar:
Test:

- can we generate negative floats in the client fuzzer?
- encoder/decoder
- clipboard tests

pbiggar:
Done

Issue #3280, opened 2021-11-28
Editor should allow negative numbers


pbiggar:
The fluid editor does not allow negative numbers.

A UX for negative numbers is described below:

- allow entering `-` at the start of an integer or float to convert it to a negative number
  - for `0`, allow entering the `-` but turn it into a partial (as it might be part of typing `-1` to turn `0`). As soon as the partial is valid again, turn it back into an int.
- allow deleting `-` from the start of an integer or float to convert it back to a positive
- if typing `-` in a position that is not a binop, start a partial (already happens). Once a partial of `-` gets a number added, turn it into an integer

pbiggar:
Also check copying and pasting negative numbers

pbiggar:
Test:

- can we generate negative floats in the client fuzzer?
- encoder/decoder
- clipboard tests

pbiggar:
Done

PR #4206, opened 2022-06-25
Remove ocaml types - part 2

pbiggar:
Main changes:

- replace `ID.t` with `id` type
- rename Patterns from `FPVariable` to `PVariable` to match backend
- add `sign` to `EFloat` and `PFloat` to match backend
- move ProgramTypes comments to the backend ProgramTypes.fs
- s/bucklescript/rescript in comments

Also added the ability to paste negative floats.


pbiggar:
Tests all pass. The only thing I'm vaguely worried about is that this might break some negative float literals. However, negative literals aren't actually supported yet, so that's fine.

StachuDotNet (approved):
All looks great. Maybe more tests could be added around negative floats, though:
- can we generate negative floats in the client fuzzer?
- encoder/decoder
- clipboard tests
Though non-blocking. Maybe we create a separate Issue to revisit and expand upon client-side testing?

pbiggar:
> All looks great. Maybe more tests could be added around negative floats, though:
> 
> * can we generate negative floats in the client fuzzer?
> * encoder/decoder
> * clipboard tests
>   Though non-blocking. Maybe we create a separate Issue to revisit and expand upon client-side testing?

added to #3987 and #3280

PR #4200, opened 2022-06-23
Libstaticassets cleanup

StachuDotNet:
Resolves #2408 and a few CLEANUP items

PR #4214, opened 2022-06-27
Fix new LibStaticAssets tests; Just/Nothing, not Some/None

StachuDotNet:
This corrects a mistake I made in #4200, currently causing `main` builds to fail currently.

Upon CI being happy, I plan to merge this.


pbiggar:
Thanks for following up

PR #4217, opened 2022-06-27
Rename Date::subtract to Date::subtractSeconds; implement Date::subtract_v1 that subtracts dates

### LINKED ISSUE #2568: Add Date::diff to the standard library
Currently to compare `Date` types, they must be converted to `Int` seconds. It would be nice to have something like `Date::diff` instead. This commonly arises when timing the duration of some execution:

```
let durationSeconds = Date::toSeconds Date::now - Date::toSeconds startTime 

let durationSeconds = Date::diff Date::now startTime
```

We could also consider `Date::secondsSince` for the specific case of comparing to `Date::now`

```
let durationSeconds = Date::secondsSince startTime
```

StachuDotNet:
This makes me want a proper TimeSpan type in Dark, but we can save that for another day.

This will resolve #2568


StachuDotNet:
For now, we could expose a Result version of this, that fails in such cases.

StachuDotNet:
> For now, we could expose a Result version of this, that fails in such cases.

but #4209 might just make this PR work as-is once it's merged in.

StachuDotNet:
I see, thanks for clarifying. I'll have to do more research to see what's wrong here, then.

fsharp-backend/src/LibExecutionStdLib/LibDate.fs
```diff
@@ -577,4 +577,19 @@ let fns : List<BuiltInFn> =
         | _ -> incorrectArgs ())
       sqlSpec = SqlFunctionWithPrefixArgs("date_trunc", [ "'day'" ])
       previewable = Pure
+      deprecated = NotDeprecated }
+
+
+    { name = fn "Date" "diff" 0
+      parameters = [ Param.make "endDate" TDate ""; Param.make "startDate" TDate "" ]
```
StachuDotNet, line 584: maybe these should just be `a` and `b`? I think there's some usefulness to having `end` and `start` in there somewhere, since we're not taking the absolute value of the result. Thoughts?

pbiggar: There's a sign in the output I think? If so, then end/start makes sense as it does matter which is which. a/b is for situations where there's no semantic meaning to the arguments.

StachuDotNet: good point thanks

StachuDotNet, line 584:
```suggestion
      parameters = [ Param.make "end" TDate ""; Param.make "start" TDate "" ]
```

pbiggar (changes requested):
Functionality looks good. I'm not sure how it fits in with the rest of the functionality of the date module (and I don't have super good suggestions of what would be better).

fsharp-backend/src/LibExecutionStdLib/LibDate.fs
```diff
@@ -256,6 +256,20 @@ let fns : List<BuiltInFn> =
...
+      returnType = TDate
+      description = "Returns a new Date `seconds` seconds before `d`"
+      fn =
+        (function
+        | _, [ DDate d; DInt s ] ->
+          d - (NodaTime.Period.FromSeconds s) |> DDate |> Ply
+        | _ -> incorrectArgs ())
+      sqlSpec = SqlBinOp "-"
+      previewable = Pure
+      deprecated = ReplacedBy(fn "Date" "subtractSeconds" 0) }
```
pbiggar, line 269: Can you use the rename stuff instead?

pbiggar (changes requested):

fsharp-backend/src/LibExecutionStdLib/LibDate.fs
StachuDotNet: Ah, forgot about that - will do.

fsharp-backend/src/LibExecutionStdLib/LibDate.fs
```diff
@@ -577,4 +591,21 @@ let fns : List<BuiltInFn> =
...
+      parameters = [ Param.make "end" TDate ""; Param.make "start" TDate "" ]
+      returnType = TInt
+      description = "Returns the difference of the two dates, in seconds"
+      fn =
+        (function
+        | _, [ DDate endDate; DDate startDate ] ->
+          let diff = (DDateTime.toInstant endDate) - (DDateTime.toInstant startDate)
+          diff.TotalSeconds |> System.Math.Round |> int64 |> DInt |> Ply
+        | _ -> incorrectArgs ())
+      sqlSpec = NotYetImplementedTODO
```
StachuDotNet, line 609: maybe this is just '-'? I can test it out to see.

StachuDotNet: the SQL here would be `EPOCH FROM (end - start)` but as far as I can see, we don't currently support such an expression (and it's beyond the scope of this PR to implement that in the sql compiler), so I'll leave it as-is, with a comment added.

StachuDotNet: Hmm maybe `"EPOCH FROM", [ "-" ]` would work.

StachuDotNet: Ah, it's `EXTRACT(EPOCH FROM (end - start))`, which would actually be more complicated.

fsharp-backend/src/LibExecutionStdLib/LibDate.fs
```diff
@@ -256,6 +256,20 @@ let fns : List<BuiltInFn> =
 
 
     { name = fn "Date" "subtract" 0
+      parameters = [ Param.make "d" TDate ""; Param.make "seconds" TInt "" ]
```
pbiggar, line 259: This should use the rename functionality: https://github.com/darklang/dark/blob/main/fsharp-backend/src/LibExecutionStdLib/StdLib.fs#L10

fsharp-backend/tests/testfiles/date.tests
```diff
@@ -238,15 +238,30 @@ Date.toHumanReadable_v0 (d "2021-07-02T17:42:36Z") = "18 days, 19 hours, 27 minu
...
+Date.subtract_v0 (d "2020-11-26T04:37:46Z") 1000000 = (d "2020-11-14T14:51:06Z")
+Date.subtract_v0 (d "2020-11-26T04:37:46Z") -10 = (d "2020-11-26T04:37:56Z")
+Date.subtractSeconds_v0 (d "2020-11-26T04:37:46Z") 0 = (d "2020-11-26T04:37:46Z")
+Date.subtractSeconds_v0 (d "2020-11-26T04:37:46Z") 1 = (d "2020-11-26T04:37:45Z")
+Date.subtractSeconds_v0 (d "2020-11-26T04:37:46Z") 10 = (d "2020-11-26T04:37:36Z")
+Date.subtractSeconds_v0 (d "2020-11-26T04:37:46Z") 1000000 = (d "2020-11-14T14:51:06Z")
+Date.subtractSeconds_v0 (d "2020-11-26T04:37:46Z") -10 = (d "2020-11-26T04:37:56Z")
+
+[tests.date difference]
+Date.subtract_v1 (d "2020-11-26T04:37:46Z") (d "2020-11-26T04:37:46Z") = 0LL
```
pbiggar, line 260: 0LL?

pbiggar (changes requested):
thanks! Could you also add some comments to document the decision making (why we chose these names)?

fsharp-backend/src/LibExecutionStdLib/LibDate.fs
StachuDotNet: Ahh, OK. When you mentioned renames previously, I thought you were referring to `Deprecation.RenamedTo`. Thanks

fsharp-backend/tests/testfiles/date.tests
StachuDotNet: typo, will fix

pbiggar (approved):

PR #4236, opened 2022-07-01
Implement Tuple::toList_v0

StachuDotNet:
I had these changes staged locally while working on #4182, but we decided to save this for a separate PR.

This PR includes a superset of #4182 - it'll look cleaner when that's merged.


pbiggar:
`Tuple::toList` doesn't seem to make sense, Tuples are supposed to be heterogenous types, and they have specific lengths - you can't have a single function be generic over tuples of all lengths.

The signature here is `a -> b -> List<'a>`, but you can't put a `'b` in a `List<'a>`. Maybe this should be called `Tuple2::toList` with the signature `'a -> 'a -> List<'a>`.

StachuDotNet:
Fair point!

I know the types don't match, but was thinking "well, we already have mixed types in Lists, this is just another case of that."

But it's not a case of that - those are examples of functions that are "legitimate" getting bad parameters.
This would be an example of an illegitimate function getting good parameters.

(by "legitimate," I mean "will still be a type-safe fn once we have type safety guarantees")

I'll close this and think through this more.

StachuDotNet:
Thinking through this more, I'm seeing more clearly now that tuple fns can't just be of type "tuple" and that they'll take a bit of special care in arg-checking. Thanks!

PR #4255, opened 2022-07-10
Remove scripts/run-fsharp-benchmark

StachuDotNet:
The project is gone now, so this doesn't work.


StachuDotNet:
alternatively, we could re-implement the project. I'm not sure why it was deleted

pbiggar:
Deleted it cause it didn't do much. The goal was to compare against the ocaml implementation. Once that was gone, the benchmark wasn't much more than a few lines of code. Easy to reimplement in whatever form we need it next.

PR #4256, opened 2022-07-11
Start moving APIs over to the vanilla serializer

pbiggar:
Switch over any ApiServer serializations where the results are the same (I got the list by generating them all and running sha1sum over all the files in backend/serialization).

This switches over both Params and return types for many of the handlers. Any handlers still using the old types have OCamlTypes in them (either Expressions or Dval).

I originally started making a v1 url for them, but I realized if things were identical I didn't need that. I left in the v1 handlers because I'm going to swap over the others in the future.


StachuDotNet:
"I got the list by generating them all and running sha1sum over all the files in backend/serialization"
love this

StachuDotNet (approved):

PR #4260, opened 2022-07-12
Fix analysis timezone offset bug

StachuDotNet:
Aimed at reproducing and then resolving #4114.

Also forces integration tests to run _not_ in UTC, via a Playwright config, so that we can actually demonstrate the issue in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -302,7 +302,7 @@ jobs:
       - run: shasum fsharp-backend/paket.lock fsharp-backend/global.json > ../checksum
       - restore_cache:
           keys:
-            - v12-fsharp-backend-{{ checksum "../checksum" }}
+            - v13-fsharp-backend-{{ checksum "../checksum" }}
             # Fails often enough that it's better not to have a fallback
       - show-large-files-and-directories
       # For tests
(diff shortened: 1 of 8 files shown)


StachuDotNet:
> Also forces integration tests to run _not_ in UTC, via a Playwright config, so that we can actually demonstrate the issue in CI.

FWIW, I've ran the tests locally in non-New_York timezones, and had different failing results. (just confirming that the setting _works)

pbiggar:
I imagined the fix would be to set the timezone for the runtime to always be UTC, and to ensure we called that in Wasm.

pbiggar:
This I think: https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.currentculture?view=net-6.0

StachuDotNet:
> This I think: https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.currentculture?view=net-6.0

I believe that's only relevant to formatting (am/pm vs 24hr). I can't see any way that it's related to the active timezone.

From that page:
> The [CultureInfo](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo?view=net-6.0) object that's returned by this property and its associated objects determine the default format for dates, times, numbers, currency values, the sorting order of text, casing conventions, and string comparisons.

StachuDotNet:
Note: the tuple changes here aren't relevant to the work done, but were rather hidden by our caching of F# builds in CI. Definitely something to address, separately from this issue, but figured we should fix it while it's causing pain.

StachuDotNet:
@pbiggar does this look good to you? Seems low-risk to me.

pbiggar (approved):
Looks good, thanks

PR #4269, opened 2022-07-13
Delete old backend/templates directory

StachuDotNet:
We can still delete some more code in `backend` that references these files, so this is a WIP.

Sometimes I stumble upon these files when I look for `ui.html` and see both copies; this is just to remove that annoyance.


pbiggar:
It seems that we incorrectly use these files in containers/gcp-fsharp-apiserver, so we should change that too.

PR #4271, opened 2022-07-13
Allow admin fns for all admin canvases

### LINKED ISSUE #4270: Update Editor so Admin-only fns are available in Admin canvases even if user is not an Admin
`LibDarkInternal` functions are executable by users who are Admins. E.g. they're exposed to the `dark` and `ops` users.

Non-admin users may have access to Admin canvases.
e.g. I have access to `ops-` and `dark-` canvases.

In such a scenario, although the user (`stachu`) is not an admin, they are able to execute Admin-only functions, by editing the Admin canvas (`ops-x`).

This is fine!

That said, since the accessing user is not an admin, they are currently restricted from getting proper autocomplete and such in the editor, because the `ALLFUNCTIONS`/`userIsAdmin` value injected in `ApiServer/UI.fs` bases the availability of such functions on the _user_ rather than the _canvas_.

At minimum, the resolution to this Issue is to change that logic, instead including all functions accessible to the _canvas_ that the user has access to.

StachuDotNet:
will resolve #4270

### DIFF
--- a/fsharp-backend/src/ApiServer/Ui.fs
+++ b/fsharp-backend/src/ApiServer/Ui.fs
@@ -153,6 +153,7 @@ let uiHandler (ctx : HttpContext) : Task<string> =
     if integrationTests && Config.allowTestRoutes then
       do! LibBackend.Canvas.loadAndResaveFromTestFile canvasInfo
 
+    // CLEANUP this results in 2 DB queries, but could be reduced to 1
     let! canAccessOperations =
       Account.usernameForUserID canvasInfo.owner
       |> Task.bind (fun u ->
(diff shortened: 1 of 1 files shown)


pbiggar:
We should probably reuse the `canAccessOperations` functionality (it's a bit overwrought, but better to have one thing than two)

fsharp-backend/src/ApiServer/Ui.fs
```diff
@@ -152,11 +153,19 @@ let uiHandler (ctx : HttpContext) : Task<string> =
     if integrationTests && Config.allowTestRoutes then
       do! LibBackend.Canvas.loadAndResaveFromTestFile canvasInfo
 
+    let! isCanvasOwnerAdmin =
```
pbiggar, line 156: Can you rename this to something like `canAccessOperations`?

This isn't really whether the canvas owner is an admin, it's whether we can use operations functions on this canvas.

pbiggar (changes requested):

pbiggar (approved):
Can you add a CLEANUP comment to the effect that this is two DB calls and could be done in one?

PR #4275, opened 2022-07-14
Remove Twitter Swagger from code/config

StachuDotNet:
I noticed that we had a `backend/swagger` folder with a single `twitter.json` document in it.

I can't think of a single place where we're using this. We don't have a twitter package currently exposed to the public - that's the only place I could think where this would have been referenced.

So, seems pretty safe to remove?


pbiggar (changes requested):
I'd prefer not to merge any changes to ocaml code. They all present some risk and there's not much value to making the changes.

I think in general it's worth it if we need to do something to the F# code and it intersects with it, but otherwise the risk of something going slightly wrong isn't worth the effort. We can delete it wholesale at the end.

StachuDotNet:
Okay. I guess seeing #4195 (and likely other PRs I'm forgetting) made me think it's reasonable to trim down OCaml code.

pbiggar:
The intent of #4195 was to trim out what was easy, so that we could then ignore the rest until we could cut it out all at once (otherwise we'd end up breaking files apart, and other stuff that was more risky than the very coarse cut in that PR).

Yeah, I think closing #4274 is a good idea.

Issue #4259, opened 2022-07-12
Rewrite Static Assets Upload in Dark


StachuDotNet:
This is a large issue that forces us to tackle other large things, so won't be done for a while.

- [x] expose more `LibDarkInternal` fns: (implemented in #4278)
  - `startStaticAssetsDeploy` that results in an `INSERT` into the relevant DB
  - `finishStaticAssetsDeploy` that results in an `UPDATE` into the relevant DB (changing status)
- [ ] Dark needs to accept Bytes in HTTP handler requests (see https://github.com/darklang/dark/discussions/4325 and document linked within)
- [ ] Dark needs a new HttpClient that can request bytes (#3840)
- [ ] Then we need to actually write the Dark code

StachuDotNet:
Note: this issue is towards replacing just the Static Assets _upload_ functionality in Dark.

Long-term, we may replace the entire feature in Dark. We can store the deploy references in a Dark Datastore and expose equivalents of the existing functions. The main wrinkle in "removing all Static Assets stuff from the Dark _repo_" project is that the Editor/Client shows the Static Assets within the sidebar, and handles some relevant Pushes. This will be an exciting thing to explore.

#3932 is the best Issue we have of tracking that larger project, though isn't currently being explored, likely ignored until the upload functionality replacement is complete.

StachuDotNet:
The relevant OCaml code can be found starting in `libbackend/webserver.ml`,
- `static_assets_upload_handler` fn [here](https://github.com/darklang/dark/blob/a47dbb37ff460275e5c64911d0e02d58bfc8e6e5/backend/libbackend/webserver.ml#L565)
- the API route [here](https://github.com/darklang/dark/blob/a47dbb37ff460275e5c64911d0e02d58bfc8e6e5/backend/libbackend/webserver.ml#L1911)

This code reaches out to various other parts of the old `backend` - worth digging in more.

StachuDotNet:
Thinking again, I'm not sure why I claimed that we need a new HttpClient that can request bytes for this Issue? If I can't remember why I claimed such, I'll remove it from the list in OP in a day.

StachuDotNet:
There are recent efforts towards a byte-friendly HTTP Client and byte-friendly HTTP handlers [here](#4633) and [here](#4353) respectively. While both still aren't complete, the client feels closer than the handler (which will require several follow-up PRs outlined in [this Discussion](https://github.com/darklang/dark/discussions/4325) and the document linked within)

[...]

PR #4278, opened 2022-07-14
Expose more internal staticassets fns for uploads

StachuDotNet:
This PR will expose and test 3 new functions for use in `dark-editor`, to support an eventual replacement of the OCaml-supported static asset upload. It's a piece of #4259.
The three functions will support `start`ing, `finish`ing, and `delete`-ing a deploy.

### DIFF
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
@@ -791,34 +791,19 @@ human-readable data."
     { name = fn "DarkInternal" "finishStaticAssetDeploy" 0
       parameters =
         [ Param.make "canvasID" TUuid ""; Param.make "deployHash" TStr "" ]
-      returnType =
-        TResult(
-          TRecord [ "deployHash", TStr
-                    "url", TStr
-                    "status", TStr
-                    "lastUpdate", TDate ],
-          TStr
-        )
+      returnType = TResult(TNull, TStr)
       description = "Marks an in-progress static asset deployment as finished"
       fn =
         internalFn (function
           | _, [ DUuid canvasID; DStr deployHash ] ->
             uply {
               let! canvasMeta = Canvas.getMetaFromID canvasID
-              let! deploy =
+              let! _updatedAt =
                 StaticAssets.finishStaticAssetDeploy
                   canvasID
                   canvasMeta.name
                   deployHash
-              return
-                DObj(
-                  Map [ "deployHash", DStr deploy.deployHash
-                        "url", DStr deploy.url
-                        "status", DStr(string deploy.status)
-                        "lastUpdate", DDate(DDateTime.fromInstant deploy.lastUpdate) ]
-                )
-                |> Ok
-                |> DResult
+              return DResult(Ok DNull)
             }
           | _ -> incorrectArgs ())
       sqlSpec = NotQueryable
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
@@ -804,6 +804,46 @@ human-readable data."
       previewable = Impure
       deprecated = NotDeprecated }
 
+
+    // TODO: re-evaluate access control of this fn. (might need some work)
+    // though likely such access control should be in the Dark code which uses this!
+    { name = fn "DarkInternal" "finishStaticAssetDeploy" 0
+      parameters =
+        [ Param.make "canvasID" TUuid ""; Param.make "deployHash" TStr "" ]
+      returnType =
+        TResult(
+          TRecord [ "deployHash", TStr
+                    "url", TStr
+                    "status", TStr
+                    "lastUpdate", TDate ],
+          TStr
+        )
+      description = "Marks an in-progress static asset deployment as finished"
+      fn =
+        internalFn (function
+          | _, [ DUuid canvasID; DStr deployHash ] ->
+            uply {
+              let! canvasMeta = Canvas.getMetaFromID canvasID
+              let! deploy =
+                StaticAssets.finishStaticAssetDeploy
+                  canvasID
+                  canvasMeta.name
+                  deployHash
+              return
+                DObj(
+                  Map [ "deployHash", DStr deploy.deployHash
+                        "url", DStr deploy.url
+                        "status", DStr(string deploy.status)
+                        "lastUpdate", DDate(DDateTime.fromInstant deploy.lastUpdate) ]
+                )
+                |> Ok
+                |> DResult
+            }
+          | _ -> incorrectArgs ())
+      sqlSpec = NotQueryable
+      previewable = Impure
+      deprecated = NotDeprecated }
+
     // ---------------------
     // Apis - 404s
     // ---------------------
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
@@ -776,13 +776,12 @@ human-readable data."
             uply {
               match! Account.getUser (UserName.create username) with
               | None ->
-                Exception.raiseInternal $"User not found" [ "username", username ]
                 return DResult(Error(DStr "User not found"))
               | Some user ->
                 let! canvasMeta = Canvas.getMetaFromID canvasID
                 let! deployHash =
                   StaticAssets.startStaticAssetDeploy user canvasID canvasMeta.name
-                return DStr deployHash |> Ok |> DResult
+                return DResult(Ok(DStr deployHash))
             }
           | _ -> incorrectArgs ())
       sqlSpec = NotQueryable
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
@@ -61,7 +61,7 @@ let internalFn (f : BuiltInFnSig) : BuiltInFnSig =
 
 let modifySchedule (fn : CanvasID -> string -> Task<unit>) =
   internalFn (function
-    | state, [ DUuid canvasID; DStr handlerName ] ->
+    | _, [ DUuid canvasID; DStr handlerName ] ->
       uply {
         do! fn canvasID handlerName
         let! s = SchedulingRules.getWorkerSchedules canvasID
@@ -106,7 +106,7 @@ let fns : List<BuiltInFn> =
 that's already taken, returns an error."
       fn =
         internalFn (function
-          | state, [ DStr username; DStr email; DStr name; DObj analyticsMetadata ] ->
+          | _, [ DStr username; DStr email; DStr name; DObj analyticsMetadata ] ->
             uply {
               let username =
                 Exception.catchError (fun () ->
@@ -200,7 +200,7 @@ that's already taken, returns an error."
       description = "Set whether a user is an admin. Returns null."
       fn =
         internalFn (function
-          | state, [ DStr username; DBool admin ] ->
+          | _, [ DStr username; DBool admin ] ->
             uply {
               let username = UserName.create username
               do! Account.setAdmin admin username
@@ -301,7 +301,7 @@ that's already taken, returns an error."
       description = "Pushes an event to Honeycomb"
       fn =
         internalFn (function
-          | state, [ DStr canvasID; DStr event; payload ] ->
+          | _, [ DStr canvasID; DStr event; payload ] ->
             (try
               Pusher.push
                 (canvasID |> System.Guid.Parse)
@@ -695,7 +695,7 @@ in OCaml; its primary purpose is to send data to honeycomb, but also gives
 human-readable data."
       fn =
         internalFn (function
-          | state, [] ->
+          | _, [] ->
             uply {
               let! tableStats = Db.tableStats ()
               // Send events to honeycomb. We could save some events by sending
@@ -764,6 +764,45 @@ human-readable data."
       previewable = Impure
       deprecated = NotDeprecated }
 
+
+    { name = fn "DarkInternal" "startStaticAssetDeploy" 0
+      parameters = [ Param.make "username" TStr ""; Param.make "canvasID" TUuid "" ]
+      returnType =
+        TResult(
+          TRecord [ "deployHash", TStr
+                    "url", TStr
+                    "status", TStr
+                    "lastUpdate", TDate ],
+          TStr
+        )
+      description = "Records an in-progress static asset deployment"
+      fn =
+        internalFn (function
+          | _, [ DStr username; DUuid canvasID ] ->
+            uply {
+              match! Account.getUser (UserName.create username) with
+              | None ->
+                Exception.raiseInternal $"User not found" [ "username", username ]
+                return DResult(Error(DStr "User not found"))
+              | Some user ->
+                let! canvasMeta = Canvas.getMetaFromID canvasID
+                let! deploy =
+                  StaticAssets.startStaticAssetDeploy user canvasID canvasMeta.name
+                return
+                  DObj(
+                    Map [ "deployHash", DStr deploy.deployHash
+                          "url", DStr deploy.url
+                          "status", DStr(string deploy.status)
+                          "lastUpdate",
+                          DDate(DDateTime.fromInstant deploy.lastUpdate) ]
+                  )
+                  |> Ok
+                  |> DResult
+            }
+          | _ -> incorrectArgs ())
+      sqlSpec = NotQueryable
+      previewable = Impure
+      deprecated = NotDeprecated }
 
     // ---------------------
     // Apis - 404s
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
@@ -767,15 +767,9 @@ human-readable data."
 
     { name = fn "DarkInternal" "startStaticAssetDeploy" 0
       parameters = [ Param.make "username" TStr ""; Param.make "canvasID" TUuid "" ]
-      returnType =
-        TResult(
-          TRecord [ "deployHash", TStr
-                    "url", TStr
-                    "status", TStr
-                    "lastUpdate", TDate ],
-          TStr
-        )
-      description = "Records an in-progress static asset deployment"
+      returnType = TResult(TStr, TStr)
+      description =
+        "Records an in-progress static asset deployment, returning the deployHash"
       fn =
         internalFn (function
           | _, [ DStr username; DUuid canvasID ] ->
@@ -786,18 +780,9 @@ human-readable data."
                 return DResult(Error(DStr "User not found"))
               | Some user ->
                 let! canvasMeta = Canvas.getMetaFromID canvasID
-                let! deploy =
+                let! deployHash =
                   StaticAssets.startStaticAssetDeploy user canvasID canvasMeta.name
-                return
-                  DObj(
-                    Map [ "deployHash", DStr deploy.deployHash
-                          "url", DStr deploy.url
-                          "status", DStr(string deploy.status)
-                          "lastUpdate",
-                          DDate(DDateTime.fromInstant deploy.lastUpdate) ]
-                  )
-                  |> Ok
-                  |> DResult
+                return DStr deployHash |> Ok |> DResult
             }
           | _ -> incorrectArgs ())
       sqlSpec = NotQueryable
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
@@ -805,8 +805,6 @@ human-readable data."
       deprecated = NotDeprecated }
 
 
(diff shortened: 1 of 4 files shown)


StachuDotNet:
Does this seem like the right path to you, @pbiggar?

Specifically, there was originally an intent to record when we're in-progress of a static asset upload - is that still important? (I believe so, but confirming.)

pbiggar:
> Does this seem like the right direction to you, @pbiggar?

Looks generally what I'd imagine it would look like.

You had mentioned storing deploys in the dark-editor canvas instead of the DB - looks like you're not going that way at the moment? 
 
> Specifically, there was originally an intent to record when we're in-progress of a static asset upload - is that still important? (I believe so, but confirming.)

Yes I think that's important.

StachuDotNet:
Thanks!

> You had mentioned storing deploys in the dark-editor canvas instead of the DB - looks like you're not going that way at the moment?

I imagine "migrate Static Asset management fully to Dark" will be a bigger story, involving:
- starting the new package
- re-implementing the existing functionality within that package, and exposing the relevant fns
- making users' existing code start using the new package
- migrating user data

So for now, I'm just focusing on replacing the _upload_ - this PR (which shouldn't take much longer) should be all that's needed to support that, besides work that needs to be done generally. (i.e. updating httpclient and handlers.)

Does that sound fair?

pbiggar:
> I imagine "migrate Static Asset management fully to Dark" will be a bigger story, involving:

We might be on a different page here.
 
> * starting the new package

It might be worth clarifying what exactly you mean by "package" here. It sounds like you mean something using the package manager.

I had thought this was going to be a set of APIs used by the cli and the editor.

> * making users' existing code start using the new package

Unclear what's meant by this?

> * ? maybe migrating the actual assets. They could continue living where they do now in our GCP billing, but they could also live somewhere else

I don't know why we'd do this?

StachuDotNet:
>  It sounds like you mean something using the package manager.

Yes.

---

To clarify:
- Right now I'm only focused on supporting the CLI-based upload, with an endpoint written in Dark.
  That demands enough sub-projects that I'm not really looking beyond that for this line of work.
- I have an _idea_ that we could replace the whole functionality with a package, in the package manager, but it's not fully thought out. Those bullets are me summarizing how it could pan out and the questions it raises, all of which I was planning to punt until the upload work is complete.

By "making users' existing code start using the new package," I meant supporting some way of migrating users from the dark-repo-supported Static Assets feature to a package-supported Static Assets feature. Again, sketching ideas.

---

> > * ? maybe migrating the actual assets. They could continue living where they do now in our GCP billing, but they could also live somewhere else
> 
> I don't know why we'd do this?

Nor do I - just questions that would be reasonably raised at time of considering the whole "static assets in a Package" idea.

FWIW I think it's pretty reasonable for static assets to be an official Dark package. In our zoom the other day I noted that my main concern there would be with the Editor showing deploys in the UI - we'd need to arrange some special way of hooking into a package, which is probably too far-fetched to consider currently.

pbiggar:
Clarified via zoom. I had gotten confused between the current project and the future dream stuff. My bad!

StachuDotNet:
There's a `branch` column in the `static_asset_deploys` table, seemingly always set to `"main"`.
What is that for? Should I just assume everything is working within "main" at this point?

pbiggar:
> There's a `branch` column in the `static_asset_deploys` table, seemingly always set to `"main"`. What is that for? Should I just assume everything is working within "main" at this point?

That's to allow deploys on different branches. I can confirm that all data in the DB has this as "main".

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
```diff
@@ -765,6 +765,112 @@ human-readable data."
...
+                  |> DResult
+            }
+          | _ -> incorrectArgs ())
+      sqlSpec = NotQueryable
+      previewable = Impure
+      deprecated = NotDeprecated }
+
+
+    // TODO should this take username as a parameter?
+    // should user B be able to finish user A's upload?
```
StachuDotNet, line 809: Outstanding question - if user A starts a static asset deploy, should other users be able to "finish" that deploy?
I don't really see any risk personally.

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
```diff
@@ -765,6 +765,112 @@ human-readable data."
...
+
+
+    { name = fn "DarkInternal" "deleteStaticAssetDeploy" 0
+      parameters =
+        [ Param.make "username" TStr ""
+          Param.make "canvasID" TUuid ""
+          Param.make "deployHash" TStr "" ]
+      returnType =
+        // TODO what should the return type be here? I basically want an empty result.
+        TResult(TNull, TStr)
```
StachuDotNet, line 855: What's a good return type for `deleteStaticAssetDeploy`? Intuitively I want it to be a `Result` of some sort, but wouldn't expect any "data" to come back. `Result<unit, string>` is what I'd normally reach for, but since we don't have `unit`, `null` is the best idea I have.

fsharp-backend/src/BackendOnlyStdLib/LibStaticAssets.fs
```diff
@@ -152,7 +152,7 @@ let fns : List<BuiltInFn> =
         (function
         | state, [] ->
           uply {
-            match! SA.latestDeployHash state.program.canvasID with
+            match! SA.getLatestDeployHash state.program.canvasID with
```
StachuDotNet, line 155: Maybe this rename is useless? I personally find it more immediately obvious, but could also see it as too verbose. :shrug:

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +95,127 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+                      "uploadedBy", Sql.uuid user.id ]
+  |> Sql.executeRowAsync (fun reader -> reader.instant "created_at")
+  |> Task.map (fun lastUpdate ->
+    { deployHash = deployHash
+      url = url canvasName deployHash Short
+      lastUpdate = lastUpdate
+      status = Deploying })
+
+
+// TODO: what should happen if the deploy hash doesn't exist?
```
StachuDotNet, line 154: Nothing? Error?

StachuDotNet: I lean towards 'Error' here.

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +95,127 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+// TODO: what if the deploy is already finished?
+let finishStaticAssetDeploy
+  (canvasID : CanvasID)
+  (canvasName : CanvasName.T)
+  (deployHash : string)
+  : Task<StaticDeploy> =
```
StachuDotNet, line 160: I'll update this to return a `Result<StaticDeploy, string>`. (though I'm not a huge fan of using `Result<T, string>` - feels like we immediately lose some useful data).

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +95,127 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+let deleteStaticAssetDeploy
+  (user : Account.UserInfo)
+  (canvasID : CanvasID)
+  (deployHash : string)
+  : Task<unit> =
+
+  // TODO the query here only allows someone to delete a deploy if they're the
+  // one who uploaded it (via `AND uploaded_by_account_id=@userId`). If more
+  // than one user are working within a canvas, shouldn't we allow anyone
+  // involved to delete the deploy?
```
StachuDotNet, line 193: Outstanding question - what do you think @pbiggar? Personally I want to remove the `AND uploaded_by_account_id=@uploadedBy` filter, as anyone who has access to a canvas should be able to delete a deploy. (this filter was stolen by our OCaml code)

StachuDotNet:
There are a few more tests and tidyings to address, but worth an intermediate review in the meantime

StachuDotNet:
I've marked several TODOs as CLEANUPs here. This at least matches status quo. I can follow up and resolve the CLEANUPs here with some input, otherwise this achieves what we need for now.

fsharp-backend/src/BackendOnlyStdLib/LibStaticAssets.fs
pbiggar: Generally we have not used to get prefix, and I don't really like it.

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
```diff
@@ -765,6 +765,110 @@ human-readable data."
...
+        )
+      description = "Records an in-progress static asset deployment"
+      fn =
+        internalFn (function
+          | _, [ DStr username; DUuid canvasID ] ->
+            uply {
+              match! Account.getUser (UserName.create username) with
+              | None ->
+                Exception.raiseInternal $"User not found" [ "username", username ]
+                return DResult(Error(DStr "User not found"))
```
pbiggar, line 786: You don't need this if you're raising an exception. Where do you want the error checking to go here? If you're planning on returning a `DResult` then you don't need an exception, but if you're planning on a `DError`, then what is the point in `DResult`

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
```diff
@@ -765,6 +765,110 @@ human-readable data."
...
+          | _, [ DUuid canvasID; DStr deployHash ] ->
+            uply {
+              let! canvasMeta = Canvas.getMetaFromID canvasID
+              let! deploy =
+                StaticAssets.finishStaticAssetDeploy
+                  canvasID
+                  canvasMeta.name
+                  deployHash
+              return
+                DObj(
```
pbiggar, line 831: I'm not sure why it returns this information which the user already has.

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +97,132 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+  |> Sql.executeAsync (fun read ->
+    let deployHash = read.string "deploy_hash"
+
+    let status, lastUpdate =
+      match read.instantOrNone "live_at" with
+      | Some datetime -> Deployed, datetime
+      | None -> Deploying, read.instant "created_at"
+
+    { deployHash = deployHash
+      url = url canvasName deployHash Short
```
pbiggar, line 120: I'm not sure this is a good spot to create the url.

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +97,132 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+      (canvas_id, branch, deploy_hash, uploaded_by_account_id)
+    VALUES (@canvasID, @branch, @deployHash, @uploadedBy)
+    RETURNING created_at"
+  |> Sql.parameters [ "canvasID", Sql.uuid canvasID
+                      "branch", Sql.string branch
+                      "deployHash", Sql.string deployHash
+                      "uploadedBy", Sql.uuid user.id ]
+  |> Sql.executeRowAsync (fun reader -> reader.instant "created_at")
+  |> Task.map (fun lastUpdate ->
+    { deployHash = deployHash
```
pbiggar, line 153: That doesn't seem to be a good reason to create this whole record here.

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +97,132 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+      url = url canvasName deployHash Short
+      lastUpdate = lastUpdate
+      status = Deployed })
 
 
+/// <summary>Removes references to a canvas' static asset deploy</summary>
+/// <remarks>
+/// Since postgres doesn't have named transactions, we just delete the db
+/// record in question. For now, we're leaving files where they are; the right
+/// thing to do here would be to shell out to `gsutil -m rm -r` - leaving  this
```
pbiggar, line 184: This comment is out of date. It's clear we should handle the deletion in the canvas.

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +97,132 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+let deleteStaticAssetDeploy
+  (user : Account.UserInfo)
+  (canvasID : CanvasID)
+  (deployHash : string)
+  : Task<unit> =
+
+  // CLEANUP the query here only allows someone to delete a deploy if they're
+  // the one who uploaded it (via `AND uploaded_by_account_id=@userId`). If
+  // more than one user are working within a canvas, shouldn't we allow anyone
+  // involved to delete the deploy? I believe so.
```
pbiggar, line 197: Yes anyone with right access to the canvas has the same permissions and so they should be able to delete these.

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +97,132 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+let allDeploysInCanvas
+  (canvasName : CanvasName.T)
+  (canvasID : CanvasID)
+  : Task<List<StaticDeploy>> =
+  Sql.query
+    "SELECT deploy_hash, created_at, live_at
+     FROM static_asset_deploys
+     WHERE canvas_id=@canvasID
+     ORDER BY created_at
+     DESC LIMIT 25"
```
pbiggar, line 109: You'll probably be happier making the limit a parameter.

fsharp-backend/src/LibBackend/StaticAssets.fs
pbiggar: That's said I don't think this is what the logic says; the logic just wants you to know who the author is. Not sure it makes much of a difference.

fsharp-backend/tests/testfiles/internal.tests
```diff
@@ -184,3 +184,67 @@ DarkInternal.unlockedDBs Test.getCanvasID = [1]
 [test.unlocked, one but locked] with DB X
 (let _ = DB.set_v1 { x = "str" } "test" X in
  DarkInternal.unlockedDBs Test.getCanvasID) = []
+
+
+// ---------------
+// static assets
+// ---------------
+[test.no static assets without deploys]
```
pbiggar, line 192: Were these tests ported from somewhere?

fsharp-backend/tests/testfiles/internal.tests
```diff
@@ -184,3 +184,67 @@ DarkInternal.unlockedDBs Test.getCanvasID = [1]
 [test.unlocked, one but locked] with DB X
 (let _ = DB.set_v1 { x = "str" } "test" X in
  DarkInternal.unlockedDBs Test.getCanvasID) = []
+
+
+// ---------------
```
pbiggar, line 189: Maybe I'm not looking in the right place but I don't see a test for fetching all the deploys.

pbiggar (changes requested):
Looking good, some small changes and we should be good to go

fsharp-backend/tests/testfiles/internal.tests
StachuDotNet: No, I made them up.

fsharp-backend/tests/testfiles/internal.tests
StachuDotNet: The tests in `internal.tests` call the 'fetch all deploys' (`DarkInternal.staticAssetsDeploys`) fn in various scenarios, and check its simple length, or map to the `status` and just check _that_ resultant list.

The result of that call returns dynamic data (`deployHash`, `url`, `lastDate`) that aren't quite useful to test here beyond checking count and `status` data. If you have other tests to recommend or ways of capturing these fields reasonably in tests, I'm open to ideas!

fsharp-backend/src/LibBackend/StaticAssets.fs
StachuDotNet: Sounds good - was just copying old OCaml code.

StachuDotNet: Actually, I didn't even implement this. I just moved it up in the file.

pbiggar: Oh, nvm then, I thought it was being exposed as an internal function

fsharp-backend/src/LibBackend/StaticAssets.fs
StachuDotNet: This is again the code that I simply moved, not implemented; ignoring.

fsharp-backend/src/LibBackend/StaticAssets.fs
StachuDotNet: Enh, I'll revert on second thought - the description alone should make it clear.

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +97,115 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+  : Task<StaticDeploy> =
+  Sql.query
+    "UPDATE static_asset_deploys
+      SET live_at = NOW()
+      WHERE canvas_id = @canvasID AND deploy_hash = @deployHash
+      RETURNING live_at"
+  |> Sql.parameters [ "canvasID", Sql.uuid canvasID
+                      "deployHash", Sql.string deployHash ]
+  |> Sql.executeRowAsync (fun reader -> reader.instant "live_at")
+  |> Task.map (fun lastUpdate ->
```
pbiggar, line 170: I'm unclear why we're creating this record here?

fsharp-backend/src/LibBackend/StaticAssets.fs
```diff
@@ -107,8 +97,115 @@ let latestDeployHash (canvasID : CanvasID) : Task<Option<string>> =
...
+      url = url canvasName deployHash Short
+      status = status
+      lastUpdate = lastUpdate })
+
+/// Creates a record of a new static asset deploy, returning a deployHash to be
+/// used by the actual upload process.
+let startStaticAssetDeploy
+  (user : Account.UserInfo)
+  (canvasID : CanvasID)
+  (canvasName : CanvasName.T)
```
pbiggar, line 129: I believe we're not using canvasName here anymore?

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
```diff
@@ -765,6 +765,85 @@ human-readable data."
...
+    { name = fn "DarkInternal" "startStaticAssetDeploy" 0
+      parameters = [ Param.make "username" TStr ""; Param.make "canvasID" TUuid "" ]
+      returnType = TResult(TStr, TStr)
+      description =
+        "Records an in-progress static asset deployment, returning the deployHash"
+      fn =
+        internalFn (function
+          | _, [ DStr username; DUuid canvasID ] ->
+            uply {
+              match! Account.getUser (UserName.create username) with
```
pbiggar, line 777: I think we don't need the user anymore?

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
```diff
@@ -765,6 +765,85 @@ human-readable data."
...
+            uply {
+              let! canvasMeta = Canvas.getMetaFromID canvasID
+              let! deploy =
+                StaticAssets.finishStaticAssetDeploy
+                  canvasID
+                  canvasMeta.name
+                  deployHash
+              return
+                DObj(
+                  Map [ "deployHash", DStr deploy.deployHash
```
pbiggar, line 815: Similar to `start`, what's the reason for this structure? Most of it is known to the caller already, right?

pbiggar (changes requested):
Looking good, a couple of small questions remain.

fsharp-backend/src/LibBackend/StaticAssets.fs
StachuDotNet: I thought I'd handled this before - thanks for reminding me.

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
StachuDotNet: Now returning a plain Result<Null, string>. We could return the update date, but I don't currently feel a need. We can always ajust.

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
StachuDotNet: We currently use it during the INSERT of a `static_asset_deploys`, in the `uploadedBy` column ("uploadedBy", Sql.uuid user.id`

fsharp-backend/src/LibBackend/StaticAssets.fs
StachuDotNet: thanks

pbiggar (approved):
Looks great, thanks!

PR #4281, opened 2022-07-14
Include `TFloat` case in `client`'s `DType`.`decodeOld`

StachuDotNet:
This will resolve https://rollbar.com/darkops/darklang/items/3412


StachuDotNet:
decodeOld is also missing TAny - is that OK?

pbiggar:
Probably needs it. Unsure.

PR #4300, opened 2022-07-18
Update/improve 'running against prod' doc

StachuDotNet:
Just some updates that would have been helpful to me as I was going through this.

### DIFF
--- a/docs/running-against-production.md
+++ b/docs/running-against-production.md
@@ -1,6 +1,11 @@
 # Running the client against production
 
-If you want to run your local client against the production server, we have a way to do that using ngrok:
+If you want to run your local client against the production server, we have a way to do that using ngrok
+
+Note: `<your-username>` below should be your Darklang username.
+It's intended to disambiguate between users so we don't run over each other's
+tunnels. It might also be a good idea to use some random digits as a suffix so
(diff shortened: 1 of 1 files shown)


docs/running-against-production.md
```diff
@@ -1,25 +1,62 @@
...
+
 - Join our ngrok account by asking Paul
 - install ngrok: `brew cask install ngrok` on OS X, or follow instructions at
   https://ngrok.com/download for Linux. (The snap installer is broken.)
 - authorize your ngrok client: `ngrok authtoken YOURTOKEN` (found at
   https://dashboard.ngrok.com/auth)
-- run your tunnel: `ngrok http 8000 --hostname=darklang-<username>.ngrok.io`
-- You can simplify this to `ngrok start darklang` by adding to your
+- run your tunnel:
+  `ngrok http darklang.localhost:8000 --hostname=darklang-<your-username>.ngrok.io`
```
pbiggar, line 13: It would be worth mentioning here that your username should be your Darklang username -- it's intended to disambiguate between users so we don't run over each other's tunnels. It might also be a good idea to use some random digits as a suffix so people can't get access to our local servers by guessing our usernames

pbiggar (approved):
Looks good, thanks!

PR #4302, opened 2022-07-18
Reference the correct HTML templates (ui, login)

### LINKED ISSUE #4280: Reference the new template html files
#4274 has some bits that are worth saving - although we don't need to change any OCaml code, there's no need for us to be referencing the old .html files

StachuDotNet:
This will resolve #4280, and is a subset of #4274 (which had some risk to it, as it changed OCaml code).


pbiggar (commented):
Can you go through how do you know that this is a safe change?

StachuDotNet:
8 files have changed:
- `backend/templates/login.html`
- `backend/templates/ui.html`
- `config/circleci`
- `config/dev`
- `config/dev_prodclone`
- `containers/gcp-fsharp-apiserver/Dockerfile`
- `containers/ocaml-container-prep.sh`
- `scripts/build/compile`

---

`scripts/build/compile` is simple: instead of waiting for changes to the old .html files and rebuilding the _old_ backend, we instead wait for changes to the new .html files, and rebuild the _new_ backend.

---

The only references to `backend/templates` and its content were via the `DARK_CONFIG_TEMPLATES_DIR` config.
I searched throughout the code for `ui.html`, `login.html`, `backend/templates`, and `templates` to verify this.

So, I changed 3/4 of the `config/` files to point to the equivalent files in `fsharp-backend/src/ApiServer/Templates` instead.

The _other_ config file, `/config/gke-builtwithdark`, has the setting set as such:

`DARK_CONFIG_TEMPLATES_DIR=/home/dark/templates`.

After some digging, I found this `templates` directory to be filled by `containers/gcp-fsharp-apiserver/Dockerfile`, so updated it to copy the right file.

That covers:
- `config/circleci`
- `config/dev`
- `config/dev_prodclone`
- `backend/templates` (both file deletions)

---

`containers/ocaml-container-prep.sh`, I'm slightly more unsure of.

When searching for `template` across the full repo, I noted a `cp -R backend/templates "$DIR/"` line in there.
It seemed reasonable enough to update this to the new Templates folder instead, and figured the resultant copied files are likely unused anyway?

pbiggar:
Yeah, maybe let's close this for now, I really don't want to change any OCaml stuff we don't have to and this feels risky.

Issue #4265, opened 2022-07-13
Allow users to create/use tuples


StachuDotNet:
As of #2545 and #4182, we have _hidden_ support for Tuples - we don't allow users to create tuples, leaving the feature "turned off" via [this switch](https://github.com/darklang/dark/blob/a47dbb37ff460275e5c64911d0e02d58bfc8e6e5/client/src/fluid/Fluid.res#L34)

Here's work to be done with tuples:
- [x] add tuple stdlib fns #3309 
- [x] hide tuples behind a feature flag rather than a constant value in-code
  (so _we_ can use them in the meantime, and see what feels good/bad)
- [x] tuple `match` pattern: #4427
- [x] fluid bugfix: #4235 
- [x] fluid bugfix: #4264
- [ ] tuple destructuring #3402 
- [ ] resolve all remaining TUPLETODOs
- [ ] #4239 (incl. backend type checking)
  maybe this should be split up?
- [ ] #4339 
- [ ] remove the experimental setting - allow tuples by default
- [x] #3310 

---
These should not block the general release of tuples, but are useful:

- [ ] #4475

StachuDotNet:
I don't believe #4339 needs to be a blocker here, but I discovered the desire while writing a draft of `docs` updates.

pbiggar:
> I don't believe #4339 needs to be a blocker here, but I discovered the desire while writing a draft of `docs` updates.

I would say this should be done before we consider the project done.

pbiggar:
Putting #3310 on your radar!

StachuDotNet:
> Putting #3310 on your radar!

Thanks, I thought it was already on the list! Adding.

pbiggar:
So it is. I missed it when reading.

StachuDotNet:
Ah right - forgot I put it in the footnote.

pbiggar:
I think we could enable to all users without #4475

pbiggar:
Tuples are out

PR #4311, opened 2022-07-19
Add StdLib fns for Tuples

### LINKED ISSUE #3309: Add functions for tuples
After tuples are added, we need library support for them. These functions should be added

```
// Access
Tuple2::first ('a,'b) -> 'a
Tuple2::second ('a,'b) -> 'b

// Creation
Tuple2::pair ('a, 'b) -> ('a,'b)

// Manipulation
Tuple2::mapFirst (('a,'b), ('a -> 'c)) -> ('c, 'b)
Tuple2::mapSecond (('a,'b), ('b -> 'c)) -> ('b, 'c)
Tuple2::mapBoth (('a,'b), ('a -> 'c), ('b -> 'd)) -> ('c, 'd)

Tuple2::swap (('a,'b)) -> ('b, 'a)
```

- [ ] Add the Tuple2 functions above
- [ ] add the equivalents for Tuple3
- [ ] Look at https://package.elm-lang.org/packages/TSFoster/elm-tuple-extra/latest/Tuple3 to see if there are useful functions there

StachuDotNet:
This will resolve #3309, and is a part of #4265.

This adds the following stdlib fns:
- `Tuple2::pair`
- `Tuple2::first`
- `Tuple2::second`
- `Tuple2::swap`
- `Tuple2::mapFirst`
- `Tuple2::mapSecond`
- `Tuple2::mapBoth`
- `Tuple3::first`
- `Tuple3::second`
- `Tuple3::third`
- `Tuple3::mapFirst`
- `Tuple3::mapSecond`
- `Tuple3::mapThird`
- `Tuple3::mapAllThree`

### DIFF
--- a/fsharp-backend/src/LibExecutionStdLib/LibTuple.fs
+++ b/fsharp-backend/src/LibExecutionStdLib/LibTuple.fs
@@ -42,6 +42,21 @@ let fns : List<BuiltInFn> =
       deprecated = NotDeprecated }
 
 
+    { name = fn "Tuple2" "swap" 0
+      parameters =
+        [ Param.make "tuple" (TTuple(TVariable "a", TVariable "b", [])) "" ]
+      returnType = TTuple(TVariable "b", TVariable "a", [])
+      description = "Returns a 2-tuple with the elements swapped."
+      fn =
+        (function
+        | state, [ DTuple (first, second, []) ] ->
+          Ply(DTuple (second, first, []))
+        | _ -> incorrectArgs ())
+      sqlSpec = NotYetImplementedTODO
+      previewable = Pure
+      deprecated = NotDeprecated }
+
+
     // Tuple3
     { name = fn "Tuple3" "first" 0
       parameters =
--- a/fsharp-backend/src/LibExecutionStdLib/LibTuple.fs
+++ b/fsharp-backend/src/LibExecutionStdLib/LibTuple.fs
@@ -13,16 +13,15 @@ let err (str : string) = Ply(Dval.errStr str)
 let incorrectArgs = Errors.incorrectArgs
 
 let fns : List<BuiltInFn> =
-  [ { name = fn "Tuple2" "first" 0
-      parameters = [ Param.make "tuple" (TTuple(TVariable "a", TVariable "b", [])) "" ]
+  [ // Tuple2
+    { name = fn "Tuple2" "first" 0
+      parameters =
+        [ Param.make "tuple" (TTuple(TVariable "a", TVariable "b", [])) "" ]
       returnType = TVariable "a"
-      description =
-        // TODO: improve this text
-        "Returns the first part of a 2-tuple."
+      description = "Returns the first part of a 2-tuple."
       fn =
         (function
-        | state, [ DTuple (first, _second, []) ] ->
-          Ply(first)
+        | state, [ DTuple (first, _second, []) ] -> Ply(first)
         | _ -> incorrectArgs ())
       sqlSpec = NotYetImplementedTODO
       previewable = Pure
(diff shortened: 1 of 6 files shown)


StachuDotNet:
`Tuple2.first (1, blank)` right now results in `blank`.
While this is "right" per our decision to evaluate a tuple with any "fake values" as the first encountered fake value, it sort of surprised me when organically using the fn.

What do you think, @pbiggar? Are you still happy with that decision? I'm fine either way, but thought I should raise the question.

pbiggar:
> What do you think, @pbiggar? Are you still happy with that decision? I'm fine either way, but thought I should raise the question

It literally couldn't work any other way, as fakeval passed to functions always return the fakeval.

Or, to put another way, incomplete functions always return incomplete regardless of the subject matter

StachuDotNet:
yeah that makes sense - just sort of caught me off-guard during testing. Thanks.

StachuDotNet:
This is at a reviewable point.

A few questions:
- should we rename `Tuple2::tuple` to `Tuple2::create`?
- should we implement a `Tuple3::create`? Elm has `Tuple3.join`, but that naming doesn't sit right with me at all.
- would you like/dislike this alternative to one giant `let fns = [...]` list in `LibTuple.fs`?
  ```
  let tuple2fns = [...]
  let tuple3fns = [...]
  let fns = tuple2fns @ tuple3fns
  ```
  (just to break apart the code a bit)
- right now this doesn't have an equivalent to `Tuple2::swap` for `Tuple3`. Elm exposes `rotateLeft`, `rotateRight`, `swapFirst`, `swapLast`, and `reverse` all in that same space. Do you find value in these? I don't, and would personally leave them to be added if/when someone actually reaches for and requests them.

StachuDotNet:
- [ ] review the fn descriptions; see if/where there are places where we can improve.

pbiggar:
> * should we rename `Tuple2::tuple` to `Tuple2::create`?
>   I don't find `::tuple` to be intuitive, and reckon elm only named it that way to not collide with their `join` fn (which is equivalent to a `Tuple3::create`)

Yeah, I think `create` would be a better name.

> * should we implement a `Tuple3::create`?

Yes!

> * would you like/dislike this alternative to one giant `let fns = [...]` list in `LibTuple.fs`?
>   ```
>   let tuple2fns = [...]
>   let tuple3fns = [...]
>   let fns = tuple2fns @ tuple3fns
>   ```

Either one big list or two files I think.

> * right now this doesn't have an equivalent to `Tuple2::swap` for `Tuple3`. Elm exposes `rotateLeft`, `rotateRight`, `swapFirst`, `swapLast`, and `reverse` all in that same space. Do you find value in these? I don't, and would personally leave them to be added if/when someone actually reaches for and requests them.

Yeah I don't see a big need for them.

> * any other stdlib fns that should be implemented before [Add functions for tuples #3309](https://github.com/darklang/dark/issues/3309) may be considered closed?

No, I think this is great.

fsharp-backend/src/LibExecutionStdLib/LibTuple.fs
```diff
@@ -0,0 +1,348 @@
...
+      sqlSpec = NotYetImplementedTODO
+      previewable = Pure
+      deprecated = NotDeprecated }
+
+
+    { name = fn "Tuple2" "mapFirst" 0
+      parameters =
+        [ Param.makeWithArgs
+            "fn"
+            (TFn([ TVariable "a" ], TVariable "c"))
```
pbiggar, line 82: Would we do better with descriptive names instead of "a" and "c"? Nothing comes to mind personally.

pbiggar (approved):
Looks great. Some minor comments mostly around formatting of the descriptions (the initial comments apply to all descriptions).

I agree that `create` is a better name, and I think Tuple3 should have a create function.

fsharp-backend/src/LibExecutionStdLib/LibTuple.fs
StachuDotNet: I thought about using `first` etc. instead of `a`, but felt it to be a bit much. We can always adjust later.

fsharp-backend/src/LibExecutionStdLib/LibTuple.fs
```diff
@@ -0,0 +1,348 @@
...
+      sqlSpec = NotYetImplementedTODO
+      previewable = Pure
+      deprecated = NotDeprecated }
+
+
+    { name = fn "Tuple2" "first" 0
+      parameters =
+        [ Param.make "tuple" (TTuple(TVariable "a", TVariable "b", [])) "" ]
+      returnType = TVariable "a"
+      description = "Returns the first value of a 2-tuple."
```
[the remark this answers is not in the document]
StachuDotNet: I went back and forth on this :sweat_smile:. Happy to change to `pair`.

StachuDotNet:
I felt the number of newlines in tuple.tests was a bit silly - condensed the file a bit.

PR #4323, opened 2022-07-21
Add http handlers that handle bytes

StachuDotNet:
WIP. untested.

Next, I need to dive into `client` code.
Maybe this is enough for one PR, though?

The new middleware can probably be thinner; will follow up there.


backend/serialization/vanilla-LibBackend-Op-AddOpEventV1-simple.json
```diff
@@ -622,7 +622,7 @@
           ]
         ],
         "spec": [
-          "HTTP",
+          "HTTPLegacy",
```
pbiggar, line 625: Depending on whether this is unserialized, it may need to read the old types as well as the new types. I think it almost certainly does.

fsharp-backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -671,7 +671,7 @@ module Handler =
     | EveryMinute
 
   type Spec =
-    | HTTP of path : string * method : string
+    | HTTPLegacy of path : string * method : string // HTTPBytesTODO: does this serialize OK?
```
pbiggar, line 674: I believe runtime types handlers are not allowed serialize afaik.

fsharp-backend/src/LibBackend/Canvas.fs
```diff
@@ -531,7 +531,12 @@ let saveTLIDs
                 Routing.routeToPostgresPattern (PTParser.Handler.Spec.toName spec),
                 PTParser.Handler.Spec.toModifier spec
               )
-            | PT.Handler.HTTPBytes _ -> None // HttpBytesTODO
+            | PT.Handler.HTTPBytes _ ->
```
pbiggar, line 534: Better to merge with previous match case

fsharp-backend/src/LibBackend/Traces.fs
```diff
@@ -94,7 +110,27 @@ let savedInputVars
 
     [ ("request", event) ] @ boundRouteVariables
 
-  | PT.Handler.HTTPBytes _ // HttpBytesTODO
+  | PT.Handler.HTTPBytes (route, _, _) ->
```
pbiggar, line 113: If the code is the same as the previous match case, they could be merged.

fsharp-backend/src/LibBackend/Traces.fs
```diff
@@ -177,7 +213,9 @@ let traceIDsForHandler (c : Canvas.T) (h : PT.Handler.T) : Task<List<AT.TraceID>
             |> Option.bind (fun matching ->
               if matching.tlid = h.tlid then Some traceID else None)
 
-          | PT.Handler.HTTPBytes _ // HttpBytesTODO
+           // HttpBytesTODO we eventually need to actually load traces
+          | PT.Handler.Spec.HTTPBytes _ -> Some traceID
```
pbiggar, line 217: I think this would be the same as the http legacy functionality?

fsharp-backend/tests/Tests/Serialization.Tests.fs
```diff
@@ -300,6 +304,10 @@ module Values =
       testUnknownHandler
       testOldWorker ]
 
+  // HttpBytesTODO for some reason adding this to the above list makes the
```
pbiggar, line 307: What's going on here?

pbiggar (commented):
I took that as a request for a review :)

It does look like a good sized chunk for a single PR. Before merging this would need:

- the lists of the goals for the project
- a discussion of the serialization changes (and possible backward compatibility)
- a binary serialization test case

backend/serialization/vanilla-LibBackend-Op-AddOpEventV1-simple.json
StachuDotNet: Can you please clarify here?
My best guess/understanding is that `client` will have to adjust to handle both "HTTPLegacy" and "HTTP" when decoding, but unsure if that's what you mean, or something else.

fsharp-backend/tests/Tests/Serialization.Tests.fs
StachuDotNet: I was previously seeing multiple `backend/serialization` files changing in a confusing way - not recalling the exact details.

It now seems no longer valid - adding the case went fine and didn't cause a large change.

StachuDotNet: Nevermind - un-commented this and tests failing in CI. Not sure yet why I wasn't seeing files change locally.

fsharp-backend/src/LibExecution/ProgramTypesParser.fs
```diff
@@ -239,7 +241,8 @@ module Handler =
 
     let toModule (s : PT.Handler.Spec) =
       match s with
-      | PT.Handler.HTTP _ -> "HTTP"
+      | PT.Handler.HTTPLegacy _ -> "HTTP"
+      | PT.Handler.HTTPBytes _ -> "HTTPBYTES" // HttpBytesTODO: consider changing to `"HTTP_BYTES"`
```
StachuDotNet, line 245: do you have any opinion here?

fsharp-backend/tests/Tests/Serialization.Tests.fs
StachuDotNet: Figured out what I was previously confused by:

We output files in the form of `backend/serialization/ocaml-ApiServer-InitialLoad-V0-T-initial.json`
A new `V_` file was added (V7). I was expecting the 'x' in `Vx` to correspond with the placement [here](https://github.com/darklang/dark/blob/4a915908123c47f19bcf68ef96db3aa01c94dfcb/fsharp-backend/tests/Tests/Serialization.Tests.fs#L295-L301), but that was wrong - they're ordered alphabetically, and I was confused by the output. OK now.

StachuDotNet:
> Before merging this would need:
> 
> * the lists of the goals for the project
Goals have been outlined [here](https://github.com/darklang/dark/discussions/4325) and we can discuss there or over Zoom.

> * a discussion of the serialization changes (and possible backward compatibility)
To make this changeset thinner, I'm tempted to revert the `HttpLegacy` rename, and leave that as a `HttpBytesTODO`, to consider later. Currently there are many files that change due to this rename, and primary use was to make sure we discover all the places we should handle the HttpBytes handlers. Besides, since it's experimental, it's not quite "legacy" at this point. WDYT?

(i.e. marking the old Http handlers as 'legacy' in some way would be most appropriate at the end of the work, just before availing the new handlers to users)

> * a binary serialization test case
Added now

StachuDotNet:
I'm going a head with a revert of `HTTPLegacy` back to `HTTP`. It adds noise and unneeded risk. Instead, for now, I'll add an `HttpBytesTODO` in those places, to review/consider later.

StachuDotNet:
Putting this back to draft status. Let's break this up in a bit.

StachuDotNet:
Closing this - this was intended as an experiment to learn what would be needed.
Some efforts here have been extracted to separate PRs, and conversation about this effort is continued here https://docs.google.com/document/d/1vO0O8F24YGcF5DVixHdmZM-c66L_2vhrzJAOTgr2quY/edit# and here https://github.com/darklang/dark/discussions/4325

StachuDotNet:
Also, the serialization changes should be less noisy now, since #4331

PR #4341, opened 2022-07-27
Consolidate HttpMiddlewareV0

StachuDotNet:
This consolidation, especially the `[<RequireQualifiedAccess>]`, makes it much
more obvious where the middleware is used within BwdServer, and prepares
BwdServer for further middlewares. This is in preparation for an upcoming
middleware that will support non-string request bodies.


StachuDotNet:
One thing that this highlights is that the V0 middleware is used for 404s - this may not be ideal when more than 1 middleware is present.

pbiggar (approved):

PR #4351, opened 2022-08-01
Implement a "List Pattern"

StachuDotNet:
WIP - will revisit once other `client` changes are merged in.


StachuDotNet:
Enough code has changed in this space, and the PR is shallow enough, that for now I'll simply abandon this branch.

Generally, these changes were followed by: look for every instance of 'Constructor' (with a capital 'c') through the codebase, and create/handle a List equivalent (there's a Constructor pattern that exists, in addition to the List pattern we want).

PR #4353, opened 2022-08-02
Create and test new 'HttpBytes' handler type

StachuDotNet:
This is a work in progress aimed towards the goals outlined in [this doc](https://docs.google.com/document/d/1vO0O8F24YGcF5DVixHdmZM-c66L_2vhrzJAOTgr2quY/edit#).

A previous attempt towards this was #4323, which has since been split up into several PRs already merged, along with this one.

- adds the HttpBytes handler spec to SerializedTypes, ProgramTypes, and RuntimeTypes
- updates a few things in `LibBackend` to correspond with the new HttpBytes variants:
  - Canvas-loading
  - Trace-loading (esp. sample data)
  - loading of "http" handlers now includes handlers with a module of `HTTP_BYTES` (for BwdServer's sake)
- adds a thin HttpMiddlewareV1 corresponding to HttpBytes
- adjusts BwdServer, run requests against the V1 middleware for requests that match `HTTPBytes` handlers
- adjusts existing BwdServer testing to be explicit about being for "Http" handlers and HttpMiddlwareV0 middleware
- adds another set of BwdServer tests around HttpBytes-specific testing
  - this was largely a copy of the existing test file/structure
  - so far this includes only 3 tests, one of which includes a non-text (image) file upload

TODOs:
- [x] consider renaming `HttpMiddlewareV0` and `V1` to better match `Http` and `HttpBytes`. (maybe `DarkMiddwareHttp`)?
- [x] add more tests
- [ ] assuming OCamlTypes work is merged in before this PR, remove adjustments/references to such (just one at this point)
- [ ] in the new BwdServer tests, consider 'injecting' the image at run-time rather than at parse-time (I did it this way because it was easier)
- [x] add tests around _returning_ bytes too (and allow 'injections' within the responses of .test files)


StachuDotNet:
I'm not sure why I was reaching for an integration test; pivoting to a BwdServer test

StachuDotNet:
@pbiggar I wouldn't consider this 'complete' yet, but certainly worth a review.
I don't know how I feel about [INJECT_DATA], so especially curious to hear what you'd recommend there.

fsharp-backend/tests/Tests/BwdServer.HttpBytes.Tests.fs
```diff
@@ -91,6 +93,14 @@ let parseTest (bytes : byte array) : Test =
       | Regex "\[http-handler (\S+) (\S+)\]" [ method; route ] ->
         (InHttpHandler,
          { result with handlers = result.handlers @ [ (method, route, "") ] })
+      | Regex "\[INJECT_DATA=(\S+)\]" [ dataFileToInject ] ->
+        // TODO Consider doing this async...
```
pbiggar, line 97: not async. better to do it during the test but fine to leave this as a todo

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
```diff
@@ -0,0 +1,134 @@
...
+  module HttpQueryEncoding = BackendOnlyStdLib.HttpQueryEncoding
+
+  let private parseQueryString (query : string) : RT.Dval =
+    // Drop leading "?"
+    let query = if query.Length > 0 then String.dropLeft 1 query else query
+    HttpQueryEncoding.ofFormEncoding query
+
+  let private parseHeaders (headers : (string * string) list) =
+    headers
+    |> List.map (fun (k, v) -> (String.toLowercase k, RT.DStr v))
```
pbiggar, line 21: Should headers be a string here (or bytes)? Should we lowercase the value too as well as the key?

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
```diff
@@ -0,0 +1,134 @@
...
+    |> Map
+    |> RT.Dval.DObj
+
+  let private cookies (headers : HttpHeaders.T) : RT.Dval =
```
pbiggar, line 25: unsure why there's cookie parsing here

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
```diff
@@ -0,0 +1,134 @@
...
+        | [] -> ("", RT.DNull) // skip empty rows
+        | [ _ ] -> ("", RT.DNull) // skip rows with only a key
+        | k :: v :: _ -> (decode k, RT.DStr(decode v)))
+      |> RT.Dval.obj
+
+    HttpHeaders.get "cookie" headers
+    |> Option.map parse
+    |> Option.defaultValue (RT.DObj Map.empty)
+
+  let private url (headers : List<string * string>) (uri : string) =
```
pbiggar, line 43: Is this necessary in a minimal framework?

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
```diff
@@ -0,0 +1,134 @@
...
+    RT.DStr(string parsed.Uri)
+
+  let fromRequest
+    (uri : string)
+    (headers : List<string * string>)
+    (query : string)
+    (body : byte array)
+    : RT.Dval =
+    [ "body", RT.DBytes body
+      "queryParams", parseQueryString query
```
pbiggar, line 63: Perhaps the query should be a string instead of parsed?

fsharp-backend/src/BwdServer/Server.fs
```diff
@@ -386,6 +386,50 @@ let runDarkHandler (ctx : HttpContext) : Task<HttpContext> =
...
+            RealExe.executeHandler
+              canvas.meta
+              (PT2RT.Handler.toRT handler)
+              (Canvas.toProgram canvas)
+              traceID
+              inputVars
+              (RealExe.InitialExecution(desc, request))
+
+          let result = HttpMiddlewareV1.Response.toHttpResponse result
+          let result = HttpMiddlewareV1.Cors.addCorsHeaders reqHeaders result
```
pbiggar, line 418: Why is cors stuff still needed?

fsharp-backend/src/LibBackend/Traces.fs
```diff
@@ -16,27 +16,42 @@ module PTParser = LibExecution.ProgramTypesParser
...
 let sampleModuleInputVars (h : PT.Handler.T) : AT.InputVars =
   match h.spec with
-  | PT.Handler.HTTP _ -> sampleRequestInputVars
+  | PT.Handler.HTTP _ -> sampleHttpRequestInputVars
+  | PT.Handler.HTTPBytes _ -> sampleHttpBytesRequestInputVars
   | PT.Handler.Cron _ -> []
   | PT.Handler.REPL _ -> []
-  | PT.Handler.UnknownHandler _ -> sampleRequestInputVars @ sampleEventInputVars
+  | PT.Handler.UnknownHandler _ ->
+    // HttpBytesTODO should this include sampleHttpBytesRequestInputVars?
```
pbiggar, line 53: Hmmm, probably not for now at least.

fsharp-backend/src/LibBackend/Traces.fs
```diff
@@ -61,18 +76,17 @@ let savedInputVars
...
+    let boundRouteVariables =
       if route <> "" then
         // Check the trace actually matches the route, if not the client
         // has made a mistake in matching the traceid to this handler, but
         // that might happen due to a race condition. If it does, carry
         // on, if it doesn't -- just don't do any bindings and inject the
         // sample variables. Communicating to the frontend that this
         // trace doesn't match the handler should be done in the future
-        // somehow.
+        // somehow. TODO
```
pbiggar, line 89: I would guess we're still going to do path parsing here - hard to avoid.

fsharp-backend/tests/Tests/BwdServer.HttpBytes.Tests.fs
```diff
@@ -0,0 +1,361 @@
+/// Tests the builtwithdark.com server (`BwdServer`),
```
pbiggar, line 1: Can having a separate test server be avoided? I don't mind the logic is a separate file or module if it's different. But we're going to have to maintain both of these long term (and more in the future)

fsharp-backend/tests/httpbytestestfiles/injected-icon-post.test
```diff
@@ -0,0 +1,20 @@
+[http-handler POST /]
+request.body |> Bytes.length_v0 |> toString |> String.toBytes_v0
+
+[request]
+POST / HTTP/1.1
+Host: HOST
+Date: Sun, 08 Nov 2020 15:38:01 GMT
+Content-Length: 685
+
+[INJECT_DATA=favicon-32x32.png]
```
pbiggar, line 10: maybe `IMPORT_DATA_FROM_FILE`?

I'm a little confused though - is this a section (implied by the square brackets), or a variable being replaced in the body.

fsharp-backend/tests/httpbytestestfiles/injected-icon-post.test
```diff
@@ -0,0 +1,20 @@
...
+
+[response]
+HTTP/1.1 200 OK
+Date: xxx, xx xxx xxxx xx:xx:xx xxx
+x-darklang-execution-id: 0123456789
+Server: darklang
+Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
+Content-Length: 3
+
+685
```
pbiggar, line 20: Let's definitely have a test that contains byte output.

pbiggar (changes requested):
Looks great!

I have comments in line; the general just being can we make it even more minimal, and can we have just one BWD server test harness.

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
StachuDotNet: Ah, I didn't think of that somehow - will do!

fsharp-backend/src/BwdServer/Server.fs
StachuDotNet: I thought there might be some localhost-testing reason to include it, but that was fuzzy. I'll remove it and see if anything breaks, thanks!

fsharp-backend/tests/httpbytestestfiles/injected-icon-post.test
StachuDotNet: It's a variable, but since I was processing it in the parsing of the test, I made it look the same as elsewhere.
I think I'll make it more like `<IMPORT_DATA_FROM_FILE=xyz.abc>`

fsharp-backend/tests/Tests/BwdServer.HttpBytes.Tests.fs
StachuDotNet: (there's still tidying up to do, but) done!

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
StachuDotNet: Removed; FWIW [this](https://github.com/darklang/dark/blob/6ec452d4fc900198e761ca2fdee9404cff41ec9b/fsharp-backend/tests/httpbytestestfiles/url-custom-scheme.test) is one outcome of this logic being removed. Does that seem OK to you?

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
StachuDotNet: I'm not sure why they would be bytes. As far as I know, and have just googled, headers (both keys and values) may only be strings.

I'm not sure why we'd lowercase either key _or_ value - I'd lean towards removing that existing `toLowercase`. Why should we do such for either?

StachuDotNet:
I think this is at another good point of review.

Note: most of the `BwdServer.Tests.fs` changes were a refactoring because I found the current code to be difficult to understand.
This feels much better to me:
image

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
pbiggar: The reason i was asking are the strings and Dark are Unicode, but I didn't expect that Unicode would be supported and HTTP headers. It looks like only printable ASCII characters are allowed, so I guess using strings is the best way to support this.

The reason to use lowercase letters is because http2 and to http3 both use lowercase headers. Since the user is likely to do a string comparison, or possibly even a map look up, they can get different answers depending on whether the request was made with http1 vs http2/3. I observed this with the old http framework, and I'd like to avoid it with the new one.

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
pbiggar: I'm not sure what I'm looking at here, can you be more specific?

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
```diff
@@ -0,0 +1,134 @@
...
+    // Expected user responses
+    | RT.DHttpResponse (RT.Redirect str) ->
+      Telemetry.addTags [ "response-type", "httpResponse redirect" ]
+      { statusCode = int 302; headers = [ "Location", str ]; body = [||] }
+
+    | RT.DHttpResponse (RT.Response (code, headers, RT.DBytes body)) ->
+      Telemetry.addTags [ "response-type", "httpResponse response" ]
+      { statusCode = int code; headers = headers; body = body }
+
+    | RT.DBytes body ->
```
pbiggar, line 57: I feel like we should only allow an RT.Response here. Wdyt?

pbiggar (changes requested):
It's looking good!

Looks like there's a few TODOs and cleanups remaining.

I think we've still got headers casing and return body/shape to consider. Is that all that's left on this iteration?

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
StachuDotNet: > The reason to use lowercase letters

Ah, that's great context, thanks!

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
StachuDotNet: In that test, the `X-forwarded-proto: ftp` is ignored.
I think that's pretty reasonable/expected, so probably not a useful question from me.

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
StachuDotNet: Should the _response_ headers also be to-lowercase'd, then?

fsharp-backend/tests/httpbytestestfiles/simple-response-headers.test
```diff
@@ -0,0 +1,25 @@
...
+ignored
+
+[response]
+HTTP/1.1 200 OK
+Date: xxx, xx xxx xxxx xx:xx:xx xxx
+Header-3: Value-3
+Header2: Value2
+x-darklang-execution-id: 0123456789
+Server: darklang
+header1: value1
```
StachuDotNet, line 21: TODO: figure out why these headers are sorted the way they are - very odd.

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
```diff
@@ -0,0 +1,64 @@
...
+  let toHttpResponse (result : RT.Dval) : HttpResponse =
+    match result with
+    // Expected user responses
+    | RT.DHttpResponse (RT.Redirect str) ->
+      Telemetry.addTags [ "response-type", "httpResponse redirect" ]
+      { statusCode = int 302; headers = [ "Location", str ]; body = [||] }
+
+    | RT.DHttpResponse (RT.Response (code, headers, RT.DBytes body)) ->
+      Telemetry.addTags [ "response-type", "httpResponse response" ]
+      { statusCode = int code; headers = headers; body = body }
```
StachuDotNet, line 55: TODO: consider tolowercase-ing the response headers here

fsharp-backend/tests/httpbytestestfiles/simple-request-headers.test
```diff
@@ -0,0 +1,32 @@
...
+Server: darklang
+Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
+Content-Length: 114
+
+{
+  header2: "value2",
+  header1: "value1",
+  header-with-casing-conflict: "test1,test2",
+  header-3: "value-3",
+}
```
StachuDotNet, line 32: Frustratingly, this test fails. Visually, the expected and actual response bodies match, but there are subtle byte-level things apparently preventing this from failing
image

I'm not sure how to deal with this; advice/assist would be appreciated.

fsharp-backend/tests/httpbytestestfiles/simple-request-headers.test
```diff
@@ -0,0 +1,32 @@
...
+Date: xxx, xx xxx xxxx xx:xx:xx xxx
+x-darklang-execution-id: 0123456789
+Server: darklang
+Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
+Content-Length: 114
+
+{
+  header2: "value2",
+  header1: "value1",
+  header-with-casing-conflict: "test1,test2",
```
StachuDotNet, line 30: Would this be your expectation?

StachuDotNet: I've no opinion.

StachuDotNet:
> Looks like there's a few TODOs and cleanups remaining.
> 
> I think we've still got headers casing and return body/shape to consider. Is that all that's left on this iteration?

Some of those TODOs/cleanups are things that are better delayed until some editor-side work is being done, I think.

Yeah, I think the PR is good enough as-is, after fixing the broken test and resolving the outstanding questions in these threads).

Regarding "return body/shape to consider" I think you were referring to the request headers instead being a list of str*str tuples; just pushed a commit out regarding that.

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
pbiggar: I had thought to remove RT.redirect here as well. But I guess since it has no body, it's fine. This is making me think that RT.redirect shouldn't exist, and should just be encoded as an RT.response.

(nothing for this PR, just thinking out loud)

fsharp-backend/tests/httpbytestestfiles/simple-request-headers.test
pbiggar: This looks good. Worth checking Date, Content-length, Host, etc?

pbiggar:
> Regarding "return body/shape to consider" I think you were referring to the request headers instead being a list of str*str tuples; just pushed a commit out regarding that.

This was referring to allowing a TBytes response, so this is taken care of.

fsharp-backend/tests/httpbytestestfiles/simple-request-headers.test
pbiggar: I'm really not sure. Maybe debug in real time?

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
StachuDotNet: > making me think that RT.redirect shouldn't exist, and should just be encoded as an RT.response.

Agreed.

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
StachuDotNet: While the keys are case-insensitive (per the spec), the values are not - so let's lowercase both request header keys on the way in, and response header keys on the way out.

fsharp-backend/tests/httpbytestestfiles/simple-request-headers.test
StachuDotNet: https://github.com/darklang/dark/pull/4353/commits/39fceabd7b55e9b161a13928289a67d4df5c52e0#diff-41aa10236dd071016d57d4da800039add6db30ca7f1450964f22da98e80b1cadR27

VS Code was trimming a space character annoyingly. This isn't the most elegant solution, but it works.

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
```diff
@@ -4,22 +4,26 @@
...
-  open LibExecution.VendoredTablecloth
+open Prelude
+open LibExecution.VendoredTablecloth
+
+module RT = LibExecution.RuntimeTypes
+module Telemetry = LibService.Telemetry
 
-  module RT = LibExecution.RuntimeTypes
+let lowercaseHeaderKeys (headers : HttpHeaders.T) =
+  headers |> List.map (fun (k, v) -> (String.toLowercase k, v))
```
pbiggar, line 14: Is it your understanding that header values can be mixed case?

fsharp-backend/tests/httptestfiles/README.md
```diff
@@ -117,6 +124,10 @@ One of the automatic changes made by text editors is the trimming of whitespace
 at the end of lines of code. To handle this, any usages of `<SPACE>` within the
 request or response are replaced with a space character (` `).
 
+Additionally, any usages of `<IMPORT_DATA_FROM_FILE=name>` will result in the
+test runner replacing this string with the contents of the referenced binary
+file - a file with the corresponding name should exist in `httpbytestestfiles/data`.
```
pbiggar, line 129: Should probably change that directory, given that it can be used by non-bytes test files.

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
```diff
@@ -0,0 +1,60 @@
...
+    (body : byte array)
+    : RT.Dval =
+    let headers =
+      headers
+      |> lowercaseHeaderKeys
+      |> List.map (fun (k, v) -> RT.DTuple(RT.DStr(k), RT.DStr(v), []))
+      |> RT.DList
+
+    [ "body", RT.DBytes body
+      "queryParams", RT.DStr query
```
pbiggar, line 30: Should we remove queryParams, like we did in HTTPClientBytes?

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
```diff
@@ -0,0 +1,60 @@
+/// Thin, bytes-friendly middleware
```
pbiggar, line 1: Is V1 a good name for this? It doesn't exactly replace V0

fsharp-backend/src/LibBinarySerialization/SerializedTypes.fs
```diff
@@ -196,6 +197,9 @@ module Handler =
     // If there's no module
     // CLEANUP: convert these into repl and get rid of this case
     | UnknownHandler of string * string * ids
+    /// Corresponds to HttpMiddlewareV1
+    /// HttpBytesTODO: rebrand to HTTPBase rather than HTTPBytes?
```
pbiggar, line 201: We should probably resolve this question before merging, as this "brand" will be saved in the database.

fsharp-backend/tests/httpbytestestfiles/simple-request-headers.test
```diff
@@ -0,0 +1,41 @@
...
+  ), (
+    "content-length", "7"
+  ), (
+    "header1", "value1"
+  ), (
+    "header2", "Value2"
+  ), (
+    "header-3", "Value-3"
+  ), (
+    "header-with-casing-conflict", "test1,test2"
```
pbiggar, line 39: I don't understand why this isn't two headers. Kind of the point of using a list of tuples here, is to allow for multiple headers of the same name.

fsharp-backend/tests/httpbytestestfiles/url-custom-scheme.test
```diff
@@ -0,0 +1,24 @@
...
+
+
+[custom-domain customdomain.myspecialdomain.com]
+
+[request]
+GET /a HTTP/1.1
+Host: customdomain.myspecialdomain.com
+Date: Sun, 08 Nov 2020 15:38:01 GMT
+Content-Length: 0
+X-forwarded-proto: ftp
```
pbiggar, line 12: What's being tested here? Maybe add a comment to say what's going on.

fsharp-backend/tests/httptestfiles/README.md
pbiggar: You're probably doing that in the followup commit I guess.

pbiggar (changes requested):
Looking really good. I have a few small questions, we should be able to get the resolved quickly.

fsharp-backend/tests/httpbytestestfiles/simple-request-headers.test
StachuDotNet: it was using a `getHeaders` in Server.fs fn that concatonated the values; fixed

fsharp-backend/src/HttpMiddleware/HttpBasic.fs
```diff
@@ -0,0 +1,61 @@
...
+  headers |> List.map (fun (k, v) -> (String.toLowercase k, v))
+
+module Request =
+  let fromRequest
+    (uri : string)
+    (headers : HttpHeaders.T)
+    (query : string)
+    (body : byte array)
+    : RT.Dval =
+    printfn "Headers %A" headers
```
StachuDotNet, line 23: whoops; fixing this shortly

fsharp-backend/src/LibBackend/Serialize.fs
```diff
@@ -158,7 +158,7 @@ let fetchReleventTLIDsForHTTP
     "SELECT tlid
      FROM toplevel_oplists
      WHERE canvas_id = @canvasID
-       AND ((module = 'HTTP'
+       AND (((module = 'HTTP' OR module = 'HTTP_BYTES')
```
pbiggar, line 161: This name should change right?

fsharp-backend/src/LibExecution/ProgramTypesParser.fs
```diff
@@ -240,6 +242,7 @@ module Handler =
     let toModule (s : PT.Handler.Spec) =
       match s with
       | PT.Handler.HTTP _ -> "HTTP"
+      | PT.Handler.HTTPBasic _ -> "HTTP_BYTES"
```
pbiggar, line 245: And here?

fsharp-backend/src/LibBackend/Serialize.fs
StachuDotNet: good catch!

fsharp-backend/src/HttpMiddleware/HttpMiddlewareV1.fs
StachuDotNet: done

pbiggar:
Looks ready to go to me.

PR #4366, opened 2022-08-04
Support non-text request/response bodies in HttpClient

StachuDotNet:
The current HttpClient:
- currently only works well with textual request+response bodies
- does a LOT out of the box, removing control from the user (de- and serialization, content-encoding, redirect behavior, etc.)

This PR is aimed at creating new HttpClient fns that do less, and work well with bytes.


fsharp-backend/src/BackendOnlyStdLib/LibHttpBytesClient.fs
```diff
@@ -0,0 +1,84 @@
+/// v5 StdLib functions in the HttpClient module
```
StachuDotNet, line 1: wrong

fsharp-backend/src/BackendOnlyStdLib/LibHttpBytesClient.fs
```diff
@@ -0,0 +1,84 @@
...
+    |> List.map (fun pair ->
+      match pair with
+      | DTuple (DStr k, DStr v, []) -> Ok(k, v)
+      | wrong ->
+        Error
+          $"Expected a (string * string), but got: {DvalReprDeveloper.toRepr wrong}")
+    |> Tablecloth.Result.values
+  | _ -> Error $"Expected a tuple, but got: {DvalReprDeveloper.toRepr dv}"
+
+// todo: move this to the consuming file probably
```
StachuDotNet, line 31: done

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
```diff
@@ -0,0 +1,199 @@
...
+          else
+            // Dark headers can only be added once, as they use a Dict. Remove them
+            // so they don't get added twice (eg via Authorization headers above)
+            req.Headers.Remove(k) |> ignore<bool>
+            let added = req.Headers.TryAddWithoutValidation(k, v)
+            // Headers are split between req.Headers and req.Content.Headers so just try both
+            if not added then
+              req.Content.Headers.Remove(k) |> ignore<bool>
+              // CLEANUP: always use lowercase headers
+              req.Content.Headers.Add(k, v))
```
StachuDotNet, line 127: Note to self: I ran into an issue here when trying to pass in a request header with an empty key. That's likely why they were previously input as Objs? Though that protection isn't obvious here, and would be good to be clearer.

fsharp-backend/src/BackendOnlyStdLib/LibHttpBytesClient.fs
```diff
@@ -0,0 +1,84 @@
...
+/// Used in both Ok and Error cases
+let headersType = TList(TTuple(TStr, TStr, []))
+let queryStryingType = TList(TTuple(TStr, TStr, []))
+
+let responseType =
+  TRecord [ "body", TBytes; "headers", headersType; "code", TInt; "error", TStr ]
+
+let returnTypeOk = responseType
+let returnTypeErr = responseType
+let returnType = TResult(returnTypeOk, returnTypeErr)
```
StachuDotNet, line 52: "error" probably doesn't fit in `returnTypeOk`, right? I forget why it's included for existing httpclient OK responses.

fsharp-backend/src/BackendOnlyStdLib/LibHttpBytesClient.fs
```diff
@@ -0,0 +1,84 @@
...
+    Param.make "headers" headersType "" ]
+
+let fns : List<BuiltInFn> =
+  [ { name = fn "HttpBytesClient" "post" 0
+      parameters = parameters
+      returnType = returnType
+      description =
+        "Make blocking HTTP POST call to `uri`. Returns a `Result` object where
+        the response object is wrapped in `Ok` if a response was successfully
+        received and parsed, and is wrapped in `Error` otherwise"
```
StachuDotNet, line 67: > successfully received and parsed

need some better wording here

StachuDotNet:
I've done a first pass in creating a bytes-friendly Http Client, and this is at a point where a review would be appreciated.

Here is a relatively exhaustive list of things that are different from the current HttpClient, and questions:

---

The type to represent request+response headers has been changed from `TDict TStr` to `TList(TTuple(TStr, TStr, []))`. If tuples aren't out by the time we release this, we can update to another type, but does this seem reasonable long-term? I think it's reasonable to error on empty keys, and have that not represented in the type explicitly as `TDict` did.

Same question as above, but for request query params. (this hasn't yet been changed - would like feedback before putting in the effort required)

The OK return type has been updated from {body, headers, raw, code, error} to {body, headers, code}. ('error' is no longer included) Does that seem reasonable?

The Error return type has been updated from `{body, headers, raw, code, error}` to `TStr`. A 4xx/5xx should result in an OK response. My initial hesitation is for errors that happen after the response has come back. Such should be extremely rare, but at that point we _have_ the response body/headers/code. I suppose the best case scenario would be a DU representing a few different types of errors, but we're not currently prepared for that. What do you think - should this type be reverted?

Should we call incorrectArgs() in the shallowest layer possible, teasing out the Dvals at the fn def level, leaving .NET types underneath? E.g. HttpBytesClient has no need to know about Dval stuff. 

The new "get" function now allows a request body, where it didn't before. Is that OK - should we allow request bodies for GET requests? If so, should we discourage it it some way?

Should these new functions exist in the HttpClient module as v6 fns, or some new module as v0 fns? Until we have some good experience with these new functions, including helpers available to 'port' existing code to use the new client, we'll likely need to keep both 'sets' of fns around. Maybe we could do `HttpClient::getBytes` etc.?

The new client does _nothing_ related to redirects. We explictly set `AllowAutoRedirect` to `false` in the .NET HttpClient and from there leave it to the consumer to handle redirect responses. Later, we may provide a fn wrapping around such, to auto-follow redirects?

To get some experience with the new client, maybe we hide the new fns under the DarkInternal module for a while? That is, if we end up merging this.

The new client explicitly uses Http 3.0, versus Http 2.0 in the current fns.

The current client has [this](https://github.com/darklang/dark/blob/fcc4ab7026a4c7818e2b5c60b299cdc88ebf50cb/fsharp-backend/src/BackendOnlyStdLib/HttpClient.fs#L149-L162) bit of code which seems to handle basic auth - I've stripped that out. Is that OK? I don't fully understand where those basic auth params came from - user passing them in as headers, maybe, and .NET interprets them automatically? Confusing.

The current client attempts to interpret the request body for content type, then serializing and setting content-length header appropriate. The new client does none of that. Same thing for the response body - we now just take the bytes and return them, otherwise ignoring the body.

The current client includes [a few](https://github.com/darklang/dark/blob/fcc4ab7026a4c7818e2b5c60b299cdc88ebf50cb/fsharp-backend/src/BackendOnlyStdLib/HttpClient.fs#L193-L195) `defaultHeaders` in responses. Should we do that in the new client? For now, I've included them.

The current client included some HTTP status header. That seemed like OCaml-replication-specific logic, so I removed it from the new client.

There are a few places where we toLowercase headers' keys and values. Is that appropriate? That sort of manipulation feels like it should be in an accessory function, not within this client.

I went back and forth on in-lining HttpBytesClient.fs into LibHttpBytesClient.fs. Open to opinions, but the outcome here may depend on what 'module' the new client fns belong in long-term.

---

Known TODOs
- [ ] handle reponses to these questions
- [ ] define the additional equivalent PUT/OPTIONS/etc. fns.
- [ ] experiment with performance impacts:
  - [ ] request round-tripping
  - [ ] traces, analysis
    - [ ] do we need to set some limits?
- [ ] keep testing the new client; see what feels good/bad; adjust
- [ ] improve the fn descriptions
- [ ] depending on if we release this to DarkInternal for now, or normally to the full userbase, consider providing some helper functions that capture logic ripped out of the current client:
  - [ ] redirects
  - [ ] ...

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
```diff
@@ -0,0 +1,212 @@
...
+// Note that the number of sockets was verified manually, with:
+// `sudo netstat -apn | grep _WAIT`
+// TODO: I don't see where "the nunber of sockets" is actually configured?
+let private socketHandler : HttpMessageHandler =
+  new SocketsHttpHandler(
+    // Avoid DNS problems
+    PooledConnectionIdleTimeout = System.TimeSpan.FromMinutes 5.0,
+    PooledConnectionLifetime = System.TimeSpan.FromMinutes 10.0,
+
+    AutomaticDecompression = System.Net.DecompressionMethods.None,
```
pbiggar, line 41: Should we do decompression? It was removed from the old version because we did it manually later, in order to keep the behavior identical. But we can't reasonably expect users to do their own decompression, especially since we barely provide the ability to do so. But maybe that's the goal?

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
```diff
@@ -0,0 +1,212 @@
...
+  (url : string)
+  (queryParams : (string * string list) list)
+  (method : HttpMethod)
+  (reqHeaders : HttpHeaders.T)
+  (reqBody : byte array)
+  : Task<Result<HttpResult, ClientError>> =
+  task {
+    use _ =
+      Telemetry.child
+        "HttpClient.call"
```
pbiggar, line 75: We should be able to tell that this event is this HTTPClient

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
```diff
@@ -0,0 +1,212 @@
...
+            Port = uri.Port,
+            Path = uri.AbsolutePath,
+            Query = queryString
+          )
+
+        use req =
+          new HttpRequestMessage(
+            method,
+            string reqUri,
+            Version = System.Net.HttpVersion.Version30,
```
pbiggar, line 103: Does this say "use exactly version three" or does it allow any HTTP version?

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
```diff
@@ -0,0 +1,212 @@
...
+        // headers
+        // TODO is it still appropriate to include these default headers?
+        let defaultHeaders =
+          Map [ "Accept", "*/*"; "Accept-Encoding", "deflate, gzip, br" ]
+
+        Map reqHeaders
+        |> Map.mergeFavoringRight defaultHeaders
+        |> Map.iter (fun k v ->
+          // TODO do we need this branch?
+          // I think so - .NET is odd with content types, to my memory
```
pbiggar, line 116: Yes, this is to work around how the .net HTTP client works.

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
```diff
@@ -0,0 +1,212 @@
...
+                Headers.MediaTypeHeaderValue.Parse(v)
+            with
+            | :? System.FormatException ->
+              Exception.raiseCode "Invalid content-type header"
+          else
+            // Dark headers can only be added once, as they use a Dict.
+            // Remove them so they don't get added twice.
+            // TODO: re-evaluate if we need this. See logic in HttpClient.
+            // Also, consider that we could (do, in this draft) return headers
+            // as a list of tuples, so the Dict claim is false currently.
```
pbiggar, line 129: I agree, this shouldn't use a dict, and we should allow headers to be set multiple times

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
```diff
@@ -0,0 +1,212 @@
...
+        use! responseStream = response.Content.ReadAsStreamAsync()
+        use memoryStream = new MemoryStream()
+        do! responseStream.CopyToAsync(memoryStream)
+        let respBody = memoryStream.ToArray()
+
+        let headers =
+          response
+          |> HttpHeaders.headersForAspNetResponse
+          // TODO Should we really be manipulating (lowercasing) these?
+          |> List.map (fun (k : string, v) -> (String.toLowercase k, v))
```
pbiggar, line 154: Should we also lowercase the values?

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
```diff
@@ -0,0 +1,212 @@
...
+  (reqHeaders : List<string * string>)
+  : Ply<Dval> =
+  uply {
+    match! httpCall uri query verb reqHeaders reqBody with
+    | Ok response ->
+      let parsedResponseHeaders =
+        response.headers
+
+        // TODO should we really be trimming these?
+        |> List.map (fun (k, v) -> (String.trim k, String.trim v))
```
pbiggar, line 196: Unsure, but probably not?

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
```diff
@@ -0,0 +1,212 @@
...
+
+        // TODO should we really filter out headers of blank keys?
+        |> List.filter (fun (k, _) -> String.length k > 0)
```
pbiggar, line 199: No, I don't think so

fsharp-backend/src/BackendOnlyStdLib/LibHttpBytesClient.fs
```diff
@@ -0,0 +1,82 @@
...
+        Error
+          $"Expected a (string * string), but got: {DvalReprDeveloper.toRepr other}")
+    |> Tablecloth.Result.values
+  | _ -> Error $"Expected a list of tuples, but got: {DvalReprDeveloper.toRepr dv}"
+
+let call (method : HttpMethod) =
+  (function
+  | _, [ DStr uri; DBytes body; query; headers ] ->
+    let headers = toStringPairs headers
+    let query = HttpQueryEncoding.toQuery query
```
pbiggar, line 37: Perhaps the query should be a string?

pbiggar (commented):
Took a look through the code. Will answer your question in a separate comment.

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
StachuDotNet: I don't understand why we should lowercase _either_, in request or response. Could you please explain? Why is it the library fn's job to adjust the casing?

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
StachuDotNet: I'm happy either way, but don't have the experience to really know which way is appropriate. So, I"ll default to your judgment here, though intuitively I'm learning _slightly_ towards automatically decompressing.

pbiggar:
> The type to represent request+response headers has been changed from `TDict TStr` to `TList(TTuple(TStr, TStr, []))`. If tuples aren't out by the time we release this, we can update to another type, but does this seem reasonable long-term? I think it's reasonable to error on empty keys, and have that not represented in the type explicitly as `TDict` did.

I think tuples are the right way to do this. I would use this as a forcing function to get tuples out, rather than compromise this to use non-tuples.

> Same question as above, but for request query params. (this hasn't yet been changed - would like feedback before putting in the effort required)

Reviewing both this PR and the HTTP bytes framework PR, it occurred to me that perhaps query params should just be a string. That would remove the logic from the client, and allow the user to control how quickly params are created/parsed.


> The `Ok`/success return type has been updated from `{body, headers, raw, code, error}` to `{body, headers, code}`. (`error` is no longer included) Does that seem reasonable?

I think you're saying that Error is no longer in the success/OK type. That's seems right.

> The `Error` return type has been updated from `{body, headers, raw, code, error}` to `TStr`. A 4xx/5xx should result in a successful response in the new client, where it does not in the current client. I have some hesitation in this type change, as some errors may happen after the response has come back. Such should be extremely rare, but at that point we _have_ the response body+headers+code, so it feels appropriate to include in the response. I suppose the best case scenario would be a DU representing a few different types of errors, but we're not currently prepared for that. What do you think - should the response type be reverted from a simple string into something like `{body, headers, code, errorMessage}`?

I agree that error status codes should use the OK result.

It might be good to have DUs, but I think that might be biting off for too much for this PR. Or, perhaps we'd like this to be the driving function for large changes, but I think we're months away at least from delivering DUs, so something to think about. Have a look at Elms http response type, it's very well thought out and could provide a good model for us.


> The new "get" function now allows a request body, where it didn't before. Is that fine - should we allow request bodies for GET requests? If so, should we discourage it it some way?

Thinking out loud, perhaps there should just be a `request` function, with the method provided by the User as a parameter parameter.

I think it's reasonable for GET to have a body, we removed it for ergonomics in the other client but that's not the point of this client. A user could reasonably want to make a HTTP request with a body, presumably to a server which supports it, even though they're not supposed to. But that's what this client is for.
 
> Should these new functions exist in the `HttpClient` module as v6 fns, or some new module as v0 fns? Until we have some good experience with these new functions, including helpers available to 'port' existing code to use the new client, we'll likely need to keep both 'sets' of fns around. Maybe we could do `HttpClient::getBytes` etc.?

If they were in the HTTPclient module, they would be v0 as they are new functions. But I think it makes more sense to have a new module, maybe "HTTPBaseClient".


> The new client does _nothing_ related to redirects. We explicitly set `AllowAutoRedirect` to `false` in the .NET HttpClient and from there leave it to the consumer to handle redirect responses. Later, we may provide a fn wrapping around such, to auto-follow redirects?

Another option, thinking out loud, would be to expose the things that are built into the .net HTTP client as options, such as compression, redirect behavior, (anything else)? Presumably, these would all be things that aren't good choices to provide as middlewares, for whatever reasons.

But otherwise that seems fine.

One data point here is that we weren't able to reuse the built-in redirection options because it did something very specific around the authentication header. If this was middleware, then the user could choose whether the redirection middleware would allow the authentication header to be forwarded.


> To get some experience with the new client, maybe we hide the new fns under the DarkInternal module for a while? That is, if we end up merging this.

Something like that for sure. Could we do something with fn attributes, possibly labeling it as pre-release or similar? We could do some sort of alpha/beta/RC thing, or users could even opt in via settings to get access, knowing that they might be cut off in the future. That's thinking ahead, but I wonder if we should start with a basic version of this now, instead of reusing darkinternal, in order to get experience with this sort of thing. If that's too much work for now, that makes sense too.

> The new client explicitly uses Http 3.0, versus Http 2.0 in the current fns.

> The current client has [this](https://github.com/darklang/dark/blob/fcc4ab7026a4c7818e2b5c60b299cdc88ebf50cb/fsharp-backend/src/BackendOnlyStdLib/HttpClient.fs#L149-L162) bit of code which seems to handle basic auth. I don't fully understand where those basic auth params came from - user passing them in as headers, maybe, and .NET interprets them automatically? Confusing, but I've stripped that out as it felt 'extra'. Any issues there? I can test on this manually, in any case.

If I recall correctly, this was to handle the possibility that the user provided the username and password in the URL. I don't currently have strong opinions on how we should handle this here, though we should probably not discard information provided.


> The current client attempts to interpret the request body for content type, then serializing and setting content-length header appropriate. The new client does none of that. Same thing for the response body - we now just take the bytes and return them, otherwise ignoring the body.

Great!

> The current client includes [a few](https://github.com/darklang/dark/blob/fcc4ab7026a4c7818e2b5c60b299cdc88ebf50cb/fsharp-backend/src/BackendOnlyStdLib/HttpClient.fs#L193-L195) `defaultHeaders` in responses. Should we do that in the new client? For now, I've included them.

These are response headers? I don't see a good reason to include them by default. What's your thought here?
 
> The current client included some HTTP status header. That seemed like OCaml-replication-specific logic, so I removed it from the new client.

🎉

> There are a few places where we `toLowercase` headers' keys and values. Is that appropriate? That sort of manipulation feels like it should be in an accessory function, not within this client.

I think we want to make requests be identical regardless of the HTTP version that the server supports. Since HTTP2 and HTTP3 use lowercase headers only, then we should do it by default for HTTP1 as well. This is more important for response headers (and in the HTTP framework request headers) as these are user visible.

> I went back and forth on in-lining `HttpBytesClient.fs` into `LibHttpBytesClient.fs`. Open to opinions, but the outcome here may depend on what 'module' the new client fns belong in long-term.

I probably would. Or at least they could both be in the same module as separate submodules/files.

>   * [ ]  traces, analysis

I would be keen to have as little difference between what's in the trace and the actual request made. (maybe that's obvious, but worth mentioning just in case)


> * [ ]  depending on if we release this to DarkInternal for now, or normally to the full userbase, consider providing some helper functions that capture logic ripped out of the current client:
>   
>   * [ ]  redirects
>   * [ ]  ...

Compression too (again depending on what we do).

fsharp-backend/src/BackendOnlyStdLib/LibHttpBytesClient.fs
StachuDotNet: perhaps it shouldn't exist, and the query should be included by the user in the URL. That's what Elm does, it seems: https://package.elm-lang.org/packages/elm/http/latest/Http

pbiggar: Yes, that makes sense to me.

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
StachuDotNet: I'm currently of the stance of 'let's not automatically decompress, but avail some decompression functions.

I'm vaguely nervous about the implications here though:
Is there any reason to think that ASP.NET does something _fancy_ such that automatic .net httpclient -level decompression would be notably more efficient than doing so 'manually' via some function?
There will certainly be more `bytes` passed around - probably not a big deal?

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
StachuDotNet: I'm not sure. I'll add a specific TODO there to research later.

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
StachuDotNet: I added some commentary around this topic in-line.

StachuDotNet:
Current status:

- this adds a single function: `DarkInternal::baseHttpClientCall`
- usage looks like this:
  `DarkInternal::baseHttpClientCall "get" "https://stachu.net" String::toBytes "" []`
- I've used it myself locally, and it feels pretty good
- it's hidden from non-admins
- I've added many in-line `HttpBaseClientTODO`s
- I haven't written tests around it

What should we do from here, @pbiggar?
- Merge the PR, with the many in-line TODOs
- leave the PR open for a bit, circling back next week
- save the code to a gist for future reference; abandon the PR in favor of a more 'collaborative' document like we have going for the base http handlers
- other?

I'm running out of steam on this topic for now, and would like to step away from it until next week. I suspect it'll be a while until this functionality is availed to users, so merging this as-is as an internal function feels fine to me, rather than letting the PR go on for a while.

fsharp-backend/src/BackendOnlyStdLib/HttpBytesClient.fs
pbiggar: A thought I had, is that automatic decompression might not work for the specific initial  use case, as you probably don't want to decompress an asset being uploaded before passing it on to CDN.

On the other hand, you may be able to prevent decompression by passing headers which don't imply the body should be decompressed. But that feels like quite a dance, which is exactly what we're trying to avoid.

So I guess I'm leaning on the side of no decompression as well. The decompression dance is not trivial, so will need to be careful for that middleware.

It may be also that we can't get it as efficient as the asp.net decompression; that would be a good thing to learn, and try to address in the language and/or runtime. If we can't fix it we can always make a new version with improved performance, if it comes to that.

pbiggar:
Agreed on all points.

> I'd much rather a simple function, and other functions that 'wrap' it, than to have a big 'config' object, or a number of 'config' params passed in. that configuration list is bound to grow and grow - I'd rather our stdlib grow, and users are able to pick and choose which functions they call accordingly. To be clear, I don't like the idea of many 'options' being passed into some base function, but I _do_ like the idea of availing a function that does the wrapping so the users don't have to define a '_reasonable_ base http client'.

I wanted to clarify for the future (not applicable here, as we're not actually exposing config parameters here): the approach of multiple functions starts to fall apart once you have three or more config parameters. That's why I like the approach of having one record with all the config options in it, and then having a default value that the user can access and partially overwrite. (which is conceptually the same as default arguments)

StachuDotNet:
I agree with the preference to have one big 'config' record param rather than many config params.
And I think that's a good option!


I was previously thinking of a different option, and didn't communicate it well.

I think both of the ideas have this basic form:
- avail a super basic `call` function (the one here) that does the absolute minimum.
- avail many other functions like 'Gzip.decompress`
- avail another function used more commonly, roughly equivalent to the current HttpClient fns - call it `callWithDefaultStuff`.
  This uses those helper functions mentioned in the second bullet.


Here's where the ideas differ a bit:

Rather than `callWithDefaultStuff` be defined based on a default `config` arg being passed to `call` like:
```
module HttpClient = 
  type HttpClientConfig = {...}
  let call (config: HttpClientConfig) method url body headers
  let callWithDefaultStuff = call HttpClientConfig.Default method url body headers
```

I was instead imagining `callWithDefaultStuff` to be defined more like:
```
module Gzip = 
  let decompress (b: byte[]): byte[] = ...

module HttpClient = 
  let call (config: HttpClientConfig) method url body headers

  let callWithDefaultStuff method url body headers =
    call method url body headers
    |> Result.map(fun response -> ... (handle redirects))
    |> Result.map(fun response -> {response with body = Gzip.decompress body }
    |> etc.
```

So no 'config' param, but rather functional composition.
Defining it this way feels more functional and extensible to me.

That said, maybe _that_ sort of falls apart for certain functionality, or is otherwise bad for some reason.

StachuDotNet:
That all said, there are some things (like timeouts) that don't really fall under the umbrella of being composable - so, some things would probably settle in a config object after all, even with my alternative.

StachuDotNet:
> Agreed on all points.

I'm not sure if this included 'merging pretty much as-is is fine' - could you please clarify?

pbiggar:
> Current status:
> 
> * this adds a single function: `DarkInternal::baseHttpClientCall`
> * usage looks like this:
>   `DarkInternal::baseHttpClientCall "get" "https://stachu.net" String::toBytes "" []`
> * I've used it myself locally, and it feels pretty good
> * it's hidden from non-admins
> * I've added many in-line `HttpBaseClientTODO`s
> * I haven't written tests around it
> 
> What should we do from here, @pbiggar?
> 
> * Merge the PR, with the many in-line TODOs
> * leave the PR open for a bit, circling back next week
> * save the code to a gist for future reference; abandon the PR in favor of a more 'collaborative' document like we have going for the base http handlers
> * other?
> 
> I'm running out of steam on this topic for now, and would like to step away from it until next week. I suspect it'll be a while until this functionality is availed to users, so merging this as-is as an internal function feels fine to me, rather than letting the PR go on for a while.

Sorry, I had missed this. I think it's safe to merge.

PR #4372, opened 2022-08-07
Fix reading InPipe on client, with test case

pbiggar:
I didn't correctly parse `InPipe` constructors, breaking trace loading for users using pipes.


pbiggar:
I removed some old tests that were hard to keep and only tested OCaml types, which we don't super care about anymore

PR #4387, opened 2022-08-13
Add tunnel host setting

pbiggar:
Up until now, only named admins could load the editor in production, using assets loaded from a tunnel (such as localtunnel or ngrok). Specifically, it required using our ngrok account with tunnels specifically allowed.

We'd like to open this up to anyone though. If APIs are going to run against the editor canvas, instead of a local server, then anyone developing the editor is going to need to tunnel the assets.

Initially I went in the direction of adding a query param to specify the tunnel. However, I realize that an attacker could then create a URL that would have are user load the attackers assets, and that wouldn't be good. So instead you now have to register your tunnel, before you use it, and you can't register it via just query params.

It looks ugly, has no animation, and it's an intuitive to use, but I'd rather ship it in its current state then fix those right now. I plan to fix those as part of the next few projects, and and this project has gotten slightly out of control already.

Other fixes done as part of this:
- refactored Sidebar, VPos to their own modules
- switched from bs-webapi to its replacement: rescript-webapi
- improve error logging during dev migrations
- fix comments on a namespace that ionide kept yelling about
- remove rustfmt file


fsharp-backend/src/LibBackend/Account.fs
```diff
@@ -345,6 +345,41 @@ let orgs (userID : UserID) : Task<List<OrgName.T>> =
   |> Sql.executeAsync (fun read -> read.string "username" |> OrgName.create)
   |> Task.map List.sort
 
+// **********************
+// Tunnels
+// We allow users to run their own JS/CSS assets, so they can help develop the
+// client. For security, uses must register the tunnel first (allowing a tunnel to be
+// set from just a url would allow an attacjer to trick users into loading assets
```
StachuDotNet, line 352:
```suggestion
// set from just a url would allow an attacker to trick users into loading assets
```

fsharp-backend/src/LibBackend/Account.fs
```diff
@@ -419,5 +454,6 @@ let initializeDevelopmentAccounts (serviceName : string) : Task<unit> =
     print $"Initing LibBackend.Account in {serviceName}"
     do! initTestAccounts ()
     do! initAdmins ()
+    print $"Inited  LibBackend.Account in {serviceName}"
```
StachuDotNet, line 457: Is `print $"Initing LibBackend.Account in {serviceName}"` here twice intentionally?

StachuDotNet (approved):

fsharp-backend/src/LibBackend/Account.fs
StachuDotNet: nevermind. I read too quickly.

pbiggar: One says "inited" and the other says "initing"

PR #4391, opened 2022-08-15
Expose internal fn to load ops for a toplevel

StachuDotNet:
The results are currently aren't ideal, but it's better than nothing.

needs tests


StachuDotNet:
Hm, Actually I don't think I can cover this in a unit test with the existing infra. (at least, not in a .test file) since we don't have fns to create/manipulate handlers/etc. WDYT, @pbiggar?

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
```diff
@@ -1170,4 +1170,27 @@ human-readable data."
...
+          | _, [ DUuid canvasID; DInt tlid ] ->
+            uply {
+              let tlid = uint64 tlid
+              let! ops =
+                let loadAmount = Serialize.LoadAmount.IncludeDeletedToplevels
+                Serialize.loadOplists loadAmount canvasID [ tlid ]
+
+              match ops with
+              | [ (_tlid, ops) ] -> return ops |> List.map (string >> DStr) |> DList
+              | _ -> return DList []
```
StachuDotNet, line 1191: `Serialize.loadOplists` does not return a `(tlid, ops)` pair for TLIDs that either don't exist or have no ops.

This (returning a `DList []` in such a case) seemed like a reasonable response to such, but maybe not?

Alternatively, we could update `Serialize.loadOplists` to return a pair regardless of whether it has ops or not.

StachuDotNet: draft refactor of `Serialize.loadOplists` (just the end)

```
  Sql.query query
  |> Sql.parameters [ "canvasID", Sql.uuid canvasID; "tlids", Sql.idArray tlids ]
  |> Sql.executeAsync (fun read -> (read.tlid "tlid", read.bytea "oplist"))
  |> Task.bind (fun list ->
    tlids
    |> Task.mapWithConcurrency 2 (fun tlid ->
      match List.tryFind (fun (t, _ops) -> tlid = t) list with
      | Some (_tlid, oplist) ->
        task { return (tlid, BinarySerialization.deserializeOplist tlid oplist) }
      | None -> task { return (tlid, []) }))
```

pbiggar: I would make the change in darkinternal. Empty list seems fine.

StachuDotNet:
currently a minor build error, which I'll fix in a bit. Short of that, seems good to me.

pbiggar (approved):

PR #4392, opened 2022-08-15
Add Savepoints to Ops in Client, to re-support Undo functionality

StachuDotNet:
TLSavepoints haven't been persisted alongside recent ops, as they have prior - this is causing Undo to not work for recently-changed code.

The relevant integration test has been working, because the TLSavepoint ops were persisted in the actual integration test _file_, not part of the tested UX. As such, a follow-up will be to add another integration test that ensures TLSavepoint is added appropriately (or just adjust the existing one).


StachuDotNet:
handling test failures - I think we just need to handle the `withSavepoints` in an above layer of logic.

pbiggar (approved):

PR #4393, opened 2022-08-16
WIP - fixing reconstruction

StachuDotNet:
given the code
```fsharp
("hello",12345)
```

if I select between the two |s:
```fsharp
("he|llo",12|345)
```

Upon hitting 'delete' I expect the remaining state to be
`"he|"`

The tuple will be killed as the deletions happen, so I expect the 12345 to disappear.
Then, I should have the `"hel"` remaining, and be on the right side of the `e`.

Instead, I get `"hel|"` - one of the "l"s remains.

---

This seems to be because how `deleteCaretRange` works:

https://github.com/darklang/dark/blob/d6d113b9dc99bf8c2401fdd5ce01f4bf40ec2e72/client/src/fluid/Fluid.res#L4930-L4953

- put the cursor at the RHS of the selection (in this case, 10)
- record the 'col' of the LHS of the selection (in this case, 4)
- hit 'backspace' until either:
  - the last backspace changed neither AST nor cursor position
  - the cursor is now at the recorded LHS

because the ( of the tuple is deleted, however, the initial LHS evaluation is sort of invalidated.
The code thinks to go to position 4, but position 4 is eventualy between the 2 "l"s, rater than between "e" and "l".

---

My first attempt here was to rewrite the algorithm pretty dramatically. It works but breaks many other tests - will review and try again tomorrow.

### DIFF
--- a/client/src/api/API.res
+++ b/client/src/api/API.res
@@ -93,7 +93,7 @@ let addOp = (m: model, focus: AppTypes.Focus.t, params: APIAddOps.Params.t): cmd
     "/v1/add_op",
     ~decoder=APIAddOps.decode,
     ~encoder=APIAddOps.Params.encode,
-    ~params,
+    ~params={...params, ops: APIAddOps.Params.withSavepoints(params.ops)},
     ~callback=x => AddOpsAPICallback(focus, params, x),
   )
 
--- a/client/src/api/APIAddOps.res
+++ b/client/src/api/APIAddOps.res
@@ -1,3 +1,5 @@
+open Prelude
+
 module PT = ProgramTypes
 module RT = RuntimeTypes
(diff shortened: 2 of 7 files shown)


StachuDotNet:
Note: I needed undo to work locally, so branched this off of my branch that fixes such.

StachuDotNet:
this strategy _might_ worth with some more tuning - `numberOfTimesToBackspace` is probably a bad metric. Will briefly try to make a less convoluted version of this logic.

pbiggar:
Haven't looked at the algorithm, but the way I would solve this is to use ASTRefs instead of positions

StachuDotNet:
this has been superseded by #4531

PR #4396, opened 2022-08-17
Remove ocamltypes

pbiggar:
This removes OCamlTypes and the OCamlCompatible serializers, which are no longer used.

To allow this:
- added a new `PackageManager.tipe` which is a cut down version of the old OCaml `tipe`, just enough to deserialize the types in the parameter column in the `packages_v0` table in the DB
- convert the test data from the old format to the new format – these were used in integration tests and some unit tests
- removed all the V0 routes


fsharp-backend/src/LibBackend/Config.fs
```diff
@@ -9,7 +9,7 @@ open LibService.ConfigDsl
 // -------------------------
 
 // -------------------------
-// Root directories - see File.fs
+// Root directories - see File.fs aand Root type below
```
StachuDotNet, line 12:
```suggestion
// Root directories - see File.fs and Root type below
```

StachuDotNet (approved):

PR #4404, opened 2022-08-19
Refactor Interpreter's pattern-matching logic to prepare for Tuple pattern

StachuDotNet:
A refactor is required as tuple patterns, unlike existing patterns, require multiple sub-patterns to match at the same time.
The `Constructor` pattern is in similar, but not exactly the same - will write up more description shortly once this is working.

A major win here will be _far_ less mutable state in this portion of the interpreter.


fsharp-backend/src/LibExecution/Interpreter.fs
```diff
@@ -48,6 +53,327 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
+              incomplete pid
+            else
+              DError(sourceID pid, $"Invalid constructor: {name}")
+
+          // Trace each argument too. TODO: recurse
+          // List.iter
+          //   (fun pat ->
+          //     let id = Pattern.toID pat
+          //     state.tracing.traceDval false id (incomplete id))
+          //   args
```
StachuDotNet, line 299: So far, I've been ignoring this now-commented out trace logic. I'm curious if we need it, and if so, how to fit it into this model. (this may be clear to me later)

StachuDotNet:
At this point, this is worth a brief review, namely of [this](https://github.com/StachuDotNet/dark/blob/a0d09489f170b0a2892c09d6487460d330b49625/fsharp-backend/src/LibExecution/Interpreter.fs#L217) new logic - is there anything surprising here?

There's still one test failing - taking a look at that.

fsharp-backend/src/LibExecution/Interpreter.fs
pbiggar: To determine if we need it, you could play around with the editor and see what patterns have analysis results available to them. If everything looks right, then that logic probably isn't needed

StachuDotNet:
Refactored a bit, given some feedback - I think it's much nicer now.

FWIW I briefly wrote a version of this which called `checkPattern` only if there's not a found case already, but that led to some (imo) really ugly nesting: https://gist.github.com/StachuDotNet/f3a22ae216246ca1b87ae3139fe7dc8f

StachuDotNet:
This has gone through another wave of refactoring - I think it's much better now, and would appreciate a review, @pbiggar.

---

I'm currently blanking on how to reasonably fuzztest the change.

I started to add a thin `EvalOptions` type, and `evalOptions: EvalOptions` param, to the Interpreter functions, but that ends up being _way_ too many changes, with many questions - seems a bad path.

Any ideas?

fsharp-backend/src/LibExecution/Interpreter.fs
```diff
@@ -48,6 +49,279 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
+      let stWithNewVars newVars = Map.mergeFavoringRight st (Map.ofList newVars)
+
+      let mutable foundMatch : Option<Dval> = None
+
+      do!
+        cases
+        |> Ply.List.iterSequentially (fun (pattern, rhsExpr) ->
+          let passes = patternPasses matchVal pattern
+
+          match foundMatch, passes, state.tracing.realOrPreview with
```
pbiggar, line 302: I think we should find a different way to represent this. There's only three cases so matching on the three values doesn't seem to represent this well. I think an `if` statement with three branches would probably work better

fsharp-backend/src/LibExecution/Interpreter.fs
```diff
@@ -48,6 +49,279 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
+          // Found a match - persist traces and set the result
+          | None, true, _ ->
+            uply {
+              let newDefs = tracePattern state.onExecutionPath matchVal pattern
```
pbiggar, line 306: The split between trace pattern pattern passes feels a little bit off to me, in a few ways:
- tracePattern fetches the matched values, which feels outside its scope (at least, given its name)
- the logic of whether something should match is repeated twice

I think if pattern passes returned the list of ids and Dvals, then tracing could just be iterating through that list, with no need to repeat the logic. I'm not sure if that would work, just an idea.

fsharp-backend/src/LibExecution/Interpreter.fs
pbiggar: The comments on each branch supports this too.

fsharp-backend/src/LibExecution/Interpreter.fs
```diff
@@ -48,6 +49,279 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
+
+      r dv pattern
+
+    uply {
+      // The value we're matching against
+      let! matchVal = eval state st matchExpr
+
+      let stWithNewVars newVars = Map.mergeFavoringRight st (Map.ofList newVars)
+
+      let mutable foundMatch : Option<Dval> = None
```
pbiggar, line 295: Perhaps `hasMatched`? I was confused by `foundMAtch`.

fsharp-backend/tests/Tests/Execution.Tests.fs
```diff
@@ -348,48 +348,60 @@ let testLambdaPreview : Test =
...
-    let varRhsId = gid ()
-
-    let astFor (arg : Expr) =
-      EMatch(
-        mid,
-        arg,
-        [ (PInteger(pIntId, 5L), EInteger(intRhsId, 17L))
+  testList
+    "test match evaluation"
+    [ let matchId = gid ()
```
pbiggar, line 361: Using a `testList` and the `yield` here is very confusing to me - I don't really understand how it works or why it's necessary. Can you do something simpler, like a `List.map`?

pbiggar (changes requested):
This is looking really nice

fsharp-backend/src/LibExecution/Interpreter.fs
StachuDotNet: Agreed - if statements look a lot better

fsharp-backend/src/LibExecution/Interpreter.fs
StachuDotNet: `hasMatched` feels like a `bool` to me, and I was trying to avoid having 2 mutable bindings (instead, just using one `option`).

I've renamed to `resultMaybe` - I think that's a clearer compromise?

fsharp-backend/tests/Tests/Execution.Tests.fs
StachuDotNet: Yeah, I get how it's confusing. Struggled with the best thing to do there.

I'm trying to balance a few things:
- making the tests clear
- making each test case a specific `testTask`, rather than all of the checks be in one huge `testTask`
- ensuring test code only gets executed when  relevant tests get run

Specifically, I'm trying to work around the fact that we generate a bunch of `id`s with `gid()`, referencing both in the test-execution code _and_ in the cases. Kinda awkward.

I have a few options:
- how it is now - with the `yield!` hack.
 (negative: confusing)
- go back to how it was
  (negative: can't easily run individual test cases, which I _really_ appreciated with this setup)
- put all of the `gid()` calls _before_ the `testList`
  (negative: "test code" is executed before anything tries to run the test)

I just pushed a commit towards that last option. I don't love it, but maybe it's better? Happy to have input here, but I really don't want to give up the ability to run individual test cases easily. In any case, it doesn't take much computational effort to run `gid()`. WDYT?

fsharp-backend/src/LibExecution/Interpreter.fs
StachuDotNet: Take another look now :)

pbiggar: Yeah, this is much nicer. Is it possible to only check the pattern if we haven't already matched?

fsharp-backend/src/LibExecution/Interpreter.fs
pbiggar: I was going the other way – I wanted to have a variable name where are the primary focus was on tracking whether the pattern had matched for the first time yet.

fsharp-backend/tests/Tests/Execution.Tests.fs
pbiggar: Yeah this works nicely.

fsharp-backend/src/LibExecution/Interpreter.fs
StachuDotNet: `foundMatch` was aimed towards that goal - it's the found match, it being None is a pretty clear message that a match hasn't been found?

fsharp-backend/src/LibExecution/Interpreter.fs
StachuDotNet: Done now. WDYT? I'm not a huge fan of the `caseIdx` use, but it achieves all of our goals, I think, and it's a small enough surface.

StachuDotNet:
> I'm currently blanking on how to reasonably fuzztest the change.
> 
> I started to add a thin `EvalOptions` type, and `evalOptions: EvalOptions` param, to the Interpreter functions, but that ends up being _way_ too many changes, with many questions - seems a bad path.
> 
> Any ideas?

Still trying to figure this out - one alternative is to (1) generate a bunch of Exprs with a Generator, (2) evaluate all using the 'old' logic; persist to disk, (3) switch to 'new logic'; evaluate all exprs again; persist to disk, (3) check diffs against files on disk. Seems a bit much.

Or, give up on the fuzztesting.

Happy to take other ideas.

pbiggar:
> > I'm currently blanking on how to reasonably fuzztest the change.
> > I started to add a thin `EvalOptions` type, and `evalOptions: EvalOptions` param, to the Interpreter functions, but that ends up being _way_ too many changes, with many questions - seems a bad path.
> > Any ideas?
> 
> Still trying to figure this out - one alternative is to (1) generate a bunch of Exprs with a Generator, (2) evaluate all using the 'old' logic; persist to disk, (3) switch to 'new logic'; evaluate all exprs again; persist to disk, (4) check diffs against files on disk. Seems a bit much.
> 
> Happy to take other ideas.

So I think the issue here is that you'd like to thread an extra boolean through, temporarily, just for fuzzing, but there's nowhere to put that variable?

What about putting it in state? Maybe state.testing?

StachuDotNet:
> What about putting it in state? Maybe state.testing?

I like this - thanks! Will try it out and follow up.

fsharp-backend/src/LibExecution/Interpreter.fs
pbiggar: I thought the old version was fine in general, and just required moving things around a bit. There's definitely a lot more moving parts in the new version, so definitely like it less.

StachuDotNet: When I went towards "only check the pattern if we haven't already matched," every solution I tried seemed bad in some way - I'll give it another short round - maybe I missed something.

pbiggar: Yeah I think so

StachuDotNet: Check again - looks better to me.

I'm not sure why we needed the Ply.List.iterSequentially - I think this is fine with a simple `for ... in` loop?

StachuDotNet:
I'm currently in the middle of fuzztesting the changes, to ensure there's no regression.
The changes started to bleed notably beyond this PR, so I'm extracting those efforts into a separate PR that I plan to merge into this branch before proceeding here.

StachuDotNet:
We need some reasonable generators to facilitate fuzztesting for regression.

Sspecifically, we need to ensure we actually generate patterns that sometimes match the values, otherwise we're likely to generate tests like:
```
match 14761237123 with
| Some Some Ok "**^&$%^" ->
  let fujqwe7fasdf = ...
| ...
```

I'm breaking much of that work off into a separate PR, as to not make _this_ PR become too large.

---

It's likely to take me ~today still to write the relevant generators (in a separate PR)
And then some more time to fuzztest this change to ensure it doesn't break any existing code.

I'm happy to either let this PR wait a day or two until I get back to it, or merge it in, after reverting the interpreter to use the old fn, `evalMatchOld` (along with a rename).

pbiggar:
When I was fuzzing before, I made it to 90% of generated testcases were valid and 10% were invalid. 

I guess you want one or more valid matches each time. Maybe what do you want is to generate a list of bools that indicate if it should match, and then map those to patterns that do match, or just any random pattern.

Given a dval, we can probably easily generate a list of patterns that match it (eg `5` is matched by `PInt` and also `PVariable`, while `Ok 5` is matched by `PConstructor (PInt)`, `PConstructor (PVariable)`, and `PVariable`.

fsharp-backend/src/LibExecution/Interpreter.fs
```diff
@@ -289,6 +289,7 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
           else
             // If we're "previewing" (analysis), persist traces for all patterns
             traces |> List.iter (fun (id, dv) -> state.tracing.traceDval false id dv)
+            state.tracing.traceDval false patternId matchVal
```
StachuDotNet, line 292: regression testing caught this

StachuDotNet:
I'd spent too much time on this task trying to ensure that traces matched between versions, in addition to the results.
They ended up only differing in inconsequential ways, and my UX has been fine.

If/when approved, there's a remaining task to be done:
- [ ] update the Interpreter to either use `evalMatchNew` by default (without state.test.useNewLogic flag), or to in-line it, removing the old code.

At this point, I'm confident in the stability. What do you think?

StachuDotNet:
Entirely open to a better name for the `state.test.useNewLogic` flag.

One idea: `type TestAB = A | B` and `testAB` prop.

StachuDotNet:
I forgot to write an actual description in the OP - will do that by the morning.

pbiggar:
I would remove the old code, and presumably the flag?

StachuDotNet:
> I would remove the old code, and presumably the flag?

This is done now

fsharp-backend/src/LibExecution/Interpreter.fs
```diff
@@ -237,164 +238,94 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
+            let! r = eval state newSymtable rhsExpr
+            matchResult <- r
+            hasMatched <- true
+          else if isRealExecution then
+            // "Real" evaluations don't need to persist non-matched traces
+            ()
+          else
+            // If we're "previewing" (analysis), persist traces for all patterns
+            traces |> List.iter (fun (id, dv) -> state.tracing.traceDval false id dv)
+            state.tracing.traceDval false id matchVal
```
pbiggar, line 325: I'm confused by this. Why is it required on this branch (or conversely, why is it not required on the other branch)?

pbiggar (changes requested):
Looking great! One final place I'm confused, and we should be good to go.

fsharp-backend/src/LibExecution/Interpreter.fs
StachuDotNet: Hmm, I forget! I'll have to check.

The fuzztesting caught this - I think this ends up _overwriting_ the trace in some cases? I'll research in a bit and follow up.

StachuDotNet: I just went back to a commit where the fuzztests were still valid (before removing old code, etc.) to test this.

I removed that line, ran fuzztests for a while, and couldn't reproduce whatever I was seeing before, and otherwise forget the details.
I suppose I'll remove it now - if something comes up where we figure this out, great.

StachuDotNet, line 325:
```suggestion
```
Remove redundant trace in interpreter

pbiggar (approved):

PR #4411, opened 2022-08-23
Consolidate F# testfiles

StachuDotNet:
Within `fsharp-backend/tests`, we currently have multiple directories with custom test files and data - it's currently a bit confusing, and this PR consolidates all of these directories into a new `fsharp-backend/testfiles` dir.

The current structure (relevant parts) is:
- `fsharp-backend/`
  - `tests`
    - `testfiles`
    - `httpclienttestfiles`
    - `httpbasictestfiles`
      - `data`
    - `httptestfiles`
    - `staticasset-tests`
    - ... actual tests

This PR results in the following structure:
- `fsharp-backend`
  - `tests`
    - ... actual tests
  - `testfiles`
    - `execution` (used to be `testfiles`)
    - `httphandler`
      - `http` (used to be `httptestfiles`)
      - `httpbasic` (used to be `httpbasictestfiles`)
    - `httpclient`
    - `staticassets`
    - `data` (currently only used by `httphandler/httpbasic`, but seems a good place to store useful bits of static data, incl. non-text binary data)

, adjusting all relevant READMEs and code comments appropriately. Namely, the new `testfiles` dir now has a new README where the `FSharpToExpr.fs` syntax notes have been migrated. I find this less confusing than the current system.

Outstanding questions:
- is the `staticassets` directory just for reference, as the content we've uploaded? If so, I'll update the readme to be clear
- `fsharp-backend/tests` also contains a `json-test-suite` file. It seems unused - do you plan on using this again, or can it go? If it should stay, I'll add a README here.


fsharp-backend/testfiles/README.md
```diff
@@ -0,0 +1,51 @@
...
+- `data` houses static assets useful in tests
+- `staticassets` files ... TODO
+
+## Syntax (FSharpToExpr.fs)
+
+All "Dark code" in tests is parsed by the F# parser, then converted from F# to
+Dark - the converseion is within `FSharpToExpr.fs`. The error messages around
+this are not very good, but there are a few things to be careful of:
+
+- all tests must be of the format `x = y` (or `x <> y`, though that's rarely
```
pbiggar, line 29: This only applies to execution tests I think

fsharp-backend/testfiles/execution/README.md
```diff
@@ -0,0 +1,60 @@
+# Dark Execution test files
+
+The files in this directory are unit tests for the Dark language and standard
+libraries. A typical test is a single line, such as:
+
+`String.length_v0 "abcdef" = 6`
+
+This is written in a Dark-like syntax, and asserts that calling the function
+`String::length_v0` on `"abcdef"` is evaluated to 6.
```
pbiggar, line 9: Perhaps refer to testfiles/README.md here

pbiggar:
> * is the `staticassets` directory just for reference, as the content we've uploaded? If so, I'll update the readme to be clear

I think so, unsure.


> * `fsharp-backend/tests` also contains a `json-test-suite` file. It seems unused - do you plan on using this again, or can it go? If it should stay, I'll add a README here.

This used to be used. Perhaps it was only used by the fuzzers. I wouldn't object to having tests with this in it again, but I'm not sure how we'd do that.

pbiggar (approved):
Looks great, thanks!

PR #4412, opened 2022-08-23
Tidy F# FuzzTests project

StachuDotNet:
- removes AllowedFuzzerErrors, which was only used to check for OCaml-backend vs. F#-backend differences
- adds some unused, in-progress ProgramTypes generators
  - I was halfway through this when Paul suggested another route for the task I was writing them for. That said, they still feel generally useful. What should I do with these?
 - misc. code and README tidyings, esp. related to OCaml server no longer being relevant


StachuDotNet:
@pbiggar I liked your suggestion [here](https://github.com/darklang/dark/pull/4404#issuecomment-1224243826), but had already written these ProgramTypes generators. What do you think we should do with them - are they useful to hang around, or waste?

pbiggar (approved):

pbiggar:
> @pbiggar I liked your suggestion [here](https://github.com/darklang/dark/pull/4404#issuecomment-1224243826), but had already written these ProgramTypes generators. What do you think we should do with them - are they useful to hang around, or waste?

Happy to keep them around

Issue #4193, opened 2022-06-22
Allow non-Chrome browsers


StachuDotNet:
Dark has had this restriction for a long time. I remember some context being mentioned a while back, but I forget the reasoning.

- [ ] check why it was restricted
- [ ] see if it needs to still be restricted
- [ ] remove restriction

pbiggar:
Closing as it doesn't apply to darklang-next

PR #4421, opened 2022-08-25
Remove Chrome requirement

benwaffle:
A quick poke around in some REPLs on firefox works great, with no issues.


pbiggar:
This is a fix for #4193. At the moment, part of the reason that we only support Chrome is that it lowers the development cost and testing burden, so out plan is to keep it that way for now.

Really appreciate this PR, will likely include this somewhere down the line, but going to close it for now.

Issue #4427, opened 2022-08-29
Add tuple `match` pattern


StachuDotNet:
(I thought we had an issue around this, but I can't find it.)

![image](https://user-images.githubusercontent.com/906686/186996974-af06b252-6171-4354-9d31-979271720211.png)

TODOs:
- [x] basic functionality (in #4425)
  - the tuple match pattern _type_ is defined
  - the interpreter respects the new pattern, and basic tests pass
  - Fluid has been updated to add new new tuple pattern to a `match` (though only tuples of size 2, currently), and have its sub-patterns filled
- [x] once #4404 is merged, update the Interpreter to handle the new Tuple match pattern in the backend
- [x] (backend) more testing
- [x] (client) hide the new tuple pattern from users, or hide behind feature flag
- [x] duplicate bindings in the same Pattern should result in the 'right-most' one winning
  e.g. `match (1,2,3) with (a, 2, a) -> a` should return `3`
- [x] (Fluid) handle `,` inserts in the tuple match pattern (draft PR #4431)
- [x] lots of (Fluid) testing
- [x] include a tuple pattern in the pattern autocomplete (#4465)
- [x] reconstruction
- [ ] relevant TUPLETODOs
---

Relates to #3402. Needed for #4265.

StachuDotNet:
Bug found:

```
match 1
  (|1, 2) -> "test"
```

Hitting 'a' keystroke here should do nothing, but it's currently replacing the 1 pattern.
(replacing the `PInteger 3` pattern with a `PString "a"` pattern)

StachuDotNet:
This is good - there are some minor todos around reconstruction, but they're already marked as TUPLETODOs and noted in #4265 as a blocker for fully enabling tuples for users

PR #4431, opened 2022-08-31
Allow tuple patterns to be expanded with ,

StachuDotNet:
This is part of #4427

Prior to this, tuple patterns were restricted to 2 elements - `,` keystrokes within them did nothing.

The blocker here was that we didn't have a great way to update the _patterns_ of Fluid expressions without context of the matchId, in the same `FluidAST.update` fashion as Expr updates are.

This removes that blocker, and implements the expected `,` functionality.

(For situations with a MatchID within context, a few functions are available in Fluid.res itself: `replacePattern` mostly, which uses  `updatePattern` and `recursePattern`. If this PR looks OK directionally, I"ll see about removing those.)

This:
- catches up on some ReScript code formatting that I hadn't performed before
- updates FluidExpression.update to use postTraversal instead of `deprecatedWalk` (preTraversal broke a ton of tests)
- updates FluidExpression.update to take both an `fExpr` as well as a new `fPattern` - extending the functionality to work with Patterns as well as traces.
- for all usages of `update`, ensure the labeled params are before the un-labeled params
- adds `insertInTuplePattern` and `insertAtTuplePatternEnd`, which are the accessory functions driving this whole change
- respect `,` inputs in a tuple match pattern, by adding elements

Some TODOs:
- [ ] get feedback - does this strategy make sense? It feels a bit annoying to pass `fExpr` and `fPattern` all over the place, but likely that's a necesarry annoyance? My other thought here was to update everything to use `fluidPatOrExpr`s, but that didn't pan out well.
- [ ] see if we can remove the `updatePattern` and `recursePattern` within Fluid.res (seems out of place)
- [ ] add tests

### DIFF
--- a/client/src/fluid/AST.res
+++ b/client/src/fluid/AST.res
@@ -300,7 +300,7 @@ let variablesIn = (ast: E.t): AnalysisTypes.avDict => {
 
 let removePartials = (expr: E.t): E.t =>
   E.postTraversal(
-    ~f=x =>
+    ~fExpr=x =>
       switch x {
       | EPartial(_, _, e)
       | ERightPartial(_, _, e)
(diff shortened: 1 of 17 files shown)


StachuDotNet:
Replaced by #4440.

PR #4433, opened 2022-08-31
Format ReScript files in scripts/formatting/format

StachuDotNet:
There are 2 outstanding TODOs - the actual "check" and the actual "format" code.

I'm not sure how to handle those, but figured this draft would get the conversation started


StachuDotNet:
I did some minor digging; notes:
- `bsb -format` seems to be what does the actual formatting under the hood.
  - I found that `./node_modules/.bin/bsc -format client/src/fluid/Fluid.res` returned a formatted version of the file. (doesn't update the file)
  - here's the source code in the `rescript-vscode` extension, which seems to handle the formatting:
    https://github.com/rescript-lang/rescript-vscode/blob/8a4fd3c12655ce960161efe2af8536c4bbdd108a/server/src/server.ts#L710
    something calls this, and assumedly replaces the file with the results of this function. Seems pretty involved.
- the 'check' that we do for other code isn't supported out of the box ReScript: https://github.com/rescript-lang/rescript-compiler/issues/4780#issue-732511591 (though we could do this ourselves by diffing source from `bsb -format` check)

based on the "seems pretty involved" above, I'm leaving this for now. Will think about it and may follow up.

StachuDotNet:
Couldn't leave it.

Here's a starting, for a single file:

Format:
`./node_modules/.bin/bsc -format client/src/fluid/Fluid.res > client/src/fluid/Fluid.res.tmp && mv client/src/fluid/Fluid.res.tmp client/src/fluid/Fluid.res`

Check:
`./node_modules/.bin/bsc -format client/src/fluid/Fluid.res > client/src/fluid/Fluid.res.tmp && diff client/src/fluid/Fluid.res.tmp client/src/fluid/Fluid.res && rm client/src/fluid/Fluid.res.tmp` (etc.)

pbiggar:
Looks good - can you use a tmpdir instead of the .tmp suffix?

StachuDotNet:
Will do.

is `diff` enough here - how does CI determine if the build should fail based on calling `./scripts/formatting/format check` - if it outputs anything? Or maybe the other `check`s (from fantomas etc) are using stderr

pbiggar:
> Will do.
> 
> is `diff` enough here for the `check` - how does CI determine if the build should fail based on calling `./scripts/formatting/format check` - if it outputs anything? Or maybe the other `check`s (from fantomas etc) are using stderr

The scripts all use exit code (or should), so I think `diff` will work

pbiggar:
> 



> I've added scripts to be used to format/check rescript files.
> 
> I haven't followed up in `scripts/formatting/format` itself.
> 
> I went to create an equivalent of this, `all_rescript_files`:
> 
> ```
> all_fsharp_files() {
>   find "$1" \
>     \( -path ".git" \
>     -o -path "_build" \
>     -o -path "./_build" \
>     -o -path "node_modules" \
>     -o -path "./node_modules" \
>     -o -path "integration-tests/node_modules" \
>     -o -path "./integration-tests/node_modules" \
>     -o -path "_esy" \
>     -o -path "./_esy" \
>     -o -path "/home/dark/.esy" \
>     -o -path "esy.lock" \
>     -o -path "fsharp-backend" \
>     -o -path "fsharp-backend/.paket" \
>     -o -path "./fsharp-backend/.paket" \
>     -o -path "fsharp-backend/Build" \
>     -o -path "./fsharp-backend/Build" \
>     \) -prune -false \
>     \
>     -o -type f \
>     \( -name "*.fs" \
>     -o -name "*.fsi" \
>     \) -print
> }
> ```
> 
> but I got confused by `find` syntax, and likely `xargs` - not sure how, but I got lost.

The `-prune` section is the stuff to ignore. `-o` means "or". Anything else I can help with?

StachuDotNet:
I've seemingly got it working now, but I'm still confused:

- if the `-prune` section is stuff to ignore, then why is `fsharp-backend` in that section for `all_fsharp_files`?
- if the `-prune` section is stuff to ignore, why does it need both `"fsharp-backend"` and subdirectories in it, like `"fsharp-backend/.paket"`?
- wouldn't it be easier to do something like `find "$1" (-path "fsharp-backend") -a \(-name "*.fs" -o -name "*.fsi" \)` - filtering by the path and then the extension?
- the `prune` section _excludes_ because of the `-false` flag after it - right or wrong?

I'm sort of assuming that the answer to the second question has to do something with the `-prune` not pruning recursively, and so `"fsharp-backend/src"` _wouldn't_ be pruned?

StachuDotNet:
Ah, right - we'll need `bsb` for it to actually work in CI - I forgot :)

pbiggar:
> I'm not sure which package gets `bsb` into `node_modules/.bin`

bs-platform

StachuDotNet:
This is working! but the install of `bs-platform` takes almost 7 minutes.

I suppose this should just be cached?

scripts/formatting/format
```diff
@@ -21,6 +25,8 @@ function print_help ()
   printf '\t%s\n' "--ocaml-only: Only check ocaml files (mli, ml)"
   printf '\t%s\n' "--fsharp-only: Only check fsharp files (fsi, fs)"
   printf '\t%s\n' "--prettier-only: Only check prettier files (js, css, etc)"
+  printf '\t%s\n' "--include-rescript: Include rescript files (res, resi)"
+  printf '\t%s\n' "--only-rescript: Only check rescript files (res, resi)"
```
StachuDotNet, line 29:
```suggestion
  printf '\t%s\n' "--rescript-only: Only check rescript files (res, resi)"
```

StachuDotNet:
because of the prohibitively long time it takes to download `bs-platform` we decided to not check/format rescript files as part of CI. We can follow up later when we upgrade rescript versions. Still, there were some useful pieces here, so I've rewritten the branch's history to reflect the useful bits.

pbiggar (approved):

PR #4438, opened 2022-09-01
Allow pressing enter in more expression bodies

pbiggar:
When you're at the start of an expressions, pressing enter should insert a new line (which at the moment means a let). That mostly works, but there were two places it didn't:


StachuDotNet:
What tool are you using that shows the keystrokes?

pbiggar:
> What tool are you using that shows the keystrokes?

Gifox

StachuDotNet (approved):

PR #4439, opened 2022-09-01
Make match arrows match the style of other arrows

pbiggar:
Before:

Screen Shot 2022-09-01 at 9 48 16 AM

After:

Screen Shot 2022-09-01 at 9 44 51 AM


StachuDotNet (approved):
Nice! I hadn't even noticed this.

PR #4447, opened 2022-09-05
Better approach to tailwind

pbiggar:
This converts avatar css to tailwind. I tried to solve some of the more challenging problems with using tailwind to make sure that they could be solved, and it feels like it's in a good place.

I added the [Rescript tailwind project](https://github.com/green-labs/res_tailwindcss) to validate tailwind class names. This caught a couple of different errors. I tried a few different approaches, and this seemed like the one that worked best – in particular the obvious approach of trying to add functions or a set of attributes failed because of just how many tailwind classes there are.

One major change is that the CSS now needs to be built before the rescript is compiled, and that rescript files are now inputs of css. Definitely a bit weird.

It can still be annoying that there isn't type safety to ensure that the macro is called, but it's a much much better situation than the alternatives.

Also renamed grey3 to grey8, and added other greys that we've been using. I also need to change the color of toggles to one from our color palette; it didn't turn out the best but it's fine for now.

There's still work to do to get all of the avatar code into tailwind, but I think this is enough for now.


StachuDotNet (approved):
This is great - should we update the `coding-guide.md` to request all tailwind styling to use this system to ensure safety?

pbiggar:
Good idea. I've added a rescript section to the coding guide with tailwind info.

CODING-GUIDE.md
```diff
@@ -145,3 +145,10 @@ module RuleEngine =
   ```
 
   Failing to do so may result in conflicting generators or unexpected behaviour.
+
+## ReScript
+
+- we are moving our cs to tailwind:
```
StachuDotNet, line 151: is this supposed to be 'css' or does cs stand for something?

pbiggar: typo, thanks

PR #4453, opened 2022-09-06
Bugfix: don't allow tuple pattern parts to be overwritten

StachuDotNet:
Given the following code involving a "tuple match pattern":
```
match (1,2)
  (|"one",2) -> "success"
```

with the cursor at the noted spot, typing `1` should do nothing, leaving the cursor in place.

Instead, prior to this change, it would turn into:
```
match (1,2)
  (1|,2) -> "success"
```

Logic has been adjusted to not overwrite non-blank patterns in such a case, and a test around this has been added.

### DIFF
--- a/client/src/fluid/Fluid.res
+++ b/client/src/fluid/Fluid.res
@@ -3954,39 +3954,38 @@ let doExplicitInsert = (
       }
 
     | (ARPattern(_, PPTuple(kind)), PTuple(_pID, first, second, theRest)) =>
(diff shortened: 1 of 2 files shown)


StachuDotNet:
I've accidentally included a commit from https://github.com/darklang/dark/pull/4451 here. If that's merged first, that should disappear - either way, not a big deal.

pbiggar (approved):

PR #4455, opened 2022-09-06
Put cursor at _start_ of blank pattern when removing elements, not end

StachuDotNet:
Given the Dark code `let x = (___,|2,3)`, hitting `backspace` changes it to `let x = (|___,3)`, with the cursor at the _start_ of the blank.

The equivalent, for the tuple match pattern, was different. It turned:
```
match (1,2)
  (___,|2,3) -> ___
```

into
```
match (1,2)
  (|___,3) -> ___
```

This was done by updating `caretTargetForEndOfPattern` to `offset` the cursor position by `0` rather than by `3`.
There was a relevant note in place, and this seemed an appropriate time to follow the note.


pbiggar (approved):
Yeah, seems nicer

PR #4467, opened 2022-09-12
(draft) rebrand 'pattern' to 'match pattern' in client code

StachuDotNet:
Decoders are set up to handle the new serializations (e.g. `MPVariable`), but encoding still favors the old `PVariable` format.

A subsequent PR, after some delay to allow clients to catch up, should handle server-side renames and encoding updates.

This is a draft - this work needs to be broken down a bit rather than be one or two giant commits.


StachuDotNet:
There are many changes required here, to prevent confusion between 'match patterns' and 'let patterns' and 'pattern matches'.

I went down the rabbit hole of making these changes for the client only, and it came with many decisions.

Quickly, here are some terms I'd like to solidify:
- `pattern match` - a `match` expr, containing an `expr` we _match against_ one or more `cases`.
- `match pattern` - one of the _patterns_ that we can match against (e.g. variable pattern)

Shortcuts around this can thus be `PM` and `MP`, respectively.

---

This is all one giant PR. Below is me taking a step back to see what commits this should be broken down into, and where some discussions might be relevant:
- in `ProgramTypes.res` and `RuntimeTypes.res`
  - update the `Pattern` module to `MatchPattern`
  - update _cases_ of the `MatchPattern` type from `PVariable` to `MPVariable`
    - update the decoders to decode these properly, to be ready for upcoming backend changes
- modules that alias the `MatchPattern` module should be updated from `module P = ...` to `module MP = ...`
- `RuntimeTokenizer.res`'s `patternToTokens` -> `matchPatternToTokens`
- The `FluidPattern` module/file `fluidPattern` is renamed to `fluidMatchPattern`, along with `TestFluidMatchPattern.res`
- fluid shortcuts and test values should be updated (e.g. `tuplePattern2WithBothBlank` -> `tupleMP2WithBothBlank`
  (TestFluidClipboard.res, TestFluidMatchPattern.res, FluidShortcuts.res, FluidFuzzer.res)
- generally favor `mp` for local variables, over `pattern` (e.g. when `List.map`ing over a list of MPs)
- autocomplete
  - `FACPattern` -> `FACMatchPattern`
  - `FPAVariable` -> `FACMPVariable` (etc.)
- AST Refs
  - `ARPattern` -> `ARMPattern`
  - `astMatchPart` -> `astPatternMatchPart`
  - `astPatternPart` -> `astMatchPatternPart`
    - `PPVariable` -> `MPPVariable`
- `PPString` -> `MPPString` (etc.)
- `fluidPatOrExpr` -> `fluidMatchPatOrExpr`
  - `Pat` -> `MatchPat`
- Token parts: `TPatternBlank` -> `TMPBlank` (etc.)
  - toTypeName: `TPatternTrue(_) => "pattern-true"` -> `TMPTrue(_) => "match-pattern-true"`
- generally, any existing functions like `updatePattern` are renamed to the form of `updateMatchPattern` or `updateMP`, depending on the fn's length and frequency of usage. (the full name can be too verbose at times, I've found). There are a number of these specifically in Fluid.res
- TODO: update `pToStructure` and `pToTestcase`

This is what's been done - some of these decisions, I'm having second thoughts on. In particular, I'm having second thoughts on all of these long acronyms. They're a bit much. Rather, maybe this is a good time to throw more of our types/fns into modules as discussed elsewhere.

---

I'm going to let this sit for a bit. Maybe we'll decide to delay this work until more of these types are in modules and we use the prefix acronyms less often. Or maybe we'll commit to continuing as-is, and simply break this down into multiple commits.

pbiggar:
This all looks pretty good. One thing I'd say is that we don't need "pattern match", as this is just a "match".

StachuDotNet:
closing in favor of #4487.

PR #4469, opened 2022-09-13
Fluid Pattern Reconstruction fixes

StachuDotNet:
- generally documents `reconstructExprFromRange`
- adds notes around limitations in reconstructing `Constructor` patterns
- adds _basic_ support for reconstructing `Tuple` patterns
- starts to fill in tests around this functionality
- fixes a bug: given a `match` with multiple patterns, if you select `match` through the first pattern, the reconstruction would mistakenly create a blank 'case' per each unselected line.

### DIFF
--- a/client/src/fluid/Fluid.res
+++ b/client/src/fluid/Fluid.res
@@ -5701,59 +5701,71 @@ let reconstructExprFromRange = (range: (int, int), astInfo: ASTInfo.t): option<
         Some(e)
       }
     | EMatch(_, cond, cases) =>
-      let newCases = List.map(cases, ~f=((pattern, expr)) => {
-        let toksToPattern = (tokens, pID) =>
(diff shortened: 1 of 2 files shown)


pbiggar:
> Maybe trying to derive these recursive Patterns via Tokens isn't a smart venture, and we should update Pattern logic to work more like the Expr logic. Not sure :)

I think this is probably right. The current pattern logic looks like a hack that deliberately avoids recursion because 
the cases it does handle don't need it. So it ends up working as a hacky, flaky, parser sorta thing, which for break if we change tokenization in any way (and likely is already broken in some ways).

StachuDotNet:
That makes sense! The current solution definitely does seem like a flaky hack.

A rewrite is warranted, but for now this PR feels like a net positive to me: we have tests, documentation, a bug fixed, and basic tuple support. What do you think?

PR #4479, opened 2022-09-18
Tone down the string background a bit

pbiggar:
I think I overdid it with the string background color. This turns it down by about half. It's still there, just more subtle. It still provides a little contrast against the background so you can see spaces on it, it just no longer overwhelms the entire handler.

Before:

Screen Shot 2022-09-18 at 9 32 22 AM

After:
Screen Shot 2022-09-18 at 9 33 56 AM

### DIFF
--- a/client/styles/_fluid.scss
+++ b/client/styles/_fluid.scss
@@ -303,7 +303,7 @@ $ff-color: adjust-hue($purple, -40deg);
   .fluid-string-ml,
   .fluid-string,
   .fluid-pattern-string {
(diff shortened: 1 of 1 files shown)


StachuDotNet (approved):
I'm currently outside and looking at this on a slightly reflective laptop screen. I can't see the text background at all in the 'after' but can in the 'before.' It's not super critical, but figured I'd mention that solving for matte screens may yield an indistinguishable contrast for reflective ones. :shrug:

Issue #3418, opened 2022-01-31
String::toUuid is in the wrong module


pbiggar:
Rename `String::toUUID` to `UUID::parse`

StachuDotNet:
resolved in #4480.

PR #4480, opened 2022-09-19
Use IDs more

pbiggar:
A lot of our internal functions use usernames and canvas names instead of IDs. That can make things a little bit ambiguous, especially in the future if we choose to allow users to rename themselves.

Instead, we want to use canvasIDs and UserID where possible. This ads two functions that help with that. It also renames `String::toUUID` to `Uuid.parse`, which is better naming.

### DIFF
--- a/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
+++ b/fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
@@ -193,6 +193,7 @@ that's already taken, returns an error."
       previewable = Impure
       deprecated = NotDeprecated }
 
+
     { name = fn "DarkInternal" "getUserID" 0
       parameters = [ Param.make "username" TStr "" ]
       returnType = TOption(TUuid)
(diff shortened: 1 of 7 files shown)


fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
```diff
@@ -523,6 +544,32 @@ that's already taken, returns an error."
...
+          | _, [ DUuid userID; DUuid canvasID ] ->
+            uply {
+              try
+                let! canvasMeta = Canvas.getMetaFromID canvasID
+                let owner = Account.ownerNameFromCanvasName canvasMeta.name
+                match! Account.usernameForUserID userID with
+                | Some username ->
+                  match! Authorization.permission owner username with
+                  | Some perm -> return DResult(Ok(DStr(string perm)))
+                  | None -> return DResult(Ok(DStr ""))
```
StachuDotNet, line 565: What's the use case of this function - anything mission-critical? (either now or later)
I'm thinking if it makes sense for this to return a None or an Error here, rather than empty string.

StachuDotNet (approved):
Great! We can close https://github.com/darklang/dark/issues/3418 with this.

fsharp-backend/src/BackendOnlyStdLib/LibDarkInternal.fs
pbiggar: I decided not to do this, to keep it matching the other permissions stuff, which uses "r", "rw", and "".

pbiggar: It will be mission critical – the intent is to replace checkPermission, which this is a near copy of (just different signature)

StachuDotNet: Cool - I suppose we can migrate to a DU some day

PR #4485, opened 2022-09-19
Refactor function representation

pbiggar:
I'm looking at improving some function data in the UI, so I did this first. This should have no user-facing changes: everything is just field and type renames, just moving things around.


StachuDotNet (approved):

pbiggar:
Can you take another glance at the last few commits here?

StachuDotNet (approved):
Looks good! glad we're using the QueryFunction sqlspec there now

PR #4501, opened 2022-09-27
Fixes to CORS

pbiggar:
Not sure about the crossorigin change - I might revert that.


pbiggar:
Actually, this is outdated.

PR #4516, opened 2022-10-06
Fix package params (+ rename "tipe")

pbiggar:
We inexplicably dropped parameters names and types when displaying package manager fns.

Also, while I was here I renamed all uses of "tipe" to "typ", unless it was part of a word in which case it became "type" or "Type".

Before:

Screen Shot 2022-10-05 at 8 25 57 PM
After:

Screen Shot 2022-10-05 at 8 50 46 PM


pbiggar:
I've checked each string usage, and the only ones that changed were CSS classnames (which I also changed)

StachuDotNet (approved):

PR #4518, opened 2022-10-06
Clarify existing client types; add ClientProgramTypes

StachuDotNet:


StachuDotNet:
I went to use the new ClientProgramTypes in ApiServer, then hit `AddOpParamsV1` which currently exists in `LibBackend`. I think this should be migrated to ClientTypes as well?

pbiggar:
That sounds right, yes.

StachuDotNet:
reminder to self: make sure I didn't add any new 'tipe' usages - if so, rename to typ/type

StachuDotNet:
I'm at some dilemma that probably has a simple solution, but I haven't figured it out:
- `LibBackend` contains `Pusher.fs`
- API-communicated types such as `AddOpResultV1` are also in `LibBackend`
- I'm trying to move these types to `ClientTypes`
- some Pusher types are dependent on these API types. (e.g. `AddOpEventV1` is dependent on `AddOpResultV1`)

Based on this, I sort of want to move the Pusher.com types _also_ to `ClientTypes`. That seems generally appropriate, beyond the specific type dependency thing.
But how?
`Pusher.fs` needs access to those types now. Moving Pusher.fs to `ApiServer` makes some sort of vague sense, but not fully, and would result in various `BackendOnlyStdLib` fns no longer have direct 'access' to `Pusher.fs`.
So it seems there might need to be some indirect calling here, or something - I'm not sure.

pbiggar:
I've run into similar problems. I solved this in Rollbar by removing those types from Rollbar.fs. It's not clear to me that those types belong in Pusher.fs, for the reasons your specify, so maybe pusher.fs should be restricted to just the `push` functions, and not specific functions for each type that gets pushed?

pbiggar:
Happy to pair anytime!

StachuDotNet:
#4542 replaces this

PR #4527, opened 2022-10-10
Move to later version of bs-tea

### LINKED ISSUE #4413: Remove custom bucklescript-tea
We switched to a custom version of bucklescript-tea in 2020, as we needed this change: https://github.com/OvermindDL1/bucklescript-tea/commit/63963ffee701b8f30ed99cce7483682193731eb9

However, bucklescript-tea didn't take this upstream, and instead did this: https://github.com/OvermindDL1/bucklescript-tea/commit/6fec48f908000d5a5e7b2a4d77a59460fe5139ff

Here's the instructions on how to use it (it's just a subscription):
https://github.com/OvermindDL1/bucklescript-tea/pull/126

pbiggar:
No changelog

Fixes #4413.

This moves off our abandoned fork of bucklescript-tea, to a much newer (but still not latest) version of our new fork, rescript-tea.

renderCallback was never accepted upstream, and they went with a much nicer approach using subscriptions. Some other functions were only in our fork and so I moved them to rescript-webapi instead (which is what should happen for all our uses of Tea.Node).

### DIFF
--- a/client/src/app/AppTypes.res
+++ b/client/src/app/AppTypes.res
@@ -609,6 +609,7 @@ module Msg = {
   type rec t<'model, 'modification> =
     | IgnoreMsg(/* debug string so you know where it came from */ string)
     | IgnoreMouseUp // for nothingMouseEvent
+    | RenderEvent
     | FluidMsg(FluidTypes.Msg.t<'model, 'modification>)
     | AppMouseDown(MouseEvent.t)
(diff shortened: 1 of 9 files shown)


pbiggar:
There are still some issues, documented in #4547, but those existed before and if anything they seem slightly better now.

StachuDotNet (approved):

PR #4537, opened 2022-10-19
Stick to gsutil

pbiggar:
No changelog

I tried to improve the speed of deploys by switching to `gcloud storage cp` from `gsutil cp`, but ended up breaking things. So this goes back to the old one.

- [x] TODO: revert dockerfile


StachuDotNet (approved):

pbiggar:
This caused the error again, so re-reverting.

PR #4538, opened 2022-10-19
I speculate this will help us debugwhen integration tests hang in CI

pbiggar:
No changelog


StachuDotNet (approved):
It's hard for me to tell if this will be triggered as-is. I wonder if we should make this timeout a few seconds shorter than 10mins, so that it's triggered before circleci times us out?

pbiggar:
I thought about this, but CircleCI only shuts it down after 10m of no output, and this is 10m total.

PR #4539, opened 2022-10-20
Revert "Stick to gsutil"

pbiggar:
Reverts darklang/dark#4537


pbiggar:
I don't think this will help, so closing.

PR #4540, opened 2022-10-20
Get a new sha for app.js

pbiggar:


pbiggar:
Confirmed there's a different hash than before

PR #4545, opened 2022-10-21
Use tailwind 3.2 nested groups

pbiggar:
No changelog

New feature I saw in the dependabot changelog. Seems much nicer, and always better to use a first-class thing rather than a plugin. We also no longer need to declare them, so they're easier to use.


StachuDotNet (approved):
[nice](https://github.com/tailwindlabs/tailwindcss/discussions/7639#discussioncomment-3925616)

PR #4546, opened 2022-10-21
Print timedout tests in Playwright

pbiggar:
No changelog

When integration tests timeout, it's hard to know the cause. This searches through the json output to find the tests which were skipped, which I take to mean "were not complete when timed out".

We shouldn't really need this, as the test should a) fail after it's own timeout and b) be reported after the global timeout imo, but those aren't happening, so here we are.

I've reported these to playwright as well: https://github.com/microsoft/playwright/issues/18250 and https://github.com/microsoft/playwright/issues/18248

### DIFF
--- a/integration-tests/package-lock.json
+++ b/integration-tests/package-lock.json
@@ -5,19 +5,19 @@
   "requires": true,
   "dependencies": {
     "@playwright/test": {
-      "version": "1.27.1",
-      "resolved": "https://registry.npmjs.org/@playwright/test/-/test-1.27.1.tgz",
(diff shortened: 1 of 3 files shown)


StachuDotNet (approved):

pbiggar:
I've tried 10 times to get this to trigger but for some reason integration tests keep passing now.

pbiggar:
Ah, because they're failing: https://app.circleci.com/pipelines/github/darklang/dark/4293/workflows/a3e9a18c-192b-4787-a79a-c1d62e2f9ebf/jobs/44636

pbiggar:
I can't get these to fail anymore, so I'm just going to merge. I've validated that the successes are real, no idea why they refuse to fail when they used to fail 80% of the time.

PR #4548, opened 2022-10-22
Move to node18

pbiggar:
npm 7 should fix our slow npm installs using git clones


pbiggar:
This broke as node >= 16 doesn't like bs-platform 9.02.

PR #4570, opened 2022-10-30
Switch to rescript 9.1.4

pbiggar:
Changelog:

```
Internals
- upgrade to later version of rescript
- reduce editor download size by X%
```

We've been falling behind on rescript versions, which has made upgrading rescript-tea (the cause of significant technical debt) challenging. This upgrades to rescript 9.1.4.

The major thing that went away in rescript 9.1.4 was the old OCaml strings formatting capabilities, such as `Printexc.to_string` and `Printf.*`. The latter broke the `bs-deriving` plugin (there was also another error), which meant we don't have the derived `show` functions anymore. This is a significant drag, but the only other alternatives were to stay at the same rescript version, or the make a new version of bs-deriving. Neither option was great imo, so here we are.

I tried to find another option a while back and [there didn't seem to be one](https://forum.rescript-lang.org/t/upgrading-to-latest-rescript-with-ppx-deriving/2139/9).

The major downside here is that debugging code will be ugly. Output now looks like this, for instance:

Screen Shot 2022-10-30 at 7 03 02 PM

Before it looked like this:

Screen Shot 2022-10-30 at 7 03 50 PM

This appears in rollbars, the debugging sidebar, and the rescript-tea debugger. I think console logs should be fine but haven't tested them.

This code has a bunch of `Obj.magic` in it. `Obj.magic` causes the rescript type-checker to assume that the value is of the expected type, without doing anything to it. This allows us to convert from `Web.Node.event` to `Dom.event`, which have the same underlying type (a JS event).

After converting to an upcoming version of rescript-tea, we'll use `Dom.event` (or more specific events) a lot more often, and this problem (and uses of `Obj.magic`) will go away.


pbiggar:
We've been falling behind on rescript versions, which has made upgrading rescript-tea (the cause of significant technical debt) challenging. This upgrades to rescript 9.1.4.

The major thing that went away in rescript 9.1.4 was the old OCaml strings formatting capabilities, such as `Printexc.to_string` and `Printf.*`. The latter broke the `bs-deriving` plugin (there was also another error), which meant we don't have the derived `show` functions anymore. This is a significant drag, but the only other alternatives were to stay at the same rescript version, or the make a new version of bs-deriving. Neither option was great imo, so here we are.

I tried to find another option a while back and [there didn't seem to be one](https://forum.rescript-lang.org/t/upgrading-to-latest-rescript-with-ppx-deriving/2139/9).

The major downside here is that debugging code will be ugly. Output now looks like this, for instance:

Screen Shot 2022-10-30 at 7 03 02 PM

Before it looked like this:

Screen Shot 2022-10-30 at 7 03 50 PM

This appears in rollbars, the debugging sidebar, and the rescript-tea debugger. I think console logs should be fine but haven't tested them.

StachuDotNet:
Reviewing now, but fwiw, the `Tag` debugging console logs are what I've _always_ seen. I'm not sure why you had a different/better experience.

pbiggar:
Since the initial version, I've added a bunch of fixes.

- Fixed calls to rollbar
- added many toDebugString calls to make it easier to debug
- reinstated pretty output in the FluidDebugger and integration tests
- removed lots of show/deriving cruft
- unbox tlid/id (makes it easier to read in devtools)

Integration tests started to fail pretty often, seemingly due to minor races in what order things were rendered. Some fixes there:
- extended timeouts for analysis
- added an increasing requestID, and only store analysis results if they are newer than the old ones
- wait for the right execution button to show (it briefly flashes a button that matched the selector, but wasn't the correct play button)
- wait for a focus

The rest of the integration test fixes I merged in #4578 - that triggered a lot more on this branch than in main.

PR #4577, opened 2022-11-05
Fix for WSL users with a space in their username

pbiggar:
Changelog:

```
Contributor experience
- Fix error when initializing devcontainer when username has a space
```


pbiggar:
This wasn't actually the issue, so maybe close this for now

PR #4579, opened 2022-11-05
Try a fresh cache each week for .net

pbiggar:
No changelog

Try to speed up the build. The .net cache can get stale, so use a new one each week. That will slow down the first build of the week, but I expect this to shave a minute or two off all subsequent builds


pbiggar:
This seems to work (saw a fresh cache and then saw it reused)

PR #4581, opened 2022-11-07
By default, don't require files to exist or run on the host

pbiggar:
Changelog:

```
Contributor experience
- do not require setup on the container host (fixes bugs for WSL users)
```

Two users recently had issues running the devcontainer on Windows/WSL. Although the issues were significantly different in both cases, the common issue was that we were trying to make things happen on the container host, and those did not work.

In one case, there were errors running the initialization script due to misconfiguration of shell files. In another, the initialization script did not actually initialize the files in question.

The solution is not to have an initialization script. Instead, just don't require the files on the host, and allow users to opt into using them by editing devcontainer.json if they want (similar to what you have to do to use a remote host)

The dockerfile change overcomes a permissions issue that happened on WSL. No idea why or how it was caused, but this seems to resolve it.


StachuDotNet (approved):

StachuDotNet (approved):
Ah, I should have caught the `initializeCommand` not being removed. This looks great! I remember having similar issues when I first got Dark set up, on linux.

PR #4588, opened 2022-11-08
Fix and Re-merge "Extend backend "ClientTypes"

StachuDotNet:
Reverts darklang/dark#4587, which is a revert of #4542.

TODOs:
- [x] find what `.allow`s are missing in CronChecker
- [x] don't fail if an `.allow` is missing, anywhere


StachuDotNet:
do we already have some pattern for rollbaring an exception in a non-blocking way?

```
let assertSerializable (t : System.Type) : unit =
  if not (isSerializable t) then
    Exception.raiseInternal
      "Invalid serialization call to type not allowed: add `Json.Vanilla.allow<type>()` to allow it to be serialized"
      [ "type", string t ]
```

This is the bit that's causing things to break. It feels like we might have an alternative to Exception.raiseInternal but I'm blanking

Maybe this construct exists 'above' prelude which is why I'm not seeing it

pbiggar:
> do we already have some pattern for rollbaring an exception in a non-blocking way?

Doesn't need to be an exception - we can call `Rollbar.info` instead. Right now, calling Exception rollbars only because we register an exception callback at startup via some stuff in Prelude - you could add similar things for Rollbar.info (as Prelude does not depend on LibService where Rollbar is defined)

StachuDotNet:
Ah I just missed that you recommended .info instead of error :construction:

fsharp-backend/src/Prelude/Prelude.fs
```diff
@@ -1083,6 +1083,7 @@ module Json =
     let assertSerializable (t : System.Type) : unit =
       if not (isSerializable t) then
         rollbarExceptionInBackground
+          false
```
pbiggar, line 1086: I didn't mean this should be an info, error is fine for `allow`. Sorry for the confusion.

fsharp-backend/src/LibService/Rollbar.fs
```diff
@@ -450,5 +450,7 @@ let init (serviceName : string) : unit =
   //   )
   // |> ignore<obj>
 
+  Prelude.rollbarExceptionInBackground <- fun msg metadata -> sendError msg metadata
```
pbiggar: I don't follow this - are these calls safe?

The name implies async but it's not, right? It's just a regular rollbar call?

pbiggar (changes requested):
Looks good but I have some questions.

fsharp-backend/src/LibService/Rollbar.fs
pbiggar: We use "background" at other time to mean async, right? I'd prefer a different name. Certainly needs a comment.

The other thing I was unclear about was whether this could/would throw an exception.

Perhaps: `Prelude.sendRollbarError`?

pbiggar:
Yay!

PR #4592, opened 2022-11-09
Add --loop flag to integration tests

pbiggar:
Changelog:

```
Internal
- allow integration tests to loop
```

By looping the integration tests with `--loop`, we can find flaky tests. Often it takes 4-15 loops to find a flaky one. It's also really useful to combine with `--pattern`, to debug one test that you know if flaky

### DIFF
--- a/integration-tests/run.sh
+++ b/integration-tests/run.sh
@@ -82,11 +82,6 @@ then
   exit 1
 fi
 
-# We need to restart the server after adding new packages. Integration tests test
-# against the dev environment, not the test one.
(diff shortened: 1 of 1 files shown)


StachuDotNet:
This is a bit nitpicky, but could we have `LOOP=0` be the default, with `1` indicating "this should loop"?

pbiggar:
> This is a bit nitpicky, but could we have `LOOP=0` be the default, with `1` indicating "this should loop"?

Ah, I can see why you ask, but alas `0` is true in bash 🤦‍♂️

StachuDotNet:
Ah, I see - confusing!

PR #4593, opened 2022-11-09
Show selected fn doc, not the doc of the fn it's an argument of

pbiggar:
Changelog:

```
Autocomplete
- When using nested function calls, display documentation for the nested function
```

As I recall, we chose the old order to prioritize more detailed documentation, so for example if you have `a 5|` it would show the docs for `a` with parameter info. However, if we have `a (b| c)` it would prioritize showing docs for `a` instead of `b`.

I need to do a little more testing to make sure this doesn't make other things worse. Maybe I'll add some tests.


pbiggar:
Abandoning for now

PR #4594, opened 2022-11-09
Remove OCaml

pbiggar:
Changelog:

```
Internal
- Remove all remnants of OCaml files and tooling
```

Moved all the OCaml stuff into https://github.com/darklang/dark-ocaml. The remaining two containers can be built and deployed over there, if needs be, but hopefully they won't be touched until we remove them.

Multiple contributors have had issues with ocaml builds, including very high memory use (and hence extremely long build times, like 10 hours) and them not working on i5s. Best to just remove them, esp since we're probably a few months from removing them otherwise.


StachuDotNet (approved):
As far as I can tell, these all seem like the appropriate removals. I assume `dark-ocaml` should stabilize (have https://github.com/darklang/dark-ocaml/pull/7 merged) before this is merged?

pbiggar:
> As far as I can tell, these all seem like the appropriate removals. I assume `dark-ocaml` should stabilize (have [darklang/dark-ocaml#7](https://github.com/darklang/dark-ocaml/pull/7) merged) before this is merged?

Yes. Should be ready now - can you review?

pbiggar:
This worked fine.

PR #4609, opened 2022-11-23
Fix ssl-cert-snakeoil.key permission denied

liweijian:
Changelog:

```
Area of change
- Add some `chmod` and `chown` in the **scripts/devcontainer/_start-background-services**
```

alternatively:

No changelog

Not sure if it's always happen, but I got this postgresql related error when trying to run `scripts/builder --compile --watch --test`

```
UTC FATAL: could not access private key file "/etc/ssl/private/ssl-cert-snakeoil.key": Permission denied
```

It sounds a lot like that's a [really tricky](https://stackoverflow.com/questions/34209661/fatal-could-not-access-private-key-file-etc-ssl-private-ssl-cert-snakeoil-key) issue, inspired by [this crazy solution](https://github.com/puntonim/docker-postgresql93/issues/2), finally I made it.

I just added some `chmod` and `chown` in the **scripts/devcontainer/_start-background-services**, I believe it won't have any other product impacts.

I didn't test it, but after adding this shell commands, I got `-- Initial compile succeeded` finally.

Just rollback this script.

No, since it's an one-off script.

### DIFF
--- a/scripts/devcontainer/_start-background-services
+++ b/scripts/devcontainer/_start-background-services
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+sudo chmod 766 /etc/ssl/private
+sudo chown postgres:postgres /etc/ssl/private/ssl-cert-snakeoil.key
+
 for name in "${@}"; do
   if [[ $name == postgresql ]]; then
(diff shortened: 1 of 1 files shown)


pbiggar:
Thanks! Have copied this into #4118

PR #4618, opened 2022-11-29
Add List::partition_v0 stdlib fn

StachuDotNet:
Changelog:

```
Standard library
- add `List::partition_v0` function to split a list into 2 lists
```

---

I was recently reaching for this function and found that we didn't have it.
For now, it's hidden behind our experimental tuple usage editor setting.

### DIFF
--- a/fsharp-backend/src/LibExecutionStdLib/LibList.fs
+++ b/fsharp-backend/src/LibExecutionStdLib/LibList.fs
@@ -1481,35 +1481,23 @@ let fns : List<BuiltInFn> =
         (function
         | state, [ DList l; DFnVal fn ] ->
           uply {
-            let abortReason = ref None
-
             let f (dv : Dval) : Ply<bool> =
               uply {
-                let run = abortReason.Value = None
+                let! result =
+                  Interpreter.applyFnVal state (id 0) fn [ dv ] NotInPipe NoRail
 
-                if run then
-                  let! result =
-                    Interpreter.applyFnVal state (id 0) fn [ dv ] NotInPipe NoRail
-
-                  match result with
-                  | DBool b -> return b
-                  | (DIncomplete _
-                  | DErrorRail _
-                  | DError _) as dv ->
-                    abortReason.Value <- Some dv
-                    return false
-                  | v ->
-                    return
-                      Exception.raiseCode (Errors.expectedLambdaType "fn" TBool v)
-                else
-                  return false
+                match result with
+                | DBool b -> return b
+                | (DIncomplete _
+                | DErrorRail _
+                | DError _) as dv -> return false
+                | v ->
+                  return Exception.raiseCode (Errors.expectedLambdaType "fn" TBool v)
               }
 
             let! (resultA, resultB) = Ply.List.partitionSequentially f l
 
-            match abortReason.Value with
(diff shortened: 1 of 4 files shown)


fsharp-backend/testfiles/execution/list.tests
```diff
@@ -186,6 +186,14 @@ List.member_v0 [1;2;3] 2 = true
 List.member_v0 [1;2;3] 4 = false
 List.member_v0 [] 1 = false
 
+List.partition_v0 [1;2;3] (fun item -> match item with | 1 -> Nothing | 2 -> false | 3 -> true) = (Test.typeError_v0 "Expected `fn` to return a Bool, but it returned `Nothing`", [])
+List.partition_v0 [true;false;true] (fun item -> "a") = (Test.typeError_v0 "Expected `fn` to return a Bool, but it returned `\"a\"`", [])
+List.partition_v0 [1;2;3] (fun item -> match item with | 1 -> true | 2 -> blank | 3 -> true) = (blank, [])
```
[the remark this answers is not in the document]
StachuDotNet: resolved

pbiggar (changes requested):
Nice.

fsharp-backend/src/LibExecutionStdLib/LibList.fs
```diff
@@ -1464,4 +1464,42 @@ let fns : List<BuiltInFn> =
...
+            let f (dv : Dval) : Ply<bool> =
+              uply {
+                let! result =
+                  Interpreter.applyFnVal state (id 0) fn [ dv ] NotInPipe NoRail
+
+                match result with
+                | DBool b -> return b
+                | (DIncomplete _
+                | DErrorRail _
+                | DError _) as dv -> return false
```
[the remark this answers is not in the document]
StachuDotNet: That makes sense!

fsharp-backend/src/LibExecutionStdLib/LibList.fs
```diff
@@ -1490,14 +1492,18 @@ let fns : List<BuiltInFn> =
                 | DBool b -> return b
                 | (DIncomplete _
                 | DErrorRail _
-                | DError _) as dv -> return false
+                | DError _) as dv ->
```
pbiggar, line 1495: I'm having a few conflicting thoughts:

1. The mechanism for returning things should be the same in both cases (fakeval and type error)
2. It should return early upon finding a faekval
3. Do we want to keep using Exceptions in stdlib code?

StachuDotNet: 1. I'm not sure what you mean here - what are the 2 'mechanisms'? Exceptions and dvals?
2. Resolved - decided to not extract out `partition` in `Prelude`
3. I think it's OK in cases like these where we see those exceptions being preventable (i.e. with type checking before we end up here)

pbiggar (approved):

fsharp-backend/src/LibExecutionStdLib/LibList.fs
pbiggar: > I'm not sure what you mean here - what are the 2 'mechanisms'? Exceptions and dvals?

Yeah, if there's a new type error than we throw. If there's an existing type error we return it. I think I'm maybe being picky here - we probably do it like this in many other places.

Issue #4489, opened 2022-09-21
Rebrand 'patterns' as 'match patterns'


StachuDotNet:
This is in preparation for #3402, where we will define and use 'let patterns'.

Generally throughout our codebase, the aim here is to make several universal renames to prevent confusion between 'let patterns' and 'match patterns':
- "Pattern" -> "MatchPattern"
  (e.g. `FluidPattern` -> `FluidMatchPattern`)
  these naming conventions are not currently serialized anywhere AFAIK.
- along with the above, update the prefixes used to name PT and RT pattern cases
  (e.g. `PString` -> `MPString`)
  these conventions _are_ serialized (to json) for client/server communication, for at least PTs (and maybe RTs?)

This rename is needed within each of `fsharp-backend` and `client`, and we need to ensure a smooth/safe transition between these naming conventions. We do not expect any issues in data within our DB, so the risk is only within API communication, as the front-end may be 'behind' the backend.

#4487 has started this effort, by completing this rebranding across the `client` codebase. It also prepares the PT and RT decoders to handle `MPString`-like match pattern cases.

From this point, I expect a few steps:
1. update backend minimally to decode `MPString`-style match pattern case serializations from `client` successfully
2. wait until users' clients are 'caught up' (no one is working on an old/incompatible client version
3. update backend _generally_ to rebrand 'patterns' as 'match patterns'
4. remove compatibility with `PString` in both backend and client, now that things have settled

(steps 2 and 3 may not have a dependency?)

pbiggar:
ClientTypes exist because we don't want to expose RuntimeTypes to the API - we want the freedom to change Runtime types as appropriate for the actual runtime of the language. (Confusingly, the client typically uses ProgramTypes everywhere, and the only place it uses RuntimeTypes is for Dval's DLambda expr field).

ProgramTypes are really supposed to be the API type. They're not executed and they're not binary serialized either, but they act as an intermediate phase between all of these.

So I think you want to convert ProgramTypes to the new names, and add an intermediate layer between it and the client for now. It's fine to leave it in, and we'll need it again in the future no doubt.

To avoid future confusion, perhaps the Dval/Expr  portion of ClientTypes should be ClientRuntimeTypes and the new thing should be APIProgramTypes? Or probably not exactly that but something along those lines?

Does that help?

StachuDotNet:
That helps, thanks! (currently on another task but will circle back and potentially have follow-ups)

PR #4622, opened 2022-12-01
Rename backend Pattern cases from PBlank to MPBlank style

StachuDotNet:
Changelog:

```
Internal
- rename backend Match Pattern cases from "PBlank"-style to "MPBlank" style in preparation for Let Patterns
```

This is part of #4489.

Match Patterns (the things we use to `match` against) were previously branded as just Patterns. In transitioning from Patterns to Match Patterns, we had [previously](#4487) updated the `client` halfway to support Match Pattern cases serializing as "MPBlank" rather than "PBlank." The types were changed, and the `decoder` was updated to handle both styles of naming.

This PR updates the backend's ClientTypes in the same way - everything is branded "Match Pattern" internally, with compatibility for both in the 'decoder.'

After this PR, the _encoders_ of both the backend and the client still use the old PVariable-style naming.

The final step here will be to remove the old-style naming convention from both client and fsharp-backend. The types, encoder/decoder, etc. can all be done in one PR in a week or so once users' clients are up to date. Several Vanilla-serialized .json files used for backend testing will need updated as part of this.


StachuDotNet:
I forgot - there are still a few serializable things not entirely in `ClientTypes` space.

Two that popped up after (only) updating the naming style of PT Patterns:
- loading a Canvas from a .json file on disk (used for integration tests only?)
- cloning a Canvas (used by F# unit tests only?)

StachuDotNet:
> loading a Canvas from a .json file on disk (used for integration tests only?)

This was indeed only used by integration tests. I updated those .json files and think it's fine to just leave them as (internal) PT types for now.

> cloning a Canvas (used by F# unit tests only?)

It turns out that we don't actually Vanilla-serialize/deserialize any "internal" types during canvas-cloning. Rather, the referenced type was only used in testing of canvas-cloning. I've adjusted the test to no longer reference the 'internal' type but to instead roundtrip the CT equivalent.

StachuDotNet:
> It might save some time if you add both MP and non-MP names in clienttypes in this PR. That will allow transitioning the types on the client in a single later deploy.

That's the plan - just taking care of internal namings first

StachuDotNet:
Both client-side and server-side decoders for the ClientTypes are prepared for `MPVariable`-style namings of these cases (in addition to the current/old-style namings).

I would go ahead and update the server-side encoder to prefer the new-style namings, but it causes the Vanilla roundtripping tests to fail.

The final step here is to remove the old-style naming convention from both `client` and `fsharp-backend`. The types, encoder/decoder, etc. can all be done in one fell swoop. Several Vanilla-serialized .json files used for backend testing will need updated as part of this.

pbiggar:
Just replaying this back so I can review:
- client can already read MP and P constructors
- client already uses MP names
- changes:
  - the backend now uses MP everywhere
  - the backend client types can now read both P and MP
- non-changes:
  - encoders both use P for now

Can you confirm my understanding?

StachuDotNet:
Yes, that's all exactly correct.

fsharp-backend/src/LibBinarySerialization/SerializedTypes.fs
```diff
@@ -91,17 +91,17 @@ module FQFnName =
...
-  | PConstructor of id * string * List<Pattern>
-  | PInteger of id * int64
-  | PBool of id * bool
-  | PCharacter of id * string
-  | PString of id * string
-  | PFloat of id * Sign * string * string
-  | PNull of id
-  | PBlank of id
-  | PTuple of id * Pattern * Pattern * List<Pattern>
+type MatchPattern =
```
pbiggar, line 94: Is renaming the type allowed?

I guess we must have tests, so it must be?

pbiggar (approved):
Nice

fsharp-backend/src/LibBinarySerialization/SerializedTypes.fs
StachuDotNet: Yes, it seems to be. The order is what matters afaict

PR #4629, opened 2022-12-05
Remove compatibility for old-style naming of Match Pattern cases

### LINKED ISSUE #4489: Rebrand 'patterns' as 'match patterns'

StachuDotNet:
Changelog:
```
Internal improvements
- Remove compatibility for old-style naming of Match Pattern cases, to prepare for Let Patterns
```

Cases are serialized as MPVariable rather than PVariable. This is the final step of an ongoing effort to rebrand Patterns as Match Patterns as we anticipate defining a Let Pattern shortly.

This will resolve #4489


StachuDotNet:
When is this appropriate to merge? A week after #4622? More/less? I don't have a good intuition here, as to when we should expects clients to be 'caught up'. A week seems pretty safe to me, if not overcautious.

pbiggar:
> 0e91e490041f06fae012f850231eb6ab

These were old OCaml files, used as part of binary serialization, and can all be removed.

pbiggar:
> When would this be appropriate to merge? A week after #4622 rolled out? More/less? I don't have a good intuition here, as to when we should expects clients to be 'caught up'. A week seems pretty safe to me, if not overcautious.

You can see the client usage in honeycomb - look at the client field of requests to ApiServer.

PR #4630, opened 2022-12-05
Remove some externals and use rescript-webapi instead

OceanOak:
This replaces most Native externals with their rescript-webapi equivalents.

### DIFF
--- a/client/src/api/APIErrorHandler.res
+++ b/client/src/api/APIErrorHandler.res
@@ -56,7 +56,7 @@ let handle = (m: AppTypes.model, apiError: t): (AppTypes.model, AppTypes.cmd) =>
     Cmd.call(_ => {
       SavedSettings.save(m)
       SavedUserSettings.save(m)
-      Native.Location.reload(true)
+      Webapi.Dom.location->Webapi.Dom.Location.reloadWithForce
     })
   } else if !ignore && APIError.shouldRollbar(apiError) {
     Cmd.call(_ => sendRollbar(m, apiError))
--- a/client/src/app/Main.res
+++ b/client/src/app/Main.res
@@ -2057,7 +2057,8 @@ let update_ = (msg: msg, m: model): modification => {
     ReplaceAllModificationsWithThisOne("HideTopbar", m => ({...m, showTopbar: false}, Cmd.none))
   | LogoutAPICallback =>
     // For some reason the Tea.Navigation.modifyUrl and .newUrl doesn't work
-    Native.Ext.redirect("/login")
+    let location = Webapi.Dom.location
+    location->Webapi.Dom.Location.replace("/login")
     NoChange
   | GoToArchitecturalView => Many(list{Deselect, MakeCmd(Url.navigateTo(Architecture))})
   | DismissErrorBar => Model.updateErrorMod(Error.clear)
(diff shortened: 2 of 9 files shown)


OceanOak:
I wasn't sure if I should use reload or reloadWithForce to replace reload.

**Native**
`@val @scope(("window", "location")) external reload: bool => unit = "reload"`

**Rescript-webapi**
```
@send external reload: t => unit = "reload"
@send external reloadWithForce: (t, @as(json`true`) _) => unit = "reload"
```

pbiggar:
> I wasn't sure if I should use reload or reloadWithForce to replace reload.
> 
> **Native** `@val @scope(("window", "location")) external reload: bool => unit = "reload"`
> 
> **Rescript-webapi**
> 
> ```
> @send external reload: t => unit = "reload"
> @send external reloadWithForce: (t, @as(json`true`) _) => unit = "reload"
> ```

Based on https://developer.mozilla.org/en-US/docs/Web/API/Location/reload, it looks like that's a non-standard parameter that only works in Firefox, so let's not use the force version.


> Also, couldn't find an equivalent for `clipboard-copy`; did I miss it, or is it under a different name?

Looks like this calls the clipboard-copy package directly. So I guess it's fine where it is (nothing to remove here)

pbiggar:
Can you edit the description to add a changelog entry? ("Internal improvements").

I've merged, but I realize I forgot to ask about testing. Did you check that everything still works in the area these changes were made? Our integration tests are not sufficient to catch bugs here unforunately.

OceanOak:
>Based on https://developer.mozilla.org/en-US/docs/Web/API/Location/reload, it looks like that's a non-standard parameter that only works in Firefox, so let's not use the force version.

So, I should make a new PR and replace reloadWithForce by reload right?

> I've merged, but I realize I forgot to ask about testing. Did you check that everything still works in the area these changes were made? Our integration tests are not sufficient to catch bugs here unforunately.

I have checked the autocomplete and commands as well as the redirect. If you can please point me to what I should check, that would be appreciated. I'm still not familiar enough with this to know exactly what I should look for.

pbiggar:
> > Based on https://developer.mozilla.org/en-US/docs/Web/API/Location/reload, it looks like that's a non-standard parameter that only works in Firefox, so let's not use the force version.
> 
> So, I should make a new PR and replace reloadWithForce by reload right?

Yes please!

> > I've merged, but I realize I forgot to ask about testing. Did you check that everything still works in the area these changes were made? Our integration tests are not sufficient to catch bugs here unforunately.
> 
> I have checked the autocomplete and commands as well as the redirect. If you can please point me to what I should check, that would be appreciated. I'm still not familiar enough with this to know exactly what I should look for.

Sure, the main this is seeing if any of the functionality below causes an error in the browser console:

- the "copy" button on live values
- the live values and anything around code still lines up properly
- commands using `Alt-x` display and work properly

OceanOak:
> Sure, the main this is seeing if any of the functionality below causes an error in the browser console:
> 
> * the "copy" button on live values
> * the live values and anything around code still lines up properly
> * commands using `Alt-x` display and work properly

All these seem to be working properly.

pbiggar:
Great, thanks!

PR #4631, opened 2022-12-05
No-op refactor of QueueWorker

StachuDotNet:
Changelog:

```
Internal
- No-op refactor of QueueWorker source
```

I've flattened `QueueWorker.fs` a bit to improve my understanding.
The core logic of `handleEvent` is still all in that fn definition.

I realize this is a bit hard to do a good 'diff' on. Given that, and potential hesitancy for refactoring here generally, what might make sense is to give [the refactored code](https://github.com/StachuDotNet/dark/blob/queueworker-redux/fsharp-backend/src/QueueWorker/QueueWorker.fs) a look in full. If it generally feels like it doesn't increase risk, I can follow up and break this down? Otherwise, I'm fine to keep the structure as it was, and just follow up by adding a few code comments.

### DIFF
--- a/docs/eventsV2.md
+++ b/docs/eventsV2.md
@@ -10,7 +10,7 @@ Goals:
 
 ## Definitions
 
-- `emit` adding an event to the queue from a Cron handler or the `emit` function
+- `emit`: adding an event to the queue from a Cron handler or the `emit` function
 - `event`: refers to the Dark value `emit`ed into the queue and the metadata around
   it (including what you might call a "job" or "message" in other systems). We say we
   "run the event" when we execute it, or "retry the event" if we requeue it.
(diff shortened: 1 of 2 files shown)


StachuDotNet:
There's a long-existing comment of:
> The algorithm below is annotated with names from the chart. `dequeueAndProcess` will block until it...
that's been hanging around the top of `processNotification` for a while.

There's no actual fn named `dequeueAndProcess`. So, I'm wondering wondering if the comment should refer to `processNotification` now? It seems to actually refer to `EQ.dequeue`, which is in `run` - that makes me think I should rephrase to "`EQ.dequeue` will block until ..."

StachuDotNet:
Wow, thanks tests. I guess this wasn't no-op :scream: 
The main point of this exercise was to learn what's going on, so I'm going to create a separate PR of smaller commits mostly focused on commentary, and follow up there. Closing this, but I'll be happy to take comments about the `dequeueAndProcess`+`EQ.dequeue` question above, here.

PR #4638, opened 2022-12-09
appsupport.js desktop support check now case-insensitve

StachuDotNet:
No changelog

### DIFF
--- a/client/src/appsupport.js
+++ b/client/src/appsupport.js
@@ -20,7 +20,7 @@ function unsupportedBrowser() {
   var isMobile = /Android|BlackBerry|iPhone|iPad|iPod|Opera Mini|IEMobile/.test(
     navigator.userAgent,
   );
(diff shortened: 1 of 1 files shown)


client/src/appsupport.js
```diff
@@ -20,7 +20,7 @@ function unsupportedBrowser() {
   var isMobile = /Android|BlackBerry|iPhone|iPad|iPod|Opera Mini|IEMobile/.test(
     navigator.userAgent,
   );
-  var isDesktopApp = /DarkLang\/Editor/.test(navigator.userAgent);
+  var isDesktopApp = /DarkLang\/Editor/i.test(navigator.userAgent);
```
ajhalili2006, line 23: Instead of doing checks based on user agent (which can be easily changed through DevTools), maybe do progressive enchantment or feature detection instead?

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent for the notes.

PR #4639, opened 2022-12-09
Merge F# backend into backend dir

StachuDotNet:
Changelog:

```
Internal
- merge F# backend code (in `fsharp-backend`) into `backend` dir
```

The first commit here is me pulling everything committed in `fsharp-backend` to the `backend` directory, and fixing things until everything worked in CI. It also updates some readmes to not reference the old folder name. This involves a large # of files changed at once, largely due to our many testfiles that moved from `fsharp-backend/testfiles` to `backend/testfiles`

Subsequent commits:
- rename scripts, script functions, etc. to rebrand from `fsharp-backend-x` and `fsharp-x` to just `backend-x`
- remove some old scripts no longer relevant (used by old OCaml backend), and references to scripts which have been previously removed

There are a still a few places where the `fsharp-` prefixes are still hanging around. Namely, I didn't touch anything related to infra (e.g. `containers` and `services` directories have been untouched). Let's hold off on those for a subsequent PR?

I'm creating a PR in `docs` alongside this, to account for the changes.

---

At the end of these changes, the `backend` directory has both `test_appdata` and `testfiles` directories, as well as the `serialization` directory. I think this is a bit messy. I propose that we migrate the contents of `test_appdata` and `serialization` into `testfiles`. What do you think?

---

While going through this venture, I found a few environment variables set in `config/..` that I think we could remove:
- `DARK_CONFIG_BIN_ROOT_DIR`
- `DARK_CONFIG_SWAGGER_DIR`
- `DARK_CONFIG_HTTP_PORT`
- `DARK_CONFIG_HEALTH_CHECK_PORT`
What do you think? Is there any reason these should hang around?
There's no rush here, so we could hold off here until a subsequent PR.
I did try removing these all in one commit, and it broke the F# build somewhow - I haven't yet investigated further.


StachuDotNet:
There are a few config values I think we're able to remove. What do you think?
- `DARK_CONFIG_BIN_ROOT_DIR`
- `DARK_CONFIG_SWAGGER_DIR`
- `DARK_CONFIG_HTTP_PORT`
- `DARK_CONFIG_HEALTH_CHECK_PORT`

Is there any risk in removing these variables? I think not, but being cautious.

StachuDotNet:
Actually, using git diff locally while ignoring renames is probably the most efficient way to review the first commit. Then just review the remaining ones in github's UI normally?

.circleci/config.yml
```diff
@@ -319,7 +319,7 @@ jobs:
       - show-large-files-and-directories
       - save_cache:
           paths:
-            - fsharp-backend/Build/obj
+            - backend/Build/obj
```
pbiggar, line 322: Would be worth changing the cache number to prevent CI problems for other branches.

Maybe remove fsharp from the cache name and the CI job name.

pbiggar (changes requested):
I'd leave the config for another day, the rest looks solid, nice job.

.circleci/config.yml
StachuDotNet: I removed `fsharp` from the job and cache names - no issue there.

I updated the cache keys:
- from `v5-fsharp-blazor` to `v1-blazor`. This feels totally safe, as I don't think we've used that before.
- from `v15-fsharp-backend` to `v1-backend`. This feels slightly less safe because we've potentially used the key start before. That said, the key also includes the `checksum` after that, so almost definitely just paranoid here.

PR #4645, opened 2022-12-15
Include dark client version in apiserver rollbars

StachuDotNet:
Changelog:

```
Internal
- include dark client version in apiserver rollbars
```

image

- [ ] it's a bit long; maybe we should truncate to ~6chars

### DIFF
--- a/backend/src/ApiServer/ApiServer.fs
+++ b/backend/src/ApiServer/ApiServer.fs
@@ -176,6 +176,7 @@ let rollbarCtxToMetadata (ctx : HttpContext) : (Rollbar.Person * Metadata) =
   let clientVersion =
     try
       string ctx.Request.Headers["x-darklang-client-version"]
+      |> String.take 7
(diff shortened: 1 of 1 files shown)


pbiggar (approved):

pbiggar:
Could shorten it I guess. Could also be a link maybe (or add another link field)

backend/src/ApiServer/ApiServer.fs
```diff
@@ -176,6 +176,7 @@ let rollbarCtxToMetadata (ctx : HttpContext) : (Rollbar.Person * Metadata) =
   let clientVersion =
     try
       string ctx.Request.Headers["x-darklang-client-version"]
+      |> String.take 7
```
StachuDotNet, line 179: I've tested this ensuring it's OK when such a header doesn't exist (i.e. was cautious of a null reference exception).

PR #4649, opened 2022-12-16
Add DarkInternal::serverBuildHash fn that returns the git hash of the server's current deploy

StachuDotNet:
Changelog:

```
Internal
- add internal function that returns the git hash of the server's current/latest deploy
```

This is to support #4648 - with this fn exposed, an endpoint in `dark-editor` may be defined to expose the current deploy version, which the `client` may then call out to, in order to see if it's at the latest build.

Questions for review:
- is `LibService.Config.buildHash` something we can/should rely on? Locally, it returns just `"dev"` but I understand it will contain the actual hash outside of local development.
- should the function instead be `::buildHash` or `::buildVersion` or `::latestBuild` or something else?
- I haven't written tests around this. I could test that it returns `"dev"`? Not sure what's best here.

### DIFF
--- a/backend/testfiles/execution/internal.tests
+++ b/backend/testfiles/execution/internal.tests
@@ -3,8 +3,22 @@
 // ---------------
 // Misc
 // ---------------
+[tests.misc]
+
+[test.correct number of tables]
 Dict.size_v0 DarkInternal.getAndLogTableSizes_v0 = 24
+
+[test.allFunctions has many functions]
 (List.length_v0 DarkInternal.allFunctions_v0 > 290) = true
+
+[test.server build hash]
+(match DarkInternal.serverBuildHash_v0 with
+ // in local dev, the value is "dev"
+ | "dev" -> true
+ // in ci, "circleci"
+ | "circleci" -> true
+ // otherwise it's the first 7 chars of the git hash
+ | hash -> (String.length hash) = 7)
 
 // ---------------
 // Grants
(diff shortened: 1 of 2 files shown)


pbiggar (approved):
Looks good. A test that it returns "dev" is probably better than no test, yeah.

backend/testfiles/execution/internal.tests
```diff
@@ -3,9 +3,21 @@
 // ---------------
 // Misc
 // ---------------
+[tests.misc]
```
StachuDotNet, line 6: Because this test turned out to be more than one line, I was sort of forced to add a [tests.] group and the [test] heading to the other 'misc' tests. Am I missing some way around this?

backend/testfiles/execution/internal.tests
```diff
@@ -3,9 +3,23 @@
...
 (List.length_v0 DarkInternal.allFunctions_v0 > 290) = true
 
+[test.server build hash]
+(match DarkInternal.serverBuildHash_v0 with
+ // in local dev, the value is "dev"
+ | "dev" -> true
+ // in ci, "circleci"
+ | "circleci" -> true
+ // otherwise it's the first 7 chars of the git hash
+ | hash -> (String.length hash) = 7)
```
StachuDotNet, line 21: This test feels a little bit silly currently. I was surprised circleci didn't have context of the git hash.

backend/testfiles/execution/internal.tests
pbiggar: That's how you do it

PR #4650, opened 2022-12-18
Add healthchecks to ensure we can deserialize canvases

pbiggar:
Changelog:

```
Internal
- Check for serialization errors before starting up servers
```

This adds a serialization check at startup. That is, do not start up if we cannot successfully deserialize 5 important canvases.

I made the list of canvases a config variable to allow us flexibility. Perhaps we break one of these canvases - we don't want to be unable to restart the servers or scale up or whatever.

Not added to cronchecker as it isn't needed and also its deployment pattern (delete old server, start new one) makes this unsafe.


StachuDotNet (approved):
Great.

We have several test canvases that we have in static files (currently in `backend/test_appdata`).
Could/should we add those to the configs of non-prod environments? Many of them will be caught in integration tests - maybe `sample-gettingstarted` is one to add in all configs? (since it's used in prod as well as the ones currently listed)

pbiggar:
> Great.
> 
> We have several test canvases that we have in static files (currently in `backend/test_appdata`). Could/should we add those to the configs of non-prod environments? Many of them will be caught in integration tests - maybe `sample-gettingstarted` is one to add in all configs? (since it's used in prod as well as the ones currently listed)

I think this doesn't bring us much because I'm worried about old data in the binary serialized format in production. If we created this data for testing, we presumably wouldn't be saving this in the old format.

Also feels like it'll be tricky to set up.

PR #4655, opened 2022-12-21
Iterate on and test HttpBaseClient

StachuDotNet:
Changelog:

```
Language and Standard Library
- add `Bytes::empty`

Internal
- iterate on `HttpBaseClient::request`
```

This iterates on `HttpBaseClient::request`. It adds some tests, but they've been disabled for now.


pbiggar:
> when we've been hiding functions behind 'contributor settings' so far, we did so with pretty good confidence we'd move forward. The question was more about timing. I think a lot of things will change before we fully avail this function, though. I'd really like a way of telling users "please don't use this in anything real. We reserve the right to break your code if you do."

I don't think we want to change functions we put out there - that's what function versioning is for. But we also don't want to get stuck maintaining code (or even having code in the codebase) that we'd rather doesn't exist.

Remind me what will change with this?

StachuDotNet:
> > when we've been hiding functions behind 'contributor settings' so far, we did so with pretty good confidence we'd move forward. The question was more about timing. I think a lot of things will change before we fully avail this function, though. I'd really like a way of telling users "please don't use this in anything real. We reserve the right to break your code if you do."
> 
> I don't think we want to change functions we put out there - that's what function versioning is for. But we also don't want to get stuck maintaining code (or even having code in the codebase) that we'd rather doesn't exist.
> 
> Remind me what will change with this?

Currently, it's a function that only "internal" users can use. So, I figured we wouldn't care if we later removed it (as we'd be cautious to not rely on it).

With this change, the function isn't quite "public" but once someone opts in there's nothing preventing them from relying on it.

backend/src/BackendOnlyStdLib/LibHttpBaseClient.fs
```diff
@@ -149,54 +159,39 @@ module HttpBaseClient =
...
-
-        return
-          [ ("body", DBytes response.body)
-            ("headers", responseHeaders)
-            ("code", DInt(int64 response.code)) ]
-          |> Dval.obj
-          |> Ok
-          |> DResult
-      | Error err -> return DResult(Error(DStr err.error))
+        return Error(Other(Exception.getMessages e |> String.concat " "))
```
StachuDotNet, line 194: The existing internal `HttpClient` does an exception type check for:
`| :? HttpRequestException as e ->`.

This is a case where we can't return an "OK" result. Naturally, the other option is to return an Error case - but since we _have_ a status code in some cases, it feels a bit inappropriate to just "drop" the status code and not return it to the user. What do you think?

pbiggar: Maybe `Other` should have a status code?

pbiggar: An optional one?

backend/src/Prelude/Prelude.fs
```diff
@@ -26,6 +26,13 @@ let failwith = failwith
                            IsHidden = true)>]
 let printf = printf
 
+// is this useful?
+[<CompilerMessageAttribute("printfn is banned, use Prelude.print instead",
+                           0,
+                           IsError = true,
+                           IsHidden = true)>]
+let printfn = printfn
```
StachuDotNet, line 33: misc - I noticed that while I couldn't `printf` when debugging, I could still use `printfn`

backend/testfiles/execution/bytes.tests
```diff
@@ -1,6 +1,9 @@
 
 // TODO: try decoding values with incorrect padding
 
+Bytes.length_v0 Bytes.empty_v0 = 0
+Bytes.empty_v0 = String.toBytes_v0 ""
```
StachuDotNet, line 5: I couldn't think of any other tests for `Bytes::empty`. ideas?

StachuDotNet:
I was originally intending to avail the new HttpClient's stdlib fn to users, but we're not quite there.
In any case, this pushes us forward a bit:
- defines `Bytes::empty` which is useful
- starts to test `HttpBaseClient::request`
  - yielded a few fixes and refactors in impl.

I've tried to tidy the commit history a bit (via `get rebase -i`), but let me know if the history is currently confusing - I can take anothe rpass.

regarding merge-ability:
- is it fine that `HttpBaseClient.Tests.fs` is separate from `HttpClient.Tests.fs`, for now, and that their webservers run on different ports (incl. a new port in `config`)? If not, I think I could consolidate them to one test module I think.
- is it reasonable that `LibExecution.Tests.fs` has a special case of running the `httpbaseclient.tests` file 'as admin'? (feels a bit 'off')
- is the `backend/testfiles/httpbaseclient/v0/todos` directory too noisy to merge in?

Issue #4663, opened 2022-12-24
Switch BwdServer to Cloud Run


pbiggar:
- [x] get bwdserver standing up
  - [x] yaml
  - [x] envvars
  - [x] secrets
- [x] come up with testing plan
- [ ] get it deployed
  - [x] ensure it's not world accessible (internal only)
  - [x] deploy via terraform
    - [x] create terraform file for deployment
    - [x] create terraform file for envvars
    - [x] add terraform to docker file
    - [x] add bash script to run terraform with the deploy vars in question
  - [x] handle envvar situation
  - [ ] add deployment via CI
  - [x] check whether shipit is appropriate
  - [x] merge 
- [ ] make production ready
  - [x] should we be using 2nd generation?
  - [x] add startup CPU boost
  - [ ] ensure performance is up to scratch
    - [ ] SQL statements
      - [ ] see if connecting to google cloud private IP is faster
      - [ ] on hold for now since we have to have some downtime to add a private IP. Will see where things sit a little later
      - [ ] ask GCP folks about faster internal tools
    - [ ] compute
      - [ ] run benchmarks with FizzBuzz to measure performance
    - [ ] httpclient
      - [ ] test default latency matches GKE version
  - [ ] disable access to metadata
    - [ ] test resolving DNS to exclude internal metadata
    - [ ] or look at restricting on the socket connection level
    - [ ] ask cloud run or support on best approach
  - [ ] use healthchecks and startupcheck
  - [ ] limits/resources
  - [ ] check it works
    - [ ] Check [this issue](https://cloud.google.com/run/docs/issues#ah)
    - [ ] ensure rollbars arrive
    - [ ] ensure pubsub events arrive
    - [ ] ensute pushers arrives
    - [ ] ensure honeycomb events arrive
    - [ ] check that it works (check http params (path, headers, url) arrive ok)
    - [ ] check httpclient works
    - [ ] check performance is similar/the same (within a few ms)
- [ ] expose publicly
  - [ ] load balancer
    - [ ] cloud armor
  - [ ] bwdserver cert
  - [ ] check it works
    - [ ] check that managed certs work
    - [ ] use separate target proxy from bwdserver
    - [ ] check that we are able to get a HTTP request instantly once deployed (no 502s in between)
- [ ] figure out deployment/migration strategy (https://cloud.google.com/run/docs/migrate/from-kubernetes#strategy)
- [ ] migrate cert-manager certs
- [ ] switch over DNS

optimizations:
[...]

PR #4668, opened 2022-12-27
Add Cloud Run

pbiggar:
No changelog (changes aren't effected yet)

We'd like to move away from kubernetes, and Google Cloud Run is a great option for that:
- does everything we need with minimal config
- only bills when CPU is in use
- autoscales
- no control plane to manage/upgrade

This is the start of the setting it up - I have it working in practice with these settings, but still more testing and stuff before it's ready to put it into production. This doesn't do any deployment, and users can't reach this.

- adds terraform
- adds terraform config to run bwdserver in cloud run
- creates an file with env vars and secrets that should replace config/gke-builtwithdark (can be used for queues as well)

This should be safe to merge as it's not actually connected to anything userfacing.

Remaining work is in #4663.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ executors:
       IN_DEV_CONTAINER: true
     docker:
       # DOCKERFILE_REPO: see Dockerfile note about how this is built.
-      - image: darklang/dark-base:89294ed
+      - image: darklang/dark-base:f68d6ec
 
 commands:
   show-large-files-and-directories:
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -27,7 +27,8 @@
     "ms-python.python",
     "ms-azuretools.vscode-docker",
     "editorconfig.editorconfig",
-    "chenglou92.rescript-vscode@1.1.3"
+    "chenglou92.rescript-vscode@1.1.3",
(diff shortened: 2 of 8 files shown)


StachuDotNet:
A lot of config vals match environment variables that we've set elsewhere. Is there a plan to remove the duplication of these?

As far as editing the .tf files goes, are you using any sort of vs code extension that prevents bad edits, or helps in any way?

StachuDotNet:
also, a basic question:
`tf/.terraform.lock.hcl` is generated by the dependencies referenced in `tf/main.tf`, correct?

pbiggar:
> also, a basic question:
> `tf/.terraform.lock.hcl` is generated by the dependencies referenced in `tf/main.tf`, correct?

That's my understanding.

pbiggar:
> A lot of config vals match environment variables that we've set elsewhere. Is there a plan to remove the duplication of these?

Yes, we should be able to remove the other ones. I wrote them this way to allow that.

It may be also possible to parse a file to create these, but I haven't done that as it seemed a bit much for now.

> As far as editing the .tf files goes, are you using any sort of vs code extension that prevents bad edits, or helps in any way?

https://marketplace.visualstudio.com/items?itemName=HashiCorp.terraform

I added this to recommended set in devcontainer.json

StachuDotNet (approved):

PR #4671, opened 2022-12-28
Rename 'Analysis' fsproj to 'Wasm'

StachuDotNet:
```
Internal
- rename "Analysis" F# project to "Wasm"
```

This updates the project name, and its resultant assembly, from "Analysis" to "Wasm."


pbiggar (approved):
> This does not change the namespace of Analysis.fs - it's still "Analysis." This makes sense to me given our so-far single use, but I'm open to renaming the namespace to Wasm as well.

Keeping module names matching filenames (I know we don't do this everywhere but probably should), seems to make more sense to me.

Feel free to do `Wasm.Analysis` if you think we'll have more Wasm in the future but that might be overkill for now.

Issue #4644, opened 2022-12-13
Followups to EInfix (4640)


pbiggar:
Followups

- [x] test on the client with flag enabled
- [ ] ~ask users to test it~
- [ ] ~wait a few days~
- [x] allow users to switch
  - [x] add refactoring command to change old version to new
  - [x] make clear to users that old BinOp `&&` and `||` are deprecated
    - [x] docs should indicate it
    - [x] not in autocomplete by default
  - [x] remove experimental flag
- [x] remove old types
  - [x] wait until old clients out of circulation
  - [x] remove EBinOp from server-side/client-side ClientTypes

pbiggar:
> I'm concerned that "short-circuiting" is not a term known to all; likely worth adding some additional explanation in both the command and DeprecatedBecause of the old fn. Unfortunately I don't have better wording than "short-circuiting" (seems better than 'minimal evaluation') - more so just recommending some redundant explanation

> I can't help but think that the Commands aren't really accessible, so users may be confused by the Deprecated notice telling them to go use a command. Maybe we can make this easier by 'linking' to the command from the DeprecatedBecause?

- [ ] clarify "short-circuiting" and find better way to explain the semantics of these operators
- [ ] make Command easier to use somehow

pbiggar:
No longer relevant in new direction

PR #4674, opened 2022-12-30
Remove EBinOp

pbiggar:
No Changelog

Remove EBinOp completely (server side EInfix followup, see #4644)

Before this PR, EBinOp still exists
- client can read but does not send EBinOp
- server can read but does not send EBinOp
- client/server send EInfix instead 
- serializers can read EInfix and EDeprecatedBinOp, and save as EDeprecatedBinOp

After this PR, EBinOp is gone
- client can no longer read EBinOp
- server can no longer read EBinOp
- serializers can read EInfix and EDeprecatedBinOp, but they are saved as EInfix


backend/src/ApiServer/IntegrationTests.fs
```diff
@@ -79,10 +80,14 @@ let loadAndResaveFromTestFile (meta : Meta) : Task<unit> =
   task {
     let oplists =
       let tls = meta |> tryLoadJsonFromDisk LibBackend.Config.Testdata
+      // Save back in the latest format
+      do tls |> Option.tap (saveJsonToDisk LibBackend.Config.Testdata meta)
```
StachuDotNet, line 84: To make sure I understand:
- every time we run an integration test, we've been loading them from disk
- this adds another operation where we re-save the oplist to disk, such that:
  - short term, we got rid of the EBinOp usages in the oplists (in a previous commit)
  - long term, we detect other changes similar to how we detect serialization changes with Vanilla
  - 
?

StachuDotNet (approved):

backend/src/ApiServer/IntegrationTests.fs
pbiggar: Yes

PR #4676, opened 2023-01-02
Use new short-circuiting and/or

pbiggar:
Changelog:

```
Editor
- default to short-circuiting `||`/`&&`
- deprecate old `||`/`&&` functions
- add `convert-to-short-circuiting` command to migrate from old `||`\`&&` functions
```

Removes the old settings, defaults to the new expressions, deprecates old expressions, and updates docs/deprecation settings. Also adds a command to convert automatically.

Deprecated old version
Screenshot 2023-01-02 at 6 22 30 PM

Doc for new version
Screenshot 2023-01-02 at 6 22 40 PM

I'm not super happy with the docstring or the name `convert-to-short-circuiting`. Advice appreciated.

### DIFF
--- a/backend/src/LibExecutionStdLib/StdLib.fs
+++ b/backend/src/LibExecutionStdLib/StdLib.fs
@@ -76,11 +76,16 @@ let infixFnMapping : Map<FQFnName.StdlibFnName, (FQFnName.StdlibFnName * Depreca
     ("String", "append", 1), (("", "++"), NotDeprecated)
     ("", "equals", 0), (("", "=="), NotDeprecated)
     ("", "notEquals", 0), (("", "!="), NotDeprecated)
-    // TODO: deprecate when there's a nice way to move from infix fn to language version
-    // ("Bool", "and", 0), (("", "&&"), DeprecatedBecause("Use `&&` instead"))
-    // ("Bool", "or", 0), (("", "||"), DeprecatedBecause("Use `||` instead"))
-    ("Bool", "and", 0), (("", "&&"), NotDeprecated)
-    ("Bool", "or", 0), (("", "||"), NotDeprecated) ]
+    ("Bool", "and", 0),
+    (("", "&&"),
+     DeprecatedBecause(
+       "Use built-in keyword `&&` instead. Use the `convert-to-short-circuiting` command to convert automatically"
+     ))
+    ("Bool", "or", 0),
+    (("", "||"),
+     DeprecatedBecause(
+       "Use built-in keyword `&&` instead. Use the `convert-to-short-circuiting` command to convert automatically"
+     )) ]
   |> List.map (fun ((module_, name, version), ((newMod, opName), deprecation)) ->
     FQFnName.stdlibFnName module_ name version,
     (FQFnName.stdlibFnName newMod opName 0, deprecation))
--- a/backend/src/LibExecutionStdLib/StdLib.fs
+++ b/backend/src/LibExecutionStdLib/StdLib.fs
@@ -84,7 +84,7 @@ let infixFnMapping : Map<FQFnName.StdlibFnName, (FQFnName.StdlibFnName * Depreca
     ("Bool", "or", 0),
     (("", "||"),
      DeprecatedBecause(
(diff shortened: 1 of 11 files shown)


backend/src/LibExecutionStdLib/StdLib.fs
```diff
@@ -76,11 +76,16 @@ let infixFnMapping : Map<FQFnName.StdlibFnName, (FQFnName.StdlibFnName * Depreca
...
-    ("Bool", "or", 0), (("", "||"), NotDeprecated) ]
+    ("Bool", "and", 0),
+    (("", "&&"),
+     DeprecatedBecause(
+       "Use built-in keyword `&&` instead. Use the `convert-to-short-circuiting` command to convert automatically"
+     ))
+    ("Bool", "or", 0),
+    (("", "||"),
+     DeprecatedBecause(
+       "Use built-in keyword `&&` instead. Use the `convert-to-short-circuiting` command to convert automatically"
```
StachuDotNet, line 87:
```suggestion
    ("Bool", "or", 0),
    (("", "||"),
     DeprecatedBecause(
       "Use built-in keyword `||` instead. Use the `convert-to-short-circuiting` command to convert automatically"
```
typo

backend/src/LibExecutionStdLib/StdLib.fs
```diff
@@ -76,11 +76,16 @@ let infixFnMapping : Map<FQFnName.StdlibFnName, (FQFnName.StdlibFnName * Depreca
...
     ("", "notEquals", 0), (("", "!="), NotDeprecated)
-    // TODO: deprecate when there's a nice way to move from infix fn to language version
-    // ("Bool", "and", 0), (("", "&&"), DeprecatedBecause("Use `&&` instead"))
-    // ("Bool", "or", 0), (("", "||"), DeprecatedBecause("Use `||` instead"))
-    ("Bool", "and", 0), (("", "&&"), NotDeprecated)
-    ("Bool", "or", 0), (("", "||"), NotDeprecated) ]
+    ("Bool", "and", 0),
+    (("", "&&"),
+     DeprecatedBecause(
+       "Use built-in keyword `&&` instead. Use the `convert-to-short-circuiting` command to convert automatically"
```
StachuDotNet, line 82: Does the average Dark user know how to 'Use [a] command'? It's never felt very discoverable to me, so I could see a user getting confused by this and not knowing what to do.

I wonder if we could include another field in `DeprecatedBecause` that would allow us to 'link' a Command that would show up when the fn is highlighted. I'm imagining it shows up in the info box, with some link that brings up the command dialog, with that option highlighted.

Maybe overkill?

StachuDotNet, line 82: I think we should call out what "short-circuiting" is _here_ (as well as in the command) - I'm not sure if "short-circuiting" is a known term to the majority of programmers.

StachuDotNet (changes requested):
Functionality looks good, but:
- I'm concerned that "short-circuiting" is not a term known to all; likely worth adding some additional explanation in both the command and `DeprecatedBecause` of the old fn. Unfortunately I don't have _better_ wording than "short-circuiting" (seems better than 'minimal evaluation') - more so just recommending some redundant explanation
- I can't help but think that the Commands aren't really accessible, so users may be confused by the Deprecated notice telling them to go use a command. Maybe we can make this easier by 'linking' to the command from the DeprecatedBecause?

backend/src/LibExecutionStdLib/StdLib.fs
pbiggar: I have intent to followup on this sort of thing, but I think this is the best we can do without increasing the scope of the PR a lot.

pbiggar: I do think we should do better to explain this concept and others, but I'd like to wait for the doc box project to finish before we iterate on this, so we can think a little bigger.

pbiggar:
Some good suggestions, but I think ones that are pretty hard to solve right now. I marked them to followup: https://github.com/darklang/dark/issues/4644#issuecomment-1370245990. Does that work for you?

StachuDotNet:
Yeah that's fine :+1:

StachuDotNet (approved):

PR #4693, opened 2023-01-11
Upgrade to .NET 7

StachuDotNet:
Changelog:

```
Internal
- upgrade backend to .NET 7
```

So far, I've seen two issues with this upgrade:
- An error occurs when calling `LibService.Telemetry.flush()`
  This occurred during the cleanup after we run migrations. 
  The specific error is `Attempted to read or write protected memory`
  I've disabled the flushing process for now.
- An error occurs when we try to load Blazor
  The specific error is:
  `DOMException: Failed to execute 'importScripts' on 'WorkerGlobalScope': The script at 'http://static.darklang.localhost:9000//blazor/dotnet.7.0.2.jfexfn3ogw.js' failed to load. '\nNote that errors related to the following exceptions can be ignored. They do no seem to cause an issue, and we believe they will go away when we upgrade to .NET 7.' " - 'System.ArgumentNullException: Value cannot be null'" " - 'System.Runtime.InteropServices.JavaScript.Runtime.ReleaseJSOwnedObjectByGCHandle(Int32 gcHandle)'"`
  I've disabled the automatic load of Blazor for now.


StachuDotNet:
I've been wondering: what's the difference between `aspnetcore`'s official `blazor.webassembly.js` for .net 6, and for .net 7?
Here's the diff: https://github.com/StachuDotNet/dotnet-7-blazorworker/commit/262c0b8ae710ac0ce09c054dc98de7bb15bbae88

I haven't yet reviewed the diff, but I've been hoping that it will show us what to change in our BlazorWorker.
If not, my plan is to to procure a similar diff for `dotnet.js`, since that's the script that actually failed in runtime.

StachuDotNet:
Thanks for the hint to apply prettier to the .net 6 and .net 7 blazor.webassembly.js versions - here's an updated diff: https://github.com/StachuDotNet/dotnet-7-blazorworker/commit/7e5141e913865bf04b1d061ae9ebaaecac0ef1e5

There are some parts of this diff that seem promising, related to loading dotnet.js, but I haven't finished analysis yet.

StachuDotNet:
Regarding Blazor work:
- The JS -> .NET communication works great at this point, and is actually tidier than before
- the .NET -> JS communication strategy that we used previously has been removed from .NET. I've been trying a few other strategies, and think I'm close to figuring it out.
- I'm currently still working on (a fork of) the official blazor.webassembly.js, compiled locally from `aspnetcore` sources after removing some lines of their [`Boot.WebAssembly.ts`](https://github.com/dotnet/aspnetcore/blob/v7.0.2/src/Components/Web.JS/src/Boot.WebAssembly.ts) related to rendering

Assuming I get the .NET -> JS communication sorted, I'll then jump back to updating our BlazorWorker appropriately - but the .NET -> JS communciation is more of a blocker at this point.

StachuDotNet:
FWIW I also spent a bit of time trying to use the static `JSHost` type exposed in `System.Runtime.InteropServices.JavaScript` (https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.javascript.jshost?view=net-7.0)

It exposes a `GlobalThis` property, so I was hoping I could reference `this.self.postMessage` or something there. While I'm able to traverse the `this` object, and _see_ things on `self`, I was unable to _call_ any fn.

```
// let typ : System.Type =
    //   let sourceAssembly : Assembly =
    //     Assembly.Load "System.Runtime.InteropServices.JavaScript"
    //   sourceAssembly.GetType "System.Runtime.InteropServices.JavaScript.JSHost"

    // System.Console.WriteLine($"type {typ}")
    // let globalThis =
    //   typ.GetProperty("GlobalThis").GetValue(typ, null)
    //   :?> System.Runtime.InteropServices.JavaScript.JSObject

    // System.Console.WriteLine($"globalThis {globalThis}")

    // let self = globalThis.GetPropertyAsJSObject("self") //.GetValue(globalThis, null)
    // //:> System.Runtime.InteropServices.JavaScript.JSObject
    // System.Console.WriteLine($"self {self}")

    // let postMessage = self.GetTypeOfProperty("postMessage") // function
    // System.Console.WriteLine($"postMessage {postMessage}")
```

I'll read more into the `JSImport`/`JSExport` docs - that seems close to `JSHost`, so that may fill in some gaps.

StachuDotNet:
Maybe the `protected` member is trimmed or something, somehow? (maybe we're able to explicitly _not_ trim it).

StachuDotNet:
https://github.com/LostBeard/SpawnDev.BlazorJS looks a bit promising

StachuDotNet:
To expand upon that link^

If it works for us, we we won't have to do our `postMesasge`/`onMessage` dance, or otherwise manage a BlazorWorker, which would be great. The library supports all sorts of _Streams_, too.
Timeboxing a tiny spike to try this out, on `main` (outside of .net 7 PR).

The main concern for me is if this will support AOT builds - iirc https://github.com/Tewr/BlazorWorker had issues with such.

StachuDotNet:
I couldn't get that SpawnDev project to work in .NET 6. For some reason (even after clearing build, docker container, etc.) paket seemed to keep tryign to download the .net 7 version?

In any case, I then started on using it within the .NET 7 branches - it's actually really nice to be able to 'call JS' from .NET really succinctly, e.g.:
```fsharp
let innerHeight = JS.Get<int> "window.innerHeight"
JS.Set("document.title", "Hello World!")
```

Still some work to go, but this is vaguely promising.

The library in question is certainly immature, and I had to do some awkward hacking to get it to work.
I'm not sure if it'll be worth using directly, or forking for our use, or just stealing whatever's making their `JS` work so seamlessly.

StachuDotNet:
This is the new magic to get us to `this`/`self`:
```fsharp
let jsRuntimeThis : IJSInProcessRuntime =
      let assemblyName = "Microsoft.AspNetCore.Components.WebAssembly"
      let typeName =
        "Microsoft.AspNetCore.Components.WebAssembly.Services.DefaultWebAssemblyJSRuntime"

      let assembly = Assembly.Load assemblyName
      let jsRuntimeType = assembly.GetType typeName
      let jsRuntimeTypeInstance =
        let flags =
          System.Reflection.BindingFlags.NonPublic
          ||| System.Reflection.BindingFlags.Static
        jsRuntimeType.GetField("Instance", flags)
      jsRuntimeTypeInstance.GetValue(null) :?> IJSInProcessRuntime
```

I'm abandoning the pursuit of `SpawnDev.BlazorJS` - it's way too bloated for our needs.

In any case, `SpawnDev.BlazorJS` taught me quite a bit about what's possible with .NET WASM + Web Workers. I think we may end up being able to do such sooner than later, but I'm leaving that venture for another day.

StachuDotNet:
I just completed a little spike trying to use `WebAssemblyHostBuilder` for running Analysis in a web worker.
This is something I saw while going through SpawnDev's solution.

Somehow, I thought this might help us manage running some wasm'd operations in a background thread, without having to update our BlazorWorker.
But, it turns out that SpawnDev also has a hacky way of loading Blazor in a WebWorker, equivalent to our BlazorWorker - see [here](https://github.com/LostBeard/SpawnDev.BlazorJS/blob/main/SpawnDev.BlazorJS.WebWorkers/wwwroot/spawndev.blazorjs.webworkers.serviceworker.faux-env.js) (which sets up a 'faux' environment with a bogus `window` object, etc) and [here](https://github.com/LostBeard/SpawnDev.BlazorJS/blob/main/SpawnDev.BlazorJS.WebWorkers/wwwroot/spawndev.blazorjs.webworkers.js) (which uses that bogus environment to load Blazor in, like our web worker).

So, I'll abandon trying to use `WebAssemblyHostBuilder`.

The next step here is back to porting a BlazorWorker.

Then the only known missing piece for this PR will be to research and resolve the OpenTelemetry runtime error mentioned early in this thread.

pbiggar:
With the new plan, I'm thinking it feels like running blazorworker in a browser webworker is unlikely to be in our critical path in the near future.

StachuDotNet:
> With the new plan, I'm thinking it feels like running blazorworker in a browser webworker is unlikely to be in our critical path in the near future.

What do you think we should do, then?
I'm nervous that if we merge this effectively as-is, the UI performance hit on some users will be very damning.

I see the following options:
1. Leave this PR hanging for some time and then circle back
2. Merge this basically as-is (after updating the dockerfile repo, and following up here so CI passes), leaving some users potentially in a bad spot
3. Pursue this to its end (incl blazorworker in webworker)
4. Abandon this PR in favor of a simpler .net 6 upgrade

pbiggar:
> Abandon this PR in favor of a simpler .net 6 upgrade

If there's security concerns, maybe let's limit it to this. I don't think we'll get anything out of the others for now.

PR #4698, opened 2023-01-11
Update playwright

pbiggar:
No Changelog

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,14 +13,21 @@ updates:
     interval: daily
   open-pull-requests-limit: 10
 
-- package-ecosystem: npm
-  directory: "/integration-tests"
-  schedule:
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -13,14 +13,21 @@ updates:
     interval: daily
   open-pull-requests-limit: 10
 
-- package-ecosystem: npm
-  directory: "/integration-tests"
+# The playwright packages need to be updated together and dependabot doesn't support that yet
```
StachuDotNet, line 16: What do you mean by 'need to be updated together'?

pbiggar: If playwright updates to 19.2.6, then both dependencies need to be set to 19.2.6 in the same PR, but playwright does 2 PRs, one for each dependency.

PR #4699, opened 2023-01-11
Don't run dependabot where it doesn't have permission

pbiggar:
No Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -484,7 +484,7 @@ workflows:
             - gcp context
           filters:
             branches:
-              ignore: /^(pull\/).*$/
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -484,7 +484,7 @@ workflows:
             - gcp context
           filters:
             branches:
-              ignore: /^(pull\/).*$/
+              ignore: /^(pull|dependabot)\/.*$/
```
StachuDotNet, line 487: this prevents dependabot complaining on outdated dependencies for dependabot-created branches?

StachuDotNet (approved):

PR #4700, opened 2023-01-12
Convert parts of canvas to tailwind

OceanOak:
Changelog:

```
Internal improvements
-  Convert the tutorial toolotips to Tailwind.
```

**Questions:**

1. I came across this note while converting the tutorial tooltip to tailwind:
 `" NOTE: This will change in the future to pretty tool tips, this is just an inbetween state"`
 If you have a general idea of the things you want to be changed please let me know. For now it is just converted to tailwind.

2. The tooltips for the sidebar elements are converted to Tailwind but they are not used currently. Re-adding them was mentioned as a followup to this PR #4511. Should I re-add them? (I am not sure how they would look like with the current popups we have)
 
### Before ->After
Frame 11 (4)

Frame 13

Frame 12 (5)

Frame 17 (3)

Frame 15

### DIFF
--- a/client/src/canvas/View.res
+++ b/client/src/canvas/View.res
@@ -511,7 +511,14 @@ let viewBackToCanvas = (currentPage: AppTypes.Page.t, showTooltip: bool): Html.h
       )
 
     Html.div(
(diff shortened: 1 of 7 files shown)


pbiggar:
Nice job, thanks!

> " NOTE: This will change in the future to pretty tool tips, this is just an inbetween state"

I think that was written a long time ago - no idea what it means.

PR #4702, opened 2023-01-12
New documentation box design PR -followup

OceanOak:
No changelog

Followup to the New documentation box design PR #4664 

### Changes:
- Styled the deprecated function name with our usual function color and changed the font to font-code.
- Used `typ` instead of `\"type"`.
- Removed the added colors.
- Used colors from our color palette, and used different colors  for the foreground and background.

**Function name:**
image

**Type error:**
image

Still experimenting with color options for the foreground and background in the types. I would appreciate your input on whether a colored or grey background would be more suitable.


pbiggar:
Looks great, nice job.

> Still experimenting with color options for the foreground and background in the types. I would appreciate your input on whether a colored or grey background would be more suitable.

The grey works nicely. I'm not against further iterating - up to you.

PR #4703, opened 2023-01-12
Move deploy lock to a darklang canvas

pbiggar:
Changelog:

```
Internal
- Move deployment lock to a darklang canvas
```

The deploy lock uses a storage bucket, which requires cloud permissions to be put almost everywhere during a deploy.

This reimplements the deploy lock as a simple dark DB, in https://darklang.com/a/ops-circleci, and rewrites the deploy lock scripts to use it.

There is a secret in that canvas already. For deploying locally, which we almost never do, we could stop using the deploy lock perhaps, or have a key locally. I'm leaving that decision to a later day, when we have more experience with how the goes.

- [ ] Followup: restrict gcp deployment permissions to only the darklang-static-assets bucket

### DIFF
--- a/scripts/deployment/_deploy-lock-request
+++ b/scripts/deployment/_deploy-lock-request
@@ -1,22 +1,22 @@
 #!/usr/bin/env bash
 . ./scripts/devcontainer/_assert-in-container "$0" "$@"
 
-# Add a lock identifier for the commit and the timestamp.
+# Make requests to deploy lock server
 
 set -euo pipefail
 
+set -x
+
 if [ ! -v DEPLOY_LOCK_TOKEN ] ;
 then
-  echo "No deploy lock token, get it from https://darklang.com/a/ops-circleci"
+  echo "No deploy lock token, get it from https://darklang.com/a/ops-circleci" >&2
   exit 1
 fi
 
 REQUEST_PATH=$1
 METHOD=$2
-shift 2
 
-curl -s "https://ops-circleci.builtwithdark.com/deploy-lock${REQUEST_PATH}" \
+curl -f -s "https://ops-circleci.builtwithdark.com/deploy-lock${REQUEST_PATH}" \
   -X "${METHOD}" \
   -H 'Content-type: application/json' \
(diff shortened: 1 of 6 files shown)


StachuDotNet:
I'll be able to more fully review this in an ~hour, but:
https://darklang.com/a/ops-circleci#handler=853689901&trace=fe7a4e3f-0eee-8319-4385-bf4ab251ccf5
the recent traces of this all seem to return a `400` with body `account_inactive` - is that expected?

pbiggar:
> I'll be able to more fully review this in an ~hour, but:
> https://darklang.com/a/ops-circleci#handler=853689901&trace=fe7a4e3f-0eee-8319-4385-bf4ab251ccf5
> the recent traces of this all seem to return a `400` with body `account_inactive` - is that expected?

That handler isn't related to this PR - that one does slack messages from builds, which no longer work because we disabled access to Slack as part of rolling creds recently. That should be switched to discord soon.

scripts/deployment/_deploy-lock-request
```diff
@@ -0,0 +1,22 @@
...
+
+if [ ! -v DEPLOY_LOCK_TOKEN ] ;
+then
+  echo "No deploy lock token, get it from https://darklang.com/a/ops-circleci"
+  exit 1
+fi
+
+REQUEST_PATH=$1
+METHOD=$2
+shift 2
```
StachuDotNet, line 16: I haven't used `shift` before. Could you explain what this is doing? I researched a bit, but confused - knowing what it does in this specific case would help.

StachuDotNet (approved):

scripts/deployment/_deploy-lock-request
pbiggar: It "shifts" 2 items (pops from the front of the list). The list is implicit, it's `$@`. So this removes $1 and $2 from $@ so that we can use $@ later.

pbiggar: I simplified the code and removed the shift

StachuDotNet: I appreciate it - somehow `shift` feels like magic to me a bit :)

PR #4704, opened 2023-01-14
Add overview doc of the DB password rotation

pbiggar:
Changelog:

```
Internal:
- make it easier to rotate DB credentials
```

This provides a guide to rotating the DB credentials. I haven't done the steps marked "done" but will before merging.

### DIFF
--- a/docs/production/db-creds-rotation.md
+++ b/docs/production/db-creds-rotation.md
@@ -55,7 +55,7 @@ alter role userNEW login password 'NEW PASSWORD';
 ```
 create table check_its_working(id int);
 select * from pg_tables where tableowner = 'postgres'; # should include `check_its_working`
-drop table check_its_working
+drop table check_its_working;
 ```
 
 - Update all uses (we update in place because it's a much bigger hassle not to)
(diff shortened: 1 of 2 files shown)


docs/production/db-creds-rotation.md
```diff
@@ -0,0 +1,116 @@
...
+- Set their default login role as `postgres`
+
+```
+alter role user1 set role postgres;
+alter role user2 set role postgres;
+```
+
+## Overview of pw rotation
+
+- existing infra will be using either `user1` or `user2`. Find out which one.
```
StachuDotNet, line 31: I don't understand. We just added these 2 users - how are they 'already' being used?

docs/production/db-creds-rotation.md
```diff
@@ -0,0 +1,116 @@
...
+- enable the other user, with a new password
+- change the secrets in k8s to the other user
+- bounce every pod
```
StachuDotNet, line 34: What does it mean to "bounce" a pod?

docs/production/db-creds-rotation.md
pbiggar: This applies to subsequent migrations (during the first migration, the postgres user is in use)

docs/production/db-creds-rotation.md
pbiggar: Just to restart it, which fetches the secrets again

docs/production/db-creds-rotation.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+- Update all uses (we update in place because it's a much bigger hassle not to)
+
+  - update the `default` namespace
+
+    - backup secret: `kubectl get secrets -n default cloudsql-db-credentials -o yaml > old-db-secret.yaml`
+    - update the password in the secret
+      - get the base64 for the new password: echo "NEW PASSWORD" | base64
+      - edit the secret: `kubectl edit secrets cloudsql-db-credentials -n default`
```
StachuDotNet, line 68: I'm not familiar with this command. This feels like it implies that this command will prompt you for the secret and its hash - is that correct?

StachuDotNet (approved):
Great explanation. Shame it's such a dance to get this chore done :)

pbiggar:
Clarified some more stuff.

PR #4708, opened 2023-01-16
Add function signature and convert viewIntrospect to Tailwind

OceanOak:
Changelog:

```
Internal improvements
- Convert ViewIntrospect to Tailwind CSS.
- Add the function signature to the documentation box.
```

### Function signature

dark-Dark-Google-Chrome-2023-01 (4)

### Convert ViewIntrospect to tailwind
before->after
Frame 23 (4)

typeView
image

**Questions:** 
- `_introspect.scss` file is not used anymore. Should I delete it?
- Should we remove the return type from the docbox now that we have a signature?

### DIFF
--- a/client/src/canvas/View.res
+++ b/client/src/canvas/View.res
@@ -267,7 +267,7 @@ let viewTL_ = (m: model, tl: toplevel): Html.html<msg> => {
     avatars,
     Html.div(
       list{
-        Attrs.classList(list{("use-wrapper", true), ("fade", hasFF)}),
+        Attrs.classList(list{(%twc("bg-transparent ml-6 flex flex-col items-stretch"), true), (%twc("opacity-30"), hasFF)}),
         // Block opening the omnibox here by preventing canvas pan start
         EventListeners.nothingMouseEvent("mousedown"),
       },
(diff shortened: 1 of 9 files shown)


StachuDotNet:
FYI in `ViewIntrospection` I see a few cases where `packageFnView` is assigned an `id` of `pkg-fn`, etc. Since there may be multiple of these (if a TL is referencing multiple DBs, multiple packages, etc.), I this might have to be updated to `class`es?

OceanOak:
> FYI in `ViewIntrospection` I see a few cases where `packageFnView`, `dbView`, etc. are assigned `id`s of `pkg-fn`, etc. Since there may be multiple of these (if a TL is referencing multiple DBs, multiple packages, etc.), I this might have to be updated to `class`es?

This completely slipped my mind since I've added the IDs just to fix the integration tests.
I've added back class names; they will be only used to make the elements selectable, and will have no styling purposes.
Is there a better solution?

StachuDotNet:
I think it's fine for these to be classes. It might be worth updating their classes to something like `introspection-db` instead of just `db`, though.

pbiggar:
Client is gone

PR #4719, opened 2023-01-23
Autocomplete dropdown

OceanOak:
Changelog:

```
Internal improvements
- Convert the autocomplete dropdown to Tailwind CSS.
- Remove the package name from the autocomplete.
```

### Before
ocean-Dark-Google-Chrome-2023-01 (2)

### After
ocean-Dark-Google-Chrome-2023-01 (1)

### Before->After
**Valid item**
valid

**Invalid return type**
invalid return

**Invalid piped argument**
invalid piped arg

**package functions :** 
Nothing I've tried looked good. In order to draw the user's attention to the function more than the package I've used slightly different color and font size but I'm not sure if this is enough.
image
image

**Questions:**
- For the package functions I'm using a regex to test if the function names begin with 'dark/' and to separate the package name from the function name. Is there a better way to acheive this? 
- Should we create all these variations of colors? I couldn't find a way to do it in tailwind. At the moment, I am using the hex value for some of them or a close match from the colors that we already have.
lighten(white2, 10%)
darken(white, 5%)
lighten($grey2, 15%)
lighten(saturate($red, 25%), 20%)
lighten(saturate($red, 20%), 40%)
...

### DIFF
--- a/client/src/fluid/FluidAutocomplete.res
+++ b/client/src/fluid/FluidAutocomplete.res
@@ -845,13 +845,26 @@ let documentationForFunction = (
     }
   }
 
-  let name = Html.span(
-    list{tw(%twc("text-grey3 text-xs text-ellipsis w-max overflow-hidden whitespace-nowrap"))},
-    list{Html.text(f.name |> FQFnName.toString)},
-  )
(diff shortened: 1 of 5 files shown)


pbiggar:
Client is gone

Issue #3902, opened 2022-05-25
Change NotImplementedTODO sqlSpec to NotQueryable where applicable.


pbiggar:

pbiggar:
Done

PR #4729, opened 2023-01-29
Address NotYetImplementedTODO

pbiggar:
Changelog:

```
Standard Library
- mark more functions as Not Queryable instead of Unknown
```

The remaining functions are queryable (but we're not doing them right now), so let's remove the TODO from the name.

No effect on the API/client.

### DIFF
--- a/backend/src/BackendOnlyStdLib/LibCrypto.fs
+++ b/backend/src/BackendOnlyStdLib/LibCrypto.fs
@@ -34,7 +34,7 @@ let fns : List<BuiltInFn> =
         (function
         | _, [ DBytes data ] -> MD5.HashData(ReadOnlySpan data) |> DBytes |> Ply
         | _ -> incorrectArgs ())
-      sqlSpec = NotYetImplementedTODO
+      sqlSpec = NotYetImplemented
       previewable = ImpurePreviewable
       deprecated = NotDeprecated }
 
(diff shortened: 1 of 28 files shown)


pbiggar:
Fixes #3902

backend/src/BackendOnlyStdLib/LibCrypto.fs
```diff
@@ -34,7 +34,7 @@ let fns : List<BuiltInFn> =
         (function
         | _, [ DBytes data ] -> MD5.HashData(ReadOnlySpan data) |> DBytes |> Ply
         | _ -> incorrectArgs ())
-      sqlSpec = NotYetImplementedTODO
+      sqlSpec = NotYetImplemented
```
StachuDotNet, line 37: Postgres seems to support an MD5 fn. https://www.postgresqltutorial.com/postgresql-string-functions/postgresql-md5/
Is this worth pursuing?

The md5 fn seems built-in; there are fns additionally available in the `pgcrypto` extension that could address some more of our crypto fns (I see at least hmacsha256 and hmacsha1 in there)

generally: if there are fns that _could_ be SqlCompile-able, but aren't functions we'd expect someone to reach for often in a `DB::query`, do you have any feel for what to do? I'm sort of doubtful somewhat will try to `Crypto::md5` _often_ in a query, but could see it.

StachuDotNet: Sorry, I somehow missed that you just removed the extra/unneeded "TODO" word here.

StachuDotNet (approved):

backend/src/BackendOnlyStdLib/LibCrypto.fs
pbiggar: > Is this worth pursuing?

Not sure if we want to stay on PG long term, and these functions might damage our ability to move to another platform. So my plan is to not add any more unless there is great user demand.

PR #4730, opened 2023-01-29
Remove FQFnName.parse

pbiggar:
Changelog:

```
Internal improvement
- remove ability to parse function names
```

This is the final step of removing parsing function names. Now we only do it as part of testing. We have long since removed it from the API and other places, so this is just moving it from being usable in the user-facing parts of the codebase, to not being usable from there.

The only remaining use case was in `/api/v1/execute_function` but this was a string being parsed that would also be a user function name, and so did not need to go through the `parse` function.


StachuDotNet (approved):

StachuDotNet:
Is this worth revisiting in new-dark? (should it be 'ported' in some way?)

pbiggar:
Yes, I think so. This should be moved into Parser

pbiggar:
Superseded

PR #4733, opened 2023-01-31
Update static assets tests (to correspond with google change)

StachuDotNet:
No changelog

Some of our backend tests broke due to a change in headers returned by google upstream.
(a new header `Vary: X-Goog-Allowed-Resources` is returned for upload requests)

This updates/fixes the tests


pbiggar (approved):
Thanks!

Issue #4732, opened 2023-01-31
Implement a new ELet supporting just a Variable pattern


StachuDotNet:
This PR comes out of #4711 and [this document]().

The goal is to update our ProgramTypes and RuntimeTypes such that `ELet`'s definition goes from:

```fsharp
type Expr =
  ...
  | ELet of id * string * Expr * Expr
```

to
```fsharp
type LetPattern =
  | LPVariable of id * name : string

type Expr =
  ...
  | ELet of id * LetPattern * Expr * Expr
```
so we may support additional LetPatterns after this work is done.

---

Here's a 4-PR strategy to migrate our simple ELets into ELets that support LetPatterns.
I think it's prohibitively awkward to maintain two kinds of ELets, especially in the client
where FluidTypes are involved, so this aims to avoid that.

- [ ] Implement new ELetWithPattern in both backend and client, but communicate in old-style `ELet`s
   Prepare decoders for `ELetWithPattern`, for once clients are caught up
  - client
    - update ELet in client to have LetPatterns (only maintain one kind of ELet in client)
    - encode as old ELet (boring string)
    - add decoder to handle (unused) ELetWithPattern
  - backend
    - STs: rename ELet to ELetWithoutPattern; add ELet with a LetPattern (ELetWithoutPattern will forever remain in STs)
    - CTs: add new ELetWithPattern
    - replace ELet with ELetWithPattern in RT, PT; update impl. generally
    - only communicate in old-style `ELet`s (i.e. translate any `ELetWithPattern`s to `ELet`s)
    - but implement decoder to handle incoming `ELetWithPattern`s (unused for now)
- [ ] (let clients catch up)
  now everyone has the new decoders
- [ ] Start communicating in ELetWithPattern instead of ELet
  - client, backend: update encoder to start serializing in ELetWithPattern
  - deal with roundtrip serialization tests, and vanilla serialization files
- [ ] (let clients catch up)
  at this point we're stable, but the serialized name is "ELetWithPattern" rather than just "ELet"
- [ ] update the ELet in ClientTypes to be the same as the ELetWithPattern one
  Since the clients are caught up at this point, the old ELet CTs aren't being used.
  Update them to match the shape of ELetWithPattern, so we can shortly remove long-named ELetWithPattern
- [ ] (let clients catch up)
- [ ] remove ELetWithPattern from ClientTypes and from client-side decoding
  - again deal with roundtrip serialization tests, and vanilla serialization files

StachuDotNet:
[...]

PR #4711, opened 2023-01-17
spike - tuple let patterns

StachuDotNet:
Changelog:

```
Language and Standard Library: 
- add tuple let patterns

Editor:
- (hidden) add tuple let patterns
```
---

Progress:
- [x] most `backend` work for a basic model: 
 ```fsharp
type LetPattern =
  | LPVariable of id * name : string
  | LPTuple of id * first : string * second : string * theRest : List<string>

type Expr = 
  | ...
  | ELet of id * string * Expr * Expr
  | ELetWithPattern of id * LetPattern * Expr * Expr
 ```
  - [x] `SerializedTypes`, `ProgramTypes`, `RuntimeTypes`, `ClientTypes`, all the mappings
  - [x] interpreter implementation
  - [x] parsing `let (a, b) = (1, 2)` from `.test` files (`FsharpToExpr`)
  - [x] basic impl for future `FuzzTests`
- [x] basic modeling in `client` side of things
- [ ] (WIP) improving upon the `backend` model, making it recursive
  from
  ```fsharp
  type LetPattern =
    | LPVariable of id * name : string
    | LPTuple of id * first : string * second : string * theRest : List<string>
  ```
  to
  ```fsharp
  type LetPattern =
    | LPVariable of id * name : string
    | LPTuple of id * first : LetPattern * second : LetPattern * theRest : List<LetPattern>
  ```
  As far as a "spike" goes maybe it would have been best to focus on a basic Fluid impl. next, but the model feels solid and I suspect it'd be very annoying to re-do all of the fluid work if we update the model to be recursive after an initial implementation.
- [ ] update `LetPattern` model in `client` (to be recursive)
- [ ] Fluid work
  - [ ] lots of tests
- [ ] fill in and revise bits of `backend` work
  - [ ] review all nontrivial implementation
    most of the impl was copy/pasted from `MatchPattern` logic, so there are likely major simplifications to be made
  - [ ] figure out what to do - the `FSharpToExpr` now interprets all test files as the _new_ kind of let. this isn't merge-able until we test that `ELet("name", ...)` is 100% equivalent to `ELetWithPattern(... LPVariable("name", ...))`
  - [ ] `SqlCompiler`
  - [ ] review `FuzzTests` impl.
  - [ ] serialization tests
    - [ ] backend
    - [ ] client-side roundtrip

PR #4735, opened 2023-02-01
Start implementing new "let" that contains Patterns on the LHS

StachuDotNet:
Changelog:

```
Internal
- start to implement a new 'let' expression that contains patterns on the left (e.g. to deconstruct tuples)
```

Currently, users may only assign a single variable in a let expression. This is the first step in replacing our existing Let expression with a new Let that supports multiple kinds of patterns on the LHS of the `=`. It's based off of #4711, and the larger plan is outlined in [this document](https://docs.google.com/document/d/10mr811ONuUSm6Q1egfv9dGtpN7QdHwhM96uyW8w0140/edit#heading=h.2dqrn78eah12).

This PR contains the work outlined in the first bullet of #4732.
At the end of this PR, both the `client` and `backend` 'think' in terms of the new Let (with a Variable pattern) but the communication is still in the old kind of Let. That said, _decoders_ on both ends have been set up for the next relevant PR, which will switch over communication to be in the _new_ kind of Let.
Relevantly, some serialization roundtripping tests were failing - the decoders currently "drop" the Ids contained within the LetPattern, and the encoders generate a new ID for the variable LetPattern. For now, I've disabled these tests, to turn them back on as part of the next PR - I've done enough manual testing to be comfortable with this, but I'm open to ideas to avoid such.


backend/src/LibExecution/Interpreter.fs
```diff
@@ -59,19 +59,21 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
     | ENull _id -> return DNull
     | ECharacter (_id, s) -> return DChar s
 
-
-    | ELet (_id, lhs, rhs, body) ->
+    | ELet (_id, pat, rhs, body) ->
       let! rhs = eval state st rhs
+
       match rhs with
-      // CLEANUP we should still preview the body
```
StachuDotNet, line 66: Use a simplified form of the code in #4711 for this (which _does_ preview the body)

backend/tests/FuzzTests/Generators.fs
```diff
@@ -192,16 +192,32 @@ module RuntimeTypes =
                       //(4, ok) // TODO: random string, with 0-5 args
                        ]
 
-  open MatchPattern
+    let rec genTuple (s, genArg) : Gen<RT.MatchPattern> =
```
StachuDotNet, line 195: Looks like I backfilled this, but didn't 'use' it below. Not top-priority but might as well use it below in the file

backend/tests/FuzzTests/Generators.fs
```diff
@@ -347,6 +398,7 @@ module RuntimeTypes =
     let recursiveExprs =
       [ genConstructor
         genLet
+        genLetWithPattern
```
StachuDotNet, line 401: We don't need duplicate defs

backend/tests/TestUtils/FSharpToExpr.fs
```diff
@@ -297,43 +297,29 @@ let rec convertToExpr' (ast : SynExpr) : PT.Expr =
 
   // When we add patterns on the left hand side of lets, the pattern below
   // could be expanded to use convertPat
+  // TODO: this doesn't allow nesting (i.e. `let (a, (b, c)) = (1, (2, 3))`. Should it?)
```
StachuDotNet, line 300: no longer true, comment may be removed

StachuDotNet:
this is generally looking good to me, but a few backend serialization roundtrip tests are failing.
I hit a similar case in equivalent client-side tests, and I solved it by forcing the IDs of the Variable LetPattern to be `0` in the test scenarios. I could do that same thing for the backend tests, but that's feeling pretty hacky.

Another solution would be to temporarily remove the `id` from the LPVariable definition _in ClientTypes_. Until we have additional patterns, this ID isn't actually useful. I'm going to try that.

StachuDotNet:
Nah, that's not enough, since we do roundtrips between 'internal' types and clienttypes as well.
(start with an internal Expr. then `expr -> toCT -> fromCT` - this loses the ID)

I started going down this road, but the many places where we perform toID make it a bit awkward.

To avoid repeated work and churn I'm thinking to:
- remove the hack in the client side
- simply disable the roundtripping tests
- rely on manual verification of roundtripping for now (my confidence is very high)
- turn the tests back on once we've migrated to the new Let
  (really, the next PR as outlined in #4732 is fine - the tests would only be disabled for a bit)

StachuDotNet:
I think dark is re-stable enough to refresh this against the updated codebase - what do you think @pbiggar?

pbiggar:
Yes, I think so. Though I'd wait for my Test changes to land - maybe tomorrow.

StachuDotNet:
Replaced

PR #4737, opened 2023-02-02
Separate the omnibox from the AC dropdowns for DB field types/ FN param types, and convert them to use Tailwind

OceanOak:
Changelog:

```
Area of change
- details
```


pbiggar:
Client is gone

PR #4739, opened 2023-02-02
Experiment: Try new multiline string tokenization

StachuDotNet:
The rendering of long strings in the editor has been annoying me lately. This is an experiment at improving our rendering.

This solution seems simpler and results in better output, imo. Is this naive in some way, or is it worth pursuing?

Before:
image

After:
image

It feels a little bit like cheating to increase the strLimit, so here's what it looks like without that increase:
image


pbiggar:
Getting rid of the client

PR #4753, opened 2023-03-01
Fix Int.random to correct work on range 0 1

ditansu:
Changelog:

```
Area of change
- Int.random_v1 
```

For a case, `Int.random_v1 0 1` always returns 0  

Increasing an upper bound on 1 for this case. 

Run several times the bunch of 10 tests `Int.random_v1 0 1 = 1` and check that at least one test get value 1. In fact, 30-60% of tests were passed. 

yes


pbiggar:
Great, thank you!

PR #4754, opened 2023-03-01
Remove remaining legacy serializer

pbiggar:
Changelog:

```
StdLib
- Remove toString
- Add Bool.toString
- Add Int.toString
- Add Char.toString
- Add Uuid.toString
- Add Float.toString

Internal:
- Remove DvalReprLegacyExternal.toEnduserReadable serializer
```


backend/src/LibExecutionStdLib/LibChar.fs
```diff
@@ -143,4 +143,20 @@ let fns : List<BuiltInFn> =
...
+      fn =
+        (function
+        | _, [ DChar c ] -> Ply(DStr(c))
+        | _ -> incorrectArgs ())
+      sqlSpec = NotQueryable
+      previewable = Pure
+      deprecated = NotDeprecated }
+
+
+    ]
```
StachuDotNet, line 162: Is this intended to be a new pattern of ending these fn lists with 2 newlines? (asking as a few files have this)

backend/src/LibExecutionStdLib/LibBool.fs
```diff
@@ -64,4 +64,22 @@ let fns : List<BuiltInFn> =
...
+      returnType = TStr
+      description = "Return {\"true\"} or {\"false\"}"
+      fn =
+        (function
+        | _, [ DBool b ] ->
+          match b with
+          | true -> Ply(DStr("true"))
+          | false -> Ply(DStr("false"))
+        | _ -> incorrectArgs ())
+      sqlSpec = NotQueryable
```
StachuDotNet, line 81: Probably doable

StachuDotNet: adding a tiny PR to mark these as NotYetImplemented

StachuDotNet: actually, what do you think? Am I right to think they should be marked as NotYetImplemented?

pbiggar: Yeah, NotYetImplemented is fine.

backend/src/LibExecutionStdLib/LibChar.fs
pbiggar: I'm sorta undecided here. It's annoying having the list end at the last entry, but ugly to add space. If you have a preference, happy to go with it.

StachuDotNet: I've been fine with the `} ]` endings. I've played with both locally, and find the empty lines between the two sort of ugly, not worth the tiny diff annoyance when adding a stdlib fn 'at the end'

PR #4755, opened 2023-03-01
Update Int::random; add List::unique

StachuDotNet:
Changelog:

```
Standard library
- update Int::random to be inclusive in more scenarios
- add List::unique_v0 - like List::uniqueBy_v0, but without a fn required as input
```

Just as I'd expect `Int::random 0 1` to return 0 or 1, I'd expect `Int::random 0 2` to expect any of {0, 1, 2}. Do you agree, @pbiggar? If so, I think we can update the impl to reflect this.

### DIFF
--- a/backend/src/LibExecutionStdLib/LibInt.fs
+++ b/backend/src/LibExecutionStdLib/LibInt.fs
@@ -290,7 +290,7 @@ let fns : List<BuiltInFn> =
           // .NET's "nextInt64" is exclusive,
           // but we'd rather an inclusive version of this function
           let correction : int64 = 1
-          
+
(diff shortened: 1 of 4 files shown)


pbiggar:
These are cool tests! I expect they'll fail about once every 1,267,650,600,228,229,401,496,703,205,376 test runs, is that right? (2^100)

StachuDotNet:
Something like that :)

Do you agree that `Int::random 0 2` should be able to yield all of {0, 1 2}?

pbiggar:
yeah

StachuDotNet:
I started towards adding `Fun::identity` but hit some snags so I'll save that for a later PR.

pbiggar (approved):

PR #4758, opened 2023-03-02
rename TDate to TDateTime

OceanOak:
Changelog:

```
Language 
-  Rename Date to DateTime
```

### DIFF
--- a/backend/src/LibExecution/RuntimeTypes.fs
+++ b/backend/src/LibExecution/RuntimeTypes.fs
@@ -293,7 +293,7 @@ and DType =
   | TError
   | THttpResponse of DType
   | TDB of DType
-  | TDate
+  | TDateTime
   | TChar
   | TPassword
   | TUuid
@@ -331,7 +331,7 @@ and DType =
     | TError -> "Error"
     | THttpResponse _ -> "Response"
     | TDB _ -> "Datastore"
-    | TDate -> "Date"
+    | TDateTime -> "Date"
     | TDict _ -> "Dict"
     | TPassword -> "Password"
     | TUuid -> "UUID"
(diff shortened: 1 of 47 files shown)


OceanOak:
While I was renaming, I came across this comment.
>   // CLEANUP The module is only there for a few functions in the Date module, such as
>   // Date::<. Making these infix wasn't a great idea, and we should remove them.

Should we remove these infix (Date::<, Date::>,etc) functions? If so, what other elements besides the functions themselves , the tests, the regex in the ProgramTypesParser? should we remove?

backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -331,7 +331,7 @@ and DType =
     | TError -> "Error"
     | THttpResponse _ -> "Response"
     | TDB _ -> "Datastore"
-    | TDate -> "Date"
+    | TDateTime -> "Date"
```
OceanOak, line 334: Is this supposed to be changed to  `TDateTime -> "DateTime"`?

pbiggar: yes

OceanOak: Fixed

pbiggar:
Nice job.

> While I was renaming, I came across this comment.
> 
> > // CLEANUP The module is only there for a few functions in the Date module, such as
> > // Date::<. Making these infix wasn't a great idea, and we should remove them.
> 
> Should we remove these infix (Date::<, Date::>,etc) functions? If so, what other elements besides the functions themselves , the tests, the regex in the ProgramTypesParser? should we remove?

Yeah, this is a good follow-up. All those things sound like the right things to remove.

PR #4759, opened 2023-03-02
Add Fun::identity fn

StachuDotNet:
Changelog:

```
Area of change
- details
```

It's broken currently - the interpreter doesn't allow you to pass a stdlib fn as the RHS of a pipe (has to be lambda or use fn, it seems). See the test in language.tests

The identityB fn would probably work (I haven't tested) but seems unideal.

I'm going to leave this for a few days while I do other work - @ anyone: feel free to pick up this PR in the meantime.

### DIFF
--- a/backend/testfiles/execution/language.tests
+++ b/backend/testfiles/execution/language.tests
@@ -129,6 +129,9 @@ List.push_v0 [] (fun x -> -4.611686018e+18) = [(fun x -> -4.611686018e+18)]
  let x = 6 in
  [1; 2; 3; 4] |> List.map_v0 y) = [ 6; 7; 8; 9 ]
 
+[test.pipe identity function]
+([1; 2; 3] |> List.map_v0 Fun.identity) = [1; 2; 3]
+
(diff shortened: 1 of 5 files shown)


backend/testfiles/execution/language.tests
```diff
@@ -129,6 +129,9 @@ List.push_v0 [] (fun x -> -4.611686018e+18) = [(fun x -> -4.611686018e+18)]
  let x = 6 in
  [1; 2; 3; 4] |> List.map_v0 y) = [ 6; 7; 8; 9 ]
 
+[test.pipe identity function]
+([1; 2; 3] |> List.map_v0 Fun.identity) = [1; 2; 3]
```
StachuDotNet, line 133: This is apparently illegal. @pbiggar could you confirm that this _should_ be legal Dark code?

pbiggar: It definitely isn't legal now, but it's also definitely an obvious thing that should be enabled/allowed.

PR #4772, opened 2023-03-07
Remove authorization

OceanOak:
Changelog:

```
Tech debt
-  Remove authorization
```
Functions were deleted from Authorization.fs one at a time, ensuring that they were only used in tests or functions that were intended to be removed. As a result, the file became empty so I have deleted it. Related tests were also deleted.


backend/tests/Tests/Authorization.Tests.fs
```diff
@@ -39,36 +22,4 @@ let testSetUserAccess =
...
-    Expect.equal (A.Read.max (A.Read)) A.Read ""
-
-    Expect.equal (A.Read.max (A.ReadWrite)) A.ReadWrite ""
-
-    Expect.equal (A.ReadWrite.max (A.Read)) A.ReadWrite ""
-
-    Expect.equal (A.ReadWrite.max (A.ReadWrite)) A.ReadWrite ""
-  }
-
-let tests = testList "A" [ testPermissionMax; testSetUserAccess ]
```
OceanOak, line 74: I've realized that I accidentally removed both tests here. Should I add `testSetUserAccess` back and remove it in a separate PR where I start removing things from `Accounts.fs`, or should I just delete the `Authorization.Tests.fs `file?

pbiggar:
Great!

PR #4773, opened 2023-03-07
Start supporting user-defined enums

StachuDotNet:
Changelog:

```
Language and standard library
- add partial support for user-defined enums
```


StachuDotNet:
There's still work to be done here, but some preliminary thoughts and questions:

I'm really unsold on needing nameID and typeID and all sorts of field-level IDs.
Taken to the extreme, we'd end up with [this](https://gist.github.com/StachuDotNet/b0976880322a74b15469ff17bec0e91e), and it gets even uglier if you consider the attributes needed at the `SerializedTypes` level.
if we need to reference some field in one of these structures, we could do that with a specific Op or API endpoint given the structure's "main" ID - not sure why we'd need additional IDs. If, however, we do need IDs along every field for some reason, we could instead tackle this via some `type ThingWithId<'a> = | Thing of id * 'a` structure so code is a bit less noisy, but that seems bad too. I think the impl this is heading towards is sufficient, but happy to discuss and add a bunch of IDs back in if needed.

Here are some problems I'm hitting:
- ERecords are a bit distracting for some reason
- the parser needs to be adjusted such that Enum _values_ can be parsed, which (I think) requires it to be recursive with the so-far-parsed UserTypes within scope

Here are some big-ticket TODOs on my mind currently, given the progress:
- [ ] rebrand current `ERecord` as `EAnonRecord`, and similar for `DObj`-> `DAnonRecord` (both to be deprecated once we've moved existing records to be typed)
- [ ] update Parser (and relevant code) so that all EAnonRecord usages are like F#'s anon records: `{| A = 1 |}`
- [ ] update Parser to handle Enum _values_; this will require some recursion such that the parser is aware of pre-defined user types
- [ ] revisit interpreter changes - I just got them to a point where they compiled, basically
- [ ] revisit type-checking
- [ ] uncomment `EUserRecord` - follow everything through such that UserRecord usages 'work' (for now, ignore the case of multiple records having the same shape)
- [ ] fill in a bunch of gaps (TODOs) noted in this PR
- [ ] lots of additional testing
- [ ] ensure we've included all of the new/adjusted things within the serialization test values
- I think I've forgotten some big-ticket items; they'll come to me after reviewing the PR again.

I don't know where this PR ends. Maybe now-ish, or after TODOs are resolved?
Or maybe this is a bigger "Rebrand Records as AnonRecords; support UserTypes properly" scope. (honestly seems reasonable to me)

What do you think? Does this PR so far meet what you were thinking? if so, keep trucking along here, or merge soon after some tidying, and then follow up in subsequent PRs?

---

Later, after this all settles, some follow-up tasks:
- [ ] extend UserTypes to really be CustomTypes, and allow them in Stdlib space
- [ ] define Stdlib-level types for existing fns; use them in Stdlib fns
  - [ ] `Option`
  - [ ] `Result`
  - [ ] others?
- [ ] replace EAnonRecords+DAnonRecords with their typed equivalents
- [ ] remove EAnonRecord+DAnonRecord
- [ ] replace `MPConstructor` with `MPCustomEnum`
- [ ] (whenever we get back to Packages) extend CustomTypes to allow them in Package space

---

Tangentially: SerializedTypesToProgramTypes and ProgramTypesToSerializedTypes currently exist as separate modules. I'm not sure why, since we have these sort of back-and-forth translations typically within just 1 module. Should we consolidate for consistency?

StachuDotNet:
FWIW I'm happy to adjust ^ totally ignoring the existing "Records" but find it a bit distracting. Since I have the motivation to move that ball forward, it seems like it's worthwhile even if not technically directly in my path

StachuDotNet:
Later, after this all settles, some follow-up tasks:
- [ ] extend UserTypes to really be CustomTypes, and allow them in Stdlib space
- [ ] define Stdlib-level types for existing fns; use them in Stdlib fns
  - [ ] `Option`
  - [ ] `Result`
  - [ ] others?
- [ ] replace EAnonRecords+DAnonRecords with their typed equivalents
- [ ] remove EAnonRecord+DAnonRecord
- [ ] replace `MPConstructor` with `MPCustomEnum`
- [ ] (whenever we get back to Packages) extend CustomTypes to allow them in Package space

StachuDotNet:
Tangentially: SerializedTypesToProgramTypes and ProgramTypesToSerializedTypes currently exist as separate modules. I'm not sure why, since we have these sort of back-and-forth translations typically within just 1 module. Should we consolidate for consistency?

pbiggar:
> Tangentially: SerializedTypesToProgramTypes and ProgramTypesToSerializedTypes currently exist as separate modules. I'm not sure why, since we have these sort of back-and-forth translations typically within just 1 module. Should we consolidate for consistency?

Yes, I was just thinking that. (though not yet, I've made changes there)

pbiggar:
> I'm really unsold on needing nameID and typeID and all sorts of field-level IDs.

So in the new version of Dark, the IDs are useful for the following reasons:
- we can assign a live value to an ID
- we can specify a particular operation as happening at a particular ID

So the question is which IDs we want to add live values to. Or, alternatively, whether we want to add another mechanism for tagging live values (eg parameter names). I'd be down to remove all IDs that we aren't going to attach a live value to (eg keep it for parameter names, but not types).

pbiggar:
> the PR feels like it's a bit wide, and I'm torn whether to keep pressing forward or split it up (I'd prefer to just keep going but would like input here)

I think the current PR is in a good shape, and making it wide risks making it unreviewable and conflicting everything. I'd recommend merging ASAP and doing followups in different PRs.

StachuDotNet:
Mergeable:
- [x] follow up on RuntimeTypes.Dval.toType
- [x] revisit RuntimeTypes.Dval.typeMatches
- [x] revisit RuntimeTypesAst.preTraversal and .postTraversal
- [x] fill in TestUtils.exprEqualityBaseFn
- [x] fill in dvalEqualityBaseFn

Later:
- [ ] update parser to keep track of defined types (recursively); revisit language.tests
- [ ] follow up in interpreter: how do we handle fake vals for EUserEnum?
- [ ] follow up in interpreter: handle analysis/preview of EUserEnum

forgot to leave this yesterday

backend/src/ClientTypes2ExecutionTypes/ProgramTypes.fs
```diff
@@ -10,6 +10,16 @@ type id = Prelude.id
 type tlid = Prelude.tlid
 type Sign = Prelude.Sign
 
+
+module UserTypeName =
+  let fromCT (u : CTPT.UserTypeName) : PT.UserTypeName =
+    { type_ = u.type_; version = u.version }
```
pbiggar, line 16: `type_` should be spelled `typ` everywhere.

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -5,11 +5,24 @@ type id = Prelude.id
...
+/// ```
+/// (steal from `FQFnName`)
+type UserTypeName = { type_ : string; version : int }
+
 /// A Fully-Qualified Function Name
 /// Includes package, module, and version information where relevant.
 module FQFnName =
 
-  /// Standard Library Function Name
+  /// Standard Library Function NameDF
```
pbiggar, line 25: typo?

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -185,18 +245,15 @@ type DType =
   | TPassword
   | TUuid
   | TOption of DType
-  | TUserType of string * int
+  // split into
```
pbiggar, line 248: incomplete?

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -231,31 +288,27 @@ module DB =
       cols : List<Col> }
 
 module UserType =
-  type RecordField = { name : string; typ : Option<DType>; nameID : id; typeID : id }
-  type Definition = Record of List<RecordField>
+  // TODO: move this to some ComplexType or CustomType type
```
pbiggar, line 291: I find it's not super helpful to have incomplete thoughts added as comments, would prefer to remove them unless they're explanations of why the code is how to it (in which case longer explanations are more useful).

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -231,31 +288,27 @@ module DB =
...
+  type RecordField = { id : id; name : string; typ : DType }
 
-  type T =
-    { tlid : tlid
-      name : string
-      nameID : id
-      version : int
-      definition : Definition }
+  type EnumField = { id : id; type_ : DType; label : Option<string> }
```
pbiggar, line 294: What's the semantics here? Is it like:

```
type X1 =
  | Y1 of { a1 : int; b1 : int }
```

```
type X2 =
  | Y2 of a2 : int * b2 : int
```

or is there so difference there?

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -152,21 +165,68 @@ type Expr =
...
+  // TODO one of these:
+  // - implement EStdlibEnum and EStdlibRecord, then EPackageEnum and EPackageRecord
+  // - implement a more generic EDefinedEnum and EDefinedRecord
+  //   that reference awith `User`, `Stdlib`, and `Package` cases
+
+  /// Given a User type of:
+  ///   `type MyEnum = A | B of int | C of int * (label: string) | D of MyEnum`
+  /// , this is the expression
+  ///   `EUserEnum(UserType.MyEnum, "C", [EInteger(1), EString("title")]`
+  | EUserEnum of id * UserTypeName * caseName : string * fields : List<Expr>
```
pbiggar, line 218: This should be integrated with EConstructor, right?

This isn't an enum, afaik. This is a constructor (the enum is the type definition, right?)

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -152,21 +165,68 @@ type Expr =
...
+  | EFeatureFlag of
+    id *
+    flagName : string *
+    cond : Expr *
+    caseA : Expr *
+    caseB : Expr
+
+
+  // TODO:
+  // - define EUser
```
pbiggar, line 200: I wouldn't use EUser here - it seems like there should be ERecord, EConstructor and EFnCall, which each have a FQName also FQName.User, FQName.Package, and FQName.Stdlib.

backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -277,6 +281,7 @@ and Dval =
   | DOption of Option<Dval>
   | DResult of Result<Dval, Dval>
   | DBytes of byte array
+  | DUserEnum of typeName : UserTypeName * caseName : string * fields : List<Dval>
```
pbiggar, line 284: I'd call this DConstructor I think?

backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -301,7 +306,7 @@ and DType =
   | TPassword
   | TUuid
   | TOption of DType
-  | TUserType of string * int
+  | TUserType of UserTypeName // this should probably be split into TUserEnum and TUserRecord
```
pbiggar, line 309: I think it's fine to just use the name and find whether it's an enum or a record from the name.

backend/testfiles/execution/language.tests
```diff
@@ -32,6 +32,30 @@ module Records =
...
+  type Card = | A | B of string | C of i: int | D of c: string * i: int
+
+  // TODO: fix parser - "A" below is currently interpreted as a variable we can't find.
+  //(let cards = [A; B "test"; C(5)]
+  //
+  // match List.getAt_v1 cards 1 with
+  // | Some b -> Ok b
+  // | None -> Error "Failure") = Ok (B "test")
+
+  // TODO: support enum deconstruction in `match` exprs
```
pbiggar, line 46: That's what MPConstructor is for.

pbiggar (changes requested):

backend/src/LibExecution/ProgramTypes.fs
StachuDotNet: I've made the labels on enum fields optional, so it's more like the second option here. If you'd rather them be required, I could update there, in which case the first option could be a fair representation later. Otherwise I think they're equivalent if labels are required.

backend/src/LibExecution/ProgramTypes.fs
StachuDotNet: You're right, it is an EConstructor. I can get those merged together.

I've had a really hard time with the name "constructor" not referring to something in OOP space - would you be open to EEnumCtor or something else, that's not plainly "EConstructor"?

backend/src/LibExecution/RuntimeTypes.fs
StachuDotNet: :+1: please see comment above about "Constructor" feeling odd, though

backend/src/LibExecution/ProgramTypes.fs
StachuDotNet: some other ideas, some provided by chatgpt: EVariant, EUnionCase, ECase. If we can't find something that works, I'd be fine sticking with EConstructor. It's just been bugging me for a while for some reason. F#'s pattern to extract the values out of enums in `match` patterns is called the "identity" pattern if that's worth anything

PR #4780, opened 2023-03-09
Remove apiserver stuff

OceanOak:
Changelog:

```
Tech debt
- Remove more apiserver stuff
```
Changes
- remove all userInfo except for UserId, and keeps only the necessary account for testing.
- remove the following internal Lib fns: getUser, getUserByEmail, getUsers, canvasesFor, usernameToUserInfo, and getCanvasList.

There are a few pieces that are not removed yet, such as:
`usernameForUserID`  used in internalFn which wraps all internal Lib functions. Should we remove it and unwrap them?
`getUser` used for tests
`getUserAndCreatedAt`: used in analytics should we remove it?


pbiggar:
> `usernameForUserID` used in internalFn which wraps all internal Lib functions. Should we remove it and unwrap them?

Yeah, a lot of internalFn is going away so match just set `canAccess` to true for now.

> `getUser` used for tests
Can be removed

> `getUserAndCreatedAt`: used in analytics should we remove it?
Yes

backend/src/LibBackend/Account.fs
```diff
@@ -22,24 +22,9 @@ type Account =
...
-  { username : UserName.T
-    name : string
-    admin : bool
-    email : string
-    id : UserID
-    createdAt : NodaTime.Instant }
-
-let userInfoToPerson (ui : UserInfo) : LibService.Rollbar.Person =
-  Some { id = ui.id; username = Some ui.username }
+type UserInfo = { id : UserID }
```
pbiggar, line 25: I think we can remove this type entirely.

backend/src/LibBackend/Account.fs
```diff
@@ -177,6 +161,7 @@ let userIDForUserName (username : UserName.T) : Task<UserID> =
     | Some v -> v
     | None -> Exception.raiseGrandUser "User not found")
 
+//used in internalFn
```
pbiggar, line 164:
```suggestion
// used in internalFn
```

backend/src/LibBackend/Account.fs
```diff
@@ -22,24 +22,9 @@ type Account =
...
-    admin : bool
-    email : string
-    id : UserID
-    createdAt : NodaTime.Instant }
-
-let userInfoToPerson (ui : UserInfo) : LibService.Rollbar.Person =
-  Some { id = ui.id; username = Some ui.username }
+type UserInfo = { id : UserID }
 
+type UserInfoAndCreatedAt = { id : UserID; createdAt : NodaTime.Instant }
```
pbiggar, line 27: Is this still used? Plausibly we should remove it (can be another PR)

pbiggar (changes requested):
Looks great. One small change and we can merge.

backend/src/LibBackend/Account.fs
OceanOak: Removing it goes way deeper
It used by `testOwner` in `TestUtils.fs` which is used by `initializeTestCanvas` and `createTestCanvas` which are used by 

- Canvas.Tests
- Fuzz tests ( Passowirds.FuzzTests, ExecutionRegression.fuzzTests.fs, ExecutePureFunctions.FuzzTests.fs)
- Cron.Tests
- Stdlib.Tests,
- Traces.Tests
- `execute` function in FSI.fs, 
etc.

backend/src/LibBackend/Account.fs
OceanOak: Removed

backend/src/LibBackend/Account.fs
OceanOak: I have tried replacing
```
`let getUser (username : UserName.T) : Task<Option<UserInfo>>
// by 
let getUser (username : UserName.T) : Task<Option<UserID>>  

// and 
(owner : Task<LibBackend.Account.UserInfo>)
// by 
(owner : Task<UserID>)

// and 
let initializeCanvasForOwner
 (owner : Account.UserInfo)
 (name : string)
// by
let initializeCanvasForOwner
 (owner : UserID)
 (name : string)
```

but the backend tests were not passing
example of the errors :
```
account_id = 77a... is not present in table accounts

Unexpected exception count in test test-htdjy
```

pbiggar: Seems like you might need to edit some of the SQL to reflect these changes.

backend/testfiles/execution/language.tests
```diff
@@ -390,9 +390,9 @@ module Shadowing =
 // ---------------------------
 // DarkInternal tests are internal
 // ---------------------------
-module Darkinternal =
-  (let _ = Test.setExpectedExceptionCount 1 in
-   DarkInternal.getAllSchedulingRules) = Test.typeError_v0 "Unknown error"
+//module Darkinternal =
```
pbiggar, line 393: Accidentally commented out?

OceanOak: Apologies for forgetting to uncomment it. It was causing an error. Could you please explain this type of error since it is not clear to me?

```
[ERR] tests.LibExecution.All.language.Darkinternal.line394 
expected:
("Exception thrown in test", [], "")
  actual:
("UNEXPECTED EXCEPTION COUNT in test test-test-qqbga",
 [("expectedExceptionCount", 1); ("actualExceptionCount", 0)],
 "   at Prelude.Exception.raiseInternal[a](String msg, FSharpList`1 tags) in /home/dark/app/backend/src/Prelude/Prelude.fs:line 173
```

pbiggar: Hmmm, I'm guessing that this test is checking permissions, and specifically that it fails?

Maybe leave it commented but with a `CLEANUP` comment because we'll need to put this back before we ship the web service.

PR #4781, opened 2023-03-09
List.repeat returns Result instead of DError 

ditansu:
Changelog:

```
Area of change
- `List.repeat `
```

Functions using `DError` (sometimes called `err`) or `Exception.raiseCode` should return a Result instead

Swap `DError` to `Result` 

Tests `List.repeat` passed


pbiggar:
Wonderful, thank you!

PR #4782, opened 2023-03-09
Continue adding support for user-defined enum types and values

StachuDotNet:
Changelog:

```
Language and Standard Library
- add base support for stdlib-defined types
- support user- and stdlib-defined enums
- add support for generic types

Internal:
- parser now handles user enum type definitions, enum constructors, and type names`
```

Here's what's been done - some of the below description reflects work in #4773.

- User (and now Stdlib) Records now require at least 1 field
- Enums now require at least 1 case
- Extracted `CustomType.T = | Enum of ... | Record of ...` out of `UserType.Definition`
    - for reusability, as these "custom types" can be defined not only by users/canvases, but also Stdlib and (future) Packages
- Defined `FQTypeName` to refer to defined/custom types
    - previously, only __user__ types existed, and our means of referring to them was an inline `name:string * version: int` pair
    - now we can robustly refer to a stdlib- or user- defined type in a consistent structured way
- Updated `EConstructor` and `DConstructor` to include an `Option`al `FQTypeName` reference.
    - when `None`, this refers to either Option or Result
    - otherwise, it refers to an enum defined either by Stdlib or a User/Canvas
- `TUserType` has been rebranded to `TCustomType`, as we now have Stdlib types to refer to
- Defined `BuiltInType`, in RT and PT
    - These types are used to define Stdlib types in our `LibX.fs` modules (i.e. one is now in `LibMaybe.fs` - see below)
- Added `stdlibTypes` field to `RuntimeTypes.Libraries`
- Updated Parser
  We now pass in `availableTypes` to the parser to be able to successfully parse type references, enum ctors, etc.
    - Added support for parsing User enums
    - type names (`FQTypeName`), using `availableTypes`
    - enum constructors (using the case names of `availableTypes` enums)
- Enum demo: define `type Maybe<'a> = | Nah | Totally 'a` in Stdlib
  Stdlib-defined Enum with at least one type argument
    - demonstrates how `FQTypeName.StdlibTypeName` may be used in _type_ definitions
    - shows a few different ways of matching against DConstructors in a stdlib fn
    - tests parsing of EConstructor, handling of MPConstructor, etc.
- Updated a few places where we serialize Dark types, vals, etc.
  `TCustomType`
    - `LibDarkInternal::?`
    - `DvalReprDeveloper`
    - `DvalReprInternalRoundtrippable`
    - `DarkInternalQueryable`


backend/testfiles/execution/language.tests
```diff
@@ -32,27 +32,49 @@ module Records =
   { col1 = 2 ; col2 = Test.typeError_v0 "test" } = Test.typeError_v0 "test"
   { col1 = 2 ; ___ = 3 } = Test.typeError "Record key is empty"
 
+
 module UserEnums =
```
StachuDotNet, line 36: am I missing classes of basic obvious tests here?

backend/src/Parser/Parser.fs
```diff
@@ -203,7 +206,31 @@ let rec convertToExpr (ast : SynExpr) : PT.Expr =
...
+          |> Option.map (fun _ -> typeName)
+
+        | PT.UserType.Record _ -> None)
+
+    match matchingEnumCases with
+    | [] -> PT.EVariable(id, name.idText)
+    | [ userFnName ] ->
+      let fields =
+        // These are filled in later..
+        // look "EUserEnum" elsewhere in the parser and you'll find it.
```
StachuDotNet, line 226: This could use a better explanation

backend/src/Parser/Parser.fs
```diff
@@ -203,7 +206,31 @@ let rec convertToExpr (ast : SynExpr) : PT.Expr =
...
+        []
+
+      PT.EUserEnum(id, userFnName, name.idText, fields)
+    | _ ->
+      Exception.raiseInternal
+        "Too many things match the identifier"
+        [ "name", name.idText ]
```
StachuDotNet, line 233: Rephrase

backend/src/Parser/Parser.fs
```diff
@@ -473,7 +508,20 @@ let parseTestFile (filename : string) : Module =
     checker.ParseFile(fsharpFilename, Text.SourceText.ofString input, parsingOptions)
     |> Async.RunSynchronously
 
-  let rec convertType (typ : SynType) : PT.DType =
+  // TODO: support non-0 versions of user-defined types
+  let matchingUserTypes
```
pbiggar, line 512: I don't think we should overlap parsing and type matching like this. It's weird if you change the name of your type definition and then the parser stops working.

pbiggar: It would be better if parsing produced the same output regardless of the type in existence.

backend/src/Parser/Parser.fs
```diff
@@ -504,49 +554,59 @@ let parseTestFile (filename : string) : Module =
...
         match argPats with
-        | SynArgPats.Pats pats -> List.map parseArgPat pats
+        | SynArgPats.Pats pats -> List.map (parseArgPat userTypes) pats
         | SynArgPats.NamePatPairs _ ->
           Exception.raiseInternal "Unsupported pattern" [ "pat", pat ]
       (name.idText, parameters)
     | _ -> Exception.raiseInternal "Unsupported pattern" [ "pat", pat ]
 
-  let parseBinding (binding : SynBinding) : PT.UserFunction.T =
+  let parseBinding userTypes (binding : SynBinding) : PT.UserFunction.T =
```
pbiggar, line 591: Similarly, I'm pretty sure we shouldn't be passing userTypes into the parser (like we don't pass userFunctions into the parser)

backend/testfiles/execution/language.tests
```diff
@@ -32,27 +32,49 @@ module Records =
...
 
-  // TODO: fix parser - "A" below is currently interpreted as a variable we can't find.
-  //(let cards = [A; B "test"; C(5)]
-  //
-  // match List.getAt_v1 cards 1 with
-  // | Some b -> Ok b
-  // | None -> Error "Failure") = Ok (B "test")
+  (let a = A
+   let b = B
+   a == b) = false
```
pbiggar, line 42: The true version of this test might have value

backend/testfiles/execution/language.tests
pbiggar: Pattern matching for variants of different sizes I think

pbiggar (changes requested):

backend/src/Parser/Parser.fs
StachuDotNet: Without context of the types, how is the parser supposed to figure out if something is an unnamed variable or an enum case?

StachuDotNet: when parsed as F#, it doesn't present it to our 'converter' as "this is an enum case" - it's just an identifier.  Passing in UserTypes allows us to determine whether it should be an EConstructor or an EVariable. I'm not sure how else we could handle this, but maybe I'm missing something.

pbiggar: I see - that's a tricky problem. I guess there's a similar problem for function names? Or can they tell from context?

StachuDotNet: Yes, there's a related situation with fns right now - `DateTime.now` is parsed as a LongIdent in F# AST, and we currently assume all LongIdents are fn calls. If/when we want to refer to a fn within a specific module, we'll face the same situation.

StachuDotNet: clarification on the above comment:

> we currently assume all LongIdents are fn calls

is not fully true.

This is a better phrasing:
Given a LongIdent:
- if it has 2 components (A.B) and the first character of the first component is uppercase, then it's a fn call\
- (kinda supported: if it has 4 components (A.B.C.D), then it's a package fn call)
- otherwise, it's some sort of field access

pbiggar: I guess in F# they're all values, and the typechecker really is the thing that checks whether it's valid to apply it

pbiggar: I suppose I'd prefer if we had a separate step after the parser to resolve what's a type and what's a function, but I'm not sure that's possible?

StachuDotNet: I'm not sure it's possible either. Maybe it's worth following up on that later? I'm sort of stumped of how we could do this another way.

backend/src/LibExecutionStdLib/LibExecutionStdLib.fsproj
```diff
@@ -23,6 +23,7 @@
     <Compile Include="LibList.fs" />
     <Compile Include="LibMath.fs" />
     <Compile Include="LibNoModule.fs" />
+    <Compile Include="LibMaybe.fs" />
```
StachuDotNet, line 26: This should be deleted, soon. I just think replacing Options/Results with a stdlib-defined enum should be a part of a separate PR, but wanted to have a demo like this for conversation.

StachuDotNet:
https://github.com/darklang/dark/pull/4782/commits/caa345a398d3c8ff6c318c80364f91e5d85c2fe3 a stdlib-defined enum is being parsed OK in tests, and tests against relevant fns pass. :)

At this point, I need to do a thorough review of the work done (both in this PR and the [previous](#4773) one), write up a proper explanation of what's changed and what's next, and get some feedback.

---

Here are some notes+TODOs I've been working against locally - I'll be reorganizing this and addressing some of the TODOs next.

```
PT.CustomType
	This is the structure of some custom/named/defined type
	Currently, there are 2 cases, Enums and Records.
	In the future, I could see additional cases (i.e. "Abbreviation"/"Alias") being a part of this


PT.FQTypeName
	How we reference any Custom type (Enum)


TCustomType
	`| TCustomType of FQTypeName`
	wherever DTypes are relevant, this refers to some Stdlib- or User-defined type


Dvals
	DConstructor
		- used to construct Stdlib enums (Options and Results)
		  if the typeName is `None`
		- 
		- once FQTypeName.T includes a Stdlib case,
		  and ,
		  then the typeName will become non-optional,
		  and DConstructors will be used to construct to Stdlib Enums "normally"

	DObj
		- currently only used for Stdlib
		- will be removed in favor of DRecord instances

	DRecord
		- doesn't exist yet, but will shortly...
		- 


PT.UserType
	{ id: tlid; name: PT.FQTypeName.UserTypeName; Definition: CustomType }
```




Mergeable
- [ ] turn above into a proper PR description
- [x] figure out how to avail stdlib types to the Parser
- [x] update parser to actually use ^ to handle enum values
- [x] review all of the bits in ProgramTypes
- [ ] review the Interpreter changes
- [ ] review the previous PR for any TODOs
- [ ] backfill any missing serialization test values
  - [ ] make sure all FQTypeName cases are represented somewhere
- [ ] add more tests (review everything in the 2 PRs and write down what tests to add, then do it)
  - [ ] probably more in langauge.tests
  - [ ] bunch of type-checking tests
- [ ] discuss "how do we handle non-unique names?" with regards to Type names, variable names, fn names, Enum cases, record structures. There's uncertainty of how to handle this in the Parser, and moreover we should prevent users from conflicting with Stdlid types and such. For now, I think it'd be fine for there to be _no duplicates_ in any user program. add tests ensuring stdlib doesn't have any duplicates, and extend that per canvas/program


Future PRs:
- [ ] consider renaming `DConstructor` - it's a value so it's already ...constructed! 
- [ ] call TypeChecker/unify in many more places
  - [ ] make sure enum cases can't be called with the wrong number of fields
- [ ] support recursively-defined types (esp. user types i.e. enums)
- [ ] expose some stdlib Record type; use it; type-check it
- [ ] expose Option type in Option module
- [ ] expose Result type in Result module

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -5,24 +5,45 @@ type id = Prelude.id
...
+
+// I'm not sure when we'll need this, but probably soon
+// Some of the FQTypeName references might need to be replaced with FQResolvedTypeName ones
+//
+/// A FQTypeName with any relevant named type arguments "applied"
+/// The StdLib-defined type "Option" is an example of a type with type arguments
+/// - the Stdlib defines `Option<a>`, whic has a type argument `a`
+///   This would look like `{ typ: Option, args: ["a"]}
+/// - Map.map is defined as `Map.map<a, b>(f: a -> b, map: Map<a>): Map<b>`
+type FQResolvedTypeName = { typ : FQTypeName.T; args : List<string> }
```
StachuDotNet, line 39: This turned out to be useless - TODO: delete.

StachuDotNet:
I need to break for a bit, but here are the TODOs I'm currently working with:

(some of the todos might need to be moved around

## Reviewable (do before requesting review)
- [ ] Deal with merge conflicts
- [ ] review ProgramContext.allTypes - probably replace usages with something else, then remove allTypes member
- [ ] Fill in RuntimeTypes.typeMatches (very empty for TCustomTypes right now)
- [ ] Review TypeChecker.unify
- [ ] Ensure we have __roundtripping__ and __persisted vanilla serialization__ tests of
    - [ ] `CustomType`
    - [ ] `FQTypeName`
    - [ ] any other new things
- [ ] consider removing redundant `Definition` from `UserType`, and just reference `CustomType` directly
    - [ ] could then flatten `UserType` module def to just a type def
- [ ] Define/test a user type that references a StdlibType
  `language.tests, UserEnum, type SomeType<x> = | HasMaybe of Maybe<x>`
- [ ] Do a final pass-through on all relevant files. anything to tidy up before requesting review?
- [ ] do a final pass through on all relevant files. Anything else that's been done I haven't noted in "done" section?
- [ ] update PR description
    - update triple-\` `Changelog`

## Mergeable (I'd like some input/help on some of these)

- [ ] review language.tests
- [ ] actually type-check \#/type of enum fields
  we have some code written against this but it's not run
- [ ] ensure we have 'id's where we need them
    - [ ] in all places where we might have live values
        - [ ] do we need them in `BuiltInType` at all?
    - [ ] where we might want to specify an operation per some thing, by `id` (i.e. change the label of enum case X, by id 123)
        - [ ] I think we're covered here, but worth thinking through
- [ ] do stdlib type names need __versions__? right now I've assumed no, but maybe?
- [ ] Should MPConstructor have a UserTypeName in it?
    - [ ] if so, then the interpreter might need to ensure that the DVal's typename matches the MP's typename?
      enh, I guess this is a type checking thing
- [ ] Review Interpreter handling of `EConstructor`
    - [ ] preview/analysis
    - [ ] fake val handling
- [ ] Review all usages of `allTypes`, and consider if they need an update
- [ ] extract out any remaining TODOs to some Issue or Notion doc

StachuDotNet:
## Later (work to be done as future PRs - or never done, some of these are just ideas)
(these should be reviewed and moved to a GitHub issue or notion doc)

- [ ] Consolidate ST2PT and PT2ST
- [ ] consider renaming "`CustomType`" to something else
    - `DefinedType`?
- [ ] Port our baked-in Enums to stdlib-defined `CustomType`s
    - [ ] `Option` -> `CustomType` enum
        - [ ] add definition to top of `LibOption.fs`
        - [ ] remove `TOption`
        - [ ] remove `DOption`
        - [ ] remove some bits from Parser
    - [ ] `Result` -> `CustomType` enum
        - Copy what was done for Option
- [ ] Improve support for Records
- [ ] Remove `Option`ality of FQTypeName references
  i.e. ensure Dark is fully typed!
    - [ ] `ECtor`
    - [ ] `ERecord`
- [ ] Extend `FQTypeName` to include Package types
- [ ] support recursively-defined custom types
  e.g. `type Tree<'a> = | Leaf | Node of 'a * Tree<'a> * Tree<'a>`
- [ ] Figure out what to do about duplicate names
  a test, for stdlib?
  type check?
    - type names
    - enum cases
    - fns
    - bindings (vars)
    - keywords (`if`)
    - reserved (`while`)
- [ ] Ensure we don't have 2 of the same `id` used in type definitions
  see `LibMaybe`
- [ ] Write typed JSON stdlib fns
    - `Json::parse<'a>`
    - `Json::serialize`
      we could write this in Darklang
- [ ] Add `Abbreviation`/`Alias` case of `CustomType`
- [ ] Expose Darklang (PT, RT) in Stdlib/package
- [ ] Write Eval in Dark
  `fn eval(expr: PT.Expr, ...): RT.Dval`
- [ ] Port F# libraries into Dark packages
    - Lexer/parser might be handy

PR #4788, opened 2023-03-11
Strict supporting ISO8601 format in Date.Parse_v2

ditansu:
```
Area of change
- Date.Parse_v2
```
`Date.parse_v2 `- has to support only that one format of dates 

Accept only one format, `"yyyy-MM-ddTHH:mm:ssZ"`

TDD approach, red tests for the expected behavior, refactoring, tests passed


backend/src/LibExecutionStdLib/LibDateTime.fs
```diff
@@ -16,97 +16,33 @@ let product lists =
...
-    Error()
+let ISO8601Format = "yyyy-MM-ddTHH:mm:ssZ"
+
+let ISO8601DateParser (s : string) : Result<DarkDateTime.T, unit> =
+
+  let culture = System.Globalization.CultureInfo.InvariantCulture
+  let styles = System.Globalization.DateTimeStyles.AssumeUniversal
+  let mutable (result : System.DateTime) = Unchecked.defaultof<System.DateTime>
+  if
+    not (s.EndsWith('z') || s.Contains("GMT"))
```
pbiggar, line 27: Do we need this condition?

backend/src/LibExecutionStdLib/LibDateTime.fs
```diff
@@ -16,97 +16,33 @@ let product lists =
...
-
-  // Cartesian product
-  [ dateFieldSeparators
-    dateTimeSeparator
-    timeSeparator
-    seconds
-    meridians
-    millisecondFormat
-    tzSuffixes ]
-  |> product
```
pbiggar, line 44: product (and maybe some other functions) I think are no longer used?

pbiggar (changes requested):
Great! Some small changes here and we should be good to go.

backend/src/LibExecutionStdLib/LibDateTime.fs
ditansu: @pbiggar Heh, you partial right!  We need only `s.Contains("GMT")` because I don't know why the `System.DateTime.TryParseExact` with` ISO8601Format = "yyyy-MM-ddTHH:mm:ssZ"` successfully parse this string:  `"2008-09-15T09:30:41GMT" ` to  DateTime `"2008-09-15T09:30:41Z"` I drop redundant condition and push the fix asap.

pbiggar: Thanks! Just thinking - do we currently allow non "Z" suffixes like "z" or "UTC" or "EST"? Would you be able to add test cases to disallow them?

pbiggar (approved):
Looks great, thanks!

I'll merge this once CI is running again.

PR #4789, opened 2023-03-11
Float clamp and Int power return Result

ditansu:
```
- Float clamp and Int power 
```

Modified tests were passed.


pbiggar:
Great, thank you!

PR #4790, opened 2023-03-11
String.padStart and String.padEnd return Result

ditansu:
```
Area of change
- String.padStart 
- String.padEnd
```

Modified tests were passed.


pbiggar:
Great, thank you!

PR #4791, opened 2023-03-11
HttpClient.request returns a Result in all cases

ditansu:
```
- HttpClient.request
```

Modified tests were passed.

### DIFF
--- a/backend/src/BackendOnlyStdLib/LibHttpClient.fs
+++ b/backend/src/BackendOnlyStdLib/LibHttpClient.fs
@@ -293,7 +293,7 @@ let fns : List<BuiltInFn> =
 
               | Error (HttpClient.BadUrl details) ->
                 // TODO: include a DvalSource rather than SourceNone
-                return DError(SourceNone, $"Bad URL: {details}")
+                return DResult(Error(DStr $"Bad URL: {details}"))
 
               | Error (HttpClient.Timeout) ->
                 return DResult(Error(DStr $"Request timed out"))
@@ -306,11 +306,11 @@ let fns : List<BuiltInFn> =
             }
 
           | Error reqHeadersErr, _ ->
-            uply { return DError(SourceNone, reqHeadersErr) }
+            uply { return DResult(Error(DStr reqHeadersErr)) }
 
           | _, None ->
             let error = "Expected valid HTTP method (e.g. 'get' or 'POST')"
-            uply { return DError(SourceNone, error) }
+            uply { return DResult(Error(DStr error)) }
 
         | _ -> incorrectArgs ())
       sqlSpec = NotQueryable
--- a/backend/src/BackendOnlyStdLib/LibHttpClient.fs
+++ b/backend/src/BackendOnlyStdLib/LibHttpClient.fs
@@ -217,6 +217,9 @@ let fn = FQFnName.stdlibFnName
 
 let headersType = TList(TTuple(TStr, TStr, []))
 
+type HeaderError =
+  | BadInput of string
+  | TypeMismatch of string
 
 
 let fns : List<BuiltInFn> =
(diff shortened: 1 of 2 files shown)


pbiggar:
@StachuDotNet can you look at this?

StachuDotNet:
@ditansu Thank you!

Most of these look great, but I'm a bit hesitant on the `HttpClient.request "get" "https://darklang.com" [1] Bytes.empty` test.

The request here involves an actual type error so a normal DError doesn't make sense here, I think. (i.e. the old test seems more correct)
That said, we've sort of been removing tests that second-guess the type-checking, so maybe we could remove the test? Or, we could leave this as-is. What do you think, @pbiggar?

backend/src/BackendOnlyStdLib/LibHttpClient.fs
```diff
@@ -306,11 +306,11 @@ let fns : List<BuiltInFn> =
             }
 
           | Error reqHeadersErr, _ ->
-            uply { return DError(SourceNone, reqHeadersErr) }
+            uply { return DResult(Error(DStr reqHeadersErr)) }
```
pbiggar, line 309: This is a type-error (which can evenually be caught statically), so this should remain a DError

pbiggar:
> That said, we've sort of been removing tests that second-guess the type-checking, so maybe we could remove the test? What do you think, @pbiggar?

This test might be fine to keep, as there's a lot of code and it's a change from the old behavior.

But you're right that type errors should be DErrors

backend/testfiles/execution/httpclient.tests
```diff
@@ -25,8 +25,8 @@ HttpClient.basicAuth_v1 "" ":" = { Authorization = "Basic Ojo=" }
 // Tests that try to make requests to the internet
 
 // type errors for bad `headers` are OK
-HttpClient.request "get" "https://darklang.com" [1] Bytes.empty = Error "Expected request headers to be a List of (string * string), but got: 1"
-HttpClient.request "get" "https://darklang.com" [("", "")] Bytes.empty = Error "Empty request header key provided"
+HttpClient.request "get" "https://darklang.com" [1] Bytes.empty = Test.typeError_v0 "Expected request headers to be a List of (string * string), but got: 1"
+HttpClient.request "get" "https://darklang.com" [("", "")] Bytes.empty = Test.typeError_v0 "Empty request header key provided"
```
StachuDotNet: The first test here now looks right, but the second one is now wrong I think - an empty string is still a string, so we're not hitting a "type error there," so such should return a `DResult(Error` instead of an `Error`.

In the implementation, where we gather `reqHeaders`, I think we need to change the implementation a bit - I've added some `//` commentary in-line, but suspect we'll need a refactor here to match the expectations.

```fsharp
let reqHeaders : Result<List<string * string>, string> =
  reqHeaders
  |> List.fold (Ok []) (fun agg item ->
    match agg, item with
    | (Error err, _) -> Error err
    | (Ok pairs, DTuple (DStr k, DStr v, [])) ->
      // TODO: what about whitespace? What else can break?
      if k = "" then
        // This case should result in a DResult - it's a valid string, just "bad" input
        Error "Empty request header key provided"
      else
        Ok((k, v) :: pairs)

    | (_, notAPair) ->
      // This case should result in an Error (not DResult)
      // (this is currently resulting in the "correct" behaviour)
      Error $"Expected request headers to be a List of (string * string), but got: {DvalReprDeveloper.toRepr notAPair}")
      |> Result.map (fun pairs -> List.rev pairs)
```

ditansu: @StachuDotNet thanks for comment! I hope understood you correctly. Please look on my last commit.  Probably my solving is naive, but it works :)

StachuDotNet: I think the implementation is great, not naive :)

StachuDotNet (approved):

PR #4793, opened 2023-03-14
LibCrypto merge

ditansu:
Changelog:

```
Area of change
- LibCrypto.fs
```

Move all hash functions from BackendOnly area to common place in backend/src/LibExecutionStdLib/LibCrypto.fs

Merge function and delete redundant backend/src/BackendOnlyStdLib/LibCrypto.fs

Tests passed


pbiggar (approved):
Looks great, thanks!

I'll merge this once CI is running again.

PR #4796, opened 2023-03-16
Cleanup items related to string interpolation

OceanOak:
No changelog

This one doesn't seem right to keep this way, but since we are deleting FuzzTests I don't think that we need to fix it.
https://github.com/darklang/dark/blob/main/backend/tests/FuzzTests/ExecutionRegression.FuzzTests.fs#L65


backend/tests/TestUtils/TestUtils.fs
```diff
@@ -566,7 +566,7 @@ module Expect =
     // expressions with no values
     | EUnit _, EUnit _ -> ()
     // expressions with single string values
-    | EString (_, v), EString (_, v') -> () //CLEANUP
+    | EString (_, v), EString (_, v') -> ()
```
pbiggar, line 569: I think this function will need to look inside string segments.

pbiggar (changes requested):
Great, thanks! I think a few of these places could do with expansion so that we take care of the interpolated strings in tests and serialization.

backend/tests/TestUtils/TestUtils.fs
```diff
@@ -566,7 +566,7 @@ module Expect =
     // expressions with no values
     | EUnit _, EUnit _ -> ()
     // expressions with single string values
-    | EString (_, v), EString (_, v') -> ()
+    | EString (_, s), EString (_, s') -> check path s s'
```
pbiggar, line 569: So what's going on here is that `s` is a list of string parts and interpolated expressions. We need to compare each of them, using `check` to check equality for the string parts, and `eq` to check equality on the interpolated expressions.

pbiggar (changes requested):
I left a comment, but you probably also need to check in new serializations generated by running tests locally.

OceanOak:
> I left a comment, but you probably also need to check in new serializations generated by running tests locally.

When running the backend tests, new serializations aren't being generated, which suggests that something is not working as expected right?

pbiggar:
> When running the backend tests, new serializations aren't being generated, which suggests that something is not working as expected right?

Yeah, that's odd. New ones will only appear if they're different from the old ones maybe?

pbiggar:
Nice job getting this over the line.

PR #4803, opened 2023-03-20
WIP Allow only callable values in pipe

ditansu:
```
-  ProgramType to RuntimeType tranformation for EPipe 
```

Allow only callable values in pipe: function, lambda, infix and/or etc.  

Mark wrong tokens as NotInPipe - super naive solution but it works now. 

Write tests 

There is the issue - a pipe execution break on last wrong value, e.g 

```F#
  (51
  |> 57
  |> 34.2
  |> (+) 23) = Test.typeError_v0 "Expected a function value, got something else: 34.2"
  // below the correct expected result but I don't know how breake pipe on the 1st wrong expression
  // |> (+) 23) = Test.typeError_v0 "Expected a function value, got something else: 57"
```

### DIFF
--- a/backend/src/LibExecution/ProgramTypesToRuntimeTypes.fs
+++ b/backend/src/LibExecution/ProgramTypesToRuntimeTypes.fs
@@ -153,7 +153,7 @@ module Expr =
               match op with
               | PT.BinOpAnd -> RT.EAnd(id, prev, toRT expr2)
               | PT.BinOpOr -> RT.EOr(id, prev, toRT expr2)
-            | other -> RT.EApply(pipeID, toRT other, [ prev ], RT.InPipe pipeID)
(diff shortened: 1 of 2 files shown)


ditansu:
@pbiggar please look at this commit and give me feedback - am I right direction or not?

pbiggar:
Directionally, the goal here is to change the internal format (ProgramTypes.Expr) of Pipes so that we can only represent types we support in pipes. So the outcome would be a definition of EPipes like this:

```
| EPipes of id * PipeExpr * PipeExpr * List<PipeExpr>
```

ditansu:
@pbiggar that exactly I wanted to do because it is more elegant, graceful and "typed".  That's why I spent so much time for research how to work the whole system of types. But when I found out how many places of code would be affected, I decided to find a more simple and safe way :) Sure now when you approve "typed" way I will rewrite solution on types. It will be an awesome experience... I hope :)

PR #4808, opened 2023-03-22
Support type arguments in stdlib fns

StachuDotNet:
Changelog:

```
Language and Standard Library
- support type arguments in stdlib fn definitions and fn calls
  i.e. support calls like `Json::parse<int> "1"`
```

I have some inline TODOs written here to still address, but open to feedback in the meantime.

- `PT.EFnCall` now includes a `typeArgs: List<DType>` field, as does `RT.EApply`
- A 'typeArgs' field is now present in our models of functions: BuiltInFn, UserFunction.T, PackageFn.T
- The interpreter passes the type args through from `PT.EApply`  function calls, and ensures that the expected # of type args have been supplied
- `fnName<TypeArg> arg` is now parse-able
- a tiny implementation of `Json::parse<T>` and `Json::serialize<T>` is included to demonstrate both definition and usage of type args

---

As part of this PR, I refactored the Parser a bit so that I could understand it better. If you feel strongly that this is in the wrong direction, I'm open to reverting, but I think that this is for the better, and the `FSharpToProgramTypes` module matches the style of our other 'mappings' throughout the codebase.


backend/src/LibExecutionStdLib/LibJson.fs
```diff
@@ -13,4 +9,38 @@ let fn = FQFnName.stdlibFnName
 
 let incorrectArgs = Errors.incorrectArgs
 
-let fns : List<BuiltInFn> = []
+let varA = TVariable "a"
+
+
+let fns : List<BuiltInFn> =
+  [ { name = fn "Json" "serialize" 0
+      typeArgs = [ TVariable "a" ]
```
StachuDotNet, line 17: I feel like there's a use to this being a `List<DType>`, but can't put my finger on why, which makes me think typeArgs here could simply be an `int` of the expected count.

backend/src/LibExecutionStdLib/LibNoModule.fs
```diff
@@ -117,8 +117,10 @@ and equalsExpr (expr1 : Expr) (expr2 : Expr) : bool =
   | EFieldAccess (_, target1, fieldName1), EFieldAccess (_, target2, fieldName2) ->
     equalsExpr target1 target2 && fieldName1 = fieldName2
   | EVariable (_, name1), EVariable (_, name2) -> name1 = name2
-  | EApply (_, name1, args1, isInPipe1), EApply (_, name2, args2, isInPipe2) ->
+  | EApply (_, name1, typeArgs1, args1, isInPipe1),
+    EApply (_, name2, typeArgs2, args2, isInPipe2) ->
     name1 = name2
+    //&& List.forall2 equalsDType typeArgs1 typeArgs2 // TODO: add equalsDType
```
StachuDotNet, line 123: TODO

backend/tests/TestUtils/TestUtils.fs
```diff
@@ -190,12 +190,13 @@ let testUserFn
     description = ""
     infix = false
     name = name
-    returnType = PT.TVariable "a"
+    typeArgs = [] // TODO: review
```
StachuDotNet, line 193: TODO

backend/src/LibExecution/Interpreter.fs
```diff
@@ -628,23 +630,42 @@ and callFn
...
+          let expectedTypeParamLength = List.length fn.typeArgs
+          let expectedArgLength = List.length fn.parameters
+
+          let actualTypeArgLength = List.length typeArgs
+          let actualArgLength = List.length argvals
+
+          let err errMsg = DError(sourceID, errMsg)
 
-          if expectedLength = actualLength then
+          match (expectedTypeParamLength = actualTypeArgLength),
```
StachuDotNet, line 642: TODO: add test cases around all 3 error messages (currently, only the arg # mismatch is tested).

backend/src/LibExecution/ProgramTypesToRuntimeTypes.fs
```diff
@@ -142,10 +178,13 @@ module Expr =
           let rec convert thisExpr =
             match thisExpr with
             // TODO: support currying
-            | PT.EFnCall (id, fnName, PT.EPipeTarget ptID :: exprs) ->
+
+            | PT.EFnCall (id, fnName, typeArgs, PT.EPipeTarget ptID :: exprs) ->
+              // TODO: consider using typeArgs
```
StachuDotNet, line 183: TODO: remove this comment (it's since been used)

backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -138,6 +138,54 @@ module DarkDateTime =
   let toIsoString (d : T) : string = (toInstant d).toIsoString ()
 
 
+/// Dark runtime type
+type DType =
```
StachuDotNet, line 142: Note: in a few files, I had to move the 'dtype' module/type def above the Expr module/type def, because EApply now references DType

backend/tests/Tests/TypeChecker.Tests.fs
```diff
@@ -84,6 +85,6 @@ let testArguments : Test =
       (("myAnyFn", RT.TVariable "a", S.eInt 5), RT.DInt 5L) ]
 
 
-
+// TODO: add tests around type args
```
StachuDotNet, line 88: TODO: add some typeChecker tests around type args (_could_ wait for another PR, if not trivial)

backend/src/LibBackend/PackageManager.fs
```diff
@@ -387,9 +387,10 @@ let allFunctions () : Task<List<PT.Package.Fn>> =
                  "Cannot parse returnType"
                  [ "type", returnType ]
           { name = name
-            body = expr
-            returnType = returnType
+            typeArgs = [] // TODO: review
```
StachuDotNet, line 390: TODO

backend/src/LibBackend/SqlCompiler.fs
```diff
@@ -206,7 +206,8 @@ let rec lambdaToSql
     lambdaToSql fns symtable paramName dbFields typ e
 
   match expr with
-  | EApply (_, FnName name, args, _) ->
+  | EApply (_, FnName name, _typeArgs, args, _) ->
+    // TODO: should the typeArgs be used for anything here?
```
StachuDotNet, line 210: TODO

backend/src/LibBackend/SqlCompiler.fs
```diff
@@ -184,7 +184,7 @@ let rec inline'
 
 let (|Fn|_|) (mName : string) (fName : string) (v : int) (expr : Expr) =
   match expr with
-  | EApply (_, FnName (FQFnName.Stdlib std), args, _) when
+  | EApply (_, FnName (FQFnName.Stdlib std), _typeArgs, args, _) when
```
StachuDotNet, line 187: TODO: see if there's any use for typeArgs here

backend/src/LibBackend/SqlCompiler.fs
```diff
@@ -482,7 +483,9 @@ let partiallyEvaluate
           name1 <> paramName && name2 <> paramName
           ->
           return! exec expr
-        | EApply (_, _, args, _) ->
+        | EApply (_, _, _typeArgs, args, _) ->
+          // TODO: consider if typeArgs is relevant here at all
```
StachuDotNet, line 487: TODO

backend/src/Parser/FSharpToProgramTypes.fs
```diff
@@ -0,0 +1,762 @@
...
+      // with type args
+      | "List", [ arg ] -> PT.TList(c arg)
+      | "Option", [ arg ] -> PT.TOption(c arg)
+
+      // Some user- or stdlib- type
+      | _ ->
+        match matchingCustomTypes availableTypes name with
+        | [ matchedType ] -> PT.TCustomType(matchedType, List.map c typeArgs)
+        | _ ->
+          Exception.raiseInternal $"Unsupported type" [ "name", name; "type", typ ]
```
StachuDotNet, line 68: TODO: note in the err msg that we didn't find exactly 1 matching type (could be 0, or 2+)

backend/src/Parser/FSharpToProgramTypes.fs
```diff
@@ -0,0 +1,762 @@
...
+    : PT.Package.Fn =
+    let availableTypes = [] // eventually, we'll likely need package types supported here
+    let userFn = UserFunction.fromSynBinding availableTypes binding
+    { name =
+        { owner = p1
+          package = p2
+          module_ = p3
+          function_ = userFn.name
+          version = 0 }
+      typeArgs = [] // TODO: review
```
StachuDotNet, line 752: TODO

backend/tests/Tests/TypeChecker.Tests.fs
StachuDotNet: leaving the equivalent call-out TODO in language.tests, and removing this TODO

backend/src/LibBackend/SqlCompiler.fs
StachuDotNet: I don't know what you mean by this - do you mean we should only allow the apply if the fn has no typeArgs (i.e. `typeArgs = []`)?

StachuDotNet: Ah, I see your comment from SqlCompiler now - came out of order in github UI

backend/src/LibExecutionStdLib/LibJson.fs
StachuDotNet: Thinking out loud:

- functions may take 0 or more type params
- the implementation of a stdlib function will expect that exact # of arguments, and won't worth if another is provided
- it's useful to have this # defined per function, so that the type checker can complain if another # of type args was provided. (currently handled in the interpreter via the "ensure we have the expected # of typeArguments _and_ arguments values" error message)
- right now, I'm modeling this via a `List<DType>`
- but that doesn't really make sense - if anything, a list here would be a list of _constraints_, if/when we support that eventually
  e.g. `let min<'T when 'T : comparison> (x:'T) (y:'T) = if x < y then x else y
- since we don't support type constraints right now, all fn type (definition) parameters are basically "any" constraints
- long term, I could see this being `typeArgs: List<TypeConstraint>`, with `type TypeConstraint = None | Comparable | ...`
- for now, we can just model we could model as an `int` capturing the expected type arg count, _or_ capture this with a tiny `List<TypeConstraint`, with `type TypeConstraint = | NoConstraint`

backend/src/Parser/Parser.fs
```diff
@@ -904,18 +241,18 @@ let parseTestFile
 
 
 let parsePTExprWithTypes
-  (availableTypes : Map<PT.FQTypeName.T, PT.CustomType.T>)
+  (availableTypes : List<PT.FQTypeName.T * PT.CustomType.T>)
   (code : string)
   : PT.Expr =
-  code |> parse |> convertToExpr availableTypes
+  code |> parse |> FS2PT.Expr.fromSynExpr availableTypes
```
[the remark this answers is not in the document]
StachuDotNet: I tried another layout between the two styles - let me know what you think?
I agree it was a bit ugly, but I was having a hard time with the old structure

backend/src/Parser/Parser.fs
```diff
@@ -904,18 +241,18 @@ let parseTestFile
 
 
 let parsePTExprWithTypes
-  (availableTypes : Map<PT.FQTypeName.T, PT.CustomType.T>)
+  (availableTypes : List<PT.FQTypeName.T * PT.CustomType.T>)
```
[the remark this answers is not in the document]
StachuDotNet: Hm. I was using it as a list so I could do this:

```fsharp
module DType =
  let rec fromNameAndTypeArgs
    ...
    let matchingCustomTypes =
      availableTypes
      |> List.choose (fun (typeName, _def) ->
        match typeName with
        | PT.FQTypeName.User u -> if u.typ = name then Some typeName else None
        | PT.FQTypeName.Stdlib t -> if t.typ = name then Some typeName else None)

    match matchingCustomTypes with
    | [] -> PT.TVariable(name)
    | [ matchedType ] -> PT.TCustomType(matchedType, List.map fromSynType typeArgs)
    | _ ->
      Exception.raiseInternal
        $"Matched against multiple custom types - not sure what to do"
        [ "name", name; "typeArgs", typeArgs ]
```

(complain if more than one custom types matched some name)

I'm not sure how I'd replace this if `availableTypes` was a `Map<PT.FQTypeName.T, PT.CustomType.T>`
I feel like I'm missing something obvious - am I?

backend/src/LibExecutionStdLib/LibJson.fs
StachuDotNet: added another commit which might be overkill, but at least makes more sense.

StachuDotNet:
some quick thoughts:
- [ ] the `typeArgs` of builtin fns as a List<String> doesn't really make sense. Maybe this is something like a TypeConstraint
- [ ] maybe TCAny of TypeConstraint needs an `of name: string` component
- [ ] can kill that one DType.fromXYZ fn in Parser.fs

backend/src/LibExecutionStdLib/LibNoModule.fs
```diff
@@ -79,6 +79,33 @@ let rec equals (a : Dval) (b : Dval) : bool =
   | DError _, _
   | DIncomplete _, _ -> Exception.raiseCode "Both values must be the same type"
 
+and equalsDType (actual : DType) (expected : DType) : bool =
```
pbiggar, line 82: Isn't this just `actual = expected`?

backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -186,6 +186,15 @@ type DType =
     | TFn _ -> true
     | _ -> false
 
+/// Used to constrain the type of a variable
+///
+/// e.g. `when 'T: comparison` in
```
pbiggar, line 191: I really don't think we should go down this road (at the very least without a plan, but maybe not even then). I think simply naming the type parameters should be plenty.

pbiggar: I agree these aren't DTypes though.

StachuDotNet: Agreed, see https://github.com/darklang/dark/pull/4808#issuecomment-1480640133 - removing

backend/src/LibExecutionStdLib/LibNoModule.fs
StachuDotNet: Yeah. I was trying (maybe in vain) to catch any future case in which DType might have an `id` component to it, to make sure comparisons wouldn't consider the ID. But, I doubt it ever will, so this can be simplified.

PR #4814, opened 2023-03-27
Typed pipe

ditansu:
@pbiggar hey, need code review and advise.  I added PipeExpr and resolve all issues with types, but probably I something did wrong, because when I try to start the tests and I get runtime error. I have no idea how to troubleshoot here. Could you advise something? 

```
dark@dark-dev:~/app$ ./scripts/run-backend-tests
Tests: no process found
Dropping testdb database
Waiting for postgresql
14/main (port 5432): online
Running PubSub emulator (/home/dark/app/rundir/logs/pub-sub-emulator.log)
Finished loading PubSub emulator
Running Cloud Storage emulator (/home/dark/app/rundir/logs/cloud-storage-emulator.log)
Finished loading Cloud Storage emulator
pkill tunnel
rm tunnel
pkill sleep
Starting tunnel
Started tunnel
Running migrations (/home/dark/app/rundir/logs/test-migrations.log)
Initing LibService in Tests
Configuring Rollbar
 Configured rollbar
Configuring Telemetry
 Configured Telemetry
 Inited LibService in Tests
Initing LibBackend in Tests
 Inited LibBackend in Tests
Initing LibRealExecution in Tests
 Inited LibRealExecution in Tests
Initing LibBackend.Account in Tests
Inited  LibBackend.Account in Tests
Outer exception: error: error deserializing toplevel
Outer exception:  tlid: 0UL
Outer exception: exceptionType: Prelude+InternalException
Outer exception:    at LibBinarySerialization.BinarySerialization.wrapSerializationException[a](UInt64 tlid, FSharpFunc`2 f) in /home/dark/app/backend/src/LibBinarySerialization/BinarySerialization.fs:line 47
   at LibBinarySerialization.BinarySerialization.serializeOplist(UInt64 tlid, FSharpList`1 oplist) in /home/dark/app/backend/src/LibBinarySerialization/BinarySerialization.fs:line 71
   at Tests.BinarySerialization.ConsistentSerializationTests.formats@51.Invoke(FSharpList`1 oplist)
   at Tests.BinarySerialization.ConsistentSerializationTests.generateTestFiles() in /home/dark/app/backend/tests/Tests/Serialization.Binary.Tests.fs:line 72
   at Tests.BinarySerialization.generateTestFiles() in /home/dark/app/backend/tests/Tests/Serialization.Binary.Tests.fs:line 120
   at Tests.All.main(String[] args) in /home/dark/app/backend/tests/Tests/Tests.fs:line 68
prefex.inner[0]: error: Failed to serialize Microsoft.FSharp.Collections.FSharpList`1[[LibBinarySerialization.SerializedTypes+Op, LibBinarySerialization, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]] value.
prefex.inner[0]: exceptionType: MessagePack.MessagePackSerializationException
prefex.inner[0]:    at MessagePack.MessagePackSerializer.Serialize[T](MessagePackWriter& writer, T value, MessagePackSerializerOptions options)
   at MessagePack.MessagePackSerializer.Serialize[T](T value, MessagePackSerializerOptions options, CancellationToken cancellationToken)
   at LibBinarySerialization.BinarySerialization.serializeOplist@71.Invoke(Unit unitVar0) in /home/dark/app/backend/src/LibBinarySerialization/BinarySerialization.fs:line 73
   at LibBinarySerialization.BinarySerialization.wrapSerializationException[a](UInt64 tlid, FSharpFunc`2 f) in /home/dark/app/backend/src/LibBinarySerialization/BinarySerialization.fs:line 43
prefex.inner[1]: error: The type initializer for 'FormatterCache`1' threw an exception.
prefex.inner[1]: exceptionType: System.TypeInitializationException
prefex.inner[1]:    at MessagePack.FSharp.DynamicUnionResolver.GetFormatter[T]()
   at MessagePack.FSharp.FSharpResolver.FormatterCache`1..cctor()
--- End of stack trace from previous location ---
   at MessagePack.FormatterResolverExtensions.Throw(TypeInitializationException ex)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.Formatters.TupleFormatter`2.Serialize(MessagePackWriter& writer, Tuple`2 value, MessagePackSerializerOptions options)
   at MessagePack.Formatters.CollectionFormatterBase`4.Serialize(MessagePackWriter& writer, TCollection value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.ExprFormatter2.Serialize(MessagePackWriter& writer, Expr value, MessagePackSerializerOptions options)
   at MessagePack.Formatters.LibBinarySerialization_SerializedTypes\.Handler\.TFormatter1.Serialize(MessagePackWriter& writer, T value, MessagePackSerializerOptions options)
   at MessagePack.FSharp.Formatters.LibBinarySerialization_SerializedTypes\.OpFormatter1.Serialize(MessagePackWriter& writer, Op value, MessagePackSerializerOptions options)
   at MessagePack.Formatters.CollectionFormatterBase`4.Serialize(MessagePackWriter& writer, TCollection value, MessagePackSerializerOptions options)
   at MessagePack.MessagePackSerializer.Serialize[T](MessagePackWriter& writer, T value, MessagePackSerializerOptions options)
prefex.inner[2]: error: can't find matched method parameter, index not found. case:EPipeFnCall parameterName:_typeArgs
prefex.inner[2]: exceptionType: MessagePack.FSharp.Internal.MessagePackDynamicUnionResolverException
prefex.inner[2]:    at MessagePack.FSharp.Internal.UnionSerializationInfo.CreateOrNull(Type type, UnionCaseInfo caseInfo)
   at MessagePack.FSharp.DynamicUnionResolver.BuildSerialize(Type type, UnionCaseInfo[] infos, MethodBuilder method, FieldBuilder[] stringByteKeysFields, ILGenerator il, Int32 firstArgIndex)
   at MessagePack.FSharp.DynamicUnionResolver.BuildType(Type type)
   at MessagePack.FSharp.DynamicUnionResolver.FormatterCache`1..cctor()
```


pbiggar:
Hmm, maybe add `[<MessagePack.MessagePackObject>]` to the definition of PipeExpr?

Code looks great!

ditansu:
> Hmm, maybe add `[<MessagePack.MessagePackObject>]` to the definition of PipeExpr?

Looks like it is already added [here](https://github.com/darklang/dark/pull/4814/files#diff-2b14100865b3962b377ccbdbaf7f7d266aa4ba205cfc982ac514a84899fd7043R176)  
 
> Code looks great!

thanks :)

pbiggar:
> > Hmm, maybe add `[<MessagePack.MessagePackObject>]` to the definition of PipeExpr?
> 
> Looks like it is already added [here](https://github.com/darklang/dark/pull/4814/files#diff-2b14100865b3962b377ccbdbaf7f7d266aa4ba205cfc982ac514a84899fd7043R176)

My thinking is that perhaps that annotation doesn't extend to the `and`ed types?

It looks like there's a deserialization issue so this is where I looked first.

Another alternative is that perhaps there is data in the DB that doesn't deserialize correctly, but this is the test DB so it should be cleared on each run.

ditansu:
> My thinking is that perhaps that annotation doesn't extend to the `and`ed types?
> 
> It looks like there's a deserialization issue so this is where I looked first.
> 
> Another alternative is that perhaps there is data in the DB that doesn't deserialize correctly, but this is the test DB so it should be cleared on each run.

How to purge the DB?

pbiggar:
I think it's cleared:

> Dropping testdb database

ditansu:
> I think it's cleared. From the test output:
> 
> > Dropping testdb database

It's definitely not a tesdb. I've dropped db and got the same error.

```shell 
dark@dark-dev:~/app$ dropdb testdb
dark@dark-dev:~/app$ dropdb testdb
dropdb: error: database removal failed: ERROR:  database "testdb" does not exist
dark@dark-dev:~/app$ DEBUG=True ./scripts/run-backend-tests
...
prefex.inner[2]: error: can't find matched method parameter, index not found. case:EPipeFnCall parameterName:_typeArgs
```

And probably it is not a serialize error. It seems like the parser doesn't understand how to work with PipeExpr  (e.g,  EPipeFnCall)  It is really weird because everywhere I convert back  PipeExpr to Expr thus PipeExpr can't appear in the parser.

pbiggar:
Yeah, that's part of the binary serialization. Did you try adding the annotation to the PipeExpr type? If it's not that, we can think about what other part of the serialization might cause this.

ditansu:
> Yeah, that's part of the binary serialization. Did you try adding the annotation to the PipeExpr type? If it's not that, we can think about what other part of the serialization might cause this.

Yes, you're right, it is a serialization problem. `PipeExpr` is a part of `Expr` (because they are mutual recursive types) and probably annotation doesn't work with `and`, see attached screenshot. 
image

Text of error message 
/home/dark/app/backend/src/LibBinarySerialization/SerializedTypes.fs(204,1): error FS0010: Unexpected keyword 'and' in definition [/home/dark/app/backend/src/LibBinarySerialization/LibBinarySerialization.fsproj]
/home/dark/app/backend/src/LibBinarySerialization/SerializedTypes.fs(233,3): error FS0010: Incomplete structured construct at or before this point in implementation file [/home/dark/app/backend/src/LibBinarySerialization/LibBinarySerialization.fsproj]
    0 Warning(s)
    2 Error(s)

ditansu:
Tests show that **MessagePack** support `and` notation 

image

pbiggar:
I think the annotation would go between the `and` and the type name.

ditansu:
> I'm really struggling to see what this might be. The obvious thing I think is it might be a bug with the MessagePack F# extensions?

Oh gosh, you're totally right.  
Code in the snippet below was crashed. It was one of my 1st attempt when you 1st time offered to use annotation. After that I went to wrong way :)  
```F#
...
and 
PipeExpr =
  | EPipeLambda of id * List<id * string> * Expr
  | EPipeInfix of id * Infix * Expr * Expr
  | EPipeFnCall of id *
...
```
But this one work is correct. The mistake is obvious - I forgot about indents 🤦🏻‍♂️
```F#
...
and 
  PipeExpr =
  | EPipeLambda of id * List<id * string> * Expr
  | EPipeInfix of id * Infix * Expr * Expr
  | EPipeFnCall of id *
...
```

ditansu:
@pbiggar hey, I've added types into pipe. All test passed, exclude one:
```fsharp
   (let x = 1 in
   (5
   |> x
   |> x
   |> (+) 3)) = Test.typeError_v0 "Expected a function value, got something else: 1"
```  
I have no idea how to fix it, need time. Anyway, if code is OK, I guess it is possible to merge it.

pbiggar:
Great! I'll take a look tomorrow.

pbiggar:
Note that there are merge conflicts. I'm not sure how hard they are but if they seem hard I can take care of them.

backend/testfiles/execution/language.tests
```diff
@@ -238,10 +238,18 @@ module Pipes =
   |> (+) 3)) = 10
 
   // check typing
+
+  // TODO: resolve this test 
+  // (let x = 1 in
+  // (5
+  // |> x
+  // |> x
+  // |> (+) 3)) = Test.typeError_v0 "Expected a function value, got something else: 1"
```
pbiggar, line 247: What's the issue with this test?

backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -423,6 +426,7 @@ module Expr =
     | ERecord (id, _, _)
     | EConstructor (id, _, _, _)
     | EFeatureFlag (id, _, _, _)
+    | EForbiddenExpr (id, _, _)
```
pbiggar, line 424: Are we able to remove the EForbiddenExpr from all places? The intent is to prevent the creating of invalid pipe expressions using the type system, and EForbiddenExpr subverts that a bit.

backend/src/LibExecution/ProgramTypesToRuntimeTypes.fs
```diff
@@ -171,55 +171,56 @@ module Expr =
...
-                RT.FnTargetExpr(toRT other),
-                typeArgs,
-                [ prev ],
-                RT.InPipe pipeID
-              )
-          convert next)
-
-        (expr2' :: rest')
+
+      let folder (prev : RT.Expr) (next : PT.Expr) : RT.Expr =
```
pbiggar, line 175: It seems like this should work on PT.PipeExprs, not PT.Exprs?

pbiggar (changes requested):
The squash merge PR makes it quite challenging to review/diff this - could you do a regular merge or a rebase instead? I think it'll make future merge conflicts easier too.

FYI, after changing types, the tests need to be rerun to regenerate the serialization files.

I think we should remove EForbiddenExpr - that's kinda subverting the point of the PR.

Also, it would be better to deal with PipeExpr directly rather than converting them to Expr (this is both in the parser, and when converting to RuntimeTypes). By converting them back and forth to Exprs, we hide errors which we want to catch statically in the type system.

ditansu:
> The squash merge PR makes it quite challenging to review/diff this - could you do a regular merge or a rebase instead? I think it'll make future merge conflicts easier too.

ok, sorry, I guess that squash commit is a better way. Honestly, the git management is not my strength skill. Next time I will do rebase. 
 
> FYI, after changing types, the tests need to be rerun to regenerate the serialization files.

Yeah, now I realized that. 

> I think we should remove EForbiddenExpr - that's kinda subverting the point of the PR.
> 
> Also, it would be better to deal with PipeExpr directly rather than converting them to Expr (this is both in the parser, and when converting to RuntimeTypes). By converting them back and forth to Exprs, we hide errors which we want to catch statically in the type system.

My main idea was to reduce an `EForbiddenExpr` expression to `EExpr` expression to decrease affection on the interpretation and to avoid adding (redundant?) complexity to `eval'`. But as I've understood you, I should to do something similar to `EMatch` with their `EMatchExpr` i.e. add to the `RT.Expr` and `eval'` the `EForbiddenExpr`, is it correct?

backend/testfiles/execution/language.tests
ditansu: The issue is that `x` is not a callable (it is literal `5`), or it is OK?

backend/src/LibExecution/ProgramTypesToRuntimeTypes.fs
ditansu: Yeah, if we decide to process the `PipeExpre` in `eval`, then we definitely should work with `PipeExpr`. I will fix it soon.

backend/src/LibExecution/RuntimeTypes.fs
ditansu: Agree, I remove it

backend/testfiles/execution/language.tests
StachuDotNet: Right - you shouldn't be able to `5 |> x` here - you can't "call" `1` with `5` as a parameter.
So the text looks correct to me.

pbiggar:
> My main idea was to reduce an `EForbiddenExpr` expression to `EExpr` expression to decrease affection on the interpretation and to avoid adding (redundant?) complexity to `eval'`. But as I've understood you, I should to do something similar to `EMatch` with their `EMatchExpr` i.e. add to the `RT.Expr` and `eval'` the `EPipeExpr`, is it correct?

Yes, the idea is to reduce edge cases. I agree with reducing complexity in the eval - I expect (and hope) that that can be done by pulling out some of that functionality as functions that can be reused.

pbiggar:
I think I may have confused you: the PipeExpr is good, and EPipeFnCall etc are good too!

ditansu:
>  I think I may have confused you: the PipeExpr is good, and EPipeFnCall etc are good too!

@pbiggar 
I spent a lot of time to understand what the F#'s type system not have enough power to resolve all edge cases like `Variable` with literals in the middle of a pipe or the lambda with `DError` inside which need to return as result. I guess only `Idris` with dependent types can handle it. Anyway I don't know how to do logic like [this](https://github.com/darklang/dark/pull/4848/files#diff-a82f65d22e0661b469040051df1d0d4a0ffd47bebeba75fdb141032bc1632dd7R521) on types 🫠
btw, I've opened new [PR](https://github.com/darklang/dark/pull/4848) to avoid the mix-up with redundant commits.

PR #4819, opened 2023-03-29
Fill in json serialize and parse

StachuDotNet:
Changelog:

```
Language and Standard Library
- implement `Json::serialize<'a>` and `Json::parse<'a>`
```

Note: I was experimenting with the idea of exposing a JsonParseError and JsonSerializationError type. Might just keep those internal for now, but I'm aware they're currently in a bad/mixed state.


StachuDotNet:
This is feeling pretty good. All happy paths seem to be working, and most error cases behave well-enough.

- [ ] deal with merge conflicts
- [ ] `let parse` should return JsonParseError as the Error type, and the stdlib fn (for now) should do the toString-ing. Make sure we're not "escaping" from this anywhere.
- [ ] update `serialize` to support a `JsonSerializationError` as the error type, and similarly ToString it, if needed, _in_ the stdlib fn
- [ ] revisit test cases - I don't think it's worth being exhaustive for now, but there may be some low-hanging fruit (i.e. package-defined enums)
- [ ] ask gpt for more test cases

backend/tests/TestUtils/FSI.fs
```diff
@@ -0,0 +1,30 @@
+/// <summary>
```
pbiggar, line 1: Did you intend to resurrect this?

StachuDotNet: nope, will re-remove

backend/src/Parser/Parser.fs
```diff
@@ -42,7 +43,10 @@ module DType =
     // with type args
     | "List", [ arg ] -> PT.TList(fromSynType arg)
     | "Option", [ arg ] -> PT.TOption(fromSynType arg)
-
+    | "Result", [ okArg; errorArg ] ->
+      PT.TResult(fromSynType okArg, fromSynType errorArg)
+    | "Tuple", first :: second :: theRest ->
```
pbiggar, line 48: This surprises me - is the F# compiler using the word "Tuple"?

pbiggar:
Would love to get this merged ASAP, even if it's not in the final shape. Maybe add some CLEANUPs for what's left to do?

backend/src/Parser/Parser.fs
StachuDotNet: Hmm
The context here is handling `Json.serialize<Tuple<int, string>>`.
I guess I could have done `Json.serialize<int * string>`, as F# might "expect".

PR #4820, opened 2023-03-29
CanvasHack improvements; improve error-reporting of non-HttpResponse results of HTTP handlers

StachuDotNet:
Changelog:

```
HTTP Handlers
- when something other than an HttpResponse is return, error reporting is improved

Internal
- migrating `dark-editor` canvas seeded by `CanvasHack` to new `canvases` directory,
  preparing to house multiple concurrent experiments
```
image

Once CanvasHack can handle multiple experiments, I'll move the bogus endpoint to a separate canvas.
(Or, we could simply remove it - the test coverage may be enough)


pbiggar:
I don't fully follow, but it seems everyone will be able to work on their own experiments with this, so seems fine.

pbiggar (approved):

StachuDotNet:
yeah I'm basically thinking each "experiment" is a folder (or maybe a single file) in the `canvases` dir

PR #4827, opened 2023-04-04
OpenTelemetry sampling

pbiggar:
Changelog:

```
Infrastructure
- enable sampling of OpenTelemetry data
```

We are spending too much on opentelemetry, and that is about to double due to our provider increasing costs. This is the start of addressing that, by allowing whole-trace sampling. I'd love to sample only errors, but that doesn't exist in dotnet OpenTelemetry.

The plan is in https://github.com/darklang/classic-dark/issues/16, merging this is step 1.

### DIFF
--- a/backend/src/LibService/LaunchDarkly.fs
+++ b/backend/src/LibService/LaunchDarkly.fs
@@ -196,6 +196,41 @@ module Internal =
     stringSetTestDefault name testDefault
     fun canvasID -> stringVar name $"canvas-{canvasID}" default_
 
+  // -------------
+  // per-service values
+  // -------------
+  let serviceBool
+    (name : string)
+    (default_ : bool)
+    (testDefault : bool)
+    : string -> bool =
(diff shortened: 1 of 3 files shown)


StachuDotNet (approved):
As far as I can tell this looks OK - have you been able to test this locally, seeing an appropriate # of samples?

pbiggar:
> I see that you've requested review here, but I've already reviewed it - did you men to re-request a review on the other PR?

AH, looks like I hadn't refreshed. I'm good, thanks!

pbiggar:
> As far as I can tell this looks OK - have you been able to test this locally, seeing an appropriate # of samples?

I tested this locally to check the logic worked, and it does now (it didn't before). However, dropping the samples does not seem to actually drop the samples in the Console exporter. I suspect this might be a ConsoleExporter thing but I can't see it anywhere. I figure we'll see what happens in production.

PR #4828, opened 2023-04-05
Support single-file canvas definitions in Parser/CanvasHack

StachuDotNet:
Changelog:

```
Area of change
- details
```

I have some notable tidying to still do.

Beyond tidying, I need to figure out how to reliably 'clear canvas' appropriately - I've introduced an 'id' field in the canvas .yml spec, and I'm thinking to somehow feed that into the `clear-canvas` script. If that's the case, then the 'script' will need to move from a .sh script to something "in dark code," I believe.

I think we can get rid of support for 'v1' of canvas definitions, and only support single-file defs (just one index/main.dark), but didn't want to break what Ocean's currently working on, so let support for now.


scripts/clear-canvas.sh
```diff
@@ -3,15 +3,17 @@
...
+echo "Clearing canvas (logs in ${DBLOG})"
 DB="${DARK_CONFIG_DB_DBNAME}"
 
 function run_sql { psql -d "$DB" -c "$@" >> "$DBLOG" ; }
 
 function fetch_sql { psql -d "$DB" -t -c "$@"; }
 
-CANVASES="${DARK_CONFIG_ALLOWED_DARK_INTERNAL_CANVAS_ID}"
+# TODO: get canvasId from name
+CANVASES=$(fetch_sql "SELECT id FROM canvases_v0;") # TODO re-add some sort of clause here WHERE name = '{$1}';")
```
pbiggar, line 15: You could use `select canvas_id from domains_v0 where domain = ...` here.

backend/src/Prelude/Prelude.fs
```diff
@@ -1589,5 +1589,7 @@ module HttpHeaders =
 
 let id (x : int) : id = uint64 x
 
+let randomId (x : int) : id = uint64 x
```
StachuDotNet, line 1592: this was a mistake - removing.

StachuDotNet:
Working on rebasing now

backend/src/Parser/CanvasV2.fs
```diff
@@ -0,0 +1,179 @@
...
+      let newFn = ProgramTypes.UserFunction.fromSynBinding availableTypes letBinding
+      { m with fns = newFn :: m.fns }
+
+    | [ attr ] ->
+      match attr with
+      | SimpleAttribute ("DB", [ name ]) ->
+        // let newDB = PT.DB(name, randomIds ())
+        // { m with dbs = newDB :: m.dbs }
+        Exception.raiseInternal
+          $"Not currently supporting DBs - probably makes sense to wait until they are based on a type rather than 'columns'"
```
pbiggar, line 77: For now, using the v1 parser is probably fine - just an annotation on a type instead of a `let`.

backend/src/Parser/CanvasV2.fs
```diff
@@ -0,0 +1,179 @@
...
+
+
+/// An F# module has a list of declarations, which can be:
+/// - a group of let bindings
+/// - a group of type definitions
+/// - a single expression
+/// - a nested module (not currently supported)
+/// - etc
+///
+/// (The declarations are presented as _groups_ presumably to support recursive definitions)
```
pbiggar, line 123: Not sure what this comment says?

backend/src/Parser/CanvasV2.fs
```diff
@@ -0,0 +1,179 @@
...
+///
+/// This builds up a CanvasModule by parsing each declaration:
+/// - type definitions are parsed as user types
+/// - let bindings with no attributes are parsed as user functions
+/// - let bindings with attributes are parsed
+///   as handlers (i.e. with `[<HttpHandler("method", "path")>]`)
+///   or as DBs (i.e. with `[<DB("name")>]`)
```
pbiggar, line 130: I'm not clear how a let would make sense to model a DB declaration?

pbiggar (approved):

backend/src/Parser/CanvasV2.fs
StachuDotNet: Initially I was expecting the "declarations" list of an F# module to be a list of
```
type Declaration =
  | Expr of ..
  | Type of ..
  | Submodule of ..
  | Let of ... 
```

But, to support recursive definitions such as
```
let Enum1 = A of Enum2
and Enum2 = B of Enum1
```
they're presented in groups.

I found this a bit unintuitive so put a note around it. that said, it's probably not worth it, as that's part of the "check the F# compiler docs" note in the README

StachuDotNet:
Merging now - we can follow up around DB definitions shortly in a follow-up PR.

My thought about DB definitions is that we might want to define a DB for a type defined _outside_ of this module.
e.g. what if I just want a data store of List<int>?

backend/src/Parser/CanvasV2.fs
StachuDotNet: My thought about DB definitions is that we might want to define a DB for a type defined outside of this module.
e.g. what if I just want a data store of List?

StachuDotNet: Or should DB definitions be restricted to only Record type definitions?

pbiggar: I think for now they could. Agree that stuff like a list of data would be a cool addition, and it probably wouldn't be too much to add that.

PR #4829, opened 2023-04-06
Handle Dicts/Records in JSON stdlib fns

StachuDotNet:
```
Language and Standard Library
- Add minimal support for records+dictionaries to Json::parse<'a> and ::serialize<'a>
```

This was blocking Ocean, so we hacked together to get out a working solution. Does this seem reasonable for now, @pbiggar?

My hesitation is that we're returning a DDict in the parser when we want to return a DRecord.
This was done as the Parser currently interprets (afaik) all 'objects' as DDicts - so, returning a DRecord as we'd _like_ to do would result in difficult testing (until we support labeling a record like `Person { Name = ... }`)

### DIFF
--- a/backend/src/LibExecutionStdLib/LibJson.fs
+++ b/backend/src/LibExecutionStdLib/LibJson.fs
@@ -135,8 +135,52 @@ let rec serialize
       w.WritePropertyName "Error"
       r errType dv)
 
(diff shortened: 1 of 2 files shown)


pbiggar (approved):
The DDict to DRecord change is still to come so this seems best for now.

PR #4831, opened 2023-04-07
Add some new features to dark-editor

OceanOak:
No changelog

- Fix the openai api call
- Add a copy result button and an editable dark prompt
- Add File.write function


backend/experiments/cli/LibCli.fs
```diff
@@ -55,8 +55,23 @@ let fns : List<BuiltInFn> =
       previewable = Impure
       deprecated = NotDeprecated }
 
-
-
-
-
-    ]
+    { name = fn "File" "write" 0
```
OceanOak, line 58: Would it be better to move the function to a separate file in order to ensure that the two experiments remain independent of each other?

canvases/dark-editor/index.dark
```diff
@@ -61,19 +66,144 @@
...
+        height:80%;
+      }
     </style>
   </head>
   <body>
     <div class="split-container">
       <div class="left-half">
         <form id="gpt4-form">
           <input type="text" name="dark-editor" value="dark-editor" />
+          <textarea id="dark-prompt" name="dark-prompt">
```
OceanOak, line 79: Maybe it is better to figure out a way to read the prompt from a file rather than pasting it here

StachuDotNet: Yeah, in a separate PR you could:
- add an http GET handler to fetch the prompt on disk
- add some JS in the front-end part of this that fetches that prompt, and inserts it there

OceanOak: That worked well, thanks!

canvases/dark-editor/index.dark
```diff
@@ -100,10 +126,70 @@
...
+          event.preventDefault();
+          const prompt = document.getElementById("dark-prompt").value;
+          const response = await fetch("/save-prompt", {
+            method: "POST",
+            headers: { "Content-Type": "text/plain" },
+            body: prompt,
+          });
+        });
+
+      const getPromptBtn = document.getElementById("get-prompt");
```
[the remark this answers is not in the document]
StachuDotNet: I think that's a great next step!

Beyond that, I've been sort of imagining a "prompt builder" UI where we could pull in _multiple_ sources and concat them together visually.

You know how people separate their prompt sections with "---" bits? I'd love to expand/collapse each of those 'sections' visually, and have them pieced together. That way, I can focus on one 'part' of the prompt at a time

Issue #2552, opened 2020-06-08
Add List::groupBy to the standard library


vkrauchunas:
See #2411 for more information about adding to the standard library

StachuDotNet:
I believe this can now be merged?

PR #4833, opened 2023-04-08
Add List.groupAdjacentBy_v0 and List.groupBy_v0

nicu-chiciuc:
Related issue: https://github.com/darklang/dark/issues/2552

Adds:
* groupAdjacentBy :  List a -> (a -> a -> Bool) -> List (List a))
* groupBy : List a -> (a -> b) -> List (Tuple (b, (List a)))

The definition of `groupBy` with a key function is more straightforward, the only edge-cases were the ordering of tuples inside the list (based on the first appearence of a key), and the ordering of elements in each tuple.

The definition for `groupAdjacentBy` is not as clear. Initially I went with the definition that's used in [Haskell](https://hackage.haskell.org/package/base-4.7.0.1/docs/src/Data-List.html)
```hs
-- | The 'groupBy' function is the non-overloaded version of 'group'.
groupBy                 :: (a -> a -> Bool) -> [a] -> [[a]]
groupBy _  []           =  []
groupBy eq (x:xs)       =  (x:ys) : groupBy eq zs
                           where (ys,zs) = span (eq x) xs
```

but this has strange behavior when the predicate is not a function that decides equality. For example:
```
List.groupAdjacentBy_v0 [1, 2, 3, 2, 3, 2] (fun x y -> x < y) 
```

Using the definition in Haskell, the response would be `[[1,2,3,2,3,2]]` since all the elements are compared to the first element in the group (1).
Another implementation would be actually checking each adjacent elements, in that case the result would be `[[1, 2, 3], [2, 3], [2]]`

The `(<)` predicate can be considered as not being valid, but I think it's important to have a good definition for such edge-cases.

**The current implementation uses the `[[1, 2, 3], [2, 3], [2]]` approach.**

Questions:
* We pass the list first and then the predicate. Haskell does the opposite. We don't yet have currying (I assume), but I was wondering if there's a decision regarding this order of arguments.

### DIFF
--- a/backend/src/LibExecutionStdLib/LibList.fs
+++ b/backend/src/LibExecutionStdLib/LibList.fs
@@ -1487,16 +1487,24 @@ let fns : List<BuiltInFn> =
                   let! key = applyFn head
 
                   if not (Dval.isFake key) then
-                    let foundGroup = List.tryFind (fun (k, _) -> k = key) groups
+                    let (updatedGroups, found) =
+                      List.foldBack
+                        (fun (k, elems) (acc, found) ->
+                          let newElems, newFound =
+                            if k = key then
+                              (elems @ [ head ], true)
+                            else
+                              elems, found
+
+                          (k, newElems) :: acc, newFound)
+                        groups
+                        ([], false)
 
                     let newGroups =
-                      match foundGroup with
-                      | Some (_, elements) ->
-                        let updatedGroup = (key, elements @ [ head ])
-                        List.map
-                          (fun (k, v) -> if k = key then updatedGroup else (k, v))
-                          groups
-                      | None -> groups @ [ (key, [ head ]) ]
+                      if found then
+                        updatedGroups
+                      else
+                        (key, [ head ]) :: updatedGroups
 
                     return! loop newGroups tail
                   else
@@ -1511,7 +1519,7 @@ let fns : List<BuiltInFn> =
               }
 
             match! loop [] l with
-            | Ok result -> return DList result
+            | Ok result -> return DList(List.rev result)
             | Error fakeDval -> return fakeDval
           }
         | _ -> incorrectArgs ())
--- a/backend/src/LibExecutionStdLib/LibList.fs
+++ b/backend/src/LibExecutionStdLib/LibList.fs
@@ -1389,74 +1389,7 @@ let fns : List<BuiltInFn> =
       deprecated = NotDeprecated }
 
 
-    { name = fn "List" "groupAdjacentBy" 0
-      typeParams = []
-      parameters =
-        [ Param.make "list" (TList varA) ""
-          Param.makeWithArgs "fn" (TFn([ varA; varA ], TBool)) "" [ "a"; "b" ] ]
-      returnType = TList(TList varA)
-      description =
-        "Groups <param list> into sublists, where each sublist contains values that
-         are equal according to <param fn>.
-
-         For example, if <param list> is {{[1, 2, 2, 3, 2, 2, 2, 1, 1, 1]}} and <param fn>
-         is {{fn a b -> a == b}}, returns {{[[1], [2, 2], [3], [2, 2, 2], [1, 1, 1]]}}.
-
-         Important: It behaves differently from Haskell's groupBy: https://stackoverflow.com/a/1316392
-         Each value is compared to the next value in the list, not the initial value in the group.
-
-         Preserves the order of values."
-      fn =
-        (function
-        | state, _, [ DList l; DFnVal fn ] ->
-          uply {
-            let group l =
-              let applyFn dvalA dvalB =
-                Interpreter.applyFnVal state fn [ dvalA; dvalB ]
-
-              let rec loop currentGroup groups l =
-                uply {
-                  match currentGroup, groups, l with
-                  | [], [], [] -> return Ok([])
-                  | _, _, hd :: tl ->
-                    match currentGroup with
-                    | [] -> return! loop [ hd ] groups tl
-                    | cgHd :: _ ->
-                      let! fnResult = applyFn cgHd hd
-
-                      match fnResult with
-                      | DBool true -> return! loop (hd :: currentGroup) groups tl
-                      | DBool false ->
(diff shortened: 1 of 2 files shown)


nicu-chiciuc:
I didn't add the Haskell-like definition in the current PR, since it doesn't seem very useful. I can add it thought if necessary.

backend/src/LibExecutionStdLib/LibList.fs
```diff
@@ -1388,6 +1388,138 @@ let fns : List<BuiltInFn> =
...
+              let applyFn dval = Interpreter.applyFnVal state fn [ dval ]
+
+              let rec loop groups l =
+                uply {
+                  match l with
+                  | hd :: tl ->
+                    let! key = applyFn hd
+
+                    if not (Dval.isFake key) then
+                      let foundGroup = List.tryFind (fun (k, _) -> k = key) groups
```
pbiggar, line 1487: I think this could be done in one pass. Right now it finds the key, then goes through to do an update. Couldn't we go through and do the check and update at the same time.

Alternatively, it feels like a Map would be a better structure for this, though you would lose the order of tuples. However, I'm not super sure the ordering of tuples is important here?

backend/src/LibExecutionStdLib/LibList.fs
```diff
@@ -1388,6 +1388,138 @@ let fns : List<BuiltInFn> =
...
+      typeParams = []
+      parameters =
+        [ Param.make "list" (TList varA) ""
+          Param.makeWithArgs "fn" (TFn([ varA; varA ], TBool)) "" [ "a"; "b" ] ]
+      returnType = TList(TList varA)
+      description =
+        "Groups <param list> into sublists, where each sublist contains values that
+         are equal according to <param fn>.
+
+         For example, if <param list> is {{[1, 2, 2, 3, 2, 2, 2, 1, 1, 1]}} and <param fn>
```
pbiggar, line 1402: This is a sort of odd function and I agree that it's unclear what someone would use it for.

I think perhaps you included this because of my comments about wanting a function wiht the signature `List<'a> -> ('a -> 'a -> bool) -> List<List<'a>>`. I think I may have been mistaken here. What I meant to say was a function whose return type was `List<List<'a>>` - I wasn't actually that concerned with the predicate.

What I had intended to suggest was

`List.groupBy: List<'a> -> ('a -> 'key) -> List<List<'a>>`
and
`List.groupByWithKey: List<'a> -> ('a -> 'key) -> List<'key * List<'a>>`

(I think a `Dict<'key, List<'a>>` would be better but we don't support that yet.)

pbiggar (changes requested):
Thanks! Code looks good. I think I have confused you with some of my comments in the past - hopefully the comments here are clearer.

backend/src/LibExecutionStdLib/LibList.fs
nicu-chiciuc: > I'm not super sure the ordering of tuples is important here
Even if it adds a little more implementation complexity, I think it's worth the quality of life to have a defined behavior for this.
* Easier to update tests if needed; otherwise the order will change when adding/removing a single element at the end
* As a user of Dark I would be confused if my output would jump around when adding new elements to a list

It's possible to use a Map and a separate list for storing the order of keys. The issue is that I can't use `Dval` as keys for a Map:
Screenshot 2023-04-10 at 13 05 52

There is the `equals` function in https://github.com/darklang/dark/blob/main/backend/src/LibExecutionStdLib/LibNoModule.fs#L15
But I would have to create a wrapper type so that it can be used by the map.

I've searched for usages of `Map` in `Lib*.fs` files, but couldn't find a similar usage.

As an alternative, I used a `found` boolean and a `List.foldr` so that there's only a single pass over the list.

backend/src/LibExecutionStdLib/LibList.fs
nicu-chiciuc: > perhaps you included this because of my comments about wanting a function wiht the signature List<'a> -> ('a -> 'a -> bool) -> List<List<'a>>.

yes, and I figured it's based on Haskell, since that's what I consider the de facto standard for anything functional programming

I assume the `groupAdjacentBy` could be conformable for someone accustomed to Haskell

I don't mind renaming the `groupBy` to `groupWithKeyBy`. 

Regarding `List.groupBy: List<'a> -> ('a -> 'key) -> List<List<'a>>`, wouldn't it be better to implement it in Dark since it would require the `groupByWithKey` and a map

nicu-chiciuc:
Also, I'm used to squashing my PRs (so I don't pay too much attention to iterative commits), but that doesn't seem the norm for this codebase. If necessary, I can reorganize/rename the commits so they look better in the graph tree

nicu-chiciuc:
the review request can be ignored for now

`List.foldr` doesn't exist in F#, for some reason Ionide wasn't showing errors, and running `clear && scripts/run-backend-tests --filter-test-list list` also didn't show any error.

nicu-chiciuc:
should be good now

backend/src/LibExecutionStdLib/LibList.fs
pbiggar: Cool, good idea.

backend/src/LibExecutionStdLib/LibList.fs
pbiggar: Yeah, good point.

pbiggar:
Do you know why the CircleCI pipeline fails? It seems to be only for your PRs. Maybe you need to sign into it and rerun this build?

nicu-chiciuc:
Ah, I should've signed in, I don't think I had an account. I assumed it didn't rely on me having an account.

pbiggar:
I think this isn't waiting on my review, right?

nicu-chiciuc:
@pbiggar it shows that there are changes requested still

nicu-chiciuc:
@pbiggar I think this can be merged

nicu-chiciuc:
* removed `groupAdjacentBy`
* renamed `groupBy` to `groupByWithKey`
* simplified implementation to O(N)

pbiggar:
Great, thank you!

PR #4834, opened 2023-04-09
Add tests and rename fixup pass to parsing

pbiggar:
Feedback on the API here and in #4832 and how best to use it would be appreciated

### DIFF
--- a/backend/src/Parser/CanvasV2.fs
+++ b/backend/src/Parser/CanvasV2.fs
@@ -158,11 +158,11 @@ let parseDecls
 
 let postProcessModule (m : CanvasModule) : CanvasModule =
   let fnNames = m.fns |> List.map (fun f -> f.name) |> Set
-  let convert = ProgramTypes.Expr.convertVariablesToFnCalls fnNames
+  let fixup = ProgramTypes.Expr.fixupPass fnNames
   { m with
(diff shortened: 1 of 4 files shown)


StachuDotNet:
Are you looking for feedback on how we're handling "post-processing" of the Parser?
I think this looks fine!

It feels to me as though we might want to, at some point, use some intermediate "ParserTypes" AST which is later transformed into PTs. This seems like a great step to handle ambiguities that the Parser faces ("is this a variable or something else?") but might not capture everything down the line?

Looks good, though - worth merging

StachuDotNet (approved):

StachuDotNet:
If you're referring to something else, though, in what feedback you're looking for, please let me know

pbiggar:
> Are you looking for feedback on how we're handling "post-processing" of the Parser? I think this looks fine!
> 
> It feels to me as though we might want to, at some point, use some intermediate "ParserTypes" AST which is later transformed into PTs. This seems like a great step to handle ambiguities that the Parser faces ("is this a variable or something else?") but might not capture everything down the line?
> 
> Looks good, though - worth merging

Yeah, we might need an intermediate parsing type. I wonder if that's what ProgramTypes is supposed to be? It was for the editor, hence EPartial etc.

PR #4835, opened 2023-04-10
Allow CanvasHack canvases to use experimental stdlib fns

StachuDotNet:
Changelog:

```
CanvasHack, Experimental
- allow CanvasHack canvases to use experimental stdlib fns
```

- add LibExperimentalStdLib with LibCLI fns
- add LibExperimentalExecution - just LibRealExecution, but with LibExperimentalStdLib types/fns included
  I was tempted to adjust RealExecution to take in an 'additionalLibraries' parameter but decided against it.
  Please let me know if you think this was the right call
- add BwdDangerServer to be an _additional_ BwdServer that serves HttpHandlers with the additional functions available
- update scripts/docs relevantly


pbiggar (approved):
Looks good!

StachuDotNet:
We _could_ house an additional .sln within Experiments, to be extra safe about experimental stuff affecting the 'main' codebase.

PR #4802, opened 2023-03-18
implement tuple let pattern

StachuDotNet:
Changelog:

```
Language and Standard Library
- add tuple deconstruction in `let` bindings (i.e. `let (a, b) = (1, "two")` works)
```

Extracted from #4711

PR #4840, opened 2023-04-11
Implement tuple destructuring let pattern

nicu-chiciuc:
Changelog:

```
Language and Standard Library
- add tuple deconstruction in `let` bindings (i.e. `let (a, b) = (1, "two")` works)
```

supersedes https://github.com/darklang/dark/pull/4802


nicu-chiciuc:
> Worth checking length of 3 and 4, and also errors
> https://github.com/darklang/dark/pull/4802#discussion_r1141228810
added checks for tuples of different lengths

> Hard to tell if this works from a read of it, but there's a test case for matchPatterns that could be copied to ensure the analysis works.
> https://github.com/darklang/dark/pull/4802#discussion_r1141229419
Not sure I understand this comment

Add implementation in the `SqlCompiler.fs` and some test for destructuring inside sql queries

StachuDotNet:
> > Hard to tell if this works from a read of it, but there's a test case for matchPatterns that could be copied to ensure the analysis works.
> > [#4802 (comment)](https://github.com/darklang/dark/pull/4802#discussion_r1141229419)
> 
> Not sure I understand this comment

In classic-dark, match cases that didn't 'match' the expression were evaluated as far as they could be, in the editor, to provide feedback on "what _if_ the branch matched?"

For example, here's a case where the first branch isn't matched, but we "preview" the evaluation anyway, such that many values can be previewed. (hope that makes some sense)
Here, my cursor is on the 'x' on the 3rd line. Although that branch didn't get matched, we can still 'inspect'/analyze the "what if"
image.
This is part of a larger process called 'analysis'.
Throughout the Interpreter, you'll find bits of logic that aim to exhaust all code branches, even if not "on the execution path."

Take a look at `backend/tests/Tests/Execution.Tests.fs`.
Inside of that, there's `testMatchPreview` which tests this sort of thing for a match expression.

I believe Paul is suggesting that we could add relevant additional test cases to ensure we do our best to 'preview' all branches of logic regardless of whether they're hit. Not sure if this is clear enough, paul may want to expand or I can reflect and follow up more later.

backend/src/LibBackend/SqlCompiler.fs
```diff
@@ -164,13 +164,33 @@ let rec inline'
...
+        (currentExpr : Expr, letPattern : LetPattern)
+        : Map<string, Expr> =
+        match letPattern with
+        | LPVariable (_id, name) -> Map.add name currentExpr symtable
+        | LPTuple (_id, first, second, theRest) ->
+          match currentExpr with
+          | ETuple (_, firstExpr, secondExpr, restExpr) ->
+            // Each expression is paired with the corresponding let pattern
+            let zipped =
+              List.zip
```
pbiggar, line 179: We should check length before List.zip as it will generate the built-in error from List.zip and we'd prefer to have one of our own errors (`Exception.raiseCode`)

backend/testfiles/execution/db.tests
```diff
@@ -487,6 +487,17 @@ module PartialEvaluation =
   (friends (fun p -> let x = { y = { z = { a = [1;2;3;4;5] }}}
                      (x.y.z.a |> List.length_v0) < p.height)) = [" Chandler "; "GrumpyCat"; "Rachel"; "Ross" ]
 
+  // tuple destructuring outside query
```
pbiggar, line 490: I think we can skip the test outside the query.

backend/tests/Tests/Serialization.TestValues.fs
```diff
@@ -231,7 +231,12 @@ module ProgramTypes =
     let e = PT.EInt(34545UL, 5)
     PT.ELet(
       14219007199254740992UL,
-      PT.LPVariable(189271UL, "x1"),
+      PT.LPTuple(
+        189271UL,
+        PT.LPVariable(18274132UL, "x0"),
```
pbiggar, line 236: Maybe have a nested LPTuples here too?

pbiggar (changes requested):
This is great!

I made some comments for some small fixes. After that, a quick test for preview/analysis logic would be helpful (see testMatchPreview in  backend/tests/Tests/Execution.Tests.fs for a template)

backend/testfiles/execution/db.tests
nicu-chiciuc: done

backend/src/LibBackend/SqlCompiler.fs
nicu-chiciuc: done

backend/tests/Tests/Serialization.TestValues.fs
nicu-chiciuc: add a nested tuple, and also a tuple in the rest part

backend/tests/Tests/Execution.Tests.fs
```diff
@@ -589,6 +596,15 @@ let testMatchPreview : Test =
           (okVarRhsVarId, "rhs", er (DString "y"))
           (okVarRhsStrId, "str", er (DString "ok: ")) ]
 
+      t
+        "tuple match"
+        (eTuple (eInt 2) (eStr "sample") [])
```
nicu-chiciuc, line 601: @StachuDotNet it was a little hard to fully understand what's going on in this test, but I think this should do the trick

pbiggar (changes requested):
Nice catch! Lookd like we had missed testing matches on tuples.

The test that we had been hoping for though was a little different. In Execution.Tests.fs, most of the tests are to test the preview of some construct: testMatchPReview, testLambdaPreview, etc. We wanted a new test called `testLetPreview` that was similar to testMatchPreview but tested Lets and LetPatterns.

nicu-chiciuc:
ah, I see. I'll get to that

backend/tests/Tests/Execution.Tests.fs
```diff
@@ -600,6 +616,68 @@ let testMatchPreview : Test =
...
+
+        let! result = createLetPattern letPattern assignExpr retExpr
+
+        Expect.equal
+          (Map.get lpID result)
+          (Some(AT.ExecutedResult(DInt 1L)))
+          "hello assigned correctly"
+      }
+
+      testTask "let (x, y) = (1, 2) in y" {
```
nicu-chiciuc, line 654: @pbiggar is this what you had in mind?

StachuDotNet: This looks like a good start to me!

What happens upon `let (x, y) = (1, 1.0/0.0) in x`?

Without care, the interpreter would likely fully halt execution upon the `1/0` exception. I suppose the intent here is that the `x` variable is still visited/previewed.

backend/tests/Tests/Execution.Tests.fs
```diff
@@ -600,6 +616,172 @@ let testMatchPreview : Test =
...
+
+        // TODO: remove
+        debuG "divID" (Map.get divID result) // DEBUG: divID (Some(ExecutedResult (DError (SourceNone, "Division by zero"))))
+        debuG "tuple" (Map.get (Expr.toID assignExpr) result) // DEBUG: tuple (Some(ExecutedResult (DError (SourceNone, "Division by zero"))))
+
+
+        // TODO: figure it out
+        // expected: Some (NonExecutedResult (DInt 1L))
+        // actual: Some (NonExecutedResult (DIncomplete (SourceID (7UL, 309035826UL))))
+        Expect.equal
```
nicu-chiciuc, line 710: https://github.com/darklang/dark/pull/4840/files#r1170419108
@StachuDotNet 
I assume the inner tuple error propagates to the tuple and the tuple also evaluates to a DError.

Does this mean that the implementation has to be changed a little bit?

backend/tests/Tests/Execution.Tests.fs
StachuDotNet: following up on a conversation had in a discord call:
my claim here was incorrect - the RHS of this expr will result in a Dval of an Error, so the "tuple-ness" will be lost, and `x` should not be expected to have the value of `1` even "in preview mode" or anything

StachuDotNet (approved):

backend/tests/Tests/Execution.Tests.fs
```diff
@@ -600,6 +616,164 @@ let testMatchPreview : Test =
...
+    (expr : Expr)
+    (finalExpr : Expr)
+    : Task<Map<id, AT.ExecutionResult>> =
+    task {
+      let ast = ELet(letID, patternMatch, expr, finalExpr)
+
+      let! results = execSaveDvals "let-preview" [] [] [] ast
+      let stuff = results |> Dictionary.toList |> Map
+
+      return stuff
```
pbiggar, line 634: Here you could just `return results |> Dictionary.toList |> Map`

backend/tests/Tests/Execution.Tests.fs
```diff
@@ -600,6 +616,164 @@ let testMatchPreview : Test =
...
       ]
 
+
+let testLetPreview : Test =
+  let letID = gid ()
+
+  let createLetPattern
+    (patternMatch : LetPattern)
+    (expr : Expr)
+    (finalExpr : Expr)
```
pbiggar, line 626: You could call finalExpr `body` I think

pbiggar (changes requested):
This looks great. I flagged some tiny changes and we can merge after!

backend/tests/Tests/Execution.Tests.fs
nicu-chiciuc: done

pbiggar:
Amazing, thanks! Really happy to get this in!

PR #4842, opened 2023-04-11
add `EnvVar.get` and `EnvVar.getAll`

nicu-chiciuc:
Changelog:

- Add `System.getEnv` and `System.getEnvVars` to `backend/experiments/LibExperimentalStdLib/LibCLI.fs`
- Add `canvases/dark-cli-nicu/main.dark`. I planned to leave it for myself, but since **I didn't add tests** I figured I'd at least provide some sample code.

The initial purpose of this was to read the OPENAI_API_KEY from the env variables.
Although it seems that it's not available there.


backend/experiments/LibExperimentalStdLib/LibCLI.fs
```diff
@@ -75,6 +75,51 @@ let fns : List<BuiltInFn> =
       previewable = Impure
       deprecated = NotDeprecated }
 
+    { name = fn "System" "getEnv" 0
```
pbiggar, line 78: How about `Environment.get`?

backend/experiments/LibExperimentalStdLib/LibCLI.fs
```diff
@@ -75,6 +75,51 @@ let fns : List<BuiltInFn> =
...
+          match envValue with
+          | null -> Ply(DOption None)
+          | _ -> Ply(DOption(Some(DString envValue)))
+        | _ -> incorrectArgs ())
+      sqlSpec = NotQueryable
+      previewable = Impure
+      deprecated = NotDeprecated }
+
+
+    { name = fn "System" "getEnvVars" 0
```
pbiggar, line 97: Maybe `Environment.getAll`?

backend/experiments/LibExperimentalStdLib/LibCLI.fs
```diff
@@ -75,6 +75,51 @@ let fns : List<BuiltInFn> =
...
+      typeParams = []
+      parameters = []
+      returnType = TList(TTuple(TString, TString, []))
```
pbiggar, line 100: Tuple list is good, but should this be a `Map<string,string>` instead?

pbiggar (changes requested):
Looks good! Small suggestions, wdyt?

backend/experiments/LibExperimentalStdLib/LibCLI.fs
nicu-chiciuc: I thought we don't yet support Maps/Dicts. Should I instead return a `TDict`?

backend/experiments/LibExperimentalStdLib/LibCLI.fs
nicu-chiciuc: I feel like `Environment.get` should contain the word "variable" or "var" somewhere. Otherwise, it seems too general. Maybe `Environment.getVar` or something.

I'll look tomorrow at how other standard libraries name such functions

backend/experiments/LibExperimentalStdLib/LibCLI.fs
pbiggar: Sorry, yeah we support TDict/DDict when the key is a string.

backend/experiments/LibExperimentalStdLib/LibCLI.fs
pbiggar: Good point. It's not really Environment, it's really Environment Variables. Maybe "EnvVar" or "getVar" might work, yeah.

nicu-chiciuc: Used `EnvVar.get` and `EnvVar.getAll`

backend/experiments/LibExperimentalStdLib/LibCLI.fs
nicu-chiciuc: I return a `TDict TString` now. We can't yet serialize Dicts, right? I've transformed the dict into a list of tuples before serializing and returning it through the http handler.

StachuDotNet: We can't directly Json.parse and .serialize a Dict currently without also providing a Record type that matches the dict's shape, but I don't think we need to.

If we have the Map, and we extract the value of the API key, we're never trying to Json-serialize

In other words:
```fsharp
let openAiApiKey = Environment.getAll |> Map.get "OPENAI_API_KEY"
type ApiRequestBody = {  apiKey: string; ... }
let requestBody = { apiKey = openAiApiKey; ... } |> Json.serialize<ApiRequestBody> |> String.toBytes
...
```

nicu-chiciuc: The reason I added the `getAll` was to see what variables are available to the canvas. Initially I only had the `get` and it didn't have access to `OPENAI_API_KEY` and I wasn't sure if that's an issue with my code or the environment.

pbiggar: The reason that OPEN_API_KEY is not available is because env vars are loaded when the process starts (which in this case means when the server process starts). So env vars likely weren't a good fit for what you were trying to do. However, they're still useful for LibCLI.

pbiggar: > I return a `TDict TString` now. We can't yet serialize Dicts, right?

Good point, fixing that at the moment.

pbiggar (changes requested):
These look great! Can you add tests for them?

backend/experiments/LibExperimentalStdLib/LibCLI.fs
```diff
@@ -76,6 +76,56 @@ let fns : List<BuiltInFn> =
...
+      description =
+        "Returns a list of tuples containing all the environment variables and their values."
+      fn =
+        (function
+        | _, _, [] ->
+          let envVars = System.Environment.GetEnvironmentVariables()
+
+          let tupleList =
+            envVars
+            |> Seq.cast<System.Collections.DictionaryEntry>
```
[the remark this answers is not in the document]
StachuDotNet: afaik, the IDictionary that `System.Environment.GetEnvironmentVariables()` returns is basically a map of `object` to `object`.
So, you have to manually cast what type you think the objects should be.
`Seq.cast` in my experience is a pretty typical way of dealing with this, unless you use something like https://github.com/demystifyfp/FsConfig that creates better-typed F# bindings as a wrapper.

backend/experiments/LibExperimentalStdLib/LibCLI.fs
```diff
@@ -76,6 +76,56 @@ let fns : List<BuiltInFn> =
...
+      parameters = [ Param.make "varName" TString "" ]
+      returnType = TOption TString
+      description =
+        "Gets the value of the environment variable with the given <param varName> if it exists."
+      fn =
+        (function
+        | _, _, [ DString varName ] ->
+          let envValue = System.Environment.GetEnvironmentVariable(varName)
+          match envValue with
+          | null -> Ply(DOption None)
```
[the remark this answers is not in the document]
StachuDotNet: `isNull` and `match x with | null -> `, I believe, are 100% equivalent.

https://stackoverflow.com/a/6818329 seems to confirm my experience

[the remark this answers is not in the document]
pbiggar: Ah, then let's us isNull

nicu-chiciuc:
> Can you add tests for them?

The other function definitions in this file didn't seem to have tests, so I wasn't sure how exactly should I test such types of functions. I'd have to set up a mocked environment?

backend/experiments/LibExperimentalStdLib/LibCLI.fs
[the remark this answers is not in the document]
nicu-chiciuc: replaced it with `isNull`

backend/experiments/LibExperimentalStdLib/LibCLI.fs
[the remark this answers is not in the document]
nicu-chiciuc: from what I remember, both ChatGPT and copilot recommended I use `Seq.cast<System.Collections.DictionaryEntry>`, so I also assumed that's the accepted way of dealing with non-generic `IDictionary`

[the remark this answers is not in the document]
pbiggar: Lol, I'm skeptical of relying entirely on the AIs for this, but seems Stachubot agrees :)

pbiggar:
For tests, I think we just need `get` works for present and missing vars, and getAll Dict.count is above some number (5?)

nicu-chiciuc:
removed the `dark-cli-nicu` canvas so it should be good to merge now

pbiggar:
Thanks!

PR #4848, opened 2023-04-13
Allow processing the only correct expressions into a pipeline. Correct error propagations. 

ditansu:
Only correct expressions processing in a pipeline. The correct expressions are:
the first expression could be: **function, lambda, variable, literal, Option, Result**
as second and next ones allowed: **function, lambda, constructor, variable with a ref to lambda only**

The logic to cover all [edge cases](https://github.com/darklang/dark/pull/4848/files#diff-30449b4569973f93055966f17bd6427c982c20bced61682f7aa9db07abc1db42R223) is complex, and F# type system doesn't have enough power and flexibility for that. That is why it needs to process all logic in the interpreter where we can evaluate the pipe's expressions and use got results to control pipe processing.

Previous attempts: 
https://github.com/darklang/dark/pull/4814 
https://github.com/darklang/dark/pull/4803


ditansu:
@pbiggar @StachuDotNet please, review this PR when you will have free time. Core changes in [eval'](https://github.com/darklang/dark/pull/4848/files#diff-a82f65d22e0661b469040051df1d0d4a0ffd47bebeba75fdb141032bc1632dd7R516) and [tests](https://github.com/darklang/dark/pull/4848/files#diff-30449b4569973f93055966f17bd6427c982c20bced61682f7aa9db07abc1db42R223)

pbiggar (changes requested):
Apologies for the delay, I started reviewing this and thought I had finished but I guess I lost the tab instead.

I think the type system has the resolution we need for this. I can lay out what the solution looks like if it's helpful.

- no change to RuntimeTypes (or interpreter or sql compiler)
- PT.EPipeTarget shouldn't be necessary anymore and can be removed
- PT.PipeExpr should include the things allowed in pipes
- ProgramTypesToRuntimeTypes should convert a PT.EPipe with PT.PipeExprs into RT.Expr (mostly RT.EApply)

ditansu:
@pbiggar thanks for answer! You could answer after weekend, I simply wasn't sure that you notice this PR :) 
I totally support types, but I don't understand how in current types we can check that a variable `x` in the middle of pipeline content a lambda in `ProgramTypesToRuntimeTypes`. Also, I can't realize how to organize the `DError` propagation through all pipeline functions. I see only two options here, either to check those cases in the interpreter - current PR or enrich `PT.Expr.EVariable` by additional meta information about Variable content, same for `PT.Expr.EFnCall` we need to know the type of content to `DError` propagation. Please advise an approach to solve this dilemma. Also, please review the tests, maybe I added unuseful test cases and that is the root cause of my mistaken task understanding.

pbiggar:
> I don't understand how in current types we can check that a variable `x` in the middle of pipeline content a lambda in `ProgramTypesToRuntimeTypes`.

Ah, I see where the issue is. We don't expect that that is caught during ProgramTypesToRuntimeTypes. That will be caught at runtime later (in the future when we have compile-time type checking, it will be caught there).

> Also, I can't realize how to organize the `DError` propagation through all pipeline functions. I see only two options here, either to check those cases in the interpreter - current PR or enrich `PT.Expr.EVariable` by additional meta information about Variable content, same for `PT.Expr.EFnCall` we need to know the type of content to `DError` propagation.

I'm not sure I follow as I don't think we actually need to propagate DErrors through th e pipeline. The pipeline is just a syntax in ProgramTypes that maps to (mostly) EApply
in function types, so I think there shouldn't be any specific error propagation code involved.

> Also, please review the tests, maybe I added unuseful test cases and that is the root cause of my mistaken task understanding.

Tests look good!

ditansu:
> Ah, I see where the issue is. We don't expect that that is caught during ProgramTypesToRuntimeTypes. That will be caught at runtime later (in the future when we have compile-time type checking, it will be caught there).

Hm, so in fact we have to allow any expression into pipeline excepting `literals (e.g. EBool, EString, etc)`,` ELet` and maybe `EUnit` am I right?  

>Tests look good!

Thanks, but as you could see there are a lot of test about [test propagation](https://github.com/darklang/dark/pull/4848/files#diff-30449b4569973f93055966f17bd6427c982c20bced61682f7aa9db07abc1db42R222-R225) in right (expected) order, if we are just folding a pipeline to the expression we can't control this. So we have to choose either we control pipeline transformation process or we have to change to `EApply` logic in `eval'` to propagate `DError`, or we change tests :) please make a choice or maybe offer another option.

pbiggar:
> > Ah, I see where the issue is. We don't expect that that is caught during ProgramTypesToRuntimeTypes. That will be caught at runtime later (in the future when we have compile-time type checking, it will be caught there).
> 
> Hm, so in fact we have to allow any expression into pipeline excepting `literals (e.g. EBool, EString, etc)`,` ELet` and maybe `EUnit` am I right?

I'd expect to support

```
  | EInfix of id * Infix * Expr * Expr
  | ELambda of id * List<id * string> * Expr
  | EVariable of id * string
  | EFnCall of id * FQFnName.T * typeArgs : List<TypeReference> * args : List<Expr>
```

In the future, I'd also support EConstructor and EFieldAccess but we don't need them yet.


 
> > Tests look good!
> 
> Thanks, but as you could see there are a lot of test about [test propagation](https://github.com/darklang/dark/pull/4848/files#diff-30449b4569973f93055966f17bd6427c982c20bced61682f7aa9db07abc1db42R222-R225) in right (expected) order, if we are just folding a pipeline to the expression we can't control this. So we have to choose either we control pipeline transformation process or we have to change to `EApply` logic in `eval'` to propagate `DError`, or we change tests :) please make a choice or maybe offer another option.

I'm not sure exactly what you're looking at here, but I think it's fine to change the exact error message that comes out in some tests.

FWIW, I'm pretty sure we already propagate DErrors ins EApply in `eval'`, though it's possible we missed a spot.

What do you mean by "control pipelines transformation process"?

ditansu:
> I'd expect to support
> ```fsharp
>   | EInfix of id * Infix * Expr * Expr
>   | ELambda of id * List<id * string> * Expr
>   | EVariable of id * string
>   | EFnCall of id * FQFnName.T * typeArgs : List<TypeReference> * args : List<Expr>
> ```
> In the future, I'd also support EConstructor and EFieldAccess but we don't need them yet.

ok, but I'm expecting what a lot of my new tests will be broken: as minimum all for `EConstructor` and probably a lot of for  `DError`. I will mark theirs as `//TODO`  

> FWIW, I'm pretty sure we already propagate DErrors ins EApply in eval', though it's possible we missed a spot.

I'm not sure but sure I'm going check it :) 

> What do you mean by "control pipelines transformation process"?

Yeah, I wrote not clear. I mean that we can check a variable content during pipelines transformation process in `eval'` when we have `Dval`. 

As for error order. For the current `EApply` implementation in `eval'`, follow testcase will fail.
```fsharp 
(1
 |> 2
 |> "error"
 |> (+) 3) = Test.typeError_v0 "Expected a function value, got something else: 2"
```
```console
expected: DError (SourceNone, "Expected a function value, got something else: 2")
  actual: DError (SourceNone, "Expected a function value, got something else: "error"")",
```
But I and probably others Dark user expect that this test would be passed. Because `2` is wrong literal at the start of pipeline. Or is it not worth it and we can score on the order of the errors?

pbiggar:
> > What do you mean by "control pipelines transformation process"?
> 
> Yeah, I wrote not clear. I mean that we can check a variable content during pipelines transformation process in `eval'` when we have `Dval`.

The pipeline transformation happens before execution/eval, so we need to rely on regular mechanisms for error handle. And I think those are already in place, so we probably don't need to do anything special for that.

 
> As for error order. For the current `EApply` implementation in `eval'`, follow testcase will fail.
> 
> ```fsharp
> (1
>  |> 2
>  |> "error"
>  |> (+) 3) = Test.typeError_v0 "Expected a function value, got something else: 2"
> ```
> 
> ```
> expected: DError (SourceNone, "Expected a function value, got something else: 2")
>   actual: DError (SourceNone, "Expected a function value, got something else: "error"")",
> ```

I think this should fail in the parser, and we'll be trying to put `2` into a `PipeExpr` which won't be valid. We can remove that test as a result (ideally, we'd have tests for bad parsing, but they don't exist right now. Maybe as a followup PR).

PR #4852, opened 2023-04-15
Draw the rest of the owl

pbiggar:
image

This builds on a handful of PR's from last week:
- https://github.com/darklang/dark/pull/4838
- https://github.com/darklang/dark/pull/4829
- https://github.com/darklang/dark/pull/4818
- https://github.com/darklang/dark/pull/4817

This handles most of the work involved in moving to proper `DRecords` instead of `DDict`s:

- removes `DType` and uses `TypeReference` everywhere
- makes DBs use a `TypeReference` instead of a weird schema thing(!!!)
  - remove related opcodes
  - fix type parameters and type checking
- remove `DType and UserDB type checking and use `TypeInference.unify` (which is not complete)
- type check return types
- parse return types
- remove old `queryExactKeys` functions that are not supported by the type system
- change test format from `[<DB>] type X = { ... }` to `type X = { ... }; [<DB>] type XDB = DB`
- Adds required type names to record and dict literals: `Dict { x = 5 }` and `Person { name = ... }`
- Disallow fieldAccess for Dicts
- restructure error messages as we no longer get types from dvals
- add `EDict` (dict literals)
- `DB.queryOIneWithKey` returns `Option<String*'a>` instead of `Option<Dict<'a>>`
- remove package manager special types
- `TypeChecker.unify` only returns the first error now

This also prepares for allowing DBs of any type, not just records!

Also:
- remove unused ClientTypes
- fix uncaught exception test error messages
- supports querying based on tuples
- disable some old stats code that could be resurrected later
- allow querying expressions where the first expression is evalutated before the query is run
- add return types to all functions

TODO:
- [ ] anything marked TYPESCLEANUP
- [ ] expand testing for DBs for any type
- [ ] look through what needs to be done for JSON parsing
- [ ] adding FQTypeNames to `DRecord`
- [ ] test querying based on tuples
- [ ] remove `HttpResponse`, `DOption` and `DResult`
- [ ] look at List tests in QueryEquality
- [ ] when creating DRecords, check the type exists (or maybe in the parser)
- [ ] use type arguments in type checking everywhere
- [ ] remove `Record.set`
- [ ] support queries where the value is not a record


pbiggar:
TODOs added to Notion: https://www.notion.so/darklang/backend-work-to-be-done-0c191de14aed45d8892a3d289d0b5885?pvs=4#2b478a1da5b4403cadcb0d1a82c4be89

PR #4853, opened 2023-04-16
Small interpreter fixes

pbiggar:
I unintentionally let this PR spiral out of control.

It started as a couple of small tidy-ups in the Interpreter, including one nice little refactor.

Then I started adding types to test files.

All of a sudden I was adding a package manager. Fortunately I managed to cut this off, and this PR only adds an initial refactoring, where we reorganize the project a bit:
- ExecHost renamed to ProdExec
- add LocalExec (same purpose, but more local-oriented functionality)
- move LibCli to its own project called StdLibCli


scripts/run-local-exec
```diff
@@ -16,9 +16,9 @@ do
 done
 
 if [[ "$PUBLISHED" == "true" ]]; then
-  EXE="backend/Build/out/ExecHost/Release/net7.0/linux-x64/publish/ExecHost"
+  EXE="backend/Build/out/LocalExec/Release/net7.0/linux-x64/publish/LocalExec"
 else
-  EXE="backend/Build/out/ExecHost/Debug/net7.0/linux-x64/ExecHost"
+  EXE="backend/Build/out/LocalExec/Debug/net7.0/LocalExec"
```
StachuDotNet, line 21: a bit old, but I just noticed that this omits the 'linux-x64' part of the directory, which seems inconsistent with other references. Just raising the point in case it was a mistake

pbiggar: That got removed from a bunch of places as we no longer only build on linux-x64 and so .NET doesn't include the directory when emitting the output executables.

So I think this was intentional

PR #4866, opened 2023-04-24
Add treesitter parser and wasm demo

StachuDotNet:
Changelog:

```
Language and Stdlib
- update Json.parse<'a> to handle Options and Records better

Internal
- add tree-sitter parser, and demo in JS/WASM
- replace `dark-serve-blazor-assets` canvas with more general `dark-serve-static` one
- add `WASM.callJSFunction` StdLib fn, only available in the WASM runtime
- simplify loading and interacting with WASM'd Dark runtime
```

Some additional details regarding the tree-sitter parser:
- add tree-sitter parser
  - very basic grammar.js
  - "build" process to yield parser in C
- compile that parser to WASM
- add new canvas to demo use of the WASM-compiled tree-sitter parser
- extend that^ demo to:
  - map the SyntaxTree output by Tree-Sitter into a "simple" syntax tree
    ```typescript
    type SyntaxTreeNode = 
      { typ: String; text: String; fieldName: Option<String>; children: 
    ```
  - send the JSON-serialized root SyntaxTreeNode to a Dark endpoint, and parse it as a PT.UserFunction
  - JSON-serialize the result


StachuDotNet:
for reference, here's the syntax tree JSON for the sample function
https://gist.github.com/StachuDotNet/686858b2ea83672afcf1d7feb65adc56

backend/src/StdLibExecution/Libs/Json.fs
```diff
@@ -410,16 +411,20 @@ let parse
 
       | CustomType.Record (firstField, additionalFields) ->
         let fieldDefs = firstField :: additionalFields
+        let enumerated = j.EnumerateObject() |> Seq.toList
+
         let dvalMap =
-          j.EnumerateObject()
-          |> Seq.map (fun jp ->
```
StachuDotNet, line 415: Previously, this would error if the JSON being parsed had any _extra_ fields, but would not error when fields were missing.
This inverts that a bit - all fields defined in the Record type are required and _extra_ fields in the JSON object are ignored.

backend/src/StdLibExecution/Libs/Json.fs
```diff
@@ -410,16 +411,20 @@ let parse
...
-            let converted = convert correspondingType jp.Value
-            (jp.Name, converted))
+          fieldDefs
+          |> List.map (fun def ->
+            let correspondingValue =
+              enumerated
+              // TODO: handle case where value isn't found for
+              // and maybe, if it's an Option<>al thing, don't complain
+              |> List.find (fun v -> v.Name = def.name)
+              |> fun field -> field.Value
```
StachuDotNet, line 424: WDYT should happen for:
```fsharp
type Person = { Age: Option<Int>; Name: String }
JSON.parse<Person> """{ "Name": "John" }"""
```
?

I'm inclined to think the Age should just be set to None.

backend/static/tree-sitter/tree-sitter-darklang.wasm
StachuDotNet, line 1: maybe this shouldn't be checked in?
is it worth including the parser-building (incl. WASM bindings) in this PR, or later?

canvases/dark-tree-sitter-demo/main.dark
```diff
@@ -0,0 +1,377 @@
...
+        <textarea id="syntax-textarea" class="textarea" disabled></textarea>
+      </div>
+
+      <div class="right-side">
+        <label>Vanilla-serialized JSON of PT.UserFunction</label>
+        <textarea id="right-textarea" class="textarea" disabled></textarea>
+      </div>
+    </div>
+    <script src="/static/tree-sitter/tree-sitter.js" />
+    <script>console.log('NOT SURE WHY I NEED THIS')</script>
```
StachuDotNet, line 74: TODO: I need to check if I still need this - for some reason, the tree-sitter bindings seemed to fail unless I added an extra script tag above the main one.

canvases/dark-tree-sitter-demo/main.dark
```diff
@@ -0,0 +1,377 @@
...
+          cursor.gotoParent();
+        }
+
+        return {
+          typ: cursor.nodeType,
+          text: cursor.nodeText,
+          fieldName: cursor.currentFieldName(),
+          children: children,
+        };
+      }
```
StachuDotNet, line 93: I tried to walk this tree in a more obvious way (recurse more organically by calling `children.map(child => simplifyTree(child)` etc) but unless you _.walk()_ the whole time, somehow the `currentFieldName` gets lost. (even if you .walk() on a node later.

canvases/dark-tree-sitter-demo/main.dark
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        function sourceCodeToSyntaxTree () {
+          const darkSourceCode = document.getElementById("input-textarea").value;
+          const darkTree = parser.parse(darkSourceCode);
+          window.tree = darkTree;
+
+          let simpleTree = simplifyTree(darkTree.rootNode.walk());
+
+          // TODO: send _this_ to (in-WASM) Dark, to be mapped into an Expr
```
StachuDotNet, line 142:
```suggestion
```
remove comment

canvases/dark-tree-sitter-demo/main.dark
```diff
@@ -0,0 +1,375 @@
...
+
+
+let gid (): Int =
+  Int.random_v1 0 9223372036854775807
+
+
+let err (msg: String): Unit =
+  // TODO: maybe something with the msg?
+  let _ = 1.0/0.0
+  ()
```
StachuDotNet, line 167: It turns out that this ends up just returning `()`, without causing a failure like I expected.

canvases/dark-tree-sitter-demo/main.dark
```diff
@@ -0,0 +1,375 @@
...
+let prettyPrintExpr (expr: PTExpr): String =
+  match expr with
+  | PTExpr.EInt (_, i) -> Int.toString i
+  | PTExpr.ELet (_, identifier, expr, body) ->
+    "let " ++ identifier ++ " = " ++ (prettyPrintExpr expr) ++ "\n" ++ (prettyPrintExpr body)
+  | PTExpr.EVariable (_, name) -> name
+  | PTExpr.EFnCall (_, fnName, args) ->
+    // TODO: wrap results of prettyPrintExpr in () if there would be ambiguity
+    let argsPart = args |> List.map (fun a -> prettyPrintExpr a) |> String.concat " "
+    fnName ++ " " ++ argsPart
```
StachuDotNet, line 193:
```suggestion
```

parser/tree-sitter-darklang/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "tree-sitter-darklang"
+description = "darklang grammar for the tree-sitter parsing library"
+version = "0.0.1"
+keywords = ["incremental", "parsing", "darklang"]
+categories = ["parsing", "text-editors"]
+repository = "https://github.com/tree-sitter/tree-sitter-darklang"
```
StachuDotNet, line 7:
```suggestion
repository = "https://github.com/darklang/dark"
```

parser/tree-sitter-darklang/README.md
```diff
@@ -0,0 +1,54 @@
...
+- `code ~/.config/tree-sitter/config.json`
+  edit the config; add the path of `dark/parser` to the `parser-directories` array
+  for example, I keep my `dark` repo in `~/code/dark`,
+  so I added "/home/stachu/code/dark/parser" there.
+- `tree-sitter dump-languages` to make sure you see `darklang` there somewhere
+- `tree-sitter generate` to generate parser from grammar
+- `tree-sitter parse ./demo.dark` to parse a hardcoded file
+- `tree-sitter test` to test the parser against a 'corpus' of test cases found in `test/corpus`
+
+Note: `export PATH=$PATH:./node_modules/.bin` may be needed at some point? I forget.
```
StachuDotNet, line 29: TODO: follow up on this - when I restarted my machine, I _did_ have to do this for something.

parser/tree-sitter-darklang/grammar.js
```diff
@@ -0,0 +1,75 @@
...
+      field('body', $.expression)
+    ),
+
+    // fns need at least one param
+    // TODO: should this rule be enforced like this _in the parser_?
+    // TODO: we could just repeat1 here
+    fn_params_def: $ => seq(
+      field("first", $.fn_param_def),
+      field("additional", repeat($.fn_param_def))
+    ),
```
StachuDotNet, line 25: I'm probably over-complicating things by trying to enforce such a rule in the grammar? Regardless, this particular set-up makes the SyntaxTree->PT.UserFunction process awkward.

parser/tree-sitter-darklang/grammar.js
```diff
@@ -0,0 +1,75 @@
+module.exports = grammar({
+  name: 'darklang',
+
+  // Annoyingly, it seems to be best to write these 'top-most first'
```
StachuDotNet, line 4:
```suggestion
```

canvases/dark-tree-sitter-demo/main.dark
```diff
@@ -0,0 +1,363 @@
...
+    </script>
+  </body>
+</html>
+  """
+
+  Http.responseWithHeaders (String.toBytes body) (Dict.fromListOverwritingDuplicates [("Content-Type", "text/html")]) 200
+
+
+let gid (): Int =
+  Int.random_v1 0 9223372036854775807
```
StachuDotNet, line 161: This is currently just returning `0`; not sure why (haven't researched)

OceanOak:
I pulled this PR locally to test the changes, but I'm having trouble loading dark-editor from disk. Are you able to load it? It seems that the issue is related to this handler: [<HttpHandler("POST", "/api/gpt4")>]. When I comment it out, it works, but I'm not sure why yet.

StachuDotNet:
I'm struggling a bit getting the parser set up well within the devcontainer.

My aim is to create a `build-parser` script that
- installs relevant npm dependencies
- builds the parser from the grammar
- builds the WASM bindings
- copies the `web-tree-sitter` artifacts (.js and .wasm) to the `backend/static/tree-sitter` dir,
  along with the WASM bindings for darklang generated in the previous step
- allows a user to quickly, afterwards, run commands such as `tree-sitter test`
  (potentially through a new `./scripts/run-parser-tests`)

---

There are a few complications/annoyances in setting this up, though:

1. tree-sitter requires a C++ compiler; installing `build-essential` seems to be recommended/easy
OK - handled by including this in the dockerfile.

2. tree-sitter assumes that you either have:
- tree-sitter installed globally `npm i -g tree-sitter-cli`
- tree-sitter installed 'locally', with an `export PATH=$PATH:./node_modules/.bin`'

the former seems best for us, done in the dockerfile - correct?

3. tree-sitter's CLI depends on a `config.json` file that includes a list of `paths` where it should look to find your parser definitions.
  
By default, the list is as follows:
`"parser-directories": [ "/home/dark/github", "/home/dark/src", "/home/dark/source" ]`

Here's the problem - this file is created by an `tree-sitter init-config` command, which yields the file at `/home/[user]/.config/tree-sitter/config.json`. I don't see any way to direct this file to be in some other place.

This file should be checked in, and the list updated to include the `parser` dir.
Since it's in an odd place, the best I can think to do would be to check in such a file _elsewhere_ and copy it to that directory when we set up (either in the dockerfile or in the `build-parser` script). I'm not sure how to handle this, though.

PR #4868, opened 2023-04-26
Experiment1 followups

OceanOak:
No Changelog

- Add prettier and eslint
- Display execution result 
- Add loading animation while waiting for API response 

Currently

https://user-images.githubusercontent.com/87861924/235358638-f13ed8f5-b735-4e88-9e3f-68195cf1f982.mov


OceanOak:
Issues :
- We need to display one result component per response, rather than one result for all responses. 
- Each result component should be rendered separately to prevent updating all components when only one is changed.

 Currently, only the last execution value is displayed in one result component. When I attempted to display a result per response, all the result components' values changed based on the last execution. 
 Any suggestion on how to address these issues is  appreciated

PR #4875, opened 2023-05-04
wasm-experimental test

StachuDotNet:
Changelog:

```
Area of change
- details
```

(only the most recent commit here is relevant)

Briefly tried out `wasm-experimental` workflow to see if we could get multithreading for free.

I got the workload installed, and code written to try and spawn a new thread, but then met with:\
```
Uncaught (in promise) Error: Failed to start platform. Reason: ReferenceError: SharedArrayBuffer is not defined
```

apparently this might be resolvable with some headers:
```
("Cross-Origin-Embedder-Policy", "require-corp")
("Cross-Origin-Opener-Policy", "same-origin")
```

but I'm not sure which requests/responses need those - maybe all, in which case serving the static assets by a different host won't work...

This sort of feels like a dead-end due to security+convenience constraints.


StachuDotNet:
image
if I load the static assets from the same dir, I hit a different issue

PR #4876, opened 2023-05-04
Fix build issues caused by emscripten

OceanOak:
No changelog

Fix the emscripten installation failed error on an arm64 machine

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -423,8 +423,9 @@ ENV PATH "$PATH:/home/dark/bin:/home/dark/.dotnet/tools"
 #############
 RUN git clone https://github.com/emscripten-core/emsdk.git \
   && cd emsdk \
-  && ./emsdk install 3.1.37 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -423,8 +423,9 @@ ENV PATH "$PATH:/home/dark/bin:/home/dark/.dotnet/tools"
 #############
 RUN git clone https://github.com/emscripten-core/emsdk.git \
   && cd emsdk \
-  && ./emsdk install 3.1.37 \
-  && ./emsdk activate 3.1.37
+  # we are using the latest version because Linux arm64 binaries aren't available in all releases
```
StachuDotNet, line 426:
```suggestion
  # TODO pin to a recent stable version (i.e. 3.1.37)
  # we are using the latest version because Linux arm64 binaries aren't available in all releases.
  # see: https://github.com/emscripten-core/emscripten/issues/19275
```

PR #4880, opened 2023-05-07
Move X509 and Password to StdLibExecution

ditansu:
Move X509 and Password stuff from `StdLibCloudExecution` to `StdLibExecution\Crypto.fs`


pbiggar (changes requested):
Thanks!

I'd prefer to keep the files named the same ("Password" and "X509"), as the filename should match the module.

I think also this PR adds the emsdk submodule, which it shouldn't.

pbiggar:
Great, thanks so much!

PR #4882, opened 2023-05-07
Enable WASM eval of full user program (types, fns, exprs) rather than just expr

StachuDotNet:
Changelog:

```
WASM Runtime
- enable WASM eval of full user program (types, fns, exprs) rather than just expr

canvases/dark-editor
- eval full user program rather than just expr

canvases/dark-repl
- eval full user program rather than just expr
```


StachuDotNet:
image

PR #4884, opened 2023-05-07
canvases/dark-editor: move logic to Dark code

StachuDotNet:
Changelog:

```
Area of change
- details
```

This PR:
- generally improves the Wasm runtime + JS integration (fewer methods available to JS - i.e. 'eval' is now a new WASM.Editor fn)
- rewrites the dark-editor (incl. dark-editor-vue) such that
  - the modeling, state, update-handling etc has been moved from Vue/TS to Dark (client.dark)
  - prompt responses are now broken down by Text and CodeSnippet sections, so we don't demand that only code is returned
  - improves the 'system prompt' a bit, regarding our more immediate uses anyway
  - some other minor improvements

Along with the improvements, dark-editor is now worse in a few ways because I rewrote it:

- styling is more basic right now (ocean already volunteered to take care of this)
- Http requests block the UI thread (they didn't before, because they weren't in Dark)
- the editing UI for the code snippets is basic again (no nice codemirror)
- fontawesome icons are gone
- the "variables" extraction/UI isn't there (I think we can live without this for a while)
- you can't save the system prompt after editing it (probably fine)

It's going on for a bit, and I'm starting to tire of the PR, so I'm thinking to merge with the few things being worse.

I think it'd be reasonable to handle fontawesome, codemirror, and styling in a separate PR (Ocean, are you up for that?)
and "don't block the UI thread on HTTP requests" in another (I'm happy to take this on ~Monday - I have some ideas for this, but open to input).

https://github.com/darklang/dark/assets/906686/3d8ccdaf-344e-42e0-a2bd-44cee54d807f

WDYT?

If that sounds reasonable, I'll do a final pass on the PR in the morning, then merge.


StachuDotNet:
Copying commentary from a particular commit:

> [Replace dark-editor JS code so its state is handled in Dark runtime](https://github.com/darklang/dark/pull/4884/commits/81cea640d465978d2259c30499aada10e8db302d) 
> [81cea64](https://github.com/darklang/dark/pull/4884/commits/81cea640d465978d2259c30499aada10e8db302d)
> I tried really hard to keep all the functionality we had but it got tough,
> and ended up deciding to clean the slate. I'll be re-adding
> functionality in the follow-up commits.

specifically I was finding it tough to transition the state-management from the way it was, to a Dark-centric management style. In any case, I'll do my best to replicate styling+functionality tomorrow.

backend/src/Wasm/DarkEditor.fs
```diff
@@ -55,13 +55,44 @@ let getStateForEval
     executingFnName = None }
 
 
+
+/// Any 'loose' exprs in the source are mapped without context of previous/later exprs
+/// so, a binding set in one 'let' will be unavailable in the next expr if evaluated one by one.
```
StachuDotNet, line 60: I should either explain this better, or figure out another solution - I got stuck on this for a while so went with the first solution I could come up with.

backend/src/Parser/ProgramTypes.fs
```diff
@@ -564,6 +562,27 @@ module Expr =
 
       PT.EFnCall(gid (), PT.FQFnName.userFqName modules name version, typeArgs, [])
 
+    // TODO: fix this before merging - it's basically a dupe of the above
```
StachuDotNet, line 565: TODO

backend/src/Wasm/Libs/HttpClient.fs
```diff
@@ -88,6 +89,14 @@ module HttpClient =
               VersionPolicy = System.Net.Http.HttpVersionPolicy.RequestVersionOrLower
             )
 
+          let strBody = System.Text.Encoding.UTF8.GetString httpRequest.body
+
+          // CLEANUP: BodyContent, used by the 'regular' HttpClient,
+          // doesn't seem to work in the Wasm runtime,
+          // so we'll have to figure out some other way of supporting
+          // non-text bodies in that environment
+          if strBody <> "" then req.Content <- new StringContent(strBody)
```
pbiggar, line 98: I took a look at the Fetch API and it seems to have support for bytes, so this isn't a fundamental problem.

PR #4891, opened 2023-05-10
List destructuring

ditansu:
The list destructuring for `match with` expression. 
Usage cases:  
```fsharp
1. head :: tail 
2. a :: b :: c... :: n :: rest
3. (a::b::ab_rest)::(x::y::xy_rest)::rest 
```

### DIFF
--- a/backend/src/LibAnalysis/ClientRuntimeTypes.fs
+++ b/backend/src/LibAnalysis/ClientRuntimeTypes.fs
@@ -233,7 +233,7 @@ type MatchPattern =
   | MPUnit of id
   | MPTuple of id * MatchPattern * MatchPattern * List<MatchPattern>
   | MPList of id * List<MatchPattern>
-  | MPListCons of id * heads : List<MatchPattern> * tail : MatchPattern
+  | MPListCons of id * head : MatchPattern * tail : MatchPattern
 
 module MatchPattern =
   let rec fromCT (p : MatchPattern) : RT.MatchPattern =
@@ -251,7 +251,7 @@ module MatchPattern =
     | MPTuple (id, first, second, theRest) ->
       RT.MPTuple(id, r first, r second, List.map r theRest)
     | MPList (id, pats) -> RT.MPList(id, List.map r pats)
-    | MPListCons (id, heads, tail) -> RT.MPListCons(id, List.map r heads, r tail)
+    | MPListCons (id, head, tail) -> RT.MPListCons(id, r head, r tail)
 
   let rec toCT (p : RT.MatchPattern) : MatchPattern =
     let r = toCT
@@ -268,7 +268,7 @@ module MatchPattern =
     | RT.MPTuple (id, first, second, theRest) ->
       MPTuple(id, r first, r second, List.map r theRest)
     | RT.MPList (id, pats) -> MPList(id, List.map r pats)
-    | RT.MPListCons (id, heads, tail) -> MPListCons(id, List.map r heads, r tail)
+    | RT.MPListCons (id, head, tail) -> MPListCons(id, r head, r tail)
 
 module CustomType =
   // TYPESCLEANUP support type parameters
--- a/backend/src/LibAnalysis/ClientRuntimeTypes.fs
+++ b/backend/src/LibAnalysis/ClientRuntimeTypes.fs
@@ -233,6 +233,7 @@ type MatchPattern =
   | MPUnit of id
   | MPTuple of id * MatchPattern * MatchPattern * List<MatchPattern>
   | MPList of id * List<MatchPattern>
+  | MPListCons of id * heads : List<MatchPattern> * tail : MatchPattern
 
 module MatchPattern =
   let rec fromCT (p : MatchPattern) : RT.MatchPattern =
@@ -250,6 +251,7 @@ module MatchPattern =
     | MPTuple (id, first, second, theRest) ->
       RT.MPTuple(id, r first, r second, List.map r theRest)
     | MPList (id, pats) -> RT.MPList(id, List.map r pats)
+    | MPListCons (id, heads, tail) -> RT.MPListCons(id, List.map r heads, r tail)
 
   let rec toCT (p : RT.MatchPattern) : MatchPattern =
     let r = toCT
@@ -266,6 +268,7 @@ module MatchPattern =
     | RT.MPTuple (id, first, second, theRest) ->
       MPTuple(id, r first, r second, List.map r theRest)
     | RT.MPList (id, pats) -> MPList(id, List.map r pats)
+    | RT.MPListCons (id, heads, tail) -> MPListCons(id, List.map r heads, r tail)
 
 module CustomType =
   // TYPESCLEANUP support type parameters
--- a/backend/src/LibAnalysis/ClientProgramTypes.fs
+++ b/backend/src/LibAnalysis/ClientProgramTypes.fs
@@ -278,6 +278,7 @@ type MatchPattern =
   | MPUnit of id
   | MPTuple of id * MatchPattern * MatchPattern * List<MatchPattern>
   | MPList of id * List<MatchPattern>
+  | MPListCons of id * heads : List<MatchPattern> * tail : MatchPattern
 
 module MatchPattern =
   let rec fromCT (pat : MatchPattern) : PT.MatchPattern =
@@ -294,6 +295,8 @@ module MatchPattern =
     | MPTuple (id, first, second, theRest) ->
       PT.MPTuple(id, fromCT first, fromCT second, List.map fromCT theRest)
     | MPList (id, pats) -> PT.MPList(id, List.map fromCT pats)
+    | MPListCons (id, heads, tail) ->
+      PT.MPListCons(id, List.map fromCT heads, fromCT tail)
 
   let rec toCT (pat : PT.MatchPattern) : MatchPattern =
     match pat with
@@ -309,7 +312,8 @@ module MatchPattern =
     | PT.MPTuple (id, first, second, theRest) ->
       MPTuple(id, toCT first, toCT second, List.map toCT theRest)
     | PT.MPList (id, pats) -> MPList(id, List.map toCT pats)
-
+    | PT.MPListCons (id, heads, tail) ->
+      MPListCons(id, List.map toCT heads, toCT tail)
 
 type BinaryOperation =
   | BinOpAnd
--- a/backend/src/LibAnalysis/ClientProgramTypes.fs
+++ b/backend/src/LibAnalysis/ClientProgramTypes.fs
@@ -278,7 +278,7 @@ type MatchPattern =
   | MPUnit of id
   | MPTuple of id * MatchPattern * MatchPattern * List<MatchPattern>
   | MPList of id * List<MatchPattern>
-  | MPListCons of id * heads : List<MatchPattern> * tail : MatchPattern
+  | MPListCons of id * head : MatchPattern * tail : MatchPattern
 
 module MatchPattern =
   let rec fromCT (pat : MatchPattern) : PT.MatchPattern =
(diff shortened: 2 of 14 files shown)


ditansu:
@pbiggar please approve [test cases](https://github.com/darklang/dark/pull/4891/files#diff-30449b4569973f93055966f17bd6427c982c20bced61682f7aa9db07abc1db42R434-R458)

pbiggar:
Tests look great!

I would also consider tests for nested destructing.

backend/src/LibAnalysis/ClientRuntimeTypes.fs
```diff
@@ -233,6 +233,7 @@ type MatchPattern =
   | MPUnit of id
   | MPTuple of id * MatchPattern * MatchPattern * List<MatchPattern>
   | MPList of id * List<MatchPattern>
+  | MPListCons of id * heads : List<MatchPattern> * tail : MatchPattern
```
StachuDotNet, line 236: Could the `tail` ever be anything other than an `MPList`?
I can't think of a case - in which case, a `List<MatchPattern>` for that as well may be ideal

pbiggar (commented):
Similar to @StachuDotNet's comments, I don't have a good intuition for why a ListCons is `| MPListCons of id * heads : List<MatchPattern> * tail : MatchPattern`.

Could you share why you modelled it like that?

ditansu:
@pbiggar and @StachuDotNet thanks, guys, for your questions. I've tried a few approaches to representing `ListCons` and, for me in the context of the task, it's the simplest way to present a list as `one`, `two`...`N` elements it is `heads` plus one `tail`.  IMO that design is very convenient for parsing and interpretation. Yeah, the tail is always a list, but it doesn't matter for the List destructuring aims. Besides, that approach avoids additional noise like `List.map toRC/fromPT/.. tail` in transformations. Anyway, if you, guys, decide that this approach is too hard to understanding, then I might redesign it. In that case, please hint me the desired `MPListCons` design.

backend/src/LibAnalysis/ClientRuntimeTypes.fs
ditansu: @StachuDotNet year you right, but what gives us - present tail as a `List<MatchPattern>`?  If you see how that changing help us to simplify the List deconstruction code, please show me, I'll be very appreciated.

StachuDotNet: Actually, I think I'm wrong. I was previously only thinking about this kind of use:
```
match [1; 2; 3; 4] with
| 1::2::[3; 4] ->
```
, which sort of shaped my thinking - namely my thinking was that "the tail of an MPListCons should be irrepresentable as anything other than a list - so the `tail` field should be a list". But, these are also valid:
```
match [1; 2; 3; 4] with
| 1::2::_ -> ...
| 1::2::theRest -> ...
```
These feel like the correct way to represent all of these:
```
match [1; 2; 3; 4] with
| 1::2::[3; 4] -> ... // MPListCons [MPInt 1; MPInt 2] MPList [MPInt 3; MPInt4] ]
| 1::2::_ -> ... // MPListCons [MPInt 1; MPInt 2] MPWildCard
| 1::2::theRest -> ... // MPListCons [MPInt 1; MPInt 2] (MPVariable "theRest")
```

(MPWildCard doesn't exist right now but I could imagine it forking out of MPVariable for some reason)

StachuDotNet:
fwiw I went back on my thinking and believe that the current setup is great. https://github.com/darklang/dark/pull/4891#discussion_r1194369669

backend/src/LibAnalysis/ClientRuntimeTypes.fs
ditansu: I'm sure I can support `MPWildCard` in this PR or next one. BTW, we need to support a check for incomplete pattern matches, but now I have no idea how to do it :)

StachuDotNet: I'm not yet sure if there's a point for MPWildCard. MPVariable with a "_" variable seems to serve the same purpose right now. The only difference that I can think of would be not adding _ to the symtable, but that's such a tiny cost that whatever?

I do keep thinking that there might be some _point_ to an MPIgnore/MPWildcard/etc, but haven't figured what it is yet. LMK if you think of one!

> BTW, we need to support a check for incomplete pattern matches, but now I have no idea how to do it :)
could you please rephrase this?
Do you mean something like:

```fsharp
let t = true
match t with
| true -> "yes"
| false -> "no"
```
this is "complete"

```
let i = 1
match 1 with
| 17 -> "yes"
```
this is "incomplete"

and we should do _something_ about it?

StachuDotNet: > BTW, we need to support a check for incomplete pattern matches, but now I have no idea how to do it :)
could you please rephrase this?
Do you mean something like:

```fsharp
let t = true
match t with
| true -> "yes"
| false -> "no"
```
this is "complete"

```fsharp
let i = 1
match 1 with
| 17 -> "yes"
```
this is "incomplete"

and we should do _something_ about it?
Maybe type-checking or something?
What do we currently do when this is the case? I forget, let's start there.

ditansu: > I do keep thinking that there might be some point to an MPIgnore/MPWildcard/etc, but haven't figured what it is yet. LMK if you think of one!

yeah, probably `MPVariable with a "_"` is good solution here, ok let's stay it as is :) 

>and we should do something about it?

yes, you totally correct understood me. As for what we should do, e.g. we should to warn user about this case like F# do it

```fsharp
> let f (a: int) : int =                                          
-   match [1;2;3;4] with
-    | a::b::rest -> a 
-    | head::tail -> a*a
- ;;
warning FS0025: Incomplete pattern matches on this expression. For example, the value '[]' may indicate a case not covered by the pattern(s).
```
Currently, as I understand, we do nothing in that case :) 

Anyway we can add all those things next time. Could you accept this PR and merge it, or changes need yet?

pbiggar:
What do we think about

` | MPListCons of id * head : MatchPattern * tail : MatchPattern`

This is what F# does roughly. My thinking is that there's an ID for each cons here (as opposed to one ID for all the head elements), and it removes the potential state of an empty cons list (which feels meaningless).

I can't come up with a better thing for the tail - I think MatchPattern is probably the right answer there.

StachuDotNet:
> What do we think about
> 
> ` | MPListCons of id * head : MatchPattern * tail : MatchPattern`
> 
> This is what F# does roughly. My thinking is that there's an ID for each cons here (as opposed to one ID for all the head elements), and it removes the potential state of an empty cons list (which feels meaningless).
> 
> I can't come up with a better thing for the tail - I think MatchPattern is probably the right answer there.

Funny, I fell asleep thinking about this last night.
So, playing this out, 

```fsharp
match [1; 2; 3; 4] with
| 1 :: 2 :: [3; 4] -> 
``
this case would be
```fsharp
MPListCons (MPInt 1) (MPListCons (MPInt 2) (MPList [MPInt 3; MPInt 4]))
```
rather than

```fsharp
MPListCons [MPInt 1; MPInt 2] (MPList [MPInt 3; MPInt 4]))
```

?

---

@ditansu I should be able to review this in about an hour for a bit - will provide feedback then! Thank you for working on this

pbiggar:
> this case would be
> 
> ```fsharp
> MPListCons (MPInt 1) (MPListCons (MPInt 2) (MPList [MPInt 3; MPInt 4]))
> ```
> 
> rather than
> 
> ```fsharp
> MPListCons [MPInt 1; MPInt 2] (MPList [MPInt 3; MPInt 4]))
> ```

yeah, that's right.

backend/src/LibAnalysis/ClientRuntimeTypes.fs
StachuDotNet: I've extracted this out to a discussion: https://github.com/darklang/dark/discussions/4900

Reviewing this PR now

StachuDotNet (changes requested):
This is looking close! A few notes:
- I think the `module Match` of language.tests is becoming large enough to break down a bit into submodules. Maybe define a submodule per type of value we can match against? (enum, int, float, list, etc)
- along the same lines, probably best to keep the MPListCons tests around the same place as the other List tests - right now I think they're elsewhere
- also, for each type of test, best to start with a super super simple test first (just `match [1; 2; 3] with | 1::2::[3]` or something) and then follow up with the more complex tests (i.e. the head/tail tests) after it (this is just a note on the _ordering_ of tests)
- could you please add some relevant test values to `backend/tests/Tests/Serialization.TestValues.fs`, and rerun `./scripts/run-backend-tests`? this should result in some committed test files being updated. Adding values here should also result in some roundtripping tests including this new MPListCons case (I don't think there's additional work on your part needed here; the adjacent Serialization.Vanilla.Tests and Serialization.Binary.Tests should include the new test value automatically)
- I think Paul's suggestion of MPListCons is a worthwhile improvement. a `::` is surrounded by 2 things - a single left part (head) and a single right part (tail). Both are good as MatchPatterns. So, sorry to iterate again, but it's now my stance that ` | MPListCons of id * head : MatchPattern * tail : MatchPattern` would be ideal
- I think ^ this change might be slightly awkward in the interpreter as a refactor, but would result in tidier code in the end. Hopefully this structure fits the model of how we handle match patterns well enough to not be _too_ awkward
- I haven't thorughly read the interpreter changes yet, but will do so after these changes

I think after these changes, and a potential final pass on the interpreter, we should be good to merge

ditansu:
>Funny, I fell asleep thinking about this last night.
>So, playing this out
@pbiggar @StachuDotNet 
Actually, my first attempt was exactly that
```fsharp
| MPListCons of id * head : MatchPattern * tail : MatchPattern
```
but then I saw things in debug like that ```MPListCons (MPInt 1) (MPListCons (MPInt 2) (MPList [MPInt 3; MPInt 4]))``` that seems ugly and probably has not good performance. So that is why I decided to continue to explore other solution.

ditansu:
> I think Paul's suggestion of MPListCons is a worthwhile improvement. a :: is surrounded by 2 things - a single left part (head) and a single right part (tail). Both are good as MatchPatterns. So, sorry to iterate again, but it's now my stance that  | MPListCons of id * head : MatchPattern * tail : MatchPattern would be ideal

@StachuDotNet yeah, I will do all your suggestions, but I'm't not sure about this ^ one.  As I've just wrote above, resulted AST looks very ugly and untidy. But ok, maybe I did something wrong in first time, and now it would be more gracefully. 😁

pbiggar:
> > I think Paul's suggestion of MPListCons is a worthwhile improvement. a :: is surrounded by 2 things - a single left part (head) and a single right part (tail). Both are good as MatchPatterns. So, sorry to iterate again, but it's now my stance that  | MPListCons of id * head : MatchPattern * tail : MatchPattern would be ideal
> 
> @StachuDotNet yeah, I will do all your suggestions, but I'm't not sure about this ^ one. As I've just wrote above, resulted AST looks very ugly and untidy. But ok, maybe I did something wrong in first time, and now it would be more gracefully. 😁

The AST often looks ugly, it's nothing to worry about. As in the PipeExpr PR, modelling the allowed states more accurately generally leads to less errors long term.

ditansu:
@StachuDotNet AST became ugly, but a code now is clear, and simply 😎  I've done all points, excluding serialization. `MPListCons` was added in `Tests.SerializationTestValues` but with no impact. Any ideas?

pbiggar:
> @StachuDotNet AST became ugly, but a code now is clear, and simply 😎 I've done all points, excluding serialization. `MPListCons` was added in `Tests.SerializationTestValues` but with no impact. Any ideas?

Serialization files are regenerated by running the tests. Looks like you've added the right values in the right place so that should hopefully fix it.

ditansu:
>Serialization files are regenerated by running the tests. Looks like you've added the right values in the right place so that should hopefully fix it.

Sure, I ran tests many times, but regeneration didn't happen. Maybe exists some bash script especially for that?

StachuDotNet (approved):
This looks great. I'll figure out what's up with the serialization files immediately after merging, along with a final pass on the interpreter. Thank you!

StachuDotNet:
Followed up a bit in #4901. Do you have time to add analysis/preview tests, as noted in that PR, in a separate PR?

PR #4897, opened 2023-05-15
wip: Make HttpClient.request work with BytesArrayContent body in WASM

StachuDotNet:
Changelog:

```
Area of change
- details
```

Our `HttpClient.request` function fails when:
- run in the WebAssembly runtime
- in a `uply { }`
- if we set the `Content` of the HttpRequest to a BytesArrayContent

It does _not_ fail if we run it outside of a `uply{}`.

I tried updating `HttpClient.fs` itself to make the call synchronous - didn't work.

Not sure what to try here next - there are a few layers of uply _above_ the actual fn impl - will try .net's HttpClient at a few levels of that stack, and see where it fails, maybe.

### DIFF
--- a/backend/src/Wasm/DarkEditor.fs
+++ b/backend/src/Wasm/DarkEditor.fs
@@ -86,3 +86,28 @@ let HandleEvent (serializedEvent : string) : Task<string> =
 
     return LibExecution.DvalReprDeveloper.toRepr result
   }
+
+
+let generateRandomBytes (length: int) =
+  let random = new Random()
+  let buffer = Array.zeroCreate<byte> length
+  random.NextBytes(buffer)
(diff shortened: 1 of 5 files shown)


pbiggar:
I didn't know we had access lower in the stack? Might be better to use that anyway if it exists (rather than some of the dance we do around content-type, etc)

StachuDotNet:
I'm not sure what you mean

pbiggar:
> will try .net's HttpClient at a few levels of that stack

Do we have other HttpClients/level of Http stack we can use? If so, we could also avoid some other hacks we have in place.

StachuDotNet:
WASM abandoned for a few months

PR #4901, opened 2023-05-18
MPListCons minor follow-ups

StachuDotNet:
No changelog.

Follow-up to #4891

- [x] tidy interpreter
- [x] serialization tests
- [ ] analysis/"preview" tests [1]

[1] @ditansu could you please take care of this one in a separate PR?
It should be similar to the tests you added [here](https://github.com/darklang/dark/pull/4840/files#diff-abd5a4fed73ecefb0afd219245616b616e722398d17c907d95ab501d199e9d9f) - there are already tests around some other match patters, albeit not super exhaustive at this point


StachuDotNet:
FYI @pbiggar @ditansu the issue with serialization tests is that we needed to add the MPListCons tests in another place in the test files. The RT.MPListCons had been added, which is used for several round-tripping tests, but we don't do any "consistency" serialization tests around those (probably rightly so). I added PT values as well, which resulted in the test-running updating the relevant files.

PR #4904, opened 2023-05-19
Run wasm runtime in webworker (to stop blocking UI thread in some edge cases)

StachuDotNet:
Changelog:

```
WebAssembly runtime
- now loads in a WebWorker
```

Loads the Darklang WASM runtime _in a webworker_.

In some cases, this un-blocks the UI thread.


StachuDotNet:
I think there's an solution to be extracted here for more general use, but don't have time to spend on that now

PR #4905, opened 2023-05-25
Web-editor spike

OceanOak:
Changelog:

```
Language 
- Add String.indexOf function

canvases/dark-editor
- Start implementing the 3 columns view (tasks and action view, chat view, and code and context view)
```

This PR: 
- Adds the 3 columns view 
- Extracts Tasks from the AI's response
- Separates the code snippets from the chat view and displays them in the code and context view
- Adds String.indexOf function
- Handles some small UI fixes (overflowing and overlaying)
- Adds more error information to parseAndSerializeProgram  (include metadata)

Current state:

https://github.com/darklang/dark/assets/87861924/4f9d0207-fb9d-45ff-9175-c09518fb85d6


OceanOak:
Questions :
- Should I fix the UI in this PR? (This includes setting the height to 100% and making the overflow scrollable within the columns ,improve the styling of the columns, etc.)
- For actions should I wire them up as if the come in a response and do the same thing as I did for tasks or should we wait until we approve the tasks model and then use a similar approach for actions?
- For actions, should I wire them up as if they come in a response (similar to tasks) ? Or should we wait until we approve the tasks model and then use a similar approach for actions?
- Did I handle anything in the Vue side that I wasn't supposed to? (specifically error handling)
- Is there a better way to prompt the AI to provide always tasks? Sometimes it forgets, and I have to explicitly include "include tasks" in the user prompt.
- Regarding error handling, the user now receives an alert when there is an error while evaluating their programs. (parsing errors. The errors that reach the screen are sent as a prompt and we are getting the fixed code snippets back).
Paul suggested using `Exception.nestedMetadata` along with `Exception.getMessages` in the `parseAndSerializeProgram` function to gather more information about the errors. However, I haven't pushed those changes yet because I'm still uncertain if I am doing it correctly. I need to experiment with it further. Here is a comparison of the errors before and after implementing the metadata changes.

before:
Screenshot 2023-05-30 at 22 45 48

after:
Screenshot 2023-05-30 at 22 47 44

StachuDotNet:
> * Should I fix the UI in this PR? (This includes setting the height to 100% and making the overflow scrollable within the columns ,improve the styling of the columns, etc.)
No, I think additional UI improvements can wait for another PR.

> * For actions, should I wire them up as if they come in a response (similar to tasks) ? Or should we wait until we approve the tasks model and then use a similar approach for actions?

I'm not sure - I'd say let's keep working with the task model and then see what feels right.

> * Did I handle anything in the Vue side that I wasn't supposed to? (specifically error handling)

Looking now, I do see some error-handling logic in Vue code that _could_ be in Dark code. Specifically, the `watch` in CodeSnippet - I suppose ideally, that logic should happen before the snippet is returned. Specifically, the `eval` should probably be one of those cases - success or failure - the Dark code can put it into one of those cases as needed, and the UI can render appropriately.

> * Is there a better way to prompt the AI to provide always tasks? Sometimes it forgets, and I have to explicitly include "include tasks" in the user prompt.

Not sure - we'll definitely have to keep playing with the prompts involved.

> * Regarding error handling, the user now receives an alert when there is an error while evaluating their programs. (parsing errors. The errors that reach the screen are sent as a prompt and we are getting the fixed code snippets back).
>   Paul suggested using `Exception.nestedMetadata` along with `Exception.getMessages` in the `parseAndSerializeProgram` function to gather more information about the errors. However, I haven't pushed those changes yet because I'm still uncertain if I am doing it correctly. I need to experiment with it further. Here is a comparison of the errors before and after implementing the metadata changes.
> 
> before: Screenshot 2023-05-30 at 22 45 48
> 
> after: Screenshot 2023-05-30 at 22 47 44

I think this good additional information, but should probably concatonate both getMessages _and_ nestedMetadata

backend/src/StdLibExecution/Libs/String.fs
```diff
@@ -972,6 +972,28 @@ let fns : List<BuiltInFn> =
...
+      typeParams = []
+      parameters =
+        [ Param.make "str" TString "The string to search in"
+          Param.make
+            "searchFor"
+            TString
+            "The string to search for within <param str>" ]
+      returnType = TInt
+      description =
+        "Returns the index of the first occurrence of <param searchFor> in <param str>, or -1 if <param searchFor> does not occur."
```
StachuDotNet, line 988: (can be done in another PR) returning an Option<Int> might make more sense for this fn

backend/testfiles/execution/string.tests
```diff
@@ -577,6 +577,12 @@ String.padEnd_v0 "🧟‍♀️🧟‍♂️" "0" 5 = Ok "🧟‍♀️🧟‍
 String.padEnd_v0 "👨‍❤️‍💋‍👨👩‍👩‍👧‍👦🏳️‍⚧️🇵🇷" "0" 10 = Ok "👨‍❤️‍💋‍👨👩‍👩‍👧‍👦🏳️‍⚧️🇵🇷000000"
 String.padEnd_v0 "鷝" "觌഻" 0 = Ok "鷝"
 
+String.indexOf_v0 "hello world" "world" = 6
+String.indexOf_v0 "hello world" "earth" = -1
+String.indexOf_v0 "" "" = 0
+String.indexOf_v0 "hello" "" = 0
+String.indexOf_v0 "" "hello" = -1
```
StachuDotNet, line 584: it might be worth adding some more complicated tests involving strings of more complicated Unicode characters - for example, borrowing from String.split tests, `String.indexOf "👨‍👩‍👧‍👦" "👩" = None`.

OK to leave this as a CLEANUP item

StachuDotNet: (worth checking out String.split tests generally, for inspiration)

StachuDotNet (changes requested):
Looks great! Just a few minor requests noted.

PR #4913, opened 2023-06-02
Handling errors 

OceanOak:
Changelog:

```
canvases/dark-editor
- Start handling errors.
```

This PR includes:
- Updating the parseAndSerializeProgram fn to use Result type for better error handling.
- Starting to handle errors ( send syntax errors and parsing errors as prompts and get the updated code snippet).
- Minor styling changes.
- Fixing the loading spinner.
- Displaying the user prompt immediately after sending it.

Current state:

https://github.com/darklang/dark/assets/87861924/29fa190f-9657-4df5-aa70-a8e304a0d9fd

### DIFF
--- a/backend/src/Wasm/Libs/Editor.fs
+++ b/backend/src/Wasm/Libs/Editor.fs
@@ -138,11 +138,14 @@ let fns : List<BuiltInFn> =
                 let inputVars = Map.empty
                 LibExecution.Execution.executeExpr state inputVars expr
 
-              return
-                LibExecution.DvalReprDeveloper.toRepr result
-                |> DString
-                |> Ok
-                |> DResult
+              match result with
+              | DError (_source, err) -> return DResult(Error(DString err))
+              | result ->
+                return
+                  LibExecution.DvalReprDeveloper.toRepr result
+                  |> DString
+                  |> Ok
+                  |> DResult
             with
             | e ->
               let error = Exception.getMessages e |> String.concat " "
--- a/backend/src/Wasm/Libs/Editor.fs
+++ b/backend/src/Wasm/Libs/Editor.fs
@@ -40,8 +40,7 @@ let fns : List<BuiltInFn> =
               let state = editor.CurrentState
               // TODO: assert that the type matches the given typeParam
               return DResult(Ok state)
-            with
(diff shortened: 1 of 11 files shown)


canvases/dark-editor/system-prompt.txt
```diff
@@ -126,13 +126,14 @@ Below all types and function defined, include at least one expression to be eval
...
 All code snippets should be wrapped in ``` tags.
 
-Here's an example of a prompt/response combination:
+Here's an example of a response to the following prompt:
 
 User: build a small program that models a list of People,
 a function that returns the age of a Person, then hardcode a list of people
 and return the age of each person
 
-You:
```
OceanOak, line 135: I made some changes to this file because I was receiving a response that looks like the screenshot below.

Screenshot 2023-06-05 at 21 37 07

backend/src/Wasm/Libs/Editor.fs
```diff
@@ -143,11 +143,14 @@ let fns : List<BuiltInFn> =
...
-                |> DResult
+              match result with
+              | DError (_source, err) -> return DResult(Error(DString err))
+              | result ->
+                return
+                  LibExecution.DvalReprDeveloper.toRepr result
+                  |> DString
+                  |> Ok
+                  |> DResult
             with
```
OceanOak, line 154: Is there a need for a try/catch block in this function? I don't believe there is a scenario where we would encounter an exception in this specific context. Or am I overlooking something?

StachuDotNet: I think you're right! Can probably be removed.

StachuDotNet:
Just reviewed, happy to merge this once the formatting thing is fixed (might as well remove the try/with you mentinoed, then)

StachuDotNet:
This turned out great!

I think we'll later provide a better way to update UI state earlier (rather than call that fn manually), but this works great for now.

backend/src/Wasm/Libs/Editor.fs
OceanOak: I'll remove it in the next PR (you merged before I got the chance to update it 🙂)

PR #4917, opened 2023-06-04
Record update

ditansu:
Support record update F# like syntax. 
```fsharp
  type RecordForUpdate = { x : Int; y : Int}
  let myRec = RecordForUpdate {x = 4; y = 1} 
  let updatedMyRec = {myRec with y = 2}) // equal RecordForUpdate {x = 4 ; y = 2}
```


backend/src/LibExecution/Interpreter.fs
```diff
@@ -240,6 +241,42 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
+        return!
+          Ply.List.foldSequentially
+            (fun r (k, v) ->
+              uply {
+                let! v = eval state st v
+                match r with
+                | DRecord(_, m) when not (Map.containsKey k m) ->
+                  return err id $"Unexpected field `{k}` in {typeStr}"
+                | DRecord(typeName, m) ->
+                  match recordMaybe typeName with
```
pbiggar, line 262: We should extract the map of types at the start, rather than once per loop

backend/src/LibExecution/Interpreter.fs
```diff
@@ -240,6 +241,42 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
+                      |> List.map (fun f -> f.name, f.typ)
+                      |> Map
+                      |> Map.find k
+                    match TypeChecker.unify [ k ] types field v with
+                    | Ok() -> return DRecord(typeName, Map.add k v m)
+                    | Error e -> return err id (TypeChecker.Error.toString e)
+                  | None ->
+                    return
+                      err id $"Expected a record but {typeStr} is something else"
+                | _ -> return err id "Expected a record"
```
pbiggar, line 275: If a previous step returns a DError, this will replace it with a new "Expected a record" DError. We should instead handle DError and DIncomplete separately to ensure they are passed through.

pbiggar (changes requested):
This looks great! I made a few small suggestions for performance and better error messages, and then we should be good to go.

backend/src/LibExecution/Interpreter.fs
ditansu: @pbiggar I'm not sure that I understand how to "handle `DError` and `DIncomplete` separately" could you please clarify what it means?

pbiggar: Similar to this: https://github.com/darklang/dark/blob/ebaaedf23f63b980cc22fb9ee604f19730869ef2/backend/src/LibExecution/RuntimeTypes.fs#L663

pbiggar: Basically, if we've already created a DError as part of some error handling, we should continue to propagate it through rahter than creating a new error with a different message on line 275.

ditansu: > Basically, if we've already created a DError as part of some error handling, we should continue to propagate it through rahter than creating a new error with a different message on line 275.

is it correct now?

ditansu:
I have no idea why after merge main branch I've gotten seven errors like below:

image
 
any suggestions?

StachuDotNet:
> I have no idea why after merge main branch I've gotten seven errors like below:
> 
> image
> any suggestions?

This was my fault and you can ignore.
(our hacky parser seems to interpret `[1, 2]` as `[1; 2]` rather than the expected `[(1, 2)]`. I "fixed" a few Http fns to include the parens to "help" parsing, but forgot to fix some corresponding tests before merging. Will fix separately.

backend/src/LibExecution/Interpreter.fs
```diff
@@ -240,6 +241,46 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
+          return!
+            Ply.List.foldSequentially
+              (fun r (k, expr) ->
+                uply {
+                  let! v = eval state st expr
+                  match r, k, v with
+                  | r, _, _ when Dval.isFake r -> return r
+                  | _, _, v when Dval.isFake v -> return v
+                  | _, "", _ -> return err id $"Empty key for value `{v}`"
+                  | DRecord(_, m), _, _ when not (Map.containsKey k m) ->
```
pbiggar, line 270: I think this should check against `expectedFields`?

backend/src/LibExecution/Interpreter.fs
```diff
@@ -240,6 +241,46 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
+          let expectedFields =
+            (expected1 :: expectedRest) |> List.map (fun f -> f.name, f.typ) |> Map
+          return!
+            Ply.List.foldSequentially
+              (fun r (k, expr) ->
+                uply {
+                  let! v = eval state st expr
+                  match r, k, v with
+                  | r, _, _ when Dval.isFake r -> return r
+                  | _, _, v when Dval.isFake v -> return v
```
pbiggar, line 268: Yeah, this is great.

pbiggar (changes requested):
Looks great, thanks! 1 question about it and we should be good to merge.

backend/src/LibExecution/Interpreter.fs
ditansu: Definitely, `expectedFields` and `m` consist of the same values. By using the `m` I wanted to highlight the fact that we are looking for a field among **existing** fields. Plus, it is easier to find with the eyes `m` than `expectedFields`, so the cognitive load becomes lower :) Btw as an option we could rename `expectedFields` to `existedFields` and that would resolve our issue. And sure I can replace `m` on `expectedFields`, choose ;)

pbiggar: I don't understand why `expectedFields` is the same as `m`? Isn't `expectedFields` made up of the type's fields, while `m` comes from the `DRecord`'s existing fields?

I'm thinking maybe you mean that we expect `m` and `expectedFields` to have the same fields to them, which I think we expect but is not guaranteed.

ditansu: Yes, you are right. I meant that both maps have the same keys, not values. They have to have common set of keys because they based on one common source i.e.`baseRecord`. And as we are checking keys (which we are updating), IMO it doesn't matter what a map we are choosing for that. Ok, or am I wrong?

pbiggar: I'm thinking that it shouldn't matter in theory. However, it is possible to have bugs in the creation of `m`, but much less likely to have bugs in `expectedFields`. If there is a bug with missing fields, the error message will be weird and hard to debug if we use `m` - better to go with the source of truth, which is `expectedFields`.

ditansu: Indeed, agree. I've replaced `m` on `expectedFields`

PR #4926, opened 2023-06-12
Extend pretty-printer package to pretty-print packages

StachuDotNet:
Changelog:

```
packages/pretty-printer
- (incomplete) extend to pretty-print packages
```


packages/darklang/stdlib/canvas.dark
```diff
@@ -0,0 +1,7 @@
+module Darklang =
```
pbiggar, line 1: I don't think this would go in the stdlib package.

pbiggar: Maybe "language tools" or similar as a package name?

scripts/build/compile
```diff
@@ -353,6 +353,9 @@ def mark(should, f):
   elif ("/backend/" in f) and ("/tests/" in f):
     should.backend_test = True
 
+  elif ("/packages/" in f) and (f.endswith(".dark")):
+    should.backend_quick_build = True
```
pbiggar, line 357: What's the goal here?

StachuDotNet: parse and re-load packages upon save of a package

might be worth
- extracting "load packages" from the run-backend-server script
- add a `should.reload_packages`
- when doing any backend build, set that^ to `true`
?

pbiggar: Yeah, thats a better (and faster) approach

PR #4928, opened 2023-06-13
Deleting Record.set and replacing usage to new record update syntax

ditansu:
1. Delete `Record.set` method  
2. Replase usage from Records.set to {.. with } e.g.: 
```fsharp
// was
Record.set_v0 response "headers" respHeaders
// became
{response with headers = respHeaders}
```


pbiggar:
Amazing, thank you!

PR #4944, opened 2023-06-20
Remove Ops

pbiggar:


pbiggar:
Only CI failure is the httpbin one, so merging.

PR #4949, opened 2023-06-28
Extract LibCLI for CLI and InternalCLI; start export/import script for packages+canvases

StachuDotNet:
Changelog:

```
Area of change
- details
```

The goal here was to write an export/import script for packages canvases.
I couldn't run it with just 'CLI' because it didn't have access to internal fns required for this specific script.
LocalExec also didn't seem appropriate, as it wasn't set up to run _other_ scripts.
Also, there seemed to be a lot of overlap between the two projects,
and "CLI" has gotten beyond an 'experimental' status.
So, I ended up extracting out a `LibCLI`, renamed `LocalExec` to `InternalCLI`, and did some other work.

At the end of this PR, the script is incomplete - it was blocked by `PrettyPrinter` changes (resolved now), and the fact that the CLI doesn't have access to packages (resolve-able but worth discussion).

This PR:
- moves `cli` from the `experimental` dir to the `src` dir
- renames `LocalExec` to `InternalCLI`, and `run-local-exec` to `run-cli-internal`
- extract `LibCLI` out of `LocalExec` and `CLI` to allow re-use
    - exposes a `DarkCLIConfig` type to be passed in for the 2 usages of the library (`CLI` and `InternalCLI`), so you can:
        - provide `extraStdlibForUserPrograms` (for Internal use)
        - `allowInternalDarkFunctions` (again, for Internal use)
- given that^ change, extracts the package-loading functionality that was in `LocalExec` from its old `main.dark` to a new `./scripts/import/packages.dark` script
    - updates `./scripts/reload-packages` to call this (it runs `./scripts/run-cli-internal ./scripts/import-packages.dark load-packages`)
- removes `StdLibExperimental` and folds it into `BwdDangerServer`
- removes several defunct/unused StdLib fns (e.g. `Experiments.parseAndExecuteExpr`)
- Adds a mutable `additionalStdlib` to `LibRealExecution.RealExecution`, so that `BwdDangerServer` may append its `Experimental.` Stdlib module
- tidies `CanvasHack` a bit
- starts to implement `./scripts/deployment/sync-packages-and-canvases.dark`

(side note: with this PR, the `experimental` directory now only contains `CanvasHack` (which _could_ be replaced with an InternalCLI script?) and `BwdDangerServer`.)

Some questions and follow-up thoughts:
- do you hate the `additionalStdlib` in `LibRealExecution.RealExecution`? Is there a reasonable way to avoid it without having a clone of `LibRealExecution`?
- I feel weird about `StdLibCLIHost` accepting arguments for its `contents`. It's unusual, but I couldn't think of another reasonable way to get it done. What do you think?
- do you think we should rename `LibRealExecution` to something like `LibCloudExecution`?
    - maybe `LibCLI` should be renamed to `LibCLIExecution`, if so
    - or, name them both to `LibExecutionCLI` and `LibExecutionCloud`
    - _or_ drop the `Lib` prefix from both and just have `ExecutionCLI` and `ExecutionCloud`
- I stopped writing the `sync..` script because Packages aren't available in the CLI. What should be done here?
- I attempted to add `alias`es of `dark=./scripts/run-cli` and `dark-internal=./scripts/run-cli-internal`, both in the Dockerfile and in the `_vscode-post-start-command` but couldn't quite get this working - worth following up on. Do you have an idea of how this should work?
- I think if a Stdlib is only used in one executable (i.e. `Experiments` available in `BwdDangerServer`), it's OK for the definition of that Stdlib to be in that .fsproj. Do you agree? If so, should we extend this idea by merging `StdLibCLIHost` into `LibCLI`?

Follow-ups:
- [ ] avail packages to the CLI script
- [ ] finish the script (add security, test it, etc)
- [ ] hook the script into our CI pipeline, so it's actually used


pbiggar:
> The goal here was to write an export/import script for packages canvases. I couldn't run it with just 'CLI' because it didn't have access to internal fns required for this specific script. LocalExec also didn't seem appropriate, as it wasn't set up to run _other_ scripts. Also, there seemed to be a lot of overlap between the two projects, and "CLI" has gotten beyond an 'experimental' status. So, I ended up extracting out a `LibCLI`, renamed `LocalExec` to `InternalCLI`, and did some other work.

I don't think this is a good idea. CLI is a product and LocalExec is an internal tool. I agree it's useful to give them access to the same libraries (BuiltInCLI or whatever) but I don't think they should share top-level code, even if it means code gets duplicated. It's better to have a simple straightforward understanding of what the product (the CLI) code does than to reuse code with an internal tool.

I think the extra level of abstraction makes it harder to understand how the "product" code works (which already wasn't the most trivial).

I'm happy to have components pulled out to be reusable when they're composable, but not the top level code. As an example of what I mean here, there are good reasons to have different error handling, notification, trace, and basically all aspects of how we "run" code, so it's better for the calling code to be simple and straightforward instead of abstracted. If we are going to abstract things here, doing so as a "library" vs a "framework" feels more appropriate.

 
> This PR:
> 
> * moves `cli` from the `experimental` dir to the `src` dir

great.

> * renames `LocalExec` to `InternalCLI`, and `run-local-exec` to `run-cli-internal`

I prefer we stick to LocalExec (whose name is analogous to ProdExec -- this one runs locally vs the other one running in production).

I think "internal" is a word we should avoid as it's unclear what its meaning is. LocalExec isn't really an internal cli, it's a tool for running things in the repo.


> * extract `LibCLI` out of `LocalExec` and `CLI` to allow re-use
>   * exposes a `DarkCLIConfig` type to be passed in for the 2 usages of the library (`CLI` and `InternalCLI`), so you can:
>     
>     * provide `extraStdlibForUserPrograms` (for Internal use)
>     * `allowInternalDarkFunctions` (again, for Internal use)

Discussed above - I would prefer to revert this change.


> * given that^ change, extracts the package-loading functionality that was in `LocalExec` from its old `main.dark` to a new `./scripts/import/packages.dark` script
>   
>   * updates `./scripts/reload-packages` to call this (it runs `./scripts/run-cli-internal ./scripts/import-packages.dark load-packages`)

It makes sense to me for this to live in LocalExec.

> * removes `StdLibExperimental` and folds it into `BwdDangerServer`

Nice. Can we remove BwdDangerServer yet? It's going away soon I expect?



> Some questions and follow-up thoughts:
> 
> * do you hate the `additionalStdlib` in `LibRealExecution.RealExecution`? Is there a reasonable way to avoid it without having a clone of `LibRealExecution`?

I really would rather not do this. Especially since DangerServer is going away soon, we don't need to create new abstractions for it to be nicer.

That is to say a clone of LibRealExecution is fine because it's going away. If you want to delete DangerServer now that's fine too, but let's not create extra abstractions to make temporary code nicer.


> * I feel weird about `StdLibCLIHost` accepting arguments for its `contents`. It's unusual, but I couldn't think of another reasonable way to get it done. What do you think?

I don't think we should have a StdLibCLIHost, as mentioned above.


> * do you think we should rename `LibRealExecution` to something like `LibCloudExecution`?

Yes, great rename. Wondering how LibBackend fits into this?


>   * maybe `LibCLI` should be renamed to `LibCLIExecution`, if so
>   * or, name them both to `LibExecutionCLI` and `LibExecutionCloud`
>   * _or_ drop the `Lib` prefix from both and just have `ExecutionCLI` and `ExecutionCloud`

I don't see what the goal is here.


> * I stopped writing the `sync..` script because Packages aren't available in the CLI. What should be done here?

The script is fine to merge in its current state and we can keep going later.


> * I attempted to add `alias`es of `dark=./scripts/run-cli` and `dark-internal=./scripts/run-cli-internal`, both in the Dockerfile and in the `_vscode-post-start-command` but couldn't quite get this working - worth following up on. Do you have an idea of how this should work?

Aliases are part of the shell, so they'd have to be written in ~/.bashrc I think. Not a huge fan of doing though. 


> * I think if a Stdlib is only used in one executable (i.e. `Experiments` available in `BwdDangerServer`), it's OK for the definition of that Stdlib to be in that .fsproj. Do you agree? 

We keep finding that StdLibs get used in other places so I would suggest we say either way is fine while things are in flux. If we've a reason to think it will never be used elsewhere, then yeah I'd default to within the executable.

StachuDotNet:
Some of our internal .dark scripts will need access to some "Internal" functions. What should run those?
- LocalExec with `local-exec ./path-to-script`?
  - with a _copy_ of the `parseAndExecuteScript` that has access to additional functions?
- the "normal" CLI, somehow?
- something else?

---

Can we remove BwdDangerServer yet? It's going away soon I expect?
I don't think so:
- `Experiments.readFromStaticDir` is still used to support dark-serve-static
- `dark-repl` and `dark-editor` are still powered largely by `parseAndSerializeProgram` and `readFromCanvases`

Maybe I'm missing a way we can handle these?

---
> > maybe LibCLI should be renamed to LibCLIExecution, if so
> > or, name them both to LibExecutionCLI and LibExecutionCloud
> > or drop the Lib prefix from both and just have ExecutionCLI and ExecutionCloud
> I don't see what the goal is here.

The goal was to have a standard 'execution' per runtime - LibCloudExecution and LibCLIExecution - but it sounds like that's not the direction we're heading.

---

> Aliases are part of the shell, so they'd have to be written in ~/.bashrc I think. Not a huge fan of doing though.

I was trying to add these to avoid havint o write out `./scripts/run-local-exec ./scripts/path-to-script` all the time, making it a bit tidier. Is there a better way to do this, or is that a bad goal for some reason?

StachuDotNet:
I've extracted the useful bits out of this, in other smaller PRs

PR #4958, opened 2023-06-30
Start adding an OpenAI package

OceanOak:
Changelog:

```
Packages
- Add an OpenAI package
```


OceanOak:
Would it be beneficial to include functions that support parameters beyond "prompt", such as model, tokens, temperature, etc.?

StachuDotNet:
> Would it be beneficial to include functions that support parameters beyond "prompt", such as model, tokens, temperature, etc.?
I think it'd be useful to have an `options` object passed in, and have a `default` constant/function available.
So,
- the CompletionOptions object would house all possible options, 
- a `default` value would contain the ..default values
- and a consumer could either create a whole CompletionOptions value from scratch, or use the `{ default with ...}` syntax to override specific properties

Can wait for a separate PR, though

StachuDotNet:
I would rename agent.dark to openai.dark - an agent logic/package wrapping this is probably worth extraction, applicable to not only this LLM but also others

StachuDotNet:
Great!

PR #4961, opened 2023-07-03
Pretty printer fixes

OceanOak:
No changelog

This PR addresses a couple pretty printer issues

### Line breaks and spaces
Issue:
```
  let indent (s: String) : String =
        s
        |> String.split "
          
          "
       |> List.map ((fun line -> 
               match line with 
              |  ->
                ""
              | line ->
                ("  ") ++ (line))) 
            |> String.join ("
            
            ")
```

Fixed:
```
 let indent (s: String) : String =
      s 
      |> String.split ("\n")
      |> List.map ((fun line -> 
         match line with 
        | "" ->
          ""
        | line ->
          ("  ") ++ (line))) 
      |> String.join ("\n")
```

### Floats
Issue:
```
{ model = "gpt-3.5-turbo"; max_tokens = 700; temperature = 0.6999999999999999555910790149937383830547332763671875 
````

Fixed:
```
{ model = "gpt-3.5-turbo"; max_tokens = 700; temperature = 0.7}
```

I've also tested these, and they are being displayed correctly
let f = 1.0
let g = 0.799999
let s = 0.7
let d = 2.2
let y = 20.33

### CODE REVIEW

File: packages/darklang/prettyPrinter/programTypes.dark [hunk 1]
```diff
@@ -458,21 +455,31 @@ module Darklang =
...
-          $"{signPart}{wholePart}{remainderPart}"
 
         | EChar(_id, c) -> "\"" ++ c ++ "\""
 
         | EString(_id, segments) ->
           match segments with
           | [] -> "\"\""
-          | [ StringText text ] -> $"\"{text}\""
+          | [ StringText text ] ->
+            let escapedText = String.replaceAll text "\n" "\\n"
```
OceanOak, line 481: I used replaceAll in this case since I couldn't use String.map  (wasn't allowed to replace a char with a string). 
Do you have any suggestions for a more efficient approach that would allow checking for things other than line breaks, such as spaces?

pbiggar: What about something like:

`String.split ... |> List.map (replace ...) |> String.concat`

packages/darklang/prettyPrinter/programTypes.dark
```diff
@@ -458,21 +455,31 @@ module Darklang =
...
+              match rem with
+              | Ok r ->
+                  if r == 0.0 then ""
+                  else (r  |> Float.toString) |> String.dropFirst 2
+              | Error e -> e
+          let wholePart =
+              match whole with
+              | "0" -> if remainderPart != "" then "0" else ""
+              | _ -> whole
+          let formattedFloat =
```
pbiggar, line 469: I'm finding it hard to understand what's happening here. Could you add comments (with examples maybe) to discuss the edge cases you're handling?

Also can you add tests? I think you should be able to add tests within backend/testfiles/execution.

packages/darklang/prettyPrinter/programTypes.dark
```diff
@@ -300,12 +300,9 @@ module Darklang =
         | MPBool(_id, b) -> Bool.toString b
         | MPInt(_id, i) -> Int.toString i
         | MPFloat(_id, sign, whole, remainder) ->
```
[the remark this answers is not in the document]
pbiggar: Yes I think those would be the same. Maybe extract the function to do both.

packages/darklang/prettyPrinter/programTypes.dark
pbiggar: That said, `String.replaceAll` might be fine. You could do something like:

```
text
|> String.replaceAll "\n" "\\n"
|> String.replaceAll "\t" "\\t"
|> String.replaceAll "\r" "\\r"
```

pbiggar (changes requested):

packages/darklang/prettyPrinter/programTypes.dark
OceanOak: We don't need this anymore. it was used when the number lies between 0 and 1, to keep the "0" in whole part.

packages/darklang/prettyPrinter/common.dark
```diff
@@ -16,4 +16,21 @@ module Darklang =
     let sign (s: PACKAGE.Darklang.Stdlib.Sign) : String =
       match s with
       | Positive -> ""
-      | Negative -> "-"
\ No newline at end of file
+      | Negative -> "-"
+
+    let escapeSpecialCharacters (s: String) : String =
```
OceanOak, line 21: Is this file the right place for these functions?

packages/darklang/prettyPrinter/common.dark
```diff
@@ -16,4 +16,21 @@ module Darklang =
...
+    let processRemainder (remainder: String) : String =
+      let rem = $".{remainder}" |> Float.parse
+
+      match rem with
+      | Ok r ->
+        if r == 0.0 then
+          "0"
+        else
+          (r |> Float.toString) |> String.dropFirst 2
+      | Error e -> e
```
OceanOak, line 36: Is it bad to return the error here?

packages/darklang/prettyPrinter/common.dark
pbiggar: Sure.

packages/darklang/prettyPrinter/common.dark
pbiggar: what would you expect to happen if you parse an invalid float?

OceanOak: I expect an error or an exception to be thrown indicating that parsing an invalid float isn't possible
The issue is if I have something like 
let z = 0.abc
let a = 3.14abc
in the pretty printer is is displayed as 
let z = 0
let a = 0

pbiggar: The difference here, if I understand it correctly, is that this is pretty printing EFloats, where it should not be possible to have invalid floats present.

OceanOak: Currently, when we send an invalid float like `1.b`, it doesn't reach the `EFloat` branch; instead, it is considered as an `EInt`, and as a result, returns 0. I don't believe this issue is related to the pretty printer. I've added some tests in 'Int.test' for this scenario (i.e., 'Int.toString 1.ab = "0"' passes)

PR #4963, opened 2023-07-03
Split RT.ExecutionState.libraries into .builtIns and .packageManager

StachuDotNet:
Changelog:

```
Area of change
- details
```
This is towards package-streaming.

Package-fetching should almost certainly be some kind of async.
In making the `packageManager.getFn` and `.getType` functions async, many functions _above_ it end up needing to be made async accordingly. I don't see a way to avoid this but maybe @pbiggar you have an idea?

This PR is nearly OK, but there is some commented-out code that still needs to be `task`'d bit by bit.

Once things are working, I'll tidy the commit history certainly before merging.

### DIFF
--- a/backend/experiments/BwdDangerServer/DangerExecution.fs
+++ b/backend/experiments/BwdDangerServer/DangerExecution.fs
@@ -18,51 +18,20 @@ module Interpreter = LibExecution.Interpreter
 
 open LibBackend
 
-let (builtInFns, builtInTypes) =
-  LibExecution.StdLib.combine
-    [ StdLibExecution.StdLib.contents
-      StdLibCloudExecution.StdLib.contents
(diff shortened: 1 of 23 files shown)


pbiggar:
> Package-fetching should almost certainly be some kind of async.
In making the packageManager.getFn and .getType functions async, many functions above it end up needing to be made async accordingly. I don't see a way to avoid this but maybe @pbiggar you have an idea?

Yeah, this is looking like a lot of "unnecessary" task code.

Another option to consider is doing a recursive code search statically at the start. But that means we need to fetch all the code at the start, recursively. Which might not be a bad idea but also will load libraries that are not needed dynamically.

I expect we'll end up with tasks in either case, so I'm thinking this is the right approach.

StachuDotNet:
superceded by #4973

PR #4964, opened 2023-07-03
Packages canvas

OceanOak:
Changelog:

```
Area of change
- allow multiple canvases to access DarkInternal
- add dark-packages-canvas, and save functions in an in-canvas database
- add a NonEmptyListConverter
```


canvases/dark-packages/main.dark
```diff
@@ -0,0 +1,35 @@
+type PackageFns = PACKAGE.Darklang.Stdlib.ProgramTypes.PackageFn.T
+
+[<DB>]
+type PackageFnsDB = PackageFns
```
OceanOak, line 4: Using
```
[<DB>]
type PackageFnsDB = PACKAGE.Darklang.Stdlib.ProgramTypes.PackageFn.T
```
instead of 
```
type PackageFns = PACKAGE.Darklang.Stdlib.ProgramTypes.PackageFn.T

[<DB>]
type PackageFnsDB = PackageFns
```
didn't work

canvases/dark-packages/main.dark
```diff
@@ -0,0 +1,35 @@
...
+        deprecated = fn.deprecated }
+
+  let key = DB.generateKey_v0 () //use id instead
+  DB.set fnToSave key PackageFnsDB
+
+  PACKAGE.Darklang.Stdlib.Http.response (String.toBytes "added") 200
+
+[<HttpHandler("GET", "/all-functions")>]
+let _handler _req =
+  let count = (DB.count PackageFnsDB) |> Int.toString
```
OceanOak, line 33: I tried the following to see what we are saving, but I encountered an `unknown error`. The `DB.count` function returns 107, indicating that we are saving something. However, I'm unsure if we are saving the right things. (The same applies to DB.getAllWithKeys)

```
[<HttpHandler("GET", "/functions")>]
let _handler _req =
   let functions = (DB.getAll PackageFnsDB) |> String.fromList 

  PACKAGE.Darklang.Stdlib.Http.response (String.toBytes functions) 200
  ```

File: backend/tests/Tests/DvalRepr.Tests.fs [hunk 3]
```diff
@@ -101,7 +101,7 @@ module ToHashableRepr =
 
     testList
       "hashv2"
-      [ t [ DBytes [||] ] "diEjLGQC8oE"; t [ DBytes [| 128uy |] ] "kQHs0urT3N4" ]
+      [ t [ DBytes [||] ] "Bu4AH9NVqA0"; t [ DBytes [| 128uy |] ] "ARIiVvuJZTo" ]
```
OceanOak, line 104: What could cause these hashes to change and are we supposed to change them?
The only change I could point it to is the one in Json.Vanilla.
Could you also please explain the error message `bad fsharp impl`?

canvases/dark-packages/main.dark
pbiggar: Could you add a commented failing test case for this?

scripts/build/load-packages-into-dark-canvas
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+. ./scripts/devcontainer/_assert-in-container "$0" "$@"
```
pbiggar, line 2: As a follow-up, would be great to get this shared code split into a `./scripts/run-local-exec` script.

canvases/dark-packages/main.dark
pbiggar: It would be good to investigate/resolve this as a followup.

canvases/dark-packages/main.dark
pbiggar: possibly as a followup.

backend/tests/Tests/DvalRepr.Tests.fs
pbiggar: `bad fsharp impl` I think was from back when we were moving over from OCaml. This would have implied that the output was not what we expected from OCaml, and so the F# implementation was incorrect somehow.

pbiggar: Yes, this change is from Json.Vanilla. Changing the format is very dangerous, which is why the test exists. Fortunately, changing this format right now is totally fine. In the future we could use a separate serializer here if we wanted to avoid this issue.

pbiggar: So I think this is fine and changing the expected answer is appropriate.

pbiggar (approved):

pbiggar:
Looks great and seems good to merge. It seems the tests are failing though.

PR #4970, opened 2023-07-11
Split ExecutionState.libraries into .builtIns and .packageManager

StachuDotNet:
Changelog:

```
Interpreter
- split RT.ExecutionState.libraries into .builtIns and .packageManager
```

towards package streaming


pbiggar:
Glad we're doing small PRs like this, I'm trying to as well.

PR #4971, opened 2023-07-12
Package-fetching redux

StachuDotNet:
Changelog:

```
Area of change
- details
```

Packages will shortly be stored/managed externally, in the `dark-packages` canvas.
We need to update our backend to fetch them as-needed.
The goal here is to fetch packages asynchronously, once, at the start of executing a `tlid`.

WIP. We still need to figure out:
- [ ] how to determine what types and functions need to be fetched
  (I think we'll need a `dependenciesFor` function, or something?)
  (so far this feels like it may take multiple http calls if managed by the consumer - so, maybe we migrate the dependency-collection logic to `dark-packages`?)
- [ ] (other things I forget right now - will follow up)

### DIFF
--- a/backend/experiments/BwdDangerServer/DangerExecution.fs
+++ b/backend/experiments/BwdDangerServer/DangerExecution.fs
@@ -30,39 +30,6 @@ let builtIns : RT.BuiltIns =
   { types = types |> Map.fromListBy (fun typ -> typ.name)
     fns = fns |> Map.fromListBy (fun fn -> fn.name) }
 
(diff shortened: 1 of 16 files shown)


pbiggar:
I think the idea in this PR is to do something super quick where we just download all functions/types, and then go forward from there. It feels like the direction that takes us is very different from the direction needed to only fetch specific functions/types as needed.

StachuDotNet:
Decided to not go this route, and rather taskify

PR #4972, opened 2023-07-13
Add some db tests

OceanOak:
No changelog


OceanOak:
Are these tests enough? and should we consider adding more tests somewhere else (like in the json.tests file)?

PR #4974, opened 2023-07-14
Name Resolvers

pbiggar:
Done:
- name resolution for most exprs as part of converting WT to PT
- error handling
- README

TODO:
- remainder of name resolution in exprs
  - variable names in pipes
- name resolution in type definitions and typerefs in exprs
- aliasing


StachuDotNet:
I think this is strictly a subset of #4979 so closing

PR #4977, opened 2023-07-24
Roundtrip (Internal <-> Dark PT) types

OceanOak:
Changelog:

```
Area of change
- Add mappings from the Dark types to internal ProgramTypes
- Add serialization tests around the PT -> DarkTypes -> PT roundtripping
```


backend/tests/Tests/Serialization.DarkTypes.Tests.fs
```diff
@@ -0,0 +1,285 @@
...
+    =
+    testTask testName {
+      let firstDT = original |> toDT
+
+      let context =
+        LibExecution.TypeChecker.Context.FunctionCallResult(
+          fnName =
+            (RT.FnName.fqPackage
+              "Darklang"
+              (NonEmptyList.ofList [ "LanguageTools"; "ProgramTypes" ])
```
OceanOak, line 47: We can replace "Darklang", "LanguageTools", "ProgramTypes", and "expr" with anything and nothing breaks. If I remember correctly, context only becomes relevant when tests fail, but I am not sure what the right thing to do here

StachuDotNet (changes requested):
This looks great. I noted a strong suspicion of why the type-checking "works" even though `empty` types are being passed in.

Maybe let's:
- move the `types` var I introduced nearer to the top of the file (above the actual tests)
- reference it in both places
- make a note in Notion to follow up and use real types soon

StachuDotNet:
Actually, this `types` might work:
```fsharp
let types: RT.Types =
    { builtIn = builtIns.types
      package = LibBackend.PackageManager.packageManager.getType
      userProgram = Map.empty }
```

backend/src/StdLibDarkInternal/Helpers/ProgramTypesToDarkTypes.fs
```diff
@@ -29,6 +40,14 @@ module Sign =
     | Positive -> DEnum(stdlibTyp [] "Sign" 0, "Positive", [])
     | Negative -> DEnum(stdlibTyp [] "Sign" 0, "Negative", [])
```
StachuDotNet, line 41: I think common.dark should move to the new LanguageTools namespace as well - I think the need for `stdlibTyp` will disappear then

PR #4979, opened 2023-07-26
Aliases step 3

pbiggar:
As part of moving aliasing into WT2PT, this moves name resolution earlier. Instead of erroring during parsing, we put error states into ProgramTypes and RuntimeTypes. The main structures here are:

- `WrittenTypes.Name` which is a `KnownBuiltin` of `Unresolved`: just a list of strings representing a name to be resolved later. This replaces all uses of `PT.FQName` in WrittenTypes
- `ProgramTypes.NameResolution` which is a `Result`, and will be `Error` if the name wasn't found during NameResolution
- `RuntimeTypes.EError` for errors that should be returned when encountered/executed
- `Runtine.FQName.Unknown` for unknown types and functions (`Result.Error`s in PT get converted into this.
- `WT.EPipeVariableOrUserFunction` for the ambiguous "is it a function or var, we don't know yet" case
- `WT2PT.NameResolver` contains all the known types/functions that a name resolution can use (both builtin and user; it doesn't support packages yet)
- `LibExecution.Errors.NameResolution`: error type for NameResolution

We implement a name resolution algorithm which looks up names in all the various places and checks that they all exist. This also looks "up" the module stack. Getting the NameResolvers right for each calling location (eg Cli vs LocalExec vs Cloud vs Tests) isn't completely right yet.

TypeChecking progress:
- recurse into Lists, Dicts, etc (find/fix lots of new errors esp in tests and DarkInternal and Dicts!)
- start to use typeArgs a bit (finds more problems)
- resolve aliases before returning errors
- return the first result, not the last one
- use typeArgs in Json type stuff

Parser changes:
- change TestModule from a nested set of modules to a flat list of modules
- change PackageParser from a nested set of modules to a flat list

Other NameResolution changes:
- split LocalExec.Libs.Packages into two files so we can have the names from Packages accessible in Packages2

Misc:
- remove unnecessary stuff from WrittenTypes
- small refactor of parsing
- write a README for the parser
- improve error information in LocalExec, TestUtils.equalDval, all LibExecutionTests, 
- split packages logs so we don't overwrite
- improve unrelated comments eg SerializedTypes
- add Tablecloth.Result.orElse
- DB tests are now sorted to be more reliable/avoid occasional ordering errors
- List.unzip uses a Tuple now
- remove tuples from newline pretty-printing


backend/src/LocalExec/StdLib.fs
```diff
@@ -17,6 +17,7 @@ let typeRenames : StdLib.TypeRenames =
 let contents : StdLib.Contents =
   StdLib.combine
     [ Libs.Packages.contents
+      Libs.Packages2.contents
```
StachuDotNet, line 20: follow-up: probably worth renaming Packages2 to something else

backend/tests/Tests/LibExecution.Tests.fs
```diff
@@ -141,40 +155,36 @@ let fileTests () : Test =
...
       testList $"skipped - {testName}" []
     else
       try
-        (baseDir + filename)
-        |> Parser.TestModule.parseTestFile
-        |> moduleToTests testName
+        let modules =
+          (baseDir + filename) |> Parser.TestModule.parseTestFile builtinResolver
+
+        // Within a module, tests have access to
```
StachuDotNet, line 165: misc, unfinished thought

backend/testfiles/execution/json.tests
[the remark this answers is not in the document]
pbiggar: Yes. Since they're now typechecked, the incorrect types just exercise the type checking logic, and not the json logic.

We could theoretically disable the runtime type checker to put this logic back, as it does contain error cases. However, they are unreachable at the moment (or should be).

backend/src/LocalExec/StdLib.fs
pbiggar: Good idea, any thoughts?

PR #4982, opened 2023-07-28
Make EApply take an Expr

pbiggar:
This changes EApply to take an Expr. This will allow returning functions that return functions, and such things, but that still has to be enabled.

Merging this now (even though a no-op) to avoid future conflicts.


backend/src/StdLibExecution/Libs/List.fs
```diff
@@ -354,7 +354,7 @@ let fns : List<BuiltInFn> =
           uply {
             let f (dv : Dval) : Ply<bool> =
               uply {
-                let! result = Interpreter.applyFnVal state fn [ dv ]
+                let! result = Interpreter.applyFnVal state 0UL fn [] [ dv ]
```
StachuDotNet, line 357: wondering - are there any negative consequences of using `0UL` for all of these internal Interpreter calls?

pbiggar: It means traces will give poor results - it won't know where to put them.

PR #4984, opened 2023-07-28
naive (temp) caching for package managers to hold us over

StachuDotNet:
This adds pretty naive caching for our package managers (both DB-based and the web-based one for CLI that accesses the `dark-packages` canvas). Still fetches in bulk, but caches results for a minute. Reasonable for now? Makes things much faster for me, anyway.


StachuDotNet:
will merge; useful improvement and we can always revert. this will be replaced soon anyway

PR #4989, opened 2023-07-30
Remove Wasm and LibAnalysis from being compiled

pbiggar:
This is old but still worthwhile I think? Should I resurrect it?


StachuDotNet:
since we've already disabled the Release build of the Wasm project (which was the more damning issue), I think these projects are fine to leave referenced by the .sln as-is; we'll need them eventually, most likely, so it'll be good to make sure they're kept up to date with other changes

pbiggar:
OK sounds good

PR #4993, opened 2023-08-01
Fix bad reformat which made tests disappear

pbiggar:
The formatting of type-alias.dark somehow hid all but one test. This reformats it properly and also fixes some error messages from recent typechecker changes.


pbiggar:
I also confirmed it didn't happen to any other test files.

PR #4994, opened 2023-08-01
Support Aliases in Enum and Dval equality when aliasing

pbiggar:


backend/src/LibExecution/Interpreter.fs
```diff
@@ -628,14 +643,17 @@ let rec eval' (state : ExecutionState) (st : Symtable) (e : Expr) : DvalTask =
...
-        let case = (case :: cases) |> List.tryFind (fun c -> c.name = caseName)
+      let! typ = Types.find typeName types
+
+      match! enumMaybe types typeName with
+      | None ->
+        match typ with
+        | None -> return err id $"There is no type named {typeStr}"
+        | Some({ definition = TypeDeclaration.Enum _ }) ->
+          return err id $"Expected a record but {typeStr} is an enum"
+        | _ -> return err id $"Expected a record but {typeStr} is something else"
```
StachuDotNet, line 654: we might as well exhaust the `match` cases here since there's just Alias and Record remaining? (helpful to catch something later)

backend/src/StdLibExecution/Libs/NoModule.fs
```diff
@@ -9,8 +9,7 @@ open LibExecution.RuntimeTypes
 open LibExecution.StdLib.Shortcuts
 
 
-let rec equals (types : Types) (a : Dval) (b : Dval) : bool =
-  let equals = equals types
```
StachuDotNet, line 13: woohoo! (no `types` input)

StachuDotNet (approved):
Looks great! Tried to get chatgpt to suggest additional notable test cases, came up with nothing useful.

backend/src/LibExecution/Interpreter.fs
pbiggar: I'm continuing to do work on it, so I'll do this as part of it. (TypeArgs aren't handled anywhere here)

PR #5003, opened 2023-08-02
Rename Just/Nothing to Some/None

OceanOak:
Changelog:

```
Area of change
- Rename Just/Nothing to Some/None
```


backend/src/LibCliExecution/PackageManager.fs
```diff
@@ -17,8 +17,8 @@ module LanguageToolsTypesFork =
   type TLID = uint64
 
   type NonStandardOption<'a> =
```
StachuDotNet, line 19:
```suggestion
  // CLEANUP can be removed (just use the standard Option)
  type NonStandardOption<'a> =
```

PR #5007, opened 2023-08-03
Refactor to remove extra fields from RT.TypeDefinition

pbiggar:
Also use NEList for TypeDefinitions in all types


backend/src/LibBinarySerialization/SerializedTypes.fs
```diff
@@ -361,8 +361,8 @@ module TypeDeclaration =
   [<MessagePack.MessagePackObject>]
   type Definition =
     | Alias of TypeReference
-    | Record of firstField : RecordField * additionalFields : List<RecordField>
-    | Enum of firstCase : EnumCase * additionalCases : List<EnumCase>
+    | Record of NEList<RecordField>
+    | Enum of NEList<EnumCase>
```
StachuDotNet, line 365: Happy we finally got to simplify these

PR #5011, opened 2023-08-05
Add List.dropLast + misc fixes

OceanOak:
Changelog:

```
Language
- Add a PACKAGE.Darklang.Stdlib.List.dropLast function 
```
This PR :
- Resolves the issue of running scripts from the CLI by adding a new StdLibCliHost project
- Fixes the pretty-printer
- Resolves the issue of saving constants to the constantsDB
- Adds a List.dropLast function


backend/src/StdLibCliHost/paket.references
```diff
@@ -0,0 +1,3 @@
+Ply
+FSharp.Core
+FSharpPlus
```
StachuDotNet, line 3: the FSharpPlus reference may be remove-able - not sure. (guessing the file was copied from the other project verbatim)

packages/darklang/stdlib/list.dark
```diff
@@ -0,0 +1,21 @@
...
+  module Stdlib =
+    module List =
+
+      let dropLast (list: List<'a>) : List<'a> =
+        match list with
+        | [] -> []
+        | _ ->
+          match (List.length_v0 list) with
+          | 1 -> []
+          | _ ->
```
StachuDotNet, line 11: rather than `match`ing on the length like this, can use the `::` match pattern

```fsharp
match list with
| [] -> []
| head :: tail -> 
  ...
```
unless I'm missing something

OceanOak: No, you're right, I just wasn't aware that I can use the `::` notation in dark files. Thanks!

OceanOak:
The show command will be introduced in a separate PR  to align with the new approach and prevent frequent rebasing of this PR with the current changes

StachuDotNet:
Great!

PR #5012, opened 2023-08-06
Use NELists more

pbiggar:
- Use NELists in more places
- Add a shorttypename for errors
- better errors for invalid types in enums:
```fsharp
Expected: MyEnum.X (String)
Actual: MyEnum.X (Int)
```

### DIFF
--- a/backend/src/LibCloud/SqlCompiler.fs
+++ b/backend/src/LibCloud/SqlCompiler.fs
@@ -510,7 +510,7 @@ let rec lambdaToSql
             : (Option<string> * NEList<string>) =
             match subExpr with
             | EFieldAccess(_, subExpr, childFieldName) ->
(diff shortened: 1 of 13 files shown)


pbiggar:
Separately, I'm thinking a bit about how we use NELists, and whether our usage of NELists to represent impossible states in ProgramTypes and WrittenTypes is valid. I believe the purpose of ProgramTypes is to represent not just valid programs, but also invalid programs as they might be typed by users. As such, when a user types eg `MyRecord {}` - it feels like that should be representable in ProgramTypes (of course it would be a runtime error in RuntimeTypes whenever the user tries to use it).

Starting to write this up here: https://www.notion.so/darklang/Language-design-guidelines-cf8f48ce32734667b7c306998ec8c9cb

I don't believe this PR falls foul of this.

PR #5013, opened 2023-08-06
Try more NELists

pbiggar:
This is a quick spike to look at what types in ProgramTypes we might consider converting to NELists. It fueled these thoughts: https://www.notion.so/darklang/Language-design-guidelines-cf8f48ce32734667b7c306998ec8c9cb

### DIFF
--- a/backend/src/LibExecution/ProgramTypes.fs
+++ b/backend/src/LibExecution/ProgramTypes.fs
@@ -366,10 +366,8 @@ type TypeReference =
   | TUuid
   | TBytes
   // A named variable, eg `a` in `List<a>`, matches anything
-  | TVariable of string // replaces TAny
-  | TFn of
-    List<TypeReference> *  // CLEANUP: NEList
-    TypeReference // replaces TLambda
+  | TVariable of string
+  | TFn of NEList<TypeReference> * TypeReference
 
   /// A type defined by a standard library module, a canvas/user, or a package
   /// e.g. `Result<Int, String>` is represented as `TCustomType("Result", [TInt, TString])`
@@ -381,7 +379,7 @@ type TypeReference =
 type Expr =
   | EInt of id * int64
   | EBool of id * bool
-  | EString of id * List<StringSegment>
+  | EString of id * NEList<StringSegment>
   /// A character is an Extended Grapheme Cluster (hence why we use a string). This
   /// is equivalent to one screen-visible "character" in Unicode.
   | EChar of id * string
@@ -396,7 +394,7 @@ type Expr =
   | EInfix of id * Infix * Expr * Expr
   // the id in the varname list is the analysis id, used to get a livevalue
   // from the analysis engine
-  | ELambda of id * List<id * string> * Expr
+  | ELambda of id * NEList<id * string> * Expr
   | EFieldAccess of id * Expr * string
   | EVariable of id * string
   | EApply of id * Expr * typeArgs : List<TypeReference> * args : List<Expr>
@@ -404,11 +402,11 @@ type Expr =
   | EList of id * List<Expr>
   | EDict of id * List<string * Expr>
   | ETuple of id * Expr * Expr * List<Expr>
-  | EPipe of id * Expr * PipeExpr * List<PipeExpr>
+  | EPipe of id * Expr * NEList<PipeExpr>
 
   // See NameResolution comment above
-  | ERecord of id * NameResolution<TypeName.T> * List<string * Expr>
-  | ERecordUpdate of id * record : Expr * updates : List<string * Expr>
+  | ERecord of id * NameResolution<TypeName.T> * NEList<string * Expr>
+  | ERecordUpdate of id * record : Expr * updates : NEList<string * Expr>
 
   // Enums include `Some`, `None`, `Error`, `Ok`, as well
   // as user-defined enums.
@@ -441,7 +439,7 @@ and StringSegment =
 
 and PipeExpr =
   | EPipeVariable of id * string
-  | EPipeLambda of id * List<id * string> * Expr
+  | EPipeLambda of id * NEList<id * string> * Expr
   | EPipeInfix of id * Infix * Expr
   | EPipeFnCall of
     id *
@@ -588,7 +586,7 @@ module UserFunction =
     { tlid : tlid
       name : FnName.UserProgram
       typeParams : List<string>
-      parameters : List<Parameter>
+      parameters : NEList<Parameter>
       returnType : TypeReference
       description : string
       deprecated : Deprecation<FnName.T>
@@ -631,7 +629,7 @@ module PackageFn =
       name : FnName.Package
       body : Expr
       typeParams : List<string>
-      parameters : List<Parameter>
+      parameters : NEList<Parameter>
       returnType : TypeReference
       description : string
       deprecated : Deprecation<FnName.T> }
--- a/backend/src/LibExecution/ProgramTypes.fs
+++ b/backend/src/LibExecution/ProgramTypes.fs
@@ -17,7 +17,7 @@ module FQName =
   /// The name of a thing in the package manager
   // TODO: We plan to use UUIDs for this, but this is a placeholder
   type Package<'name> =
-    { owner : string; modules : NEList<string>; name : 'name; version : int }
+    { owner : string; modules : NEList<string>; name : 'name; version : int } // NEList
 
   type T<'name> =
     | BuiltIn of BuiltIn<'name>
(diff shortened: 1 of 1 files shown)


backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -366,10 +366,8 @@ type TypeReference =
   | TUuid
   | TBytes
   // A named variable, eg `a` in `List<a>`, matches anything
-  | TVariable of string // replaces TAny
-  | TFn of
-    List<TypeReference> *  // CLEANUP: NEList
-    TypeReference // replaces TLambda
+  | TVariable of string
+  | TFn of NEList<TypeReference> * TypeReference
```
pbiggar, line 370: I can't think of a case where we wouldnt have an NEList here.

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -381,7 +379,7 @@ type TypeReference =
 type Expr =
   | EInt of id * int64
   | EBool of id * bool
-  | EString of id * List<StringSegment>
+  | EString of id * NEList<StringSegment>
```
pbiggar, line 382: We could represent `""` as either `EString([])` or `EString([""])`. I'm ok with either.

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -396,19 +394,19 @@ type Expr =
   | EInfix of id * Infix * Expr * Expr
   // the id in the varname list is the analysis id, used to get a livevalue
   // from the analysis engine
-  | ELambda of id * List<id * string> * Expr
+  | ELambda of id * NEList<id * string> * Expr
```
pbiggar, line 397: A user might typo and forget to add parameters here. It feels like this should be representable and that we don't want an NEList.

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -396,19 +394,19 @@ type Expr =
...
   | EFieldAccess of id * Expr * string
   | EVariable of id * string
   | EApply of id * Expr * typeArgs : List<TypeReference> * args : List<Expr>
   | EFnName of id * NameResolution<FnName.T>
   | EList of id * List<Expr>
   | EDict of id * List<string * Expr>
   | ETuple of id * Expr * Expr * List<Expr>
-  | EPipe of id * Expr * PipeExpr * List<PipeExpr>
+  | EPipe of id * Expr * NEList<PipeExpr>
```
pbiggar, line 405: NEList seems fine here, as we already had one in practice.

One place this wouldn't support is:
```fsharp
5 |> // user omitted the code to the right
```

this code won't parse. If `|>` was a simple infix fn application, than the code would parse. As such, I think going back to a List here might be appropriate.

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -396,19 +394,19 @@ type Expr =
...
 
   // See NameResolution comment above
-  | ERecord of id * NameResolution<TypeName.T> * List<string * Expr>
-  | ERecordUpdate of id * record : Expr * updates : List<string * Expr>
+  | ERecord of id * NameResolution<TypeName.T> * NEList<string * Expr>
```
pbiggar, line 408: Suppose the user types:

```fsharp
MyRecord {}
```

obviously that's wrong, but it should be allowed to parse and be represented in ProgramTypes.

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -396,19 +394,19 @@ type Expr =
...
+  | ERecordUpdate of id * record : Expr * updates : NEList<string * Expr>
```
pbiggar, line 409: Suppose the user types:

```fsharp
{ myRecord with /* nothing */ }
```

Struggling to make any sense of this, so NEList feels OK,

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -441,7 +439,7 @@ and StringSegment =
 
 and PipeExpr =
   | EPipeVariable of id * string
-  | EPipeLambda of id * List<id * string> * Expr
+  | EPipeLambda of id * NEList<id * string> * Expr
```
pbiggar, line 442: See ELambda. Should probably be a list.

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -588,7 +586,7 @@ module UserFunction =
     { tlid : tlid
       name : FnName.UserProgram
       typeParams : List<string>
-      parameters : List<Parameter>
+      parameters : NEList<Parameter>
```
pbiggar, line 589: If there's no parameters, than presumably this will be a constant? Though what if there's typeParams and no params? What would we do with that?

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -631,7 +629,7 @@ module PackageFn =
       name : FnName.Package
       body : Expr
       typeParams : List<string>
-      parameters : List<Parameter>
+      parameters : NEList<Parameter>
```
pbiggar, line 632: See UserFn

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -17,7 +17,7 @@ module FQName =
   /// The name of a thing in the package manager
   // TODO: We plan to use UUIDs for this, but this is a placeholder
   type Package<'name> =
-    { owner : string; modules : NEList<string>; name : 'name; version : int }
+    { owner : string; modules : NEList<string>; name : 'name; version : int } // NEList
```
pbiggar, line 20: Are users allowed type `@paul.MyType`? Probably they should, even if we don't allow types in that namespace (which I actually think might be fine?).

Similarly, maybe we shouldn't assert so heavily on function (and other) name regexes?

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -506,10 +504,10 @@ module TypeDeclaration =
     | Alias of TypeReference
 
     /// `type MyRecord = { a : int; b : string }`
-    | Record of NEList<RecordField>
+    | Record of NEList<RecordField> // NEList
```
pbiggar, line 507: These are record type definitions. Is the user allowed type:

```fsharp
type MyRecord = {}
```

If we don't allow that, than presumably the type won't be added to the program, which will result with a NameResolution error when it's used. This might be OK? Not sure. Perhaps we'll learn from experience here.

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -506,10 +504,10 @@ module TypeDeclaration =
...
 
     /// `type MyEnum = A | B of int | C of int * (label: string)`
-    | Enum of NEList<EnumCase>
+    | Enum of NEList<EnumCase> // NEList
```
pbiggar, line 510: What could they even type here?

```fsharp
type X = | // maybe this?
```

Dunno, feels too far.

backend/src/LibExecution/ProgramTypes.fs
StachuDotNet: Agreed - the 'smallest' possible thing would be a `Unit -> Unit`

backend/src/LibExecution/ProgramTypes.fs
StachuDotNet: I have a slight leaning towards `EString([])` being OK, and not doing NEList here

backend/src/LibExecution/ProgramTypes.fs
StachuDotNet: do you mean like `List.map ( -> item.prop)`?

StachuDotNet: That sort of feels like "should be parseable in WT, but fail in WT2PT - reminds me of https://github.com/darklang/dark/pull/5013#discussion_r1285275273

backend/src/LibExecution/ProgramTypes.fs
StachuDotNet: thinking back to the lambda example, I suppose someone could type:
`let myFn (f: -> Int)`?

StachuDotNet:
I guess we'll need some construct of "this PT.Expr is _totally reasonable_ and not at all broken"
i.e. doesn't include anything like a `MyRecord {}`

Not sure why it'd be useful, besides giving a dark dev things to fix

backend/src/LibExecution/ProgramTypes.fs
pbiggar:
```elm
List.map (\ -> item.prop)
```

backend/src/LibExecution/ProgramTypes.fs
pbiggar: That's true. On the other hand, we might clean that up in the parser too, something like `let myFn (f: _ -> Int)`

pbiggar:
Done in #5023

PR #5017, opened 2023-08-09
Add Show command

OceanOak:
Changelog:

```
Cli
- Add show command
```

currently:
`show @Owner.Modules.functionName` =>  pretty-prints a function
`show @Owner.Modules.TypeName` =>  pretty-prints a type
`show @Owner.Modules.constantName` =>  pretty-prints a constant

### DIFF
--- a/backend/src/StdLibExecution/Libs/String.fs
+++ b/backend/src/StdLibExecution/Libs/String.fs
@@ -998,30 +998,6 @@ let fns : List<BuiltInFn> =
         | _ -> incorrectArgs ())
       sqlSpec = NotYetImplemented
       previewable = Pure
-      deprecated = NotDeprecated }
-
-
-    { name = fn "ellipsis" 0
-      typeParams = []
-      parameters =
-        [ Param.make "string" TString ""; Param.make "targetLength" TInt "" ]
-      returnType = TString
-      description =
-        "Truncates <param string> if its length is more than <param targetLength>. The truncated string will have an ellipsis ('...') appended.
-        If the <param string> is shorter than or equal to <param targetLength>, returns the unchanged <param string>."
-      fn =
-        (function
-        | _, _, [ DString s; DInt l ] ->
-          let targetLength = int l
-
-          if String.lengthInEgcs s > targetLength then
-            let truncatedString = s.Substring(0, targetLength)
-            truncatedString + "..." |> DString |> Ply
-          else
-            s |> DString |> Ply
-        | _ -> incorrectArgs ())
-      sqlSpec = NotYetImplemented
-      previewable = Pure
       deprecated = NotDeprecated } ]
 
 let contents = (fns, types, constants)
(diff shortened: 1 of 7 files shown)


canvases/dark-packages/main.dark
```diff
@@ -121,4 +171,33 @@ let _handler _req =
     |> unwrap
     |> String.toBytes
 
-  PACKAGE.Darklang.Stdlib.Http.response respBody 200
\ No newline at end of file
+  PACKAGE.Darklang.Stdlib.Http.response respBody 200
+
+
+[<HttpHandler("GET", "/byName/:name")>]
```
pbiggar, line 177: Can you come up with something more RESTful for these endpoints?

packages/darklang/cli/cli.dark
```diff
@@ -3,17 +3,180 @@ module Darklang =
...
+
+        result
+
+
     let parseArguments (args: List<String>) : PACKAGE.Darklang.Cli.Command =
       match args with
       | [ "-h" ] -> PACKAGE.Darklang.Cli.Command.Help
       | [ "--help" ] -> PACKAGE.Darklang.Cli.Command.Help
       | [ "--prompt"; prompt ] -> PACKAGE.Darklang.Cli.Command.Infer prompt
+      | [ "--show"; fnName ] ->
```
pbiggar, line 172: This should be `darklang show`, not `darklang --show`

backend/src/StdLibExecution/Libs/String.fs
```diff
@@ -993,6 +993,30 @@ let fns : List<BuiltInFn> =
         | _ -> incorrectArgs ())
       sqlSpec = NotYetImplemented
       previewable = Pure
+      deprecated = NotDeprecated }
+
+
+    { name = fn "ellipsis" 0
```
pbiggar, line 999: I see this is also in a package. Can you remove it from Builtins?

pbiggar (changes requested):
Looks great! Should be good to merge with a few small changes.

backend/src/StdLibExecution/Libs/String.fs
OceanOak: The package version treats EGCs as a single character rather than breaking them apart. Is that OK?

canvases/dark-packages/main.dark
```diff
@@ -122,4 +171,33 @@ let _handler _req =
...
+  let owner = fullName |> List.head |> unwrap
+  let name = fullName |> List.last |> unwrap
+
+  let modules = fullName |> (List.drop 1) |> PACKAGE.Darklang.Stdlib.List.dropLast
+  let modulesWithoutItem = fullName |> (List.drop 1)
+
+  let getall = DB.getAll NamesDB
+
+  let res =
+    getall
```
StachuDotNet, line 189: Could we please rename this to `allItems` or in-line this (and wrap in parens ()) ?

StachuDotNet (changes requested):
Great!
It may be worth renaming `whatIsIt` to `category` to fall in line with the http handler's route.
Similarly, `NameToItem` may be clearer with a name such as `NameToCategoryLookup` or something.

Beyond those, I think this is good to merge.

StachuDotNet:
Ah - could we also add a CLEANUP to the `DB.getAll ... |> List.filter`? Eventually this should be a single DB.query but we're limited by our SqlCompiler's ignorance of some type args

PR #5019, opened 2023-08-09
if with no else

ditansu:
Doing `else` is optional part


ditansu:
@pbiggar probably the PR is ready to review :)

backend/src/LibExecution/Interpreter.fs
```diff
@@ -652,21 +652,24 @@ let rec eval'
...
       | DBool false ->
         do! preview tat st thenbody
-        return! eval state tat st elsebody
+        match elsebody with
+        | None -> return DUnit
+        | Some eb -> return! eval state tat st eb
       | DBool true ->
         let! result = eval state tat st thenbody
-        do! preview tat st elsebody
+        do! preview tat st elsebodyUnwrapped
```
pbiggar, line 664: Using the id for elseBody will save `()` as the return value for the `if` statement. I don't think that's right for this branch.

backend/src/LibExecution/Interpreter.fs
```diff
@@ -652,21 +652,24 @@ let rec eval'
...
         return result
       | cond when Dval.isFake cond ->
         do! preview tat st thenbody
-        do! preview tat st elsebody
+        do! preview tat st elsebodyUnwrapped
```
pbiggar, line 668: Same for this branch.

backend/src/LibExecution/Interpreter.fs
```diff
@@ -652,21 +652,24 @@ let rec eval'
...
         return cond
       | _ ->
         do! preview tat st thenbody
-        do! preview tat st elsebody
+        do! preview tat st elsebodyUnwrapped
```
pbiggar, line 672: Same for this branch.

backend/src/StdLibDarkInternal/Helpers/ProgramTypesToDarkTypes.fs
```diff
@@ -715,8 +715,12 @@ module Expr =
 
 
       // control flow
-      | PT.EIf(id, cond, ifTrue, ifFalse) ->
-        "EIf", [ DInt(int64 id); toDT cond; toDT ifTrue; toDT ifFalse ]
+      | PT.EIf(id, cond, thenExpr, elseExpr) ->
+        let elseExpr =
+          match elseExpr with
+          | Some elExp -> elExp
+          | None -> PT.EUnit id
```
pbiggar, line 722: For Dark types, we use full proper options. See for example `EnumField.label`

backend/tests/TestUtils/TestUtils.fs
```diff
@@ -563,10 +563,16 @@ module Expect =
...
       eq ("rhs" :: path) rhs rhs'
       eq ("body" :: path) body body'
-    | EIf(_, con, thn, els), EIf(_, con', thn', els') ->
+    | EIf(id, con, thn, els), EIf(_, con', thn', els') ->
       eq ("cond" :: path) con con'
       eq ("then" :: path) thn thn'
-      eq ("else" :: path) els els'
+      match els, els' with
+      | Some el, Some el' -> eq ("else" :: path) el el'
+      | None, None -> eq ("else" :: path) (EUnit id) (EUnit id)
```
pbiggar, line 571: No need for checking equality on DUnits, you can just do `()`

pbiggar (changes requested):
Looks great, thanks! Should be able to merge with a few small changes.

backend/src/LibExecution/Interpreter.fs
ditansu: I'm not sure that I understand what is the right option here. Could you please to add details?

backend/src/StdLibDarkInternal/Helpers/ProgramTypesToDarkTypes.fs
ditansu: hm, it is not clear for me what does mean "full proper options" here. 
if you refer to [EnumField toDT](https://github.com/darklang/dark/blob/main/backend/src/StdLibDarkInternal/Helpers/ProgramTypesToDarkTypes.fs#L972) then it seems like we need simple unwrap the optional - that is we exactly to do here. Please clarify.

StachuDotNet: I think Paul is suggesting something like (maybe exactly):
```fsharp
let elseExpr =  elseExpr |> Option.map Expr.toDT |> Dval.option
```

backend/src/LibExecution/Interpreter.fs
StachuDotNet: And here, neither the `thenBody` or `elseBody` seem appropriate since we're dealing with a fake val - we should probably return something else - not sure what

backend/src/StdLibDarkInternal/Helpers/ProgramTypesToDarkTypes.fs
ditansu: aa ok, got it, thanks @StachuDotNet

backend/src/LibExecution/Interpreter.fs
pbiggar: We make calls to `preview` so that we have analysis results for the path that wasn't executed.

That is, if we have

```
if true
  let x = 5
  6
else
  let y = "str"
  7
```

then we'd like to know that y would have been 7 if we had executed the other branch. That's what `preview` handles (it stores the result of expressions using `NotExecuted`).

If there is no else body, there is no preview to execute. Previewing using elseBodyUnwrapped adds an extra analysis result using `id` (which is the `id` for the `if`, so that's wrong). This will overwrite the result of the `if` statement with `NotExecuted DUnit`.

So when `cond` is true and there's no `elseBody`, there's nothing to preview.

backend/src/LibExecution/Interpreter.fs
pbiggar: We still want to preview both thenBody and elseBody here (unless there is no elseBody).

backend/src/LibExecution/Interpreter.fs
ditansu: @pbiggar thanks for explanation! As I've understood, we use `preview` to show value (7) on canvas for `NotExecuted` branch (of course, if the branch exists). Hope I understand correctly how to apply preview for `elseBody`, please check the last commit.

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -776,6 +776,15 @@ let partiallyEvaluate
     let rec postTraversal (expr : Expr) : Ply.Ply<Expr> =
       let r = postTraversal
 
+      let mapAsync f opt =
```
pbiggar, line 779: This would be better inlined into the code

backend/src/StdLibDarkInternal/Helpers/ProgramTypesToDarkTypes.fs
```diff
@@ -827,8 +828,12 @@ module Expr =
       PT.EVariable(uint64 id, varName)
 
     // control flow
-    | DEnum(_, _, "EIf", [ DInt id; cond; ifTrue; ifFalse ]) ->
-      PT.EIf(uint64 id, fromDT cond, fromDT ifTrue, fromDT ifFalse)
+    | DEnum(_, _, "EIf", [ DInt id; cond; thenExpr; elseExpr ]) ->
```
pbiggar, line 831: I think this is the cause of the errors in CI - we need to read from options in `fromDT` like we write to them in `toDT`

pbiggar (changes requested):
Looking good. I left comments on the two remaining changes and the nwe can merge.

StachuDotNet (commented):
I'm a bit concerned seeing this:

```
(if 1 > 3 then
   4) = ()
```
It feels a bit inappropriate to me for two branches to return different types - one an Int and one (invisible branch) a Unit. Is that expected/OK, @pbiggar? Maybe I've missed or forgotten some conversation around this or am thinking about this wrong.

pbiggar:
> It feels a bit inappropriate to me for two branches to return different types - one an Int and one (invisible branch) a Unit. Is that expected/OK, @pbiggar? Maybe I've missed or forgotten some conversation around this or am thinking about this wrong.

That's a good point. The question is what do we do about it.

We could give a runtime error if it's not a unit perhaps? Though it feels like we'd want to wait until an actual type mismatch happens, according to our principles of runtime type checking.

One thing here is that I don't think we're really done here given there's no good way to use it (since there's no statements in dark, so you can't really have an expr that does something and then is ignores, like a `debug` statement). So there's definitely followup needed here either way.

backend/testfiles/execution/language/eif.dark
```diff
@@ -6,4 +6,25 @@
...
+(if 1 < 3 then
+   4) = 4
+
+(if 1 < 3 then
+   "msg") = "msg"
+
+(if () then
+   "msg") = Test.derrorMessage "If only supports Booleans"
+
+(if true then
```
pbiggar, line 26: Maybe a test here for `if false then Test.runtimeError`

pbiggar (approved):
I've got two minor comments, however, I'm going to merge and we can do those as followups.

pbiggar:
Thanks, this was great!

PR #5024, opened 2023-08-13
Add a copy of List.iter and List.flatten to LocalExec

OceanOak:
No changelog

#5020 follow-up
This PR adds a copy of both List.iter and List.flatten to LocalExec. Reason being, we are unable to use PACKAGE functions during the 'load from disk into DB' process.

### DIFF
--- a/backend/src/LocalExec/Libs/List.fs
+++ b/backend/src/LocalExec/Libs/List.fs
@@ -19,7 +19,7 @@ let fns : List<BuiltInFn> =
       typeParams = []
       parameters =
         [ Param.make "list" (TList varA) ""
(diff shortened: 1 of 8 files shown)


backend/src/LocalExec/Libs/List.fs
```diff
@@ -0,0 +1,71 @@
...
+module Interpreter = LibExecution.Interpreter
+
+let varA = TVariable "a"
+
+let types : List<BuiltInType> = []
+let constants : List<BuiltInConstant> = []
+
+
+let fns : List<BuiltInFn> =
+  [ { name = fn [ "LocalExec"; "StdLib"; "List" ] "iter" 0
```
pbiggar, line 18: Maybe a different word from StdLib. Perhaps Builtins or a similar synonym, to indicate why we're using these.

PR #5025, opened 2023-08-14
WIP: Lambda arg destructuring

ditansu:
Lambda argument **destructuring** support
Like that ` fun (num, str) ->...` in example below

```fsharp
(let pairs = [(1, "one"); (2, "two"); (3, "three")] in
 let transformer = fun (num, str) -> $"{Int.toString num}:{str}" in
 List.map_v0 pairs transformer) = ["1:one"; "2:two"; "3:three"]
```


ditansu:
@pbiggar hey! Please review this PR. Is it right way to make a lambda args destructuring in the interpreter? What else cases need to handle - e.g. args list destrucructuring?

StachuDotNet:
It's neat you were able to tackle this without introducing any new language constructs - I expected we'd end up adding a `LambdaArgPattern` or re-use the `LetPattern` in the lambda to support this functionality. Maybe this is fine, though? I'll defer to the requested Paul to review this :)

backend/testfiles/execution/language/elambda.dark
```diff
@@ -32,7 +32,37 @@ PACKAGE.Darklang.Stdlib.List.push_v0 [] (fun x -> -4.611686018e+18) = [ (fun x -
...
- let x = 2
- List.map [ 1; 2; 3 ] f) = [ 1; 1; 1 ]
\ No newline at end of file
+(let f = fun myarg1 -> myarg1 + 1
+ List.map [ 1; 2; 3 ] f) = [ 2; 3; 4 ]
+
+List.fold_v0 [ "a"; "b" ] "x" (fun accum curr -> accum ++ curr) = "xab"
+
+(let pairs = [ (1, "one"); (2, "two") ] in
+ let transformer = fun num str -> $"{Int.toString num}:{str}" in
```
pbiggar, line 41:
```suggestion
 let transformer = fun (num, str) -> $"{Int.toString num}:{str}" in
```

backend/testfiles/execution/language/elambda.dark
```diff
@@ -32,7 +32,37 @@ PACKAGE.Darklang.Stdlib.List.push_v0 [] (fun x -> -4.611686018e+18) = [ (fun x -
...
+ List.map_v0 pairs transformer) = [ "1:one"; "2:two" ]
+
+(let triplets = [ (1, "one", true); (2, "two", false) ] in
+
+ let transformer =
+   fun num str flag ->
```
pbiggar, line 47:
```suggestion
   fun (num, str, flag) ->
```

pbiggar (changes requested):
I think we want to reuse LetPattern here. Tuples are single values so they should destructure using a tuple pattern:

```
List.map (fun (k, v) -> k ++ v) [("1", "2")]
```

backend/src/LibExecution/Interpreter.fs
```diff
@@ -829,26 +830,51 @@ and executeLambda
...
-    let expectedLength = NEList.length l.parameters
-    let actualLength = NEList.length args
-    if expectedLength <> actualLength then
+
+    let paraList = parameters |> NEList.toList
+    let argsList = args |> NEList.toList
+
+    // Check destructuring
+    match argsList with
+    | [ DTuple(_, _, rest) ] when
```
pbiggar, line 839: Can the destructuring be integrated with ELet destructuring and using the same code? It's not just about allowing tuples in lambdas, it's about having `let (x, y) =` do the same thing as `fun (x,y) -> `

pbiggar (changes requested):
Can the destructuring be integrated with ELet destructuring and using the same code? It's not just about allowing tuples in lambdas, it's about having let (x, y) = do the same thing as fun (x,y) ->

ditansu:
> Can the destructuring be integrated with ELet destructuring and using the same code? It's not just about allowing tuples in lambdas, it's about having let (x, y) = do the same thing as fun (x,y) ->

@pbiggar so does it mean we have to change the `ELambda` definition on this:
```fsharp
ELambda of id * args: NEList<LetPattern> * body: Expr
```
in `/home/dark/app/backend/src/LibParser/WrittenTypes.fs`?

pbiggar:
> > Can the destructuring be integrated with ELet destructuring and using the same code? It's not just about allowing tuples in lambdas, it's about having let (x, y) = do the same thing as fun (x,y) ->
> 
> @pbiggar so does it mean we have to change the `ELambda` definition on this:
> 
> ```fsharp
> ELambda of id * args: NEList<LetPattern> * body: Expr
> ```
> 
> in `/home/dark/app/backend/src/LibParser/WrittenTypes.fs`?

Yes, that looks like it.

ditansu:
too hard to sync

ditansu:
too hard to sync

PR #5028, opened 2023-08-17
Enable the execution of package functions from the CLI

OceanOak:
Changelog:

```
Cli
- Enable the execution of package functions from the CLI
```

### DIFF
--- a/backend/src/StdLibCliHost/Libs/Cli.fs
+++ b/backend/src/StdLibCliHost/Libs/Cli.fs
@@ -203,9 +203,7 @@ let fns : List<BuiltInFn> =
                       }
 
                     | _ ->
-                      Exception.raiseInternal
-                        "Unknown function should have been converted to EError by PT2RT"
-                        [ "fn", fn ]
+                      Exception.raiseInternal "Only package functions are executable from the CLI" [ "fn", fn ]
 
                   match fn with
                   | None -> return DString "fn not found"
--- a/backend/src/StdLibCliHost/Libs/Cli.fs
+++ b/backend/src/StdLibCliHost/Libs/Cli.fs
@@ -202,8 +202,7 @@ let fns : List<BuiltInFn> =
                         return Option.map packageFnToFn fn
                       }
 
-                    | _ ->
-                      Exception.raiseInternal "Only package functions are executable from the CLI" [ "fn", fn ]
+                    | _ -> Exception.raiseInternal "Error constructing package function name" [ "fn", fn ]
 
                   match fn with
                   | None -> return DString "fn not found"
--- a/backend/src/StdLibCliHost/Libs/Cli.fs
+++ b/backend/src/StdLibCliHost/Libs/Cli.fs
@@ -212,14 +212,16 @@ let fns : List<BuiltInFn> =
                   | Some f ->
                     let types = RT.ExecutionState.availableTypes state
 
-                    let expectedTypes = (NEList.toList f.parameters) |> List.map (fun p -> p.typ)
+                    let expectedTypes =
+                      (NEList.toList f.parameters) |> List.map (fun p -> p.typ)
 
                     let stringArgs =
                       args
                       |> List.map (fun arg ->
                         match arg with
                         | DString s -> s
-                        | e -> Exception.raiseInternal "Expected string" [ "arg", e ])
+                        | e ->
+                          Exception.raiseInternal "Expected string" [ "arg", e ])
 
                     let! args =
                       Ply.List.mapSequentially
--- a/backend/src/StdLibCliHost/Libs/Cli.fs
+++ b/backend/src/StdLibCliHost/Libs/Cli.fs
@@ -229,10 +229,9 @@ let fns : List<BuiltInFn> =
                         (fun (typ, (str : string)) ->
                           uply {
                             let str =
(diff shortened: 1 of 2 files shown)


backend/src/StdLibCliHost/Libs/Cli.fs
```diff
@@ -143,6 +145,119 @@ let fns : List<BuiltInFn> =
...
+              let fnName =
+                LibParser.NameResolver.FnName.resolve
+                  resolver
+                  []
+                  (WT.Unresolved name)
+
+              match fnName with
+              | Ok fnName ->
+                match
+                  NEList.find Dval.isFake (NEList.ofList args.Head args.Tail)
```
StachuDotNet, line 192: the args.Head might fail - we could do this to avoid it:
```
match fnName, args with
| Ok fnName, firstArg :: additionalArgs -> 
  match NEList.find Dval.isFake (NEList.ofList firstArg additionalArgs) with
  | ...
   ...
| _ -> incorrectArgs
`

File: backend/src/StdLibCliHost/Libs/Cli.fs [hunk 2]
```diff
@@ -143,6 +145,119 @@ let fns : List<BuiltInFn> =
...
+
+                  match fn with
+                  | None -> return DString "fn not found"
+                  | Some f ->
+                    let types = RT.ExecutionState.availableTypes state
+
+                    let expectedType =
+                      List.append
+                        [ f.parameters.head.typ ]
+                        (f.parameters.tail |> List.map (fun p -> p.typ))
```
StachuDotNet, line 218: I could be missing something but I think this might work:
`let expectedTypes = (NEList.toList f.parameters) |> List.map (fun p -> p.typ)`

backend/src/StdLibCliHost/Libs/Cli.fs
```diff
@@ -143,6 +145,119 @@ let fns : List<BuiltInFn> =
...
+
+                    let stringArgs =
+                      args
+                      |> List.map (fun arg ->
+                        match arg with
+                        | DString s -> s
+                        | e -> $"Expected string, got {e}")
```
StachuDotNet, line 225: this is unexpected, correct? Let's either return this as an Error case, or raise an exception here

backend/src/StdLibCliHost/Libs/Cli.fs
```diff
@@ -143,6 +145,119 @@ let fns : List<BuiltInFn> =
...
+
+                    let! args =
+                      Ply.List.mapSequentially
+                        (fun (typ, (str : string)) ->
+                          uply {
+                            let str =
+                              if str.StartsWith("\"") && str.EndsWith("\"") then str
+                              else if typ = TString then $"\"{str}\""
+                              else str
```
StachuDotNet, line 234: could you put a small // comment around here noting why we need to do this?

packages/darklang/cli/cli.dark
```diff
@@ -164,6 +165,20 @@ module Darklang =
         result
 
 
+    let runFunction (fnName: String) (args: List<String>) : Int =
+      let fn = fnName |> String.dropFirst_v0 1 |> String.split "."
+      let name = fnName |> String.dropFirst_v0 1
```
StachuDotNet, line 170: let's move this logic to _just_ after we do the "starts with @" check, so the context is visible there

StachuDotNet (changes requested):
Looking good! Just requested a few changes then I think we should be OK to merge

backend/src/StdLibCliHost/Libs/Cli.fs
```diff
@@ -143,6 +145,120 @@ let fns : List<BuiltInFn> =
...
+                    | FQName.Package pkg ->
+                      uply {
+                        let! fn = state.packageManager.getFn pkg
+                        return Option.map packageFnToFn fn
+                      }
+
+                    | _ ->
+                      Exception.raiseInternal
+                        "Unknown function should have been converted to EError by PT2RT"
+                        [ "fn", fn ]
```
StachuDotNet, line 208:
```suggestion
                    | _ ->
                      Exception.raiseInternal "Only package functions are executable from the CLI" [ "fn", fn ]
```

StachuDotNet: I'm not sure if this will be OK by the formatter

backend/src/StdLibCliHost/Libs/Cli.fs
```diff
@@ -143,6 +145,118 @@ let fns : List<BuiltInFn> =
...
+                  let! fn =
+                    match desc with
+                    | FQName.Package pkg ->
+                      uply {
+                        let! fn = state.packageManager.getFn pkg
+                        return Option.map packageFnToFn fn
+                      }
+
+                    | _ ->
+                      Exception.raiseInternal "Only package functions are executable from the CLI" [ "fn", fn ]
```
StachuDotNet, line 206: (I just updated this error message) maybe this isn't the best error message actually - the issue here is that we just tried to create a Package fn name, and _somehow_ it's not

StachuDotNet, line 206:
```suggestion
                    | _ -> Exception.raiseInternal "Error constructing package function name" [ "fn", fn ]
```

StachuDotNet:
Good to merge with a final re-format

PR #5031, opened 2023-08-18
Move several constants from builtins to package space

OceanOak:
Changelog:

```
Packages
- Move constants to packages
```
This PR moves constants to packages.

We are unable to move `List.empty`, `Bytes.empty`, and `Dict.empty` since our current support is limited to types with literal representations, excluding List and Dict constants due to their lack of implicit types

### DIFF
--- a/backend/src/StdLibExecution/Libs/Math.fs
+++ b/backend/src/StdLibExecution/Libs/Math.fs
@@ -18,15 +18,7 @@ let fn = fn modules
 let constant = constant modules
 
 let constants : List<BuiltInConstant> =
-  [ { name = constant "pi" 0
-      typ = TFloat
-      description =
-        "Returns an approximation for the mathematical constant {{π}},
-        the ratio of a circle's circumference to its diameter."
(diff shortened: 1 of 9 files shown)


backend/src/StdLibExecution/Libs/Math.fs
```diff
@@ -18,15 +18,7 @@ let fn = fn modules
...
-  [ { name = constant "pi" 0
-      typ = TFloat
-      description =
-        "Returns an approximation for the mathematical constant {{π}},
-        the ratio of a circle's circumference to its diameter."
-      body = DFloat System.Math.PI
-      deprecated = NotDeprecated }
-
-    { name = constant "tau" 0
+  [ { name = constant "tau" 0
```
pbiggar, line 21: I think you meant to remove tau from here?

OceanOak: Yes, I wasn't sure if using a direct assignment of the tau value instead of `2*pi` was acceptable. I'll remove it in a subsequent PR

PR #5032, opened 2023-08-21
call OpenAI from the CLI 

OceanOak:
Changelog:

```
Cli
- Enable the Cli to interact with the OpenAI API. 
```
This will enable users to send prompts through the CLI and save the generated output into a specified script

### DIFF
--- a/canvases/dark-editor/system-prompt-cli.txt
+++ b/canvases/dark-editor/system-prompt-cli.txt
@@ -1,0 +1,151 @@
+You are a tool that is used by Dark developers to write and manipulate their Dark
+programs.
+
+Dark is a programming language used to write backends. In addition to the language,
+it has built-in support for DBs, REPLs, CRONs, Workers, etc. The term "toplevels" is
+used to refer to these distinct parts of a user's program - a user DB, a REPL, a
+CRON, etc.
+
+Dark is very similar to F#, including syntax and semantics, with these exceptions:
+
+- Dark does not support "record deconstruction" in `match` expressions
+- field and function names use camelCase in Dark, not snake_case
+- Dark does not support mutation.
+- all names must be fully-qualified (eg `String.trimLeft`) and imports (or `open`) are not supported
+- string interpolation replaces `sprintf`, for example `$"My name is {name}"`
+
+---
+
+# Dark features:
+
+## Types:
+
+Dark is statically typed, and supports type definitions. Types are good and should be used often.
+
+- records are supported, such as `type MyType = { x : int }`
+- enums are supported, such as `type MyType = | OptionOne of Int | SecondOption | Another of Int * String * Bool`
+
+## Functions:
+
+Dark supports functions at the top level, but not within other functions. Functions are good and should be used often.
+
+- functions are defined using `let`
+- all parameter types must be specified (i.e. `String` or `Int`) and are capitalized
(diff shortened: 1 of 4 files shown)


canvases/dark-packages/main.dark
```diff
@@ -254,4 +254,12 @@ let _handler _req =
     | Some Module -> "module"
     | None -> "not found"
 
-  PACKAGE.Darklang.Stdlib.Http.response (respBody |> String.toBytes) 200
\ No newline at end of file
+  PACKAGE.Darklang.Stdlib.Http.response (respBody |> String.toBytes) 200
+
+
+
+[<HttpHandler("GET", "/openai-chat/:prompt")>]
```
OceanOak, line 261: It seems like this HTTP handler doesn't belong in this canvas. Maybe we should move it to a separate canvas?

StachuDotNet: I'm not sure why we need this endpoint - why not do the `(PACKAGE.OpenAI.ChatCompletion.completion prompt) |> unwrap` in the CLI code?

OceanOak: In the CLI, we use `CLI.HttpClient.request`, while `PACKAGE.OpenAI.ChatCompletion.completion` uses `HttpClient.request`. Attempting to use it directly results in a 'Function HttpClient.request is not found' error. This is why I needed to add the endpoint.

canvases/dark-editor/system-prompt-cli.txt
```diff
@@ -0,0 +1,151 @@
...
+Here are some examples of a response to the following prompt:
+
+prompt: build a small program that models a list of People,
+a function that returns the age of a Person, then hardcode a list of people
+and return the age of each person
+
+Response:
+
+type Person = {
+name: String
```
pbiggar, line 131: I think some spacing got missed here.

canvases/dark-packages/main.dark
pbiggar: Yes, i agree with Stachu. I guess we need to detangle HttpClient.request from LibBackend.

packages/darklang/cli/cli.dark
```diff
@@ -246,19 +246,63 @@ module Darklang =
         1
 
 
+    let generateCode (prompt: String) (scriptPath: String) : Int =
+
+      let prompt =
+        match File.read "canvases/dark-editor/system-prompt-cli.txt" with
```
pbiggar, line 252: The prompt should be a constant.

pbiggar:
There's a few things I'd change, but no reason to hold this up. I'll make a few more comments inline.

canvases/dark-editor/system-prompt-cli.txt
pbiggar: Maybe throughout this whole file in fact.

packages/darklang/cli/cli.dark
```diff
@@ -246,19 +246,63 @@ module Darklang =
...
+        | Ok darkPrompt -> prompt ++ "\n" ++ (darkPrompt |> String.fromBytes)
+        | Error e ->
+          print e
+          prompt
+
+      print prompt
+
+      let request =
```
pbiggar, line 260: I would call this `response`

PR #5033, opened 2023-08-22
Allow package functions to be used in User DB queries

OceanOak:
Changelog:

```
Language and Standard Library
- Allow package functions to be used in queries
```


backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -258,12 +258,21 @@ let (|Fn|_|) (mName : string) (fName : string) (v : int) (expr : Expr) =
 type CompiledSqlQuery =
   { sql : string; vars : List<string * SqlValue>; actualType : TypeReference }
 
+
+let sqlOpForPackageFunction (fnName : FnName.Package) : SqlSpec =
```
StachuDotNet, line 262: I think this turned out well!

OceanOak: Yes seems better. Thanks for the improvement suggestions!

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -303,97 +312,204 @@ let rec lambdaToSql
...
-                match Map.get name actualTypes with
-                | Some typ -> TList typ
-                | None -> error "Could not find return type"
-              | typ -> typ
-
-            typecheck nameStr returnType expectedType
+          match! Function.find fnName fns with
+          | Some fn ->
+            match fn with
+            | FQName.BuiltIn b ->
```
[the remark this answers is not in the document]
pbiggar: This logic looks funny to me. We should know from `fnName` whether it's a package or builtin.

backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -1209,17 +1209,48 @@ and ExecutionState =
...
+  let availableFunctions (state : ExecutionState) : Functions =
+    { builtIn = state.builtIns.fns
+      package = state.packageManager.getFn
+      userProgram = state.program.fns }
+
+module Function =
+  let empty =
+    { builtIn = Map.empty; package = (fun _ -> Ply None); userProgram = Map.empty }
+
+  let find (name : FnName.T) (functions : Functions) : Ply<Option<FnName.T>> =
```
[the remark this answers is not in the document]
pbiggar: Actually this is weird. We look up the function using the name, and then rebuild the name from the result?

pbiggar (changes requested):
A nice start, thanks! With some refactoring, we should be able to get this merged in its current state, and we can do inlining and package constants later.

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -289,111 +298,222 @@ let rec lambdaToSql
...
+              let random = randomString 8
+              let newname = $"{nameStr}_{random}"
+              let (sqlValue, actualType) = dvalToSql expectedType c.body
+              return ($"(@{newname})", [ newname, sqlValue ], actualType)
+            | None -> return error $"No built-in constant {nameStr} found"
+
+          | FQName.Package p ->
+            let! constants = constants.package p
+            match constants with
+            | Some c ->
```
OceanOak, line 318: Do we need type checking here? Here's a package constant example:
```
c ({ tlid = 752978710UL
  name = { owner = "Darklang"
           modules = ["Test"; "Constants"]
           name = ConstantName "boolConst"
           version = 0 }
  body = DBool true }) 
   ```

Unlike built-in constants where we used c.typ, this field isn't viable for package constants. And, I don't think we can get a TypeReference from the constant body using PT2DT.TypeRefrence.fromDT in this case, to use it in the typechecker

PR #5035, opened 2023-08-24
Call OpenAI from the cli - follow-ups

OceanOak:
No changelog

This PR tackles the feedback received on #5032.

### DIFF
--- a/packages/openai.dark
+++ b/packages/openai.dark
@@ -194,8 +194,9 @@ module OpenAI =
       | Ok reqBody ->
         let headers = PACKAGE.OpenAI.Config.getHeaders ()
 
+        //CLEANUP: Use HttpClient.request instead of CLI.HttpClient.request once we have the httpclient sorted out
         let openAIResponse =
-          HttpClient.request
+          CLI.HttpClient.request
             "POST"
             "https://api.openai.com/v1/chat/completions"
             headers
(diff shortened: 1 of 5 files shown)


packages/openai.dark
```diff
@@ -194,8 +194,9 @@ module OpenAI =
       | Ok reqBody ->
         let headers = PACKAGE.OpenAI.Config.getHeaders ()
 
+        //CLEANUP: Use HttpClient.request instead of CLI.HttpClient.request once we have the httpclient sorted out
         let openAIResponse =
-          HttpClient.request
+          CLI.HttpClient.request
```
OceanOak, line 199: None of the following solution I've tried seemed good.
- When moving the endpoint to the dark-editor canvas, we have to load the canvas each time, and we get an`OpenAI API request failed: Request timed out` error most of the times.

- The renaming of `CLI.HttpClient.request` to `HttpClient.request` resulted in the failure of 62 tests. I wasn't sure about commenting all of these out.

The solution I opted for involved using the `CLI.HttpClient.request` function within the `completion` OpenAI package function. This approach ensures that we avoid encountering the `HttpClient.request not found` error.

wdyt?

pbiggar: I just merged a fix so you can just use `HttpClient.request`. Let me know if that works for you?

pbiggar (changes requested):
Looks good after the HttpClient.request change (assuming it works)

packages/openai.dark
OceanOak: it seems to be working well, thanks!

PR #5045, opened 2023-08-30
Stop fetching all package items

StachuDotNet:
Changelog:

```
Area of change
- details
```

This updates our F# `RT.PackageManager` impls -- both the `LibCloud` one (which accesses our internal `devdb` sql DB) and the `LibCli` one (which fetches data from the `dark-packages` canvas):
- caching of package stuff is temporarily abandoned (will rework caching back in shortly)
- loads package items one-by-one by name, rather than all at once and `List.filter`-ing after the fact
  (this combined with caching has been causing all sorts of issues lately)
- takes ~40s locally to `reload-packages`
  while terrible, this seems to be comparable or better to the timing for `main`
- takes ~20s locally to `run-cli --help`
  while terrible, we don't really do this often
- takes 15s locally to run `run-backend-tests`
  (I think this is a lot faster and more consistent than `main`)
- is stable. namely, more stable than `main` seems to be, anyway.

The `dark-packages` canvas got new endpoints like `/type/by-name/{name}`, to support the web-based PM.
I previously intended to wait on some work in the SqlCompiler to be complete to tackle this, but it turns out that `DB.query` is so darn slow (in my experience tonight anyway) that it's not worth waiting for. So, I did some `DB.getAll`s followed by `List.filter`s in Dark code - ironic? ( -- Room for caching here too!)

---

I'd like to merge this soon, and follow up tomorrow:
- re-impl caching in the F# PMs -- it should be safer now
- breaking those timings down a bit (i.e. it seems the time is spent pretty evenly across the CLI side and the BWD side)
- add some kinda caching in `dark-packages`

P.S.
- I've been really feeling like we should tackle parallelization in things like `List.map` soon -- stop using `List.mapSequentially`?
- this deletes the BuiltIn fn `DarkInternal.Packages.all` -- I think this makes sense, and we should reimpl in package space sometime


pbiggar:
> I've been really feeling like we should tackle parallelization in things like List.map soon -- stop using List.mapSequentially?

Yeah, in my head, we can do this once we remove DError.

StachuDotNet:
Sorry - I'd meant to make this Ready for Review after updating the description.

StachuDotNet:
> fyi facing some issue with this after rebase

I think this is related to the if/else changes recently

pbiggar (approved):

packages/darklang/stdlib/string.dark
```diff
@@ -8,10 +8,11 @@ module Darklang =
...
       let ellipsis (s: String) (length: Int) : String =
-        if String.length s > length then
-          (String.slice_v0 s 0 length) ++ "..."
-        else
-          s
+        // if String.length s > length then
+        //   (String.slice_v0 s 0 length) ++ "..."
+        // else
+        //   s
+        "..."
```
StachuDotNet, line 15: whoops, reverting

PR #5047, opened 2023-08-30
Move all Tuple Built-in functions to packages and some of the Int, Float and Char functions

OceanOak:
Changelog:

```
Area of change
- Move all Tuple2 and Tuple3 Built-in functions to packages
- Move some of the Int, Float and Char built-in functions to packages
```


pbiggar:
Nice!

packages/darklang/stdlib/tuple2.dark
```diff
@@ -0,0 +1,45 @@
...
+
+
+      /// Transform both values in a pair
+      let mapBoth
+        (fnFirst: 'a -> 'c)
+        (fnSecond: 'b -> 'd)
+        (tuple: Tuple<'a, 'b>)
+        : Tuple<'c, 'd> =
+        let (first, second) = tuple
+        ((first |> fnFirst), (second |> fnSecond))
```
StachuDotNet, line 45: this feels like over-use of pipes to me

compare:
- `(fnFirst first, fnSecond second)`
- `((first |> fnFirst), (second |> fnSecond))`

OceanOak: I think we can't do this yet, waiting for EApply changes

StachuDotNet: ah, you're right

PR #5055, opened 2023-08-31
Fix Process.run builtin and package fn; add File.delete builtin; fix dark-packages ingest

StachuDotNet:
- fix some regressions preventing packages getting to the dark-packages canvas locally
- fix `Process.run` function -- I was unable to use the Process.run as-is
- adds a `File.delete` in StdLibCli, as I reached for it when writing a personal script

### DIFF
--- a/packages/darklang/stdlib/process.dark
+++ b/packages/darklang/stdlib/process.dark
@@ -5,17 +5,12 @@ module Darklang =
       /// The result of running a process, typically with `Process.run`
       type Result =
         { exitCode: Int
-          stdout: Bytes
-          stderr: Bytes }
 
-      // Return Ok stdout+stderr, or Error exitCode+stdout+stderr
-      let runCommand
-        (command: String)
-        (input: String)
-        : PACKAGE.Darklang.Stdlib.Result.Result<String, Result> =
-        let result = Builtin.Process.run command input
+          // TODO bytes, not strings -- see internal note in Process.fs
+          stdout: String
+          stderr: String }
 
-        if result.exitCode == 0 then
-          PACKAGE.Darklang.Stdlib.Result.Result.Ok(stdout ++ stderr)
-        else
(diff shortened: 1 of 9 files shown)


StachuDotNet:
Paul - what do you think about:
- adding the `File.delete`?
- having OS-/environment-specific BuiltIns for things like `Process.run` rather than try to have one fn across envs?

pbiggar:
> adding the File.delete?
yes

> having OS-/environment-specific BuiltIns for things like Process.run rather than try to have one fn across envs?
Dislike a lot

pbiggar (approved):
I dislike the linux specific stuff, but maybe let's iterate and see how it goes.

StachuDotNet:
I removed the linux-specific `Process.Linux.run` _function_ but it seems we do need some level of linux-specific stuff in our impl - seems like a good compromise to me. Please let me know if this is OK or if I should revert in a subsequent PR. (I've tested only on linux at this point, but according to online sources this should work OK on mac+windows as well.)

packages/darklang/stdlib/process.dark
```diff
@@ -5,17 +5,12 @@ module Darklang =
...
-        : PACKAGE.Darklang.Stdlib.Result.Result<String, Result> =
-        let result = Builtin.Process.run command input
+          // TODO bytes, not strings -- see internal note in Process.fs
+          stdout: String
+          stderr: String }
 
-        if result.exitCode == 0 then
-          PACKAGE.Darklang.Stdlib.Result.Result.Ok(stdout ++ stderr)
-        else
-          PACKAGE.Darklang.Stdlib.Result.Result.Error result
```
StachuDotNet, line 21: This was trying to be a bit fancy, but didn't work (and _won't work_ even more once stdout and stderr are Bytes again) and I suspect there's a better way to handle this. Made it less fancy, and the package fn is a simple pass-through now

PR #5043, opened 2023-08-29
Move stdlib fns

OceanOak:
Changelog:

```
Area of change
- details
```

alternatively:

No changelog

## What are the changes here? How do they solve the problem and what other product impacts do they cause?
*Please include before/after screenshots/gifs if appropriate*

PR #5058, opened 2023-09-04
Move Some List and String builtins to packages

OceanOak:
No changelog

This PR implements the requested changes in #5043


backend/testfiles/execution/language/elambda.dark
```diff
@@ -2,43 +2,52 @@ PACKAGE.Darklang.Stdlib.List.push_v0 [] (fun x -> -4.611686018e+18) = [ (fun x -
...
 
-(let y = (fun x -> x + 1) in Builtin.List.map_v0 [ 1; 2; 3; 4 ] y) = [ 2; 3; 4; 5 ]
+(let y = (fun x -> x + 1) in PACKAGE.Darklang.Stdlib.List.map_v0 [ 1; 2; 3; 4 ] y) = [ 2
+                                                                                       3
+                                                                                       4
+                                                                                       5 ]
 
 (let y = (fun msg -> Builtin.Test.runtimeError msg) in
- Builtin.List.map_v0 [ "1"; "2"; "3"; "4" ] y) = Builtin.Test.derrorMessage "1"
+ PACKAGE.Darklang.Stdlib.List.map_v0 [ "1"; "2"; "3"; "4" ] y) = Builtin.Test.derrorMessage
```
OceanOak, line 16: re-check this test

OceanOak:
> I think the right way to approach it is this:
> 
> use a builtin for slice, as that can be made pretty fast (1 slow thing)
> if you can, see if you can convert functions into slice. For example, trimFront could be implemented by finding the index of first non-whitespace element, and then calling slice
> if you need to do something to every element, converting to a list and iterating is the right approach

I tried to follow this approach for String.trimStart and String.trimEnd, but I couldn't find a way to do it without using loops or turning the string into a list and use list functions.

I've tried several approaches including 
- Using String.map, adding a String.charAt (I got errors like `Outer exception: error: The surrogate pair (0xD83E, 0x1B) is invalid. A high surrogate character (0xD800 - 0xDBFF) must always be paired with a low surrogate character (0xDC00 - 0xDFFF).`)

The only approach that seemed to work is the following:

**a helper function** 

```
      let findFirstNonWhitespace (idx: Int) (str: String) : Int =
          if (Builtin.Int.greaterThan_v0 idx (Builtin.String.length str)) then
            -1
          else
            let charAtIdx = Builtin.String.slice_v0 str idx (idx + 1)
            match charAtIdx with
            | " " -> PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace (idx + 1) str
            | "\t" -> PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace (idx + 1) str
            | "\n" -> PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace (idx + 1) str
            | "\u2003" -> PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace (idx + 1) str
            | _ -> idx

```   
    
**String.trimStart and String.trimEnd**

    
```
          let trimStart (str: String) : String =
            let firstNonWhitespace = PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace 0 str
            if firstNonWhitespace == -1 then
              str
            else
              Builtin.String.slice_v0 str firstNonWhitespace (Builtin.String.length str)
            
            
        let trimEnd (str: String) : String =
          let reversedString = Builtin.String.reverse_v0 str
          (PACKAGE.Darklang.Stdlib.String.trimStart reversedString) |> Builtin.String.reverse_v0
```

Only one tests failed using this approach : 


I couldn't refactor the helper function since we don't support the following in dark 
```
          match charAtIdx with
          | " "
          | "\t"
          | "\n"
          | "\u2003" ->
            PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace (idx + 1) str
          | _ -> idx
         
```

or 

```
          match charAtIdx with
          | charAtIdx when Char.isWhitespace ...
          | _ -> idx
```

OceanOak:
> I think the right way to approach it is this:
> 
> use a builtin for slice, as that can be made pretty fast (1 slow thing)
> if you can, see if you can convert functions into slice. For example, trimFront could be implemented by finding the index of first non-whitespace element, and then calling slice
> if you need to do something to every element, converting to a list and iterating is the right approach

I tried to follow this approach for String.trimStart and String.trimEnd, but I couldn't find a way to do it without using loops or turning the string into a list and use list functions.

These are three solutions that have worked, but one of the tests failed using these approaches:
`PACKAGE.Darklang.Stdlib.String.trimEnd_v0 "🇺🇸🇷🇺🇸 🇦🇫🇦🇲🇸" = "🇺🇸🇷🇺🇸 🇦🇫🇦🇲🇸"`

### Approach 1 :
**a helper function** 

```
      let findFirstNonWhitespace (idx: Int) (str: String) : Int =
          if (Builtin.Int.greaterThan_v0 idx (Builtin.String.length str)) then
            -1
          else
            let charAtIdx = Builtin.String.slice_v0 str idx (idx + 1)
            match charAtIdx with
            | " " -> PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace (idx + 1) str
            | "\t" -> PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace (idx + 1) str
            | "\n" -> PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace (idx + 1) str
            | "\u2003" -> PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace (idx + 1) str
            | _ -> idx

```   
    
**String.trimStart and String.trimEnd**

    
```
          let trimStart (str: String) : String =
            let firstNonWhitespace = PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace 0 str
            if firstNonWhitespace == -1 then
              str
            else
              Builtin.String.slice_v0 str firstNonWhitespace (Builtin.String.length str)
            
            
        let trimEnd (str: String) : String =
          let reversedString = Builtin.String.reverse_v0 str
          (PACKAGE.Darklang.Stdlib.String.trimStart reversedString) |> Builtin.String.reverse_v0
```

### Approach 2: 
```
      let trimStart (str: String) : String =
        if
          (PACKAGE.Darklang.Stdlib.String.startsWith str " ")
          || (PACKAGE.Darklang.Stdlib.String.startsWith str "\t")
          || (PACKAGE.Darklang.Stdlib.String.startsWith str "\n")
          || (PACKAGE.Darklang.Stdlib.String.startsWith str "\u2003")
        then
          PACKAGE.Darklang.Stdlib.String.trimStart (
            PACKAGE.Darklang.Stdlib.String.dropFirst str 1
          )
        else
          str
```

### Approach 3: a better version of approach 2

```
let trimStart (str: String) : String =
        if str == "" then
          ""
        else
          if
            (Builtin.Char.isWhitespace (
              match (Builtin.String.head str) with
              | Some c -> c
              | None -> ' '
            ))
          then
            PACKAGE.Darklang.Stdlib.String.trimStart (
              PACKAGE.Darklang.Stdlib.String.dropFirst str 1
            )
          else
            str
           
   //Char.fs
   { name = fn "isWhitespace" 0
      typeParams = []
      parameters = [ Param.make "c" TChar "" ]
      returnType = TBool
      description = "Return whether <param c> is a whitespace character"
      fn =
        (function
        | _, _, [ DChar c ] when c <> null ->
          Ply(DBool(System.Char.IsWhiteSpace(c[0])))
        | _ -> incorrectArgs ())
      sqlSpec = NotQueryable
      previewable = Pure
      deprecated = NotDeprecated }
```

Feedback is appreciated

backend/testfiles/execution/language/elambda.dark
pbiggar: Yeah, your implementation does the last items first, but i think the previous implementation did the first items first.

I think it would make more sense for it to do the first elements of the list first, but there may be reasons that isn't feasible?

packages/darklang/stdlib/list.dark
```diff
@@ -151,6 +162,247 @@ module Darklang =
...
+          PACKAGE.Darklang.Stdlib.List.push_v0
+            (PACKAGE.Darklang.Stdlib.List.map tail fn)
+            (fn head)
+
+
+      // Calls <param f> on every <var val> in <param list>, returning a list of only
+      // those values for which {{fn val}} returns {{true}}.
+      // Preserves the order of values that were not dropped. Consider <fn
+      // List.filterMap> if you also want to transform the values.
+      let filter (list: List<'a>) (f: 'a -> Bool) : List<'a> =
```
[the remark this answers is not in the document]
pbiggar: This also relates to the ordering of List.map.

pbiggar:
> Approach 1 :
> a helper function

This creates a full copy of the string for each whitespace character, in findFirstNonWhitespace. This is fine for 1 character but does duplicate work for more, and could be an N^2 implementation (eg for a string that's entirely whitespace).

```
          let trimStart (str: String) : String =
            let firstNonWhitespace = PACKAGE.Darklang.Stdlib.String.findFirstNonWhitespace 0 str
            if firstNonWhitespace == -1 then
              str
            else
              Builtin.String.slice_v0 str firstNonWhitespace (Builtin.String.length str)
```

This looks good.
            
```
        let trimEnd (str: String) : String =
          let reversedString = Builtin.String.reverse_v0 str
          (PACKAGE.Darklang.Stdlib.String.trimStart reversedString) |> Builtin.String.reverse_v0
```

This is good except for the reversed bit. We could have a `lastNonWhitespace` that does the same thing but goes from the back.

First non-whitespace could be done with zero copies, using something like `String.findFirstIndex : (Char -> Bool) -> String -> Int`. `Char.IsWhitespace` from approach 3 would work nicely for it. However, the approaches you've used here are all recursive, and recursive approaches don't work well for strings as they involve breaking the whole string (and thus copying it).

Strings instead have O(1) random access, meaning you can access any element of the string immediately using the index (a property lists don't have). So you can iterate over the string using just the integer.

pbiggar (commented):
Looks good! Only a few comments apart from String.trimLeft (might be easier to do that in a followup PR).

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -263,6 +263,10 @@ let sqlOpForPackageFunction (fnName : FnName.Package) : SqlSpec =
   | "Darklang", [ "Stdlib"; "Bool" ], FnName.FnName "and" -> SqlBinOp "AND"
   | "Darklang", [ "Stdlib"; "Bool" ], FnName.FnName "or" -> SqlBinOp "OR"
   | "Darklang", [ "Stdlib"; "Bool" ], FnName.FnName "not" -> SqlFunction "NOT"
+  | "Darklang", [ "Stdlib"; "String" ], FnName.FnName "contains" ->
+    SqlCallback2(fun lookingIn searchingFor ->
+      // strpos returns indexed from 1; 0 means missing
+      $"strpos({lookingIn}, {searchingFor}) > 0")
```
StachuDotNet, line 269: Neat - this taught me the `strpos()` fn, thanks!

packages/darklang/stdlib/list.dark
```diff
@@ -151,6 +162,247 @@ module Darklang =
...
+
+      ///Returns a list containing <param val> repeated <param times> times
+      let repeat
+        (times: Int)
+        (value: 'a)
+        : PACKAGE.Darklang.Stdlib.Result.Result<List<'a>, String> =
+        if times < 0 then
+          PACKAGE.Darklang.Stdlib.Result.Result.Error
+            $"Expected `times` to be positive, but it was `{times |> Builtin.Int.toString}`"
+        else if times > 2147483647 then
```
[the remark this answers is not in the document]
OceanOak: if we remove it the following test will raise an exception 
```
PACKAGE.Darklang.Stdlib.List.repeat_v0 11844674407356L 1 = PACKAGE.Darklang.Stdlib.Result.Result.Error
  "Expected `times` to be less than 2147483647, but it was `11844674407356`"
```

Exception: 
```
at Ply.TplPrimitives+ContinuationStateMachine`1[[System.__Canon, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=...
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(System.Threading.Thread, System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1+AsyncStateMachineBox`1[[System.__Canon, System.Private.CoreLib, ...
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool+WorkerThread.WorkerThreadStart()
./scripts/run-backend-tests: line 117: 64030 Aborted      

DARK_CONFIG_TELEMETRY_EXPORTER=none DARK_CONFIG_ROLLBAR_ENABLED=n DARK_CONFIG_DB_HOST=localhost DARK_CONFIG_DB_DBNAME=..." "${EXPECTO_ARGS[@]}"
```
Unless we want to delete that test, and we don't care about that edge case?

backend/testfiles/execution/stdlib/int.dark
```diff
@@ -44,31 +44,55 @@ Builtin.Int.remainder_v0 -15 6 = PACKAGE.Darklang.Stdlib.Result.Result.Ok -3
...
-                                    PACKAGE.Darklang.Stdlib.Result.Result.Ok 0
-                                    PACKAGE.Darklang.Stdlib.Result.Result.Ok 1
-                                    PACKAGE.Darklang.Stdlib.Result.Result.Ok 2
-                                    PACKAGE.Darklang.Stdlib.Result.Result.Ok 3
-                                    PACKAGE.Darklang.Stdlib.Result.Result.Ok 0
-                                    PACKAGE.Darklang.Stdlib.Result.Result.Ok 1 ]
+PACKAGE.Darklang.Stdlib.List.map_v0
+  (PACKAGE.Darklang.Stdlib.List.range_v0 -5 5)
+  (fun v -> Builtin.Int.remainder_v0 v -4) = [ PACKAGE.Darklang.Stdlib.Result.Result.Ok
+                                                 -1
```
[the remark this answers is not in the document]
OceanOak: I think it is fine (annoying but fine), is there something specific that appears odd to you?

packages/darklang/stdlib/list.dark
```diff
@@ -193,28 +193,22 @@ module Darklang =
...
       // List.filterMap> if you also want to transform the values.
       let filter (list: List<'a>) (f: 'a -> Bool) : List<'a> =
-        match list with
-        | [] -> []
-        | head :: tail ->
-          if (head |> f) then
-            PACKAGE.Darklang.Stdlib.List.push_v0
-              (PACKAGE.Darklang.Stdlib.List.filter tail f)
-              head
+        (PACKAGE.Darklang.Stdlib.List.fold list [] (fun acc elem ->
```
OceanOak, line 207: I believe this implementation is more performant than the previous one. wdyt?

backend/testfiles/execution/language/elambda.dark
OceanOak: This was only happening with that error case; I didn't have to change the results of the other tests. Updating List.map to use List.fold solved this issue

packages/darklang/stdlib/list.dark
```diff
@@ -151,6 +162,247 @@ module Darklang =
...
+          PACKAGE.Darklang.Stdlib.Result.Result.Error
+            $"Expected `times` to be positive, but it was `{times |> Builtin.Int.toString}`"
+        else if times > 2147483647 then
+          PACKAGE.Darklang.Stdlib.Result.Result.Error
+            $"Expected `times` to be less than 2147483647, but it was `{times |> Builtin.Int.toString}`"
+        else if times == 0 then
+          PACKAGE.Darklang.Stdlib.Result.Result.Ok []
+        else
+          match PACKAGE.Darklang.Stdlib.List.repeat (times - 1) value with
+          | Ok list ->
```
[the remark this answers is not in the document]
OceanOak: I couldn't find a way of doing it without wrapping and unwrapping

packages/darklang/stdlib/list.dark
OceanOak: Is it worth renaming our List.fold to List.foldRight and add List.foldLeft?

pbiggar: Maybe - is that what they are? I forget the difference between foldLeft and foldRight. I wouldn't mind having fold be a synonym of whichever one is appropriate as well.

backend/testfiles/execution/stdlib/int.dark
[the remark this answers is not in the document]
pbiggar: Turns out this was just how github was displaying it to me (it looked like there was a blank line between each list element).

packages/darklang/stdlib/list.dark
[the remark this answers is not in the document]
pbiggar: We can delete that test. Eventually we run out of memory, not a lot can be done about that, right?

packages/darklang/stdlib/list.dark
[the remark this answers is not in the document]
pbiggar: Make one version that returns a list (not a result), and assumes valid inputs. Call that from an outer function that checks the inputs and uses Results.

packages/darklang/stdlib/list.dark
pbiggar: > I believe this implementation is more performant than the previous one. wdyt?

Honestly no idea.

pbiggar:
Looking good. There are I think two remaining changes, maybe more? It would make sense to merge now and do those as a followup.

PR #5061, opened 2023-09-06
Move Some List and String builtins to packages - followup

OceanOak:
No changelog

Follow-up #5058

### DIFF
--- a/packages/darklang/stdlib/list.dark
+++ b/packages/darklang/stdlib/list.dark
@@ -68,7 +68,6 @@ module Darklang =
             PACKAGE.Darklang.Stdlib.List.findFirst tail fn
 
 
-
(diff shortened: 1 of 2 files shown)


packages/darklang/stdlib/list.dark
```diff
@@ -80,6 +80,13 @@ module Darklang =
             PACKAGE.Darklang.Stdlib.List.``member`` tail value
 
 
+      let repeatUnsafe (times: Int) (value: 'a) : List<'a> =
```
pbiggar, line 83: Can this be a lambda inside `repeat`? That way it's not accessible to other functions/users

pbiggar (changes requested):
One small question.

PR #5062, opened 2023-09-06
Wrap string builtin fns

OceanOak:
No changelog

This creates wrapper functions around Builtin String functions

### DIFF
--- a/packages/darklang/stdlib/string.dark
+++ b/packages/darklang/stdlib/string.dark
@@ -120,9 +120,13 @@ module Darklang =
       /// sections](https://www.rfc-editor.org/rfc/rfc4648.html)
       /// [4](https://www.rfc-editor.org/rfc/rfc4648.html#section-4) and
       /// [5](https://www.rfc-editor.org/rfc/rfc4648.html#section-5).
-      let base64Decode (s: String) : PACKAGE.Darklang.Stdlib.Result.Result<String, String> =
+      let base64Decode
+        (s: String)
+        : PACKAGE.Darklang.Stdlib.Result.Result<String, String> =
         match (Builtin.Bytes.base64Decode s) with
-        | Ok bytes -> (Builtin.String.fromBytes bytes) |> PACKAGE.Darklang.Stdlib.Result.Result.Ok
+        | Ok bytes ->
+          (Builtin.String.fromBytes bytes)
+          |> PACKAGE.Darklang.Stdlib.Result.Result.Ok
         | Error err -> PACKAGE.Darklang.Stdlib.Result.Result.Error err
 
 
@@ -159,17 +163,21 @@ module Darklang =
 
       /// Checks if <param lookingIn> contains <param searchingFor>
       let contains (lookingIn: String) (searchingFor: String) : Bool =
-        PACKAGE.Darklang.Stdlib.Option.isSome (Builtin.String.indexOf lookingIn searchingFor)
+        PACKAGE.Darklang.Stdlib.Option.isSome (
+          Builtin.String.indexOf lookingIn searchingFor
+        )
 
 
       /// Returns the substring of <param string> between the <param from> and <param to> indices.
       /// Negative indices start counting from the end of <param string>.
       let slice (string: String) (from: Int) (``to``: Int) : String =
         let len = Builtin.String.length string
-        let normalize =  (fun i ->
-          (if i < 0 then len + i else i)
-          |> PACKAGE.Darklang.Stdlib.Int.min len
-          |> PACKAGE.Darklang.Stdlib.Int.max 0)
+
+        let normalize =
+          (fun i ->
+            (if i < 0 then len + i else i)
+            |> PACKAGE.Darklang.Stdlib.Int.min len
+            |> PACKAGE.Darklang.Stdlib.Int.max 0)
 
         let f = normalize from
         let l = normalize ``to``
@@ -334,7 +342,9 @@ module Darklang =
         if count <= 0 then
           ""
         else
-          Builtin.String.join ((PACKAGE.Darklang.Stdlib.List.repeat count string) |> Builtin.unwrap) ""
+          Builtin.String.join
+            ((PACKAGE.Darklang.Stdlib.List.repeat count string) |> Builtin.unwrap)
+            ""
 
 
       /// Converts an integer <param n> to its ordinal representation as a string.
--- a/packages/darklang/stdlib/string.dark
+++ b/packages/darklang/stdlib/string.dark
@@ -40,9 +40,59 @@ module Darklang =
       let isEmpty (s: String) : Bool = s == ""
 
 
+      /// Iterate over each Character (EGC, not byte) in the string, performing the
+      /// operation in <param fn> on each one.
+      let map (s: String) (fn: Char -> Char) : String = Builtin.String.map s fn
+
+
+      /// Returns the list of Characters (EGC, not byte) in the string
+      let toList (s: String) : List<Char> = Builtin.String.toList s
+
+
+      /// Replace all instances on <param searchFor> in <param s> with <param replaceWith>
+      let replaceAll (s: String) (searchFor: String) (replaceWith: String) : String =
+        Builtin.String.replaceAll s searchFor replaceWith
+
+
+      /// Returns the string, uppercased
+      let toUppercase (s: String) : String = Builtin.String.toUppercase s
+
+
+      /// Returns the string, lowercased
+      let toLowercase (s: String) : String = Builtin.String.toLowercase s
+
+
+      /// Returns the length of the string
+      let length (s: String) : Int = Builtin.String.length s
+
+
+      /// Concatenates the two strings by appending <param s2> to <param s1> and returns the joined string.
+      let append (s1: String) (s2: String) : String = $"{s1}{s2}"
+
+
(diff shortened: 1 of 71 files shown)


packages/darklang/stdlib/string.dark
```diff
@@ -40,11 +40,61 @@ module Darklang =
...
+
+
       /// Concatenates the two strings by prepending <param s2> to <param s1> and
       /// returns the joined string
       let prepend (s1: String) (s2: String) : String = s2 ++ s1
 
 
+      /// Turns a string into a prettified slug, including only lowercased
+      /// alphanumeric characters, joined by hyphens
+      let slugify (s: String) : String = Builtin.String.slugify s
```
pbiggar, line 80: I'd rather this be deleted than builtin

pbiggar: Can it be implemented without regex? If not, maybe we comment it out until we have regex?

packages/darklang/stdlib/string.dark
```diff
@@ -56,6 +106,68 @@ module Darklang =
       let fromChar (c: Char) : String = PACKAGE.Darklang.Stdlib.Char.toString c
 
 
+      /// URLBase64 encodes a string without padding. Uses URL-safe encoding with
+      /// {{-}} and {{_}} instead of {{+}} and {{/}}, as defined in
+      /// [RFC 4648 section 5](https://www.rfc-editor.org/rfc/rfc4648.html#section-5)
+      let base64Encode (s: String) : String = Builtin.String.base64Encode s
```
pbiggar, line 112: I think this could be done using Bytes.base64encode?

packages/darklang/stdlib/string.dark
```diff
@@ -56,6 +106,68 @@ module Darklang =
...
+      /// Base64 decodes a string. The returned value is wrapped in {{Result}}.
+      /// Works with both the URL-safe and standard Base64
+      /// alphabets defined in [RFC 4648
+      /// sections](https://www.rfc-editor.org/rfc/rfc4648.html)
+      /// [4](https://www.rfc-editor.org/rfc/rfc4648.html#section-4) and
+      /// [5](https://www.rfc-editor.org/rfc/rfc4648.html#section-5).
+      let base64Decode
+        (s: String)
+        : PACKAGE.Darklang.Stdlib.Result.Result<String, String> =
+        Builtin.String.base64Decode s
```
pbiggar, line 124: I think this could be done using Bytes.base64Decode?

packages/darklang/stdlib/string.dark
```diff
@@ -56,6 +106,68 @@ module Darklang =
...
+
+
+      /// Take a string and hash it to a cryptographically-secure digest.
+      /// Don't rely on either the size or the algorithm.
+      let digest (s: String) : String = Builtin.String.digest s
```
pbiggar, line 129: This feels a little off to have as a builtin. Can it be done using Bytes or directly call the hashing function?

packages/darklang/stdlib/string.dark
```diff
@@ -56,6 +106,68 @@ module Darklang =
...
+
+
+      /// Generate a <type String> of length <param length> from random characters
+      let random
+        (length: Int)
+        : PACKAGE.Darklang.Stdlib.Result.Result<String, String> =
+        Builtin.String.random length
```
pbiggar, line 136: I think this could be implemented in dark?

packages/darklang/stdlib/string.dark
```diff
@@ -56,6 +106,68 @@ module Darklang =
...
+          | '"' -> "&quot;"
+          // &apos; doesn't work in IE....
+          | ''' -> "&#x27;"
+          | _ -> PACKAGE.Darklang.Stdlib.String.fromChar c)
+        |> Builtin.String.join ""
+
+
+      /// Checks if <param lookingIn> contains <param searchingFor>
+      let contains (lookingIn: String) (searchingFor: String) : Bool =
+        match Builtin.String.indexOf lookingIn searchingFor with
```
pbiggar, line 160: Could we call `Option.isSome` here? (doesn't exist but should be added)

packages/darklang/stdlib/string.dark
```diff
@@ -56,6 +106,68 @@ module Darklang =
...
+        | Some _ -> true
+        | None -> false
+
+
+      /// Returns the substring of <param string> between the <param from> and <param to> indices.
+      /// Negative indices start counting from the end of <param string>.
+      let slice (string: String) (from: Int) (``to``: Int) : String =
+        Builtin.String.slice_v0 string from ``to``
```
pbiggar, line 168: It feels like the negative index stuff should be done directly here, vs in the builtin function.

packages/darklang/stdlib/string.dark
```diff
@@ -176,6 +309,23 @@ module Darklang =
...
+        Builtin.String.indexOf str searchFor
+
+
+      /// Returns the <param string> repeated for a specified <param count> times.
+      /// If <param count> is less than or equal to 0, returns an empty string.
+      let repeat (string: String) (count: Int) : String =
+        if count <= 0 then
+          ""
+        else
+          string ++ PACKAGE.Darklang.Stdlib.String.repeat string (count - 1)
```
pbiggar, line 326: This is an O(N^2) function. You can make it O(N) by using a list and calling String.join.

pbiggar (changes requested):
This is looking good. I'm open to merging now and doing updates in a later PR - let me know what you prefer.

packages/darklang/stdlib/string.dark
```diff
@@ -157,15 +159,24 @@ module Darklang =
...
+        let normalize =  (fun i ->
+          (if i < 0 then len + i else i)
+          |> PACKAGE.Darklang.Stdlib.Int.min len
+          |> PACKAGE.Darklang.Stdlib.Int.max 0)
+
+        let f = normalize from
+        let l = normalize ``to``
+        let l = if f > l then f else l
+
+        (PACKAGE.Darklang.Stdlib.String.dropLast string (len - l))
```
OceanOak, line 186: TODO: find a different way to implement slice, dropLast uses slice (infinite recursion)

packages/darklang/stdlib/string.dark
```diff
@@ -206,8 +208,11 @@ module Darklang =
         let l = normalize ``to``
         let l = if f > l then f else l
 
-        (PACKAGE.Darklang.Stdlib.String.dropLast string (len - l))
-        |> PACKAGE.Darklang.Stdlib.String.dropFirst f
+        string
+        |> String.toList
```
OceanOak, line 212: This isn't a good way to implement slice in dark (converting it to a list...)

I had this implementation, but dropLast and dropFirst use slice (infinite recursion)
```
(PACKAGE.Darklang.Stdlib.String.dropLast string (len - l))
  |> PACKAGE.Darklang.Stdlib.String.dropFirst f
 ```
 Any suggestions for implementing it without converting it into a list?

File: packages/darklang/stdlib/string.dark [hunk 11]
```diff
@@ -118,10 +120,10 @@ module Darklang =
       /// sections](https://www.rfc-editor.org/rfc/rfc4648.html)
       /// [4](https://www.rfc-editor.org/rfc/rfc4648.html#section-4) and
       /// [5](https://www.rfc-editor.org/rfc/rfc4648.html#section-5).
-      let base64Decode
-        (s: String)
-        : PACKAGE.Darklang.Stdlib.Result.Result<String, String> =
-        Builtin.String.base64Decode s
+      let base64Decode (s: String) : PACKAGE.Darklang.Stdlib.Result.Result<String, String> =
+        match (Builtin.Bytes.base64Decode s) with
+        | Ok bytes -> (Builtin.String.fromBytes bytes) |> PACKAGE.Darklang.Stdlib.Result.Result.Ok
```
pbiggar: I see here that String.fromBytes ignores errors. If we're putting something in a result I feel we should not do that.

Perhaps we need to add two kinds of String.fromBytes (like we do in Prelude.UTF) so that we don't always just ignore errors.

packages/darklang/stdlib/string.dark
pbiggar: I meant that PACKAGE.slice should call Builtin.slice, which does the efficient string slicing. PACKAGE.slice does the normalization, and Builtin.slice just does the bits we can only do in dotnet.

packages/darklang/stdlib/string.dark
```diff
@@ -194,31 +397,6 @@ module Darklang =
...
-
-      /// Checks if <param lookingIn> contains <param searchingFor>
-      let contains (lookingIn: String) (searchingFor: String) : Bool =
-        match Builtin.String.indexOf lookingIn searchingFor with
-        | Some _ -> true
-        | None -> false
\ No newline at end of file
+      /// Returns {{Some char}} of the first character of <param str>, or returns {{None}} if <param str> is empty.
+      let head (str: String) : PACKAGE.Darklang.Stdlib.Option.Option<Char> =
+        Builtin.String.head str
```
pbiggar, line 402: This could be done in Dark

pbiggar (changes requested):
Looking good, a few small things left.

packages/darklang/stdlib/string.dark
pbiggar: (I just noticed that Builtin.slice is not efficient - but we can resolve that)

PR #5063, opened 2023-09-06
#5060 -Requested changes

OceanOak:
No changelog

This implements the requested changes on #5060


packages/darklang/stdlib/list.dark
OceanOak, line 1: This is removed in #5061
https://github.com/darklang/dark/blob/eacb8c9e3bffbc8310031ba53ff8afa06559a7d0/packages/darklang/stdlib/list.dark#L89

backend/src/BuiltinExecution/Libs/Char.fs
```diff
@@ -45,78 +45,71 @@ let fns : List<BuiltInFn> =
       deprecated = NotDeprecated }
 
 
-    { name = fn "isDigit" 0
+    { name = fn "isASCII" 0
```
pbiggar, line 48: As a followup, could isASCII be implemented in Dark?

packages/darklang/stdlib/list.dark
```diff
@@ -47,9 +47,7 @@ module Darklang =
         match list with
         | [] -> []
         | head :: tail ->
-          Builtin.List.append_v0
-            (PACKAGE.Darklang.Stdlib.List.reverse tail)
-            (Builtin.List.singleton_v0 head)
+          Builtin.List.append_v0 (PACKAGE.Darklang.Stdlib.List.reverse tail) [ head ]
```
pbiggar, line 50: I think this function is O(N^2)

packages/darklang/stdlib/list.dark
```diff
@@ -71,13 +69,11 @@ module Darklang =
...
       let ``member`` (list: List<'a>) (value: 'a) : Bool =
-        match list with
-        | [] -> false
-        | head :: tail ->
-          if head == value then
-            true
-          else
-            PACKAGE.Darklang.Stdlib.List.``member`` tail value
+        match
+          (PACKAGE.Darklang.Stdlib.List.findFirst list (fun elem -> elem == value))
```
pbiggar, line 73: You can use `Option.isSome` here.

pbiggar (changes requested):
Looks good

packages/darklang/stdlib/list.dark
OceanOak: I've tried another implementation but it isn't any better it is O(N^2)
```
let reverse (list: List<'a>) : List<'a> =
   PACKAGE.Darklang.Stdlib.List.fold list [] (fun acc item -> (PACKAGE.Darklang.Stdlib.List.append [item] acc))

```
Should we keep it as a Builtin?

backend/src/BuiltinExecution/Libs/Char.fs
OceanOak: I couldn't find a good way to say that it should be between NUL and DEL or 0 and 127.
I think we'll need a new Builtin function that returns the ASCII value of a character.
Which means adding a new builtin function to remove one. Thoughts?
the functions is something like :
```
(if c.Length = 1 then System.Convert.ToInt64 c[0] else -1) |> DInt |> Ply
```
And the dark version of isAscii  will be 
```
      let isASCII (c: Char) : Bool =
        let ascii = Builtin.Char.asciiOfChar c
        (ascii >= 0) && (ascii <= 127)
```

pbiggar: Yeah, that's a good function to add.

I'd call it Char.toAsciiCode and it should return an option as characters might not be ascii at all.

packages/darklang/stdlib/list.dark
pbiggar: I think this could be O(N) if we used push instead of append?

pbiggar: I asked chatgpt:

```
let reverseList lst =
    let rec reverseHelper acc = function
        | [] -> acc
        | head :: tail -> reverseHelper (head :: acc) tail
    reverseHelper [] lst
```

OceanOak: I think we can't do this because of the recursive lambda issue? I am not sure I'll try

pbiggar:
`List.push` should be O(1), no?

OceanOak:
> `List.push` should be O(1), no?

Yes it is, I think I was wrong

backend/testfiles/execution/stdlib/char.dark
```diff
@@ -134,6 +134,25 @@ PACKAGE.Darklang.Stdlib.Char.isASCII_v0 ' ' = true
...
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode (smiley ()) = PACKAGE.Darklang.Stdlib.Option.Option.None
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode 'ż' = PACKAGE.Darklang.Stdlib.Option.Option.Some 380
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode 'Ż' = PACKAGE.Darklang.Stdlib.Option.Option.Some 379
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode 'ó' = PACKAGE.Darklang.Stdlib.Option.Option.Some 243
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode 'Ó' = PACKAGE.Darklang.Stdlib.Option.Option.Some 211
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode 'ł' = PACKAGE.Darklang.Stdlib.Option.Option.Some 322
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode 'Ł' = PACKAGE.Darklang.Stdlib.Option.Option.Some 321
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode (hand ()) = PACKAGE.Darklang.Stdlib.Option.Option.None
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode 'ჾ' = PACKAGE.Darklang.Stdlib.Option.Option.Some 4350
+PACKAGE.Darklang.Stdlib.Char.toAsciiCode 'Ჾ' = PACKAGE.Darklang.Stdlib.Option.Option.Some 7358
```
pbiggar: The ascii chars are only the first 256 numbers.

pbiggar (changes requested):

backend/src/BuiltinExecution/Libs/Char.fs
```diff
@@ -54,8 +54,9 @@ let fns : List<BuiltInFn> =
       fn =
         function
         | _, _, [ DChar c ] ->
-          if c.Length = 1 then
-            Dval.optionSome (DInt(System.Char.ConvertToUtf32(c, 0))) |> Ply
+          let charValue = System.Char.ConvertToUtf32(c, 0)
```
pbiggar, line 57: While this works, it feels like overkill. I asked chatgpt about how best to do this, and it suggested `(int)c`. That seems to me like it would work.

Conversation here: https://chat.openai.com/share/2a6f10f0-d45d-4e71-ac57-584570baeda8

pbiggar (changes requested):

PR #5064, opened 2023-09-07
Add a new fromBytes that doesn't ignore errors

OceanOak:
No changelog

This adds a new fromBytes that doesn't ignore errors and uses it in String.base64Decode

### DIFF
--- a/backend/src/BuiltinExecution/Libs/String.fs
+++ b/backend/src/BuiltinExecution/Libs/String.fs
@@ -421,6 +421,26 @@ let fns : List<BuiltInFn> =
       deprecated = NotDeprecated }
 
 
+    { name = fn "fromBytesRecoverable" 0
+      typeParams = []
+      parameters = [ Param.make "bytes" TBytes "" ]
+      returnType = TypeReference.option TString
+      description =
+        "Converts the UTF8-encoded byte sequence into a string. Errors will be ignored by replacing invalid characters"
+      fn =
+        (function
+        | _, _, [ DBytes bytes ] ->
+          try
+            let str = System.Text.UTF8Encoding(false, true).GetString bytes
+            Ply(Dval.optionSome (DString str))
+          with e ->
+            Ply(Dval.optionNone)
+        | _ -> incorrectArgs ())
+      sqlSpec = NotYetImplemented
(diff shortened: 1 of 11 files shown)


backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -421,6 +421,26 @@ let fns : List<BuiltInFn> =
       deprecated = NotDeprecated }
 
 
+    { name = fn "fromBytesRecoverable" 0
```
pbiggar, line 424: I think this is backwards.

Errors will be ignored in the function that returns a String, and errors will not be ignored in the function that returns `Option<String>`.

I had intended fromBytesRecoverable to be the name for the function that returned a string (and hence recovers). But I think it's a poor name.

OceanOak: Please correct me if I miss-understood what you mean here:
- The original fromBytes we had (which returns a String) should be renamed to something that indicates that it recovers)
- The newly added fromBytes (currently named fromBytesRecoverable) should be renamed to fromBytes
- We should then update all the references to the original 'fromBytes' method with its new name

Name suggestions:
- fromBytesWithReplacement
- fromBytesToSafeString
- fromBytesIgnoreErrors

pbiggar: I like `fromBytesWithReplacement`

Yes, this sounds right. The default should be that the Bytes might not be strings and the user should opt into replacement.

backend/testfiles/execution/stdlib/string.dark
```diff
@@ -252,15 +252,14 @@ module Base64Decode =
     "👱👱🏻👱🏼👱🏽👱🏾👱🏿"
 
 
-  // These produce strings of bytes which are technically legal it seems
-  PACKAGE.Darklang.Stdlib.String.base64Decode_v0 "-p" = PACKAGE.Darklang.Stdlib.Result.Result.Ok
-    "�"
```
pbiggar, line 257: Can you add tests for fromBytesWithReplacement?

pbiggar (approved):
One note for followup.

PR #5065, opened 2023-09-08
Wrap Crypto, DateTime, Password, X509, NoModule, and HttpClient built-in functions

OceanOak:
No changelog

This PR creates wrapper functions around Builtin Crypto, DateTime, Password, X509, NoModule, and HttpClient functions


packages/darklang/stdlib/json.dark
```diff
@@ -0,0 +1,12 @@
+module Darklang =
+  module Stdlib =
+    module Json =
```
OceanOak, line 3: TODO: recheck this module, it doesn't work

packages/darklang/stdlib/noModule.dark
```diff
@@ -0,0 +1,12 @@
...
+    /// Returns true if the two value are equal
+    let equals (a: 'a) (b: 'a) : Bool = Builtin.equals a b
+
+
+    /// Returns true if the two value are not equal
+    let notEquals (a: 'a) (b: 'a) : Bool = Builtin.notEquals a b
+
+
+    /// Unwrap an Option or Result, returning the value or a DError if None
+    let unwrap (value: 'a) : 'a = Builtin.unwrap value
```
OceanOak, line 12: TODO: check if this can be rewritten in dark

pbiggar: I'm wondering if we should even expose this, or if it should only exist as a `!` once we have darklang syntax for it?

packages/darklang/stdlib/httpclient.dark
```diff
@@ -2,46 +2,19 @@ module Darklang =
...
-        ("content-type", "text/plain; charset=utf-8")
-
-      /// Returns a header <type (String*String))> with {{'content-type'}} set for html requests
-      let htmlContentType () : (String * String) =
-        ("content-type", "text/html; charset=utf-8")
\ No newline at end of file
+      /// The response from a HTTP request
+      type Response =
+        { statusCode: Int
+          headers: List<Tuple<String, String>>
```
pbiggar, line 8: `Tuple` is odd here. It should be `String * String`

packages/darklang/stdlib/httpclientAuth.dark
```diff
@@ -0,0 +1,47 @@
+module Darklang =
+  module Stdlib =
+    module HttpClientAuth =
```
pbiggar, line 3: These shouldn't be in a new module, they should be in httpclient.

packages/darklang/stdlib/noModule.dark
pbiggar: It can't be rewritten in Dark though

pbiggar:
Looking good.

pbiggar:
Nice, thanks!

PR #5071, opened 2023-09-11
Refactor typechecking to use value types

StachuDotNet:
only the last commit here is new work - needs a rebase

Then, circle back here when all Dvals that need ValueTypes have them


StachuDotNet:
#5406 now contains a superset of this.

PR #5072, opened 2023-09-11
Misc changes made when trying to improve dark-packages

pbiggar:
Note: after this merges, you'll have to restart your devcontainer

A real grab bag of commits:

## Startup Performance/iteration speed
- add timing info to processes on startup, so we can see what takes so long
- do some startup stuff in parallel so things don't take so long (very little progress made, saved at most 0.5 seconds)
- run migrations in the background
- reload canvases when a canvas changes

## Understanding errors/programs
- add http logging in development
- print DErrors in Http servers and tests
- Convert some of errors to RTEs in the interpreter
- find where the error is happening in http servers
- format error messages in tests better
- refactor some darklang error printing
- when there's a string in an interpolated field, error

## Match behaviour
- DError when we try to match against the wrong type (eg match a string against an int). This is a type error, equivalent to `if "x" == 8`
  - this is only on the shape in MatchPattern (eg an Enum or count of Enum fields, but it won't error if you match `Ok` against `None` (Result vs Option)).
  - enums with wildcards will now match for any number of fields
- massively expand and organize ematch.dark tests

## Package manager
- Add local cache using concurrentDictionary


scripts/build/compile
```diff
@@ -277,6 +277,15 @@ def reload_all_packages():
   return run_backend(start, f"scripts/build/reload-packages {configuration}")
 
 
+def reload_canvas(f):
+  start = time.time()
+  # make relative to cwd
+  f = os.path.normpath(f)
+  f = os.path.relpath(f, "canvases")  # remove /canvases
+  (f, _) = os.path.split(f)  # remove filename
+  return run_backend(start, f"scripts/run-canvas-hack {f}")
```
StachuDotNet, line 286: love this

PR #5074, opened 2023-09-14
Inline package functions in the sqlCompiler

OceanOak:
Changelog:

```
Area of change
- Inline package functions in the SQLCompiler
```

This PR:
- Inline package functions in the SQLCompiler
- Add an async version of RuntimeTypesAst.postTraversal function
- Temporarily add Builtin.Bool.not back until we expose `ENot`
- add some db tests for package constants and roundtrip tests for tuple, record, and option.

### DIFF
--- a/backend/src/LibCloud/SqlCompiler.fs
+++ b/backend/src/LibCloud/SqlCompiler.fs
@@ -223,52 +223,43 @@ let rec inline'
 
       match expr with
       | EApply(_, EFnName(_, (fnName)), [], args) ->
-          uply {
-            let! arguments = Ply.List.mapSequentially (inline' fns paramName symtable) (args |> NEList.toList)
-            let nameStr = FnName.toString fnName
-            match fnName with
-            | FQName.Package p ->
-                match! fns.package p with
-                | Some fn ->
-                  let parameters =
-                    fn.parameters
-                    |> NEList.toList
-                    |> List.map (fun p -> p.name, p.typ)
-                  let body = fn.body
-                  let paramToArgMap =
-                    List.zip parameters arguments
-                    |> List.map (fun ((name, _), arg) -> name, arg)
-                    |> Map.ofList
-                  let! inlinedBody = inline' fns paramName paramToArgMap body
-                  return inlinedBody
-                | None ->
-                  return expr
-              | _ ->
-                return expr
-          }
-      | ELet (_, lpVariable, expr, body) ->
+        uply {
+          let! arguments =
+            Ply.List.mapSequentially
+              (inline' fns paramName symtable)
+              (args |> NEList.toList)
+          let nameStr = FnName.toString fnName
+          match fnName with
+          | FQName.Package p ->
+            match! fns.package p with
+            | Some fn ->
+              let parameters =
+                fn.parameters |> NEList.toList |> List.map (fun p -> p.name, p.typ)
+              let body = fn.body
+              let paramToArgMap =
+                List.zip parameters arguments
+                |> List.map (fun ((name, _), arg) -> name, arg)
+                |> Map.ofList
+              let! inlinedBody = inline' fns paramName paramToArgMap body
+              return inlinedBody
+            | None -> return expr
+          | _ -> return expr
+        }
+      | ELet(_, lpVariable, expr, body) ->
         let newMap = mapLetPattern symtable (expr, lpVariable)
         inline' fns paramName newMap body
       | EVariable(_, name) as expr when name <> paramName ->
         match Map.get name symtable with
-        | Some found ->
-          uply {
-            return found
-          }
+        | Some found -> uply { return found }
         | None ->
-           // the variable might be in the symtable, so put it back to fill in later
-            uply {
-              return expr
-            }
-      | EFieldAccess (id, expr, fieldname) ->
+          // the variable might be in the symtable, so put it back to fill in later
+          uply { return expr }
+      | EFieldAccess(id, expr, fieldname) ->
         uply {
           let! newexpr = inline' fns paramName symtable expr
-          return EFieldAccess (id,newexpr, fieldname)
+          return EFieldAccess(id, newexpr, fieldname)
         }
-      | expr ->
-          uply {
-            return expr
-          }
+      | expr -> uply { return expr }
 
     )
     identityPly
@@ -695,7 +686,7 @@ let rec lambdaToSql
           | _ ->
             return
               error $"We do not support this type of DB field yet: {dbFieldType}"
-        | EIf (_, cond, then_, else_) ->
+        | EIf(_, cond, then_, else_) ->
           let! cond = lts TBool cond
           let! then_ = lts expectedType then_
           let! else_ =
@@ -706,9 +697,9 @@ let rec lambdaToSql
           typecheck "if then" then_.actualType expectedType
           typecheck "if else" else_.actualType expectedType
           return
-            ( $"(CASE WHEN {cond.sql} THEN {then_.sql} ELSE {else_.sql} END)"
-            , cond.vars @ then_.vars @ else_.vars
-            , expectedType )
+            ($"(CASE WHEN {cond.sql} THEN {then_.sql} ELSE {else_.sql} END)",
+             cond.vars @ then_.vars @ else_.vars,
+             expectedType)
 
         | _ -> return error $"We do not yet support compiling this code: {expr}"
       }
--- a/backend/src/LibCloud/SqlCompiler.fs
+++ b/backend/src/LibCloud/SqlCompiler.fs
@@ -240,8 +240,7 @@ let rec inline'
                 List.zip parameters arguments
                 |> List.map (fun ((name, _), arg) -> name, arg)
                 |> Map.ofList
-              let! inlinedBody = inline' fns paramName paramToArgMap body
-              return inlinedBody
+              return! inline' fns paramName paramToArgMap body
             | None -> return expr
           | _ -> return expr
         }
@@ -250,16 +249,16 @@ let rec inline'
         inline' fns paramName newMap body
       | EVariable(_, name) as expr when name <> paramName ->
         match Map.get name symtable with
-        | Some found -> uply { return found }
+        | Some found -> Ply found
         | None ->
           // the variable might be in the symtable, so put it back to fill in later
-          uply { return expr }
+          Ply expr
       | EFieldAccess(id, expr, fieldname) ->
         uply {
           let! newexpr = inline' fns paramName symtable expr
           return EFieldAccess(id, newexpr, fieldname)
         }
-      | expr -> uply { return expr }
+      | expr -> Ply expr
 
     )
     identityPly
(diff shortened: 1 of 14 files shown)


backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -190,11 +190,13 @@ let escapeFieldname (str : string) : string =
...
 //  we can solve that in the next version.
 let rec inline'
+  (fns : Functions)
   (paramName : string)
   (symtable : Map<string, Expr>)
   (expr : Expr)
-  : Expr =
-  RuntimeTypesAst.postTraversal
+  : Ply<Expr> =
+  let identityPly x = uply { return x }
```
OceanOak, line 198: I'm uncertain whether there's a more effective way to achieve the same result, or if there's a more suitable place for this function

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -739,6 +686,21 @@ let rec lambdaToSql
           | _ ->
             return
               error $"We do not support this type of DB field yet: {dbFieldType}"
+        | EIf(_, cond, then_, else_) ->
```
OceanOak, line 689: I needed support for `EIf` here to be able to inline `Package.*.Bool.not/and/or` and `Package.*.String.contains`. Is there a reason why we didn't support it before?

backend/tests/Tests/SqlCompiler.Tests.fs
```diff
@@ -143,17 +153,35 @@ let inlineWorksAtRoot =
...
 
     let! expected = p "3 + 5"
-    let result = C.inline' "value" Map.empty expr
-    return Expect.equalExprIgnoringIDs result expected
+    let result =
+      uply {
+        let! result = C.inline' fns "value" Map.empty expr
+        return Expect.equalExprIgnoringIDs result expected
+      }
+    return (result |> Ply.toTask |> Async.AwaitTask |> Async.RunSynchronously)
```
OceanOak, line 161: I am not happy with this solution, any suggestions please?
Perhaps one solution could be to make `equalExprIgnoringIDs` return a Ply<unit> ?

backend/src/LibCloud/SqlCompiler.fs
pbiggar: `Bool.and` should be implemented as `a && b` and then I think the inlining will be simpler. Same for `Bool.or`: `a || b`.

backend/tests/Tests/SqlCompiler.Tests.fs
```diff
@@ -143,17 +153,35 @@ let inlineWorksAtRoot =
...
 
     let! expected = p "3 + 7"
-    let result = C.inline' "value" Map.empty expr
-    return Expect.equalExprIgnoringIDs result expected
+    let result =
+      uply {
+        let! result = C.inline' fns "value" Map.empty expr
+        return Expect.equalExprIgnoringIDs result expected
+      }
+    return (result |> Ply.toTask |> Async.AwaitTask |> Async.RunSynchronously)
```
pbiggar, line 184:
```suggestion
    return! result |> Ply.toTask
```

I think that will do it

packages/darklang/stdlib/string.dark
```diff
@@ -190,9 +190,10 @@ module Darklang =
 
       /// Checks if <param lookingIn> contains <param searchingFor>
       let contains (lookingIn: String) (searchingFor: String) : Bool =
-        PACKAGE.Darklang.Stdlib.Option.isSome (
-          Builtin.String.indexOf lookingIn searchingFor
-        )
+        if (Builtin.String.indexOf lookingIn searchingFor) == -1 then
```
pbiggar, line 193: No need for the `if`. You can just do `Builtin.String.indexOf lookingIn searchingFor) != -1`

backend/src/Prelude/Ply.fs
```diff
@@ -233,3 +233,24 @@ module Map =
         })
       Map.empty
       dict
+
+// TODO : Add more functions
```
pbiggar, line 237: No need for the TODO - we'll add them as they come.

backend/src/Prelude/Ply.fs
```diff
@@ -233,3 +233,24 @@ module Map =
...
+module Result =
+  let mapSequentially
```
pbiggar, line 239: I think I'd call this `map` - since there's only one, there's nothing to do seqentially.

backend/src/Prelude/Ply.fs
```diff
@@ -233,3 +233,24 @@ module Map =
...
+    (f : 'a -> Ply<'b>)
+    (result : Result<'a, 'err>)
+    : Ply<Result<'b, 'err>> =
+    match result with
+    | Ok v -> map (fun v -> Ok v) (f v)
+    | Error err -> Ply (Error err)
+
+// TODO : Add more functions
+module Option =
```
pbiggar, line 248: Same notes as above.

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -219,28 +221,62 @@ let rec inline'
...
-
+        inline' fns paramName newMap body
       | EVariable(_, name) as expr when name <> paramName ->
-        (match Map.get name symtable with
-         | Some found -> found
-         | None ->
+        match Map.get name symtable with
+        | Some found ->
+          uply {
+            return found
```
pbiggar: This can be written as just `Ply found` (same as `uply { return found }`)

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -219,28 +221,62 @@ let rec inline'
...
-           expr)
-
-      | expr -> expr)
-    identity
-    identity
-    identity
-    identity
-    identity
-    identity
+            uply {
```
pbiggar: Same here

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -219,28 +221,62 @@ let rec inline'
...
+              return expr
+            }
+      | EFieldAccess (id, expr, fieldname) ->
+        uply {
+          let! newexpr = inline' fns paramName symtable expr
+          return EFieldAccess (id,newexpr, fieldname)
+        }
+      | expr ->
+          uply {
```
pbiggar: Same here

backend/src/LibCloud/SqlCompiler.fs
pbiggar: I think this is fine

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -219,28 +221,62 @@ let rec inline'
...
+                  let parameters =
+                    fn.parameters
+                    |> NEList.toList
+                    |> List.map (fun p -> p.name, p.typ)
+                  let body = fn.body
+                  let paramToArgMap =
+                    List.zip parameters arguments
+                    |> List.map (fun ((name, _), arg) -> name, arg)
+                    |> Map.ofList
+                  let! inlinedBody = inline' fns paramName paramToArgMap body
```
pbiggar: You can just do `return' inline' ...`

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -739,6 +695,21 @@ let rec lambdaToSql
           | _ ->
             return
               error $"We do not support this type of DB field yet: {dbFieldType}"
+        | EIf (_, cond, then_, else_) ->
```
pbiggar: This is pretty cool!

pbiggar (commented):
This looks really good.

I think we should add some more complex tests. Especially ones where we call functions, and with complex expressions in the arguments. We've opened up to a lot more code so we should check a bunch of different patterns being inlined to see what happens.

I'm interested in what the code looks like after you inline nested function

backend/testfiles/execution/cloud/db.dark
```diff
@@ -779,6 +792,17 @@ module CompiledFunctions =
...
+  (friends (fun p -> PACKAGE.Darklang.Stdlib.Int.subtract p.height 3 == 69)) = [ " Chandler " ]
+
+  (friends (fun p -> PACKAGE.Darklang.Stdlib.Int.multiply_v0 p.height 2 == 146)) = [ "Ross" ]
+
+  (friends (fun p -> PACKAGE.Darklang.Stdlib.Int.divide_v0 p.height 2 == 5)) = [ "GrumpyCat" ]
+
+  // TODO: This test should pass, but it is causing an infinite recursion
+  // (friends (fun p -> (PACKAGE.Darklang.Stdlib.Int.power p.height 2) == PACKAGE.Darklang.Stdlib.Result.Result.Ok(100))) = [ "GrumpyCat" ]
+
   (friends (fun p -> PACKAGE.Darklang.Stdlib.String.toLowercase p.name == "rachel")) = [ "Rachel" ]
```
OceanOak, line 806: I couldn't add a test for xor ( to test more nested package functions). 
  Error : We do not yet support compiling this code: EIf (This is intentional)
  Please let me now if we need more tests

backend/tests/Tests/SqlCompiler.Tests.fs
```diff
@@ -143,17 +153,35 @@ let inlineWorksAtRoot =
       |> Ply.toTask
 
     let! expected = p "3 + 5"
-    let result = C.inline' "value" Map.empty expr
-    return Expect.equalExprIgnoringIDs result expected
+    let result =
+      uply {
```
pbiggar, line 157: I think this extra uply isn't necessary?

backend/tests/Tests/SqlCompiler.Tests.fs
```diff
@@ -143,17 +153,35 @@ let inlineWorksAtRoot =
...
+        Map.empty
+    let fns = ExecutionState.availableFunctions state
     let! expr = p "let x = 5 in (let x = 6 in (3 + (let x = 7 in x)))"
 
     let! expected = p "3 + 7"
-    let result = C.inline' "value" Map.empty expr
-    return Expect.equalExprIgnoringIDs result expected
+    let result =
+      uply {
+        let! result = C.inline' fns "value" Map.empty expr
```
pbiggar, line 181: I think this could be done inline and we don't need the extra uply?

pbiggar (approved):
Great job, thanks!

I think this is worth merging in its current state. however, there's a couple of followups I think we should do. I made 2 comments on some minor changes. Here's some other followups (they might not actually be doable until user function support lands, which is fine):

- add tests in SqlCompiler.Tests that test the inlining. I think we should have a test where we can actually see the inlined code, if possible. (a pattern match against the inlined expr might be the best we can do)

- add a complicated inlining test. A thought I had was tests around the variable names from the arguments. For example:


```
let add a b = a + b
query (fun v -> 
          let a = 5
          let b = 6
          v.number = add 7 8
            &&
          v.otherNumber = b)
```
The idea here is testing that the `let b = 6` definition doesn't interfere with the `b` parameter name in `add`.


Another test would be functions that are inlined 2-3 functions deep.

A small handful of weird tests like this would greatly increase my confidence.

PR #5076, opened 2023-09-14
Add `typeArg: ValueTypes` field to `Dval.DEnum` definition; require in construction via `Dval.enum`

StachuDotNet:
Still WIP - many of these tests won't pass until `Types` is passed into Dval.enum -- and used, in order to gleam the correct typeArgs count (even if all Unknowns, the # needs to be correct)


StachuDotNet:
This PR is being replaced shortly - closing.

PR #5080, opened 2023-09-18
Migrate base64encode and base64decode from Bytes to a Base64 module

OceanOak:
Changelog:

```
Stdlib improvements
- Migrate Bytes.base64encode and Bytes.base64decode to a separate Base64 module
- Add Bytes.append and Bytes.containsSegment
```

### DIFF
--- a/backend/src/BuiltinExecution/Libs/Base64.fs
+++ b/backend/src/BuiltinExecution/Libs/Base64.fs
@@ -60,7 +60,7 @@ let fns : List<BuiltInFn> =
       deprecated = NotDeprecated }
 
 
-    { name = fn "encode" 0
+    { name = fn "urlEncode" 0
       typeParams = []
       parameters = [ Param.make "bytes" TBytes "" ]
       returnType = TString
--- a/backend/src/BuiltinExecution/Libs/Base64.fs
+++ b/backend/src/BuiltinExecution/Libs/Base64.fs
@@ -1,0 +1,83 @@
+module BuiltinExecution.Libs.Base64
+
+open System
+open System.Text
+open System.Text.RegularExpressions
+
+open Prelude
+open LibExecution.RuntimeTypes
+open LibExecution.Builtin.Shortcuts
+
+module Dval = LibExecution.Dval
+
+
+let types : List<BuiltInType> = []
+let constants : List<BuiltInConstant> = []
+
+let modules = [ "Base64" ]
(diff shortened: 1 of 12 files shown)


backend/src/BuiltinExecution/Libs/Bytes.fs
```diff
@@ -127,6 +64,45 @@ let fns : List<BuiltInFn> =
...
+        (function
+        | _, _, [ DBytes bytes1; DBytes bytes2 ] ->
+          bytes2 |> Array.append bytes1 |> DBytes |> Ply
+        | _ -> incorrectArgs ())
+      sqlSpec = NotYetImplemented
+      previewable = Pure
+      deprecated = NotDeprecated }
+
+
+    { name = fn "containsSegment" 0
```
OceanOak, line 85: What do you think about renaming `containsSegment` to just `contains`?

backend/src/BuiltinExecution/Libs/Base64.fs
```diff
@@ -0,0 +1,83 @@
...
+              |> Ply
+            with e ->
+              Ply(Dval.resultError (DString("Not a valid base64 string")))
+        | _ -> incorrectArgs ())
+      sqlSpec = NotYetImplemented
+      previewable = Pure
+      deprecated = NotDeprecated }
+
+
+    { name = fn "encode" 0
```
OceanOak, line 63: I am not sure if this should be `urlEncode` instead of `encode`

pbiggar: Yeah, I guess that would make sense!

backend/src/BuiltinExecution/Libs/Bytes.fs
pbiggar: I prefer `contains` as well. We use `List.contains`, right?

pbiggar (changes requested):
Looks good with those changes!

backend/src/BuiltinExecution/Libs/Bytes.fs
OceanOak: I believe we call it `List.member`, we have` String.contains`

pbiggar (approved):

pbiggar:
I tried solving the conflict with the github conflict editor. I think I got it right? I'll merge once this goes green.

PR #5083, opened 2023-09-19
Add some db tests

OceanOak:
No changelog

#5074  followups

### DIFF
--- a/backend/testfiles/execution/cloud/db.dark
+++ b/backend/testfiles/execution/cloud/db.dark
@@ -761,8 +761,13 @@ module FindAll =
   let userLessThan (user: Person) (height: Int) : Bool =
     Builtin.Int.lessThan_v0 user height
 
-  let useAdd (a: Int) (b: Int) : Int = a + b
+  let userAdd (a: Int) (b: Int) : Int = a + b
 
+  (friends (fun p -> userAnd (true || false) (p.height > 66))) = [ " Chandler "
+                                                                   "Ross" ]
+
+  (friends (fun p -> let a = true in (userAnd (a && a) (p.height > 66)))) = [ " Chandler "
+                                                                              "Ross" ]
 
   (friends (fun p -> userAnd p.human true)) = [ " Chandler "; "Rachel"; "Ross" ]
 
@@ -773,7 +778,26 @@ module FindAll =
   (friends (fun p ->
     let a = 1 in
     let b = 9 in
-    (p.height == useAdd 6 4) && (p.age == b))) = [ "GrumpyCat" ]
+    (p.height == (userAdd 1 b) && (p.age == b)))) = [ "GrumpyCat" ]
+
+  (friends (fun p ->
+    let a = 1 in
+    let b = 9 in
+    (p.height == (userAdd (b - 8) b) && (p.age == b)))) = [ "GrumpyCat" ]
+
+  // TODO: this should work, but it doesn't. There is an error when we try to inline the function's parameters/arguments
+  // (friends (fun p -> let a = 1 in let b = 9 in (p.height == userAdd 6 (b - 5)))) = [ "GrumpyCat" ]
+
+  (friends (fun p ->
+    let a = 1 in
+    let b = 9 in
+    (p.height == (userAdd (b - 5) b) && (p.age == b)))) = []
+
+  (friends (fun p ->
+    let a = 1 in
+    let b = 9 in
+    let c = userAdd 6 4 in
+    (p.height == c) && (p.age == b))) = [ "GrumpyCat" ]
 
 
   // Tests for functions that are inlined 2-3 functions deep
--- a/backend/testfiles/execution/cloud/db.dark
+++ b/backend/testfiles/execution/cloud/db.dark
@@ -668,7 +668,7 @@ let addFriend
   (human: Bool)
   (dob: String)
   (income: Float)
-  (age : Int)
+  (age: Int)
   : Person =
   let dob = (PACKAGE.Darklang.Stdlib.DateTime.parse dob) |> Builtin.unwrap in
 
@@ -753,8 +753,7 @@ module FindAll =
                                                                                       "Ross" ]
 
   // inlining user functions
-  let userAnd (a : Bool) (b: Bool) : Bool =
-    PACKAGE.Darklang.Stdlib.Bool.and_v0 a b
+  let userAnd (a: Bool) (b: Bool) : Bool = PACKAGE.Darklang.Stdlib.Bool.and_v0 a b
 
   let userStringContains (user: Person) (str: String) : Bool =
     PACKAGE.Darklang.Stdlib.String.contains_v0 user.name str
@@ -762,7 +761,7 @@ module FindAll =
   let userLessThan (user: Person) (height: Int) : Bool =
     Builtin.Int.lessThan_v0 user height
 
-  let useAdd (a : Int) (b : Int) : Int = a + b
+  let useAdd (a: Int) (b: Int) : Int = a + b
 
 
   (friends (fun p -> userAnd p.human true)) = [ " Chandler "; "Rachel"; "Ross" ]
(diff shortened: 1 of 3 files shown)


backend/tests/Tests/SqlCompiler.Tests.fs
```diff
@@ -184,6 +184,57 @@ let inlineWorksWithNested =
...
+        Map.empty
+        Map.empty
+        Map.empty
+    let fns = ExecutionState.availableFunctions state
+    let! expr =
+      p
+        """let x = "package" in (let y = "e" in (PACKAGE.Darklang.Stdlib.String.replaceAll_v0 x y "es"))"""
+
+    let! expected = p """Builtin.String.replaceAll_v0 "package" "e" "es" """
+    let result =
```
OceanOak, line 229: > add tests in SqlCompiler.Tests that test the inlining. I think we should have a test where we can actually see the inlined code, if possible. (a pattern match against the inlined expr might be the best we can do)

I've added a couple of tests, but I am not sure about what you meant by "a pattern match against the inlined expr" could you please elaborate?

backend/testfiles/execution/cloud/db.dark
```diff
@@ -750,26 +753,46 @@ module FindAll =
...
+  let useAdd (a: Int) (b: Int) : Int = a + b
+
 
   (friends (fun p -> userAnd p.human true)) = [ " Chandler "; "Rachel"; "Ross" ]
 
   (friends (fun p -> userStringContains p "Chandler")) = [ " Chandler " ]
 
   (friends (fun p -> userLessThan p.height 65)) = [ "GrumpyCat" ]
 
+  (friends (fun p ->
```
OceanOak, line 773: This test is meant to be the equivalent of 

> 
> add a complicated inlining test. A thought I had was tests around the variable names from the arguments. For example:
> ```
> let add a b = a + b
> query (fun v -> 
>           let a = 5
>           let b = 6
>           v.number = add 7 8
>             &&
>           v.otherNumber = b)
>   ```
> The idea here is testing that the let b = 6 definition  doesn't interfere with the b parameter name in add.

backend/tests/Tests/SqlCompiler.Tests.fs
```diff
@@ -184,6 +184,57 @@ let inlineWorksWithNested =
...
+        Map.empty
+        Map.empty
+    let fns = ExecutionState.availableFunctions state
+    let! expr =
+      p
+        "let x = true in (let y = false in (PACKAGE.Darklang.Stdlib.Bool.and_v0 x y))"
+
+    let! expected = p "true && false"
+    let result =
+      uply {
```
OceanOak, line 205: I couldn't not use a uply here.
this is what I've tried 

```
 let! result = C.inline' fns "value" Map.empty expr
 return (Expect.equalExprIgnoringIDs result expected) |> Ply.toTask
```
```
 let result = C.inline' fns "value" Map.empty expr
 return! (Expect.equalExprIgnoringIDs result expected) |> Ply.toTask
```
```
 let! result = C.inline' fns "value" Map.empty expr
 return! (Expect.equalExprIgnoringIDs result expected) |> Ply.toTask
```

```
 let result = C.inline' fns "value" Map.empty expr
  return (Expect.equalExprIgnoringIDs result expected) |> Ply.toTask
```

```
 let result = Expect.equalExprIgnoringIDs (C.inline' fns "value" Map.empty expr) expected
 return! result |> Ply.toTask
```
```
 let! result = Expect.equalExprIgnoringIDs (C.inline' fns "value" Map.empty expr) expected
 return! result |> Ply.toTask
```

backend/tests/Tests/SqlCompiler.Tests.fs
pbiggar: This is better actually

backend/testfiles/execution/cloud/db.dark
```diff
@@ -750,26 +753,42 @@ module FindAll =
...
+  let useAdd (a : Int) (b : Int) : Int = a + b
+
 
   (friends (fun p -> userAnd p.human true)) = [ " Chandler "; "Rachel"; "Ross" ]
 
   (friends (fun p -> userStringContains p "Chandler")) = [ " Chandler " ]
 
   (friends (fun p -> userLessThan p.height 65)) = [ "GrumpyCat" ]
 
+  (friends (fun p -> let a = 1 in let b = 9 in (p.height == useAdd 6 4) && (p.age == b) )) = [ "GrumpyCat" ]
```
pbiggar: So for this test, I'm trying to see if the `b` parameter of `useAdd` shadows the `b = 9` definition after inlining. However, I don't think this test would trigger it.

Maybe try the following:

```
(friends (fun p -> let a = 1 in let b = 9 in (p.height == useAdd 6 (b - 5)) = [ "GrumpyCat" ]
```

```
(friends (fun p -> let a = 1 in let b = 9 in let c = useAdd 6 4 in (p.height == c) && (p.age == b) )) = [ "GrumpyCat" ]
```

backend/tests/Tests/SqlCompiler.Tests.fs
```diff
@@ -256,4 +303,4 @@ let partialEvaluation =
 let tests =
   testList
     "SqlCompiler"
-    [ inlineWorksAtRoot; inlineWorksWithNested; partialEvaluation; compileTests ]
+    [ inlineWorksAtRoot; inlineWorksWithNested; inlineWorksWithPackageFunctionsSqlBinOp; inlineWorksWithPackageFunctionsSqlFunction; partialEvaluation; compileTests ]
```
pbiggar: These are a good start. Can you add more complex tests here, maybe the same complex ones above (using anotherNestedUserFn and `let a = 1 in let b = 9 in let c = useAdd 6 4 in (p.height == c) && (p.age == b) )`)

pbiggar (changes requested):
These are helpful but I'd like to expose a little more complexity.

To be clear, I'm trying to check what happens when function  parameters/arguments are inlined, and I'm trying to get confidence we've captured the edge cases.

backend/testfiles/execution/cloud/db.dark
```diff
@@ -750,26 +753,70 @@ module FindAll =
...
+    let b = 9 in
+    (p.height == (userAdd 1 b) && (p.age == b)))) = [ "GrumpyCat" ]
+
+  (friends (fun p ->
+    let a = 1 in
+    let b = 9 in
+    (p.height == (userAdd (b - 8) b) && (p.age == b)))) = [ "GrumpyCat" ]
+
+  // TODO: this should work, but it doesn't. There is an error when we try to inline the function's parameters/arguments
+  // (friends (fun p -> let a = 1 in let b = 9 in (p.height == userAdd 6 (b - 5)))) = [ "GrumpyCat" ]
```
OceanOak, line 789: I think the problem here extends beyond just checking if the 'b' parameter of 'useAdd' shadows the value b = 9, as even the following test fails
``` 
  (friends (fun p ->
    let a = 1 in
    let c = 9 in
    (p.height == userAdd 6 (c - 5)))) = [ "GrumpyCat" ]
   ```
Still trying to figure out how to fix it. All I know so far is where the infinite recursion is happening ↓
   https://github.com/darklang/dark/blob/879e84cf8bb94ae0a25ceac1f16fe349bf1f1154/backend/src/LibExecution/RuntimeTypesAst.fs#L341-L348

File: backend/tests/Tests/SqlCompiler.Tests.fs [hunk 7]
```diff
@@ -184,6 +184,155 @@ let inlineWorksWithNested =
...
+        Map.empty
+        Map.empty
+        Map.empty
+        Map.empty
+    let fns = ExecutionState.availableFunctions state
+    let! expr =
+      p
+        "let a = 1 in let b = 9 in let c = userAdd 6 4 in (p.height == c) && (p.age == b)"
+
+    let! expected = p "p.height == userAdd 6 4 && p.age == 9"
```
OceanOak, line 254: I don't think we can define a function (`let userAdd = ...`) in `let! expr` . So, I am not sure where to put the definition of `userAdd` to be able to inline it.

pbiggar (approved):

PR #5087, opened 2023-09-20
Use type errors in more places

pbiggar:
The goal here was to remove Errors.fs, and replace it with the new error handling. I went through each place that used it and replaced it with either:

- inlining the string value (for user-facing Results)
- using an `Exception.raiseInternal` (if the type checker should have caught it)
- adding `TypeChecker.raiseFnValResultNotExpectedType` for `Interpreter.applyFnVal` calls where the type is wrong (I don't have a good sense of how the error should look here, maybe as we use it, it will become clearer)

I also extended error-stringifying in some cases as needed, with some minor error message upgrades.

While I was there, I ended up greatly simplifying some builtin functions (List, String, and Dict) which had been made really unpleasant as a result of trying to thread DErrors through them, they're much nicer now.


backend/src/LibExecution/TypeChecker.fs
```diff
@@ -179,6 +184,18 @@ let raiseValueNotExpectedType
   : 'a =
   raiseRTE source (ValueNotExpectedType(dv, typ, context) |> Error.toRuntimeError)
 
+let raiseFnValResultNotExpectedType
+  (source : DvalSource)
+  (dv : Dval)
+  (typ : TypeReference)
```
StachuDotNet, line 190: small nitpick, maybe swap dv (actual) and typ (expected) here?

StachuDotNet (approved):

backend/src/LibExecution/TypeChecker.fs
pbiggar: Actual comes before expected, no?

StachuDotNet: Seems to differ in different worlds. I'm just really used to the .NET world where it's generally expected first. That's also what I've seen in F# libraries. https://chat.openai.com/share/91365a6c-3f59-4d2a-8b51-b710f1921415 for reference

PR #5088, opened 2023-09-20
Fix the HttpClient.basicAuth function

OceanOak:
No changelog

- Delete HttpClientAuth
- Fix `PACKAGE.Darklang.Stdlib.HttpClient.basicAuth` function
- Use `PACKAGE.Darklang.Stdlib.HttpClient.basicAuth` in tests


backend/testfiles/execution/stdlib/httpclient.dark
```diff
@@ -18,17 +18,44 @@ module NoInternal =
...
 
-  Builtin.HttpClient.basicAuth "" "-" = ("authorization", "basic Oi0=")
-  Builtin.HttpClient.basicAuth ":" "" = ("authorization", "basic Ojo=")
-  Builtin.HttpClient.basicAuth "" ":" = ("authorization", "basic Ojo=")
+  PACKAGE.Darklang.Stdlib.HttpClient.basicAuth "" "-" = PACKAGE
+    .Darklang
+    .Stdlib
+    .Result
+    .Result
+    .Ok(("authorization", "basic Oi0"))
```
OceanOak, line 43: the change from `Ojo=` to `Ojo` is because of our String.base64Encode it replaces `=` with `""`

backend/testfiles/execution/stdlib/httpclient.dark
```diff
@@ -18,17 +18,44 @@ module NoInternal =
...
+    .Result
+    .Result
+    .Ok(("authorization", "basic Og"))
+
+  PACKAGE.Darklang.Stdlib.HttpClient.basicAuth "-" "" = PACKAGE.Darklang.Stdlib.Result.Result.Error
     "Username cannot contain a hyphen"
 
-  Builtin.HttpClient.basicAuth "" "-" = ("authorization", "basic Oi0=")
-  Builtin.HttpClient.basicAuth ":" "" = ("authorization", "basic Ojo=")
-  Builtin.HttpClient.basicAuth "" ":" = ("authorization", "basic Ojo=")
```
OceanOak, line 31: the change from Ojo= to Ojo is because of our String.base64Encode it replaces = with ""

StachuDotNet (approved):
Looks great! I'd merge, but I'm not sure if it's concerning that this changes how we handle `"` instances in the string.

pbiggar:
Ok, so according to the [RFC](https://datatracker.ietf.org/doc/html/rfc7617#section-2), basic auth uses "standard" base 64 encoding: https://en.wikipedia.org/wiki/Base64#Variants_summary_table, which uses `+` and `/` as encoding chars, with optional padding using `=`. So maybe the best approach is to have a test for that.

The RFC uses the example:
>    If the user agent wishes to send the user-id "Aladdin" and password
>   "open sesame", it would use the following header field:
>
>      Authorization: Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==

I'm struggling to come up with an example where the base64-encoded string has `+` or `/` in it. it seems like we'll have to reverse engineer that.

OceanOak:
> I'm struggling to come up with an example where the base64-encoded string has + or / in it. it seems like we'll have to reverse engineer that.

I don't think we can get a base64-encoded string that has + or / in it because we are using `Base64.urlEncode` which replaces `+` with `-` and `/` with `_`
`System.Convert.ToBase64String(bytes).Replace('+', '-').Replace('/', '_')`

pbiggar:
OK, sounds like we should be using standard encoding and not url-encoding.

pbiggar:
Nice!

PR #5089, opened 2023-09-20
Type-check some DEnums upon construction

StachuDotNet:
- tidy ValueType-creation helpers - i.e. expose `ValueType.unknown`, `ValueType.int`, etc. 
- add `typeArgs: List<ValueType>` field to `RT.DEnum` definition, both in `RuntimeTypes.fs` and `runtimeTypes.dark`
- update `Dval.enum`, `Dval.optionX` and `Dval.resultX` functions accordingly
- when creating DEnums:
  - where we know that there are no typeArgs for some enum, explicitly fill them in as `[]`
  - where we know we'll need non-0 typeArgs, fill them in with [] by way of the `VT.uknownTypeArgsTODO` and `uknownTypeArgsTODO'` functions, so we are encouraged to follow up there. The Interpreter's current partial support yields test failures if we fill in actual type args (say, for Results and Options) in the meantime, as there's no type-lookup in `Dval.enum` yet
- when deconstructing DEnums
  - where we know there are no typeArgs, we match against the field with `[]`
  - otherwise, match with `_typeArgsDEnumTODO`, to be followed up on once Dval.enum and the Interpreter handles type args more fully


backend/tests/TestUtils/LibTest.fs
```diff
@@ -102,14 +104,17 @@ let fns : List<BuiltInFn> =
       returnType = TypeReference.option TChar
       description = "Turns a string of length 1 into a character"
       fn =
+        let optType = VT.char
```
pbiggar, line 107: This would be more readable inlined imo

backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -400,13 +400,75 @@ and [<RequireQualifiedAccess>] ValueType =
...
-/// CLEANUP eventually remove this binding
-let valueTypeTODO = ValueType.Unknown
-let valueTypesTODO = []
+[<RequireQualifiedAccess>]
+module ValueType =
+  // some helpers to reduce typing elsewhere
+  let unknown = ValueType.Unknown
+  let unknownTODO = ValueType.Unknown
+  let unknownDbTODO = ValueType.Unknown
+  let uknownTypeArgsTODO = []
```
pbiggar, line 409: What does uknownTypeArgs mean?

backend/src/BuiltinExecution/Libs/Math.fs
```diff
@@ -92,14 +93,15 @@ let fns : List<BuiltInFn> =
 
          This function is the inverse of <fn Math.cos>."
       fn =
+        let optType = VT.float
```
pbiggar, line 96: Inlining would be more readable here too

backend/src/BuiltinExecution/Libs/List.fs
```diff
@@ -665,15 +674,16 @@ let fns : List<BuiltInFn> =
          selected value in <param list>. Returns {{None}} if <param list> is
          empty."
       fn =
+        let optType = VT.unknownTODO
```
pbiggar, line 677: Is optType not the same as the first param in the DList?

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -425,13 +426,14 @@ let fns : List<BuiltInFn> =
       description =
         "Converts the UTF8-encoded byte sequence into a string. Errors will be ignored by replacing invalid characters"
       fn =
+        let optType = VT.string
```
pbiggar, line 429: Inline imo

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -467,13 +469,14 @@ let fns : List<BuiltInFn> =
       description =
         "Returns {{Some char}} of the first character of <param str>, or returns {{None}} if <param str> is empty."
       fn =
+        let optType = VT.char
```
pbiggar, line 472: Inline imo

pbiggar:
Very nice! Good mergeable step.

backend/src/BuiltinExecution/Libs/Math.fs
StachuDotNet: yeah I was on the fence about a few of these Option ones - will update in subsequent PR

backend/src/LibExecution/RuntimeTypes.fs
StachuDotNet: I just renamed this binding to `typeArgsTODO` in another PR. I think that's more correct. The intention here is to imply that some type (like Option) needs some typeArgs but we haven't typed them out yet when constructing the Enum/Record - so these placeholders are a call-out to follow up later, filling them in.

backend/src/BuiltinExecution/Libs/List.fs
StachuDotNet: I think you're implying that we could update the implementation of List.randomElement to something like:

```fsharp
let optType = VT.unknownTODO
        (function
        | _, _, [ DList(_, []) ] -> Ply(Dval.optionNone optType)
        | _, _, [ DList(_, l) ] ->
          // Will return <= (length - 1)
          // Maximum value is Int64.MaxValue which is half of UInt64.MaxValue, but
          // that won't affect this as we won't have a list that big for a long long
          // long time.
          let index = RNG.GetInt32(l.Length)
          match List.tryItem index l with
          | Some item -> Dval.optionSome (Dval.toValueType item) item |> Ply
          | None -> Ply(Dval.optionNone optType)
```
?

PR #5095, opened 2023-09-22
remove HttpClient string errors by defining a custom error type in DarkLang

OceanOak:
No changelog

This PR replaces the HttpClient string errors with custom errors

### DIFF
--- a/backend/src/BuiltinExecution/Libs/HttpClient.fs
+++ b/backend/src/BuiltinExecution/Libs/HttpClient.fs
@@ -447,43 +447,54 @@ let fns (config : Configuration) : List<BuiltInFn> =
         | state,
           _,
           [ DString method; DString uri; DList(_, reqHeaders); DBytes reqBody ] ->
-          let reqHeaders : Result<List<string * string>, HeaderError.HeaderError> =
-            reqHeaders
-            |> List.fold
-              (fun agg item ->
-                match agg, item with
-                | (Error err, _) -> Error err
-                | (Ok pairs, DTuple(DString k, DString v, [])) ->
-                  // TODO: what about whitespace? What else can break?
-                  if k = "" then
-                    Error HeaderError.HeaderError.EmptyKey
-                  else
-                    Ok((k, v) :: pairs)
+          uply {
+            let! (reqHeaders : Result<List<string * string>, HeaderError.HeaderError>) =
+              reqHeaders
+              |> Ply.List.foldSequentially
+                (fun agg item ->
+                  uply {
+                    match agg, item with
+                    | (Error err, _) -> return (Error err)
+                    | (Ok pairs, DTuple(DString k, DString v, [])) ->
+                      // TODO: what about whitespace? What else can break?
+                      if k = "" then
+                        return Error HeaderError.HeaderError.EmptyKey
+                      else
+                        return Ok((k, v) :: pairs)
 
-                | (_, notAPair) ->
-                  let context = TypeChecker.Context.FnValResult(TUnit, None)
-                  TypeChecker.raiseValueNotExpectedType
-                    state.caller
-                    notAPair
-                    (TList(TTuple(TString, TString, [])))
-                    context
-                  |> Ply.toTask
-                  |> Async.AwaitTask
-                  |> Async.RunSynchronously
+                    | (_, notAPair) ->
+                      let context =
+                        TypeChecker.Context.FunctionCallParameter(
+                          (FnName.fqPackage
+                            "Darklang"
+                            [ "Stdlib"; "HttpClient" ]
+                            "request"
+                            0),
+                          ({ name = "headers"; typ = headersType }),
+                          2,
+                          None
+                        )
+                      return!
+                        TypeChecker.raiseValueNotExpectedType
+                          state.caller
+                          notAPair
+                          (TList(TTuple(TString, TString, [])))
+                          context
 
-              )
-              (Ok [])
-            |> Result.map (fun pairs -> List.rev pairs)
+                  }
 
-          let method =
-            try
-              Some(HttpMethod method)
-            with _ ->
-              None
+                )
+                (Ok [])
+              |> Ply.map (Result.map (List.rev))
 
-          match reqHeaders, method with
-          | Ok reqHeaders, Some method ->
-            uply {
+            let method =
+              try
+                Some(HttpMethod method)
+              with _ ->
+                None
+
+            match reqHeaders, method with
+            | Ok reqHeaders, Some method ->
               let request =
                 { url = uri; method = method; headers = reqHeaders; body = reqBody }
 
@@ -519,18 +530,14 @@ let fns (config : Configuration) : List<BuiltInFn> =
                 let! err = err |> RequestError.toDT
                 return (err |> resultError)
 
-            }
-
-          | Error reqHeadersErr, _ ->
-            uply {
+            | Error reqHeadersErr, _ ->
               let! reqHeadersErr = reqHeadersErr |> HeaderError.toDT
               return reqHeadersErr |> resultError
-            }
 
-          | _, None ->
-            let error = "Expected valid HTTP method (e.g. 'get' or 'POST')"
-            uply { return resultErrorStr error }
-
+            | _, None ->
+              let error = "Expected valid HTTP method (e.g. 'get' or 'POST')"
+              return resultErrorStr error
+          }
         | _ -> incorrectArgs ())
       sqlSpec = NotQueryable
       previewable = Impure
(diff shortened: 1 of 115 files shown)


backend/testfiles/execution/stdlib/httpclient.dark
```diff
@@ -178,271 +190,354 @@ module Disallowed =
...
 
   PACKAGE.Darklang.Stdlib.HttpClient.request
     "get"
     "http://[::ffff:a9fe:a9fe]"
     []
-    Builtin.Bytes.empty = PACKAGE.Darklang.Stdlib.Result.Result.Error
-    "Could not connect ([::ffff:169.254.169.254]:80) Could not connect"
+    Builtin.Bytes.empty = PACKAGE.Darklang.Stdlib.Result.Result.Error(
+    PACKAGE.Darklang.Stdlib.HttpClient.RequestError.Other
+      "Could not connect ([::ffff:169.254.169.254]:80) Could not connect"
```
pbiggar, line 355: This could be a separate case instead of an `Other`.

Actually, can you remove `Other` and make each different error a different case? You might look at the different dotnet/C# error types that can be thrown and make a different error for each one maybe.

pbiggar (changes requested):
Nice! I think we can fix even more though (see comments)

OceanOak:
I've moved the BadUrl options to a separate enum, However, I'm still uncertain about the `other` exception because we don't have any tests for these cases and I don't know if all the exceptions are covered. So I have to look more into it.

pbiggar:
> I've moved the BadUrl options to a separate enum, However, I'm still uncertain about the `other` exception because we don't have any tests for these cases and I don't know if all the exceptions are covered.
> 
> Exceptions that can be thrown when the SendAsync method fails: (https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.sendasync?view=net-7.0) ArgumentNullException InvalidOperationException HttpRequestException TaskCanceledException we only handled the last two. Were the other 2 exceptions unhandled intentionally? and should we add something like `System.Exception` to catch the remaining exceptions in case we missed one?

So the key thing with these exceptions is to translate them from the F#/dotnet way of doing things to the darklang way of doing things so that it feels natural. We don't want to just create a new case named after the exception, but rather look at why the exceptions happen and create/adopt a case for that.

The docs for these exceptions say:
> [InvalidOperationException](https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=net-7.0)
The request message was already sent by the [HttpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-7.0) instance.

This should never happen as it would mean we're using dotnet HttpClient wrong. For exceptions that should never happen, we can either catch them and use `Exception.raiseInternal`, or just don't catch them (the latter is best probably)

> [HttpRequestException](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httprequestexception?view=net-7.0)
The request failed due to an underlying issue such as network connectivity, DNS failure, server certificate validation or timeout.

This is a meaty exception. The ideal would be to have `ConnectionUnavailable`, `DnsError`, `CertificateValidationError`, and `Timeout` as separate cases here. However, it might be hard to extract these from the HttpClient exception, I'm not sure. Perhaps these are all just `ConnectionError`s?

> [TaskCanceledException](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcanceledexception?view=net-7.0)
.NET Core and .NET 5 and later only: The request failed due to timeout.

This would be `Timeout`.

Certainly I would try and add test cases for all of these if you can.

OceanOak:
> This is a meaty exception. The ideal would be to have ConnectionUnavailable, DnsError, CertificateValidationError, and Timeout as separate cases here. However, it might be hard to extract these from the HttpClient exception, I'm not sure. Perhaps these are all just ConnectionErrors?

Tried to make the HttpRequestException errors separate but it seemed wrong doing it the following way:
```
       if e.Message = "The SSL connection could not be established, see inner exception." then
          return
            Error(
              RequestError.RequestError.CertificateValidationError(
                Exception.getMessages e |> String.concat " "
              )
            )
        else if ((e.Message).Contains "Name or service not known") then
          return
          return
            Error(
              RequestError.RequestError.DnsError(
                Exception.getMessages e |> String.concat " "
              )
            )
        else if ((e.Message).Contains "Could not connect") then
          return
            Error(
              RequestError.RequestError.ConnectionUnavailable(
                Exception.getMessages e |> String.concat " "
              )
            )
        else
          return
            Error(
              RequestError.RequestError.ConnectionError(
                Exception.getMessages e |> String.concat " "
              )
            )

File: packages/darklang/stdlib/httpclient.dark [hunk 2]
```diff
@@ -29,7 +29,7 @@ module Darklang =
         | HeaderError details ->
           match details with
           | EmptyKey -> "Empty key"
-        | IOException details -> $"URI format exception: {details}"
+        | IOError details -> $"URI format exception: {details}"
```
pbiggar, line 32: That string doesn't match

pbiggar:
It's not the worst way - it's kinda unfortunate that this is what's provided but if we don't have anything better we should use it.

No need to include the lower-level messages though, if they don't have anything useful to say. I'd be inclined not to include any lower-level messages as strings at all if it could be avoided - we don't want our users to have to parse out useful information from strings.

backend/testfiles/execution/stdlib/httpclient.dark
```diff
@@ -178,271 +202,402 @@ module Disallowed =
...
 
   PACKAGE.Darklang.Stdlib.HttpClient.request
     "get"
     "http://[0000:0000:0000:0000:0000:0000:0000:0001]"
     []
-    Builtin.Bytes.empty = PACKAGE.Darklang.Stdlib.Result.Result.Error
-    "Could not connect ([::1]:80) Could not connect"
+    Builtin.Bytes.empty = PACKAGE.Darklang.Stdlib.Result.Result.Error(
+    PACKAGE.Darklang.Stdlib.HttpClient.RequestError.ConnectionError
+      "Could not connect ([::1]:80) Could not connect"
```
StachuDotNet, line 310: Where does this text come from? (I'm curious if we can get rid of the redundant `Could not connect`

pbiggar: Deep within dotnet - as I recall we have no access to it except in an Exception.Message.

backend/src/BuiltinExecution/Libs/HttpClient.fs
```diff
@@ -429,24 +511,18 @@ let fns (config : Configuration) : List<BuiltInFn> =
...
-
-          | Error reqHeadersErr, _ ->
-            uply {
-              match reqHeadersErr with
-              | BadInput details -> return resultErrorStr details
-              | TypeMismatch details ->
-                return raiseUntargetedRTE (RuntimeError.oldError details)
-            }
+              | Error(err) ->
+                let err = err |> RequestError.toDT
```
pbiggar, line 515: I find the multiple errors confusing. Could you consolidate them into a single place before calling RequestError.toDT.

backend/src/BuiltinExecution/Libs/HttpClient.fs
```diff
@@ -24,15 +24,66 @@ type Request = { url : string; method : Method; headers : Headers.T; body : Body
...
+      | InvalidHost -> "InvalidHost", []
+      | InvalidUri -> "InvalidUri", []
+      | InvalidRequest -> "InvalidRequest", []
+
+    let typeName =
+      TypeName.fqPackage "Darklang" [ "Stdlib"; "HttpClient" ] "BadUrlDetails" 0
+    Dval.DEnum(typeName, typeName, [], caseName, fields)
+
+module RequestError =
+  // forked from Elm's HttpError type
```
pbiggar, line 59: Maybe say "inspired by" or something here - the old comment I felt was a bit off.

backend/src/BuiltinExecution/Libs/HttpClient.fs
```diff
@@ -24,15 +24,66 @@ type Request = { url : string; method : Method; headers : Headers.T; body : Body
...
+  // https://package.elm-lang.org/packages/elm/http/latest/Http#Error
+  type RequestError =
+    | BadUrl of BadUrl.BadUrlDetails
+    | Timeout
+    | HeaderError of HeaderError.HeaderError
+    | IOError
```
pbiggar, line 65: I'm not seeing a real difference that we can point to for users between IOError and ConnectionError. Should we consolidate these into NetworkError (like Elm).

pbiggar (changes requested):

backend/src/BuiltinExecution/Libs/HttpClient.fs
pbiggar: These aren't all the same type, AFAICT. BadMethod is a RequestError so that's fine, but BadHeader is a separate type (it should be wrapped in a RequestHeader.BadHeader enum).

The reason there's no typeerror is because they're all converted to Dvals using toDT. This is why I'm suggesting having a `let result = ....` where the f# type of `result` is `Result<Dval, RequestError>`. Then you have the type safety of knowing all the branches have the same type in F#, and can call `toDT` at the end.

pbiggar:
This is great, thank you!

PR #5099, opened 2023-09-25
Make dval result and option fns ply

StachuDotNet:
Final(?) preamble step before adding Types argument to Dval.enum and Dval.record


backend/tests/TestUtils/TestUtils.fs
```diff
@@ -1152,10 +1152,32 @@ let interestingDvals : List<string * RT.Dval * RT.TypeReference> =
...
+     TypeReference.option TInt)
+    ("option2",
+     DEnum(
+       Dval.optionType,
+       Dval.optionType,
+       Dval.ignoreAndUseEmpty [ VT.int ],
+       "Some",
+       [ Dval.int 15 ]
+     ),
+     TypeReference.option TInt)
```
StachuDotNet, line 1172: Small note here - the Dval.option and Dval.result fns here became async (Ply) so don't work well here, and the point of the tests is to check _roundtripping_, not whether the Dval fns work correctly - so I opted to just create the DEnums manually here

pbiggar:
Very nice change!

PR #5102, opened 2023-09-26
Optimize String.slice

OceanOak:
No changelog

Optimize String.slice

### DIFF
--- a/backend/src/BuiltinExecution/Libs/String.fs
+++ b/backend/src/BuiltinExecution/Libs/String.fs
@@ -305,26 +305,38 @@ let fns : List<BuiltInFn> =
         (function
         | _, _, [ DString s; DInt first; DInt last ] ->
           let strLength = String.length s
+          // Adjust first and last indices if they are negative (to start counting from the end)
           let first = if first < 0 then strLength + int first else int first
           let last = if last < 0 then strLength + int last else int last
 
+          // Return empty string if first index is out of bounds or greate than or equal to the last index
           if first >= strLength || first >= last then
-            "" |> DString |> Ply
+            Ply(DString "")
           else
+            // Create a TextElementEnumerator to iterate through the string elements
+            // This is used to handle EGCs correctly
             let textElemEnumerator =
               System.Globalization.StringInfo.GetTextElementEnumerator(s)
             let mutable startIndex = 0
             let mutable endIndex = 0
             let mutable index = 0
 
+            // Iterate through the string elements using the enumerator
+            // When the enumerator reaches `first`
+            // set the startIndex to the starting position of the current text element in the original string
+            // similarly, when the enumerator reaches `last`
             while textElemEnumerator.MoveNext() do
               if index = first then startIndex <- textElemEnumerator.ElementIndex
               if index = last then endIndex <- textElemEnumerator.ElementIndex
               index <- index + 1
 
+            // Set endIndex to the length of the original string.
+            // If it is still 0 after the loop (the last index is out of bounds)
             if endIndex = 0 then endIndex <- s.Length
-            let length = endIndex - startIndex
-            s.Substring(startIndex, length) |> DString |> Ply
+
+            // Calculate the length of the substring
+            let substringLength = endIndex - startIndex
+            s.Substring(startIndex, substringLength) |> DString |> Ply
         | _ -> incorrectArgs ())
       sqlSpec = NotYetImplemented
       previewable = Pure
--- a/backend/src/BuiltinExecution/Libs/String.fs
+++ b/backend/src/BuiltinExecution/Libs/String.fs
@@ -304,37 +304,31 @@ let fns : List<BuiltInFn> =
       fn =
         (function
         | _, _, [ DString s; DInt first; DInt last ] ->
-          let strLength = String.length s
-          // Adjust first and last indices if they are negative (to start counting from the end)
+          // let strLength = String.length s
+          let strLength = System.Globalization.StringInfo(s).LengthInTextElements
+          // Handle negative indexes (which allow counting from the end)
           let first = if first < 0 then strLength + int first else int first
           let last = if last < 0 then strLength + int last else int last
 
-          // Return empty string if first index is out of bounds or greate than or equal to the last index
           if first >= strLength || first >= last then
             Ply(DString "")
           else
-            // Create a TextElementEnumerator to iterate through the string elements
-            // This is used to handle EGCs correctly
+            // Create a TextElementEnumerator to handle EGCs
             let textElemEnumerator =
               System.Globalization.StringInfo.GetTextElementEnumerator(s)
             let mutable startIndex = 0
             let mutable endIndex = 0
             let mutable index = 0
 
-            // Iterate through the string elements using the enumerator
-            // When the enumerator reaches `first`
-            // set the startIndex to the starting position of the current text element in the original string
-            // similarly, when the enumerator reaches `last`
+            // Iterate through EGCs and record the byte indexes of the start and end
             while textElemEnumerator.MoveNext() do
               if index = first then startIndex <- textElemEnumerator.ElementIndex
               if index = last then endIndex <- textElemEnumerator.ElementIndex
               index <- index + 1
 
-            // Set endIndex to the length of the original string.
-            // If it is still 0 after the loop (the last index is out of bounds)
+            // Check out of bounds
             if endIndex = 0 then endIndex <- s.Length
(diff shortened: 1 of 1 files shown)


OceanOak:
The original String.slice function takes a string, converts it into a sequence, then into a list, performs the slicing, and finally concatenates the result. 
```
          String.toEgcSeq s
          |> Seq.toList
          |> FSharpPlus.List.drop (int first)
          |> List.truncate (int (last - first))
          |> String.concat ""
          |> DString
          |> Ply
  ```
  
I attempted to skip the conversion to a list step by using sequence functions, but using `seq.fold `ultimately offset the benefits gained from eliminating the list conversion step.

```
          let sliceSeq start stop seq =  seq |> Seq.skip start |> Seq.take (stop - start)
          let chars = String.toEgcSeq s
          chars
          |> sliceSeq (int first) (int last)
          |> Seq.fold (fun acc char -> acc + char) ""
          |> DString
          |> Ply
```

Then I've tried to use `string.Substring`, but it didn't work well with EGCs
```
          let length = last - first
          s.Substring(int first, int length)
          |> DString
          |> Ply
```

So, I went back to using sequences but avoided using `seq.fold`
```
          let charSeq = String.toEgcSeq s
          charSeq
          |> Seq.skip (int from)
          |> Seq.take (int (to_ - from))
          |> String.concat ""
          |> DString
          |> Ply
```

**Comment by pbiggar** [comment]
I think you were right on String.substring, the issue was the indexes are wrong. I believe you can solve this by iterating through the EGCs and counting the length of each textelement. You might want to use a while loop instead of trying to keep it functional, as that seems much simpler.

**Review by pbiggar — CHANGES_REQUESTED** [review]
I can't tell how this works - would you be able to see if you can make the code easier to follow (and/or add comments)?

File: backend/src/BuiltinExecution/Libs/String.fs [hunk 1]
```diff
@@ -299,20 +299,44 @@ let fns : List<BuiltInFn> =
...
         (function
         | _, _, [ DString s; DInt first; DInt last ] ->
-          String.toEgcSeq s
-          |> Seq.toList
-          |> FSharpPlus.List.drop (int first)
-          |> List.truncate (int (last - first))
-          |> String.concat ""
-          |> DString
-          |> Ply
+          let strLength = String.length s
```
pbiggar, line 307: This should use the length in EGCs, not the length in characters.

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -299,20 +299,44 @@ let fns : List<BuiltInFn> =
...
+          // Adjust first and last indices if they are negative (to start counting from the end)
```
pbiggar, line 308:
```suggestion
          // Handle negative indexes (which allow counting from the end)
```

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -299,20 +299,44 @@ let fns : List<BuiltInFn> =
...
+          let first = if first < 0 then strLength + int first else int first
+          let last = if last < 0 then strLength + int last else int last
+
+          // Return empty string if first index is out of bounds or greate than or equal to the last index
```
pbiggar, line 312: Comment isn't necessary (it repeats the code)

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -299,20 +299,44 @@ let fns : List<BuiltInFn> =
...
+          if first >= strLength || first >= last then
+            Ply(DString "")
+          else
+            // Create a TextElementEnumerator to iterate through the string elements
```
pbiggar, line 316: "TextElementEnumerator to handle EGCs"

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -299,20 +299,44 @@ let fns : List<BuiltInFn> =
...
+            // This is used to handle EGCs correctly
+            let textElemEnumerator =
+              System.Globalization.StringInfo.GetTextElementEnumerator(s)
+            let mutable startIndex = 0
+            let mutable endIndex = 0
```
pbiggar, line 321: I'd probably use `-1` for this.

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -299,20 +299,44 @@ let fns : List<BuiltInFn> =
...
+            let mutable index = 0
+
+            // Iterate through the string elements using the enumerator
```
pbiggar, line 324: This comments explains exactly what is happening, but it doesn't tell us any more than the code does.

Maybe "Iterate through EGCs and record the byte indexes of the start and end"

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -299,20 +299,44 @@ let fns : List<BuiltInFn> =
...
+            // When the enumerator reaches `first`
+            // set the startIndex to the starting position of the current text element in the original string
+            // similarly, when the enumerator reaches `last`
+            while textElemEnumerator.MoveNext() do
+              if index = first then startIndex <- textElemEnumerator.ElementIndex
+              if index = last then endIndex <- textElemEnumerator.ElementIndex
+              index <- index + 1
+
+            // Set endIndex to the length of the original string.
+            // If it is still 0 after the loop (the last index is out of bounds)
```
pbiggar, line 334: "Check out of bounds"

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -299,20 +299,44 @@ let fns : List<BuiltInFn> =
...
+            if endIndex = 0 then endIndex <- s.Length
+
+            // Calculate the length of the substring
```
pbiggar, line 337: Not necessary

pbiggar (changes requested):
The way to write comments is to describe why, as opposed to what. You've got a bit of both and it could be improved. Examples inline.

backend/src/BuiltinExecution/Libs/String.fs
OceanOak: I am not sure how this will work, I'll look more into it

OceanOak:
> The way to write comments is to describe why, as opposed to what. You've got a bit of both and it could be improved. Examples inline.

Thanks for the feedback, really appreciate it!

backend/src/BuiltinExecution/Libs/String.fs
pbiggar: I just mean that sentinel values like this are great, and usually we'd use a value that can't be a real value, such as -1 (while here you use 0). While zero can't actually come up, you have to read the code to understand why, while -1 more clearly indicates it.

pbiggar (changes requested):

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -299,20 +299,37 @@ let fns : List<BuiltInFn> =
...
         (function
         | _, _, [ DString s; DInt first; DInt last ] ->
-          String.toEgcSeq s
-          |> Seq.toList
-          |> FSharpPlus.List.drop (int first)
-          |> List.truncate (int (last - first))
-          |> String.concat ""
-          |> DString
-          |> Ply
+          let strLength = System.Globalization.StringInfo(s).LengthInTextElements
```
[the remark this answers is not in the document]
OceanOak: It is also used to check if first is out of bounds

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -304,12 +304,16 @@ let fns : List<BuiltInFn> =
...
           // Handle negative indexes (which allow counting from the end)
-          let first = if first < 0 then strLength + int first else int first
-          let last = if last < 0 then strLength + int last else int last
+          let first =
+            if first < 0 then getLengthInTextElements (s) + int first else int first
+          let last =
+            if last < 0 then getLengthInTextElements (s) + int last else int last
 
-          if first >= strLength || first >= last then
+          if first >= getLengthInTextElements (s) || first >= last then
```
pbiggar, line 316: I hadn't noticed before but this calculates string length every time. Is this case necessary (that is, if we remove it does the `else` case come up with the same answer)? If not, better to remove it so we aren't working out the length when it's not necessary.

pbiggar (changes requested):

pbiggar:
Great, thank you! Thanks for getting this over the line!

PR #5103, opened 2023-09-27
Add support for guard clause in a match pattern

OceanOak:
Changelog:

```
Language
- Add support for guard clause in a match pattern

  match Some 1 with
  | Some x when x % 1 == 2 -> "odd"
  | _ -> "even"
```


backend/src/LibBinarySerialization/ProgramTypesToSerializedTypes.fs
```diff
@@ -430,11 +430,11 @@ module Expr =
         List.map toST fields
       )
     | PT.EMatch(id, mexpr, cases) ->
-      ST.EMatch(
-        id,
-        toST mexpr,
-        List.map (Tuple2.mapFirst MatchPattern.toST << Tuple2.mapSecond toST) cases
-      )
+      let convertCase (case : PT.MatchCase) : ST.MatchCase =
```
OceanOak, line 433: Maybe use a lambda instead of creating a convertCase function?

backend/tests/TestUtils/TestUtils.fs
```diff
@@ -670,9 +670,9 @@ module Expect =
       eq ("matchCond" :: path) e e'
 
       NEList.iter2
-        (fun ((p, v) : MatchPattern * Expr) (p', v') ->
-          matchPatternEqualityBaseFn checkIDs path p p' errorFn
-          eq (string p :: path) v v')
+        (fun branch branch' ->
```
OceanOak, line 673: case instead of branch?

backend/testfiles/execution/language/ematch.dark
```diff
@@ -645,4 +644,20 @@ module TypeErrors =
...
+
+  (match -5 with
+   | x when x > 0 -> true
+   | _ -> false) = false
+
+  (match 6 with
+   | x when x == 5 -> "fail"
+   | -6 -> "fail"
+   | 6 -> "pass"
+   | var -> "fail") = "pass"
```
OceanOak, line 663: add more tests

backend/src/LibParser/WrittenTypesToProgramTypes.fs
```diff
@@ -240,18 +240,27 @@ module Expr =
...
           Ply.List.mapSequentially
-            (fun (mp, expr) ->
+            (fun (case : WT.MatchCase) ->
               uply {
-                let mp = MatchPattern.toPT mp
-                let! expr = toPT expr
-                return (mp, expr)
+                let mp = MatchPattern.toPT case.pat
+                let! expr = toPT case.rhs
+                let! whenCondition =
```
OceanOak, line 251: whenCondition before expr

backend/src/LibExecution/RuntimeTypesToDarkTypes.fs
```diff
@@ -685,13 +696,27 @@ module Expr =
         cases
         |> List.collect (fun case ->
           match case with
-          | DTuple(pattern, expr, _) ->
-            [ (MatchPattern.fromDT pattern, fromDT expr) ]
+          | DRecord(_, _, _, fields) ->
+            let whenCondition =
+              match Map.get "whenCondition" fields with
```
OceanOak, line 701: Use map.tryFind instead of Map.get

backend/src/LibExecution/RuntimeTypesAst.fs
```diff
@@ -388,19 +394,21 @@ let rec postTraversalAsync
...
             Ply.NEList.mapSequentially
-              (fun (pattern, expr) ->
+              (fun case ->
                 uply {
-                  let! pattern = postTraverseMatchPattern pattern
-                  let! expr = r expr
-                  return (pattern, expr)
+                  let! pattern = postTraverseMatchPattern case.pat
+                  let! expr = r case.rhs
+                  let! whenCondition = Ply.Option.map r case.whenCondition
```
OceanOak, line 406: whenCondition before expr

backend/src/BuiltinExecution/Libs/NoModule.fs
```diff
@@ -152,7 +152,8 @@ and equalsExpr (expr1 : Expr) (expr2 : Expr) : bool =
     equalsExpr target1 target2
     && NEList.length cases1 = NEList.length cases2
     && NEList.forall2
-      (fun (p1, e1) (p2, e2) -> equalsMatchPattern p1 p2 && equalsExpr e1 e2)
+      (fun case1 case2 ->
+        equalsMatchPattern case1.pat case2.pat && equalsExpr case1.rhs case2.rhs)
```
StachuDotNet, line 156: probably need an additional check here for `equalsSomethingOptional case1.whenCondition case2.whenCondition`?

backend/src/LibBinarySerialization/SerializedTypes.fs
```diff
@@ -281,7 +281,7 @@ type Expr =
   | EUnit of id
   | EConstant of id * NameResolution<ConstantName.ConstantName>
   | ELet of id * LetPattern * Expr * Expr
-  | EIf of id * cond : Expr * thenExpr : Expr * elseExpr : option<Expr>
+  | EIf of id * cond : Expr * thenExpr : Expr * elseExpr : Option<Expr>
```
StachuDotNet, line 284: Nice - it's good for us to be consistent here (Option, not option)

backend/tests/TestUtils/TestUtils.fs
```diff
@@ -670,9 +670,9 @@ module Expect =
...
+          matchPatternEqualityBaseFn checkIDs path branch.pat branch'.pat errorFn
+          eq (string branch.pat :: path) branch.rhs branch'.rhs)
```
StachuDotNet, line 675: similar to in NoModule.fs, we likely should check for equal whenConditions here

backend/src/LibExecution/Interpreter.fs
```diff
@@ -633,14 +633,23 @@ let rec eval
...
-            let! r = eval state tlid tst newSymtable rhsExpr
+          let! whenCondition =
+            uply {
+              match case.whenCondition with
+              | Some whenCondition ->
+                match! eval state tlid tst newSymtable whenCondition with
+                | DBool b -> return b
+                | _ -> return false
+              | None -> return true
+            }
```
StachuDotNet, line 650: This worked out tidily! Maybe rename from `whenCondition` to `passesWhenCondition`?
and rename `passes` to `passesPattern`?

backend/tests/TestUtils/TestUtils.fs
```diff
@@ -670,9 +670,15 @@ module Expect =
...
-          eq (string p :: path) v v')
+        (fun branch branch' ->
+          matchPatternEqualityBaseFn checkIDs path branch.pat branch'.pat errorFn
+          match branch.whenCondition, branch'.whenCondition with
+          | Some cond, Some cond' -> eq ("whenCondition" :: path) cond cond'
+          | None, None -> ()
+          | _ ->
+            errorFn ("whenCondition" :: path) (string actual) (string expected)
+            ()
+          eq (string branch.pat :: path) branch.rhs branch'.rhs)
```
OceanOak, line 681: I am not sure what path is used for, could you please explain why are we using branch.pat instead of "rhs"?

StachuDotNet: the `path` is used in error-reporting to show more exactly where 2 things weren't equal.
I think your intuiton is right, though, that the specific impl here is a bit weird.
I think what we really want is a path of `expr -> case [x] -> [part where it failed]`.

Here's an adjustment towards this that I think might help clarify

```
NEList.iteri2
  (fun i branch branch' ->
    let path = "Case {i}" :: path

    matchPatternEqualityBaseFn checkIDs ("pat" :: path) branch.pat branch'.pat errorFn

    match branch.whenCondition, branch'.whenCondition with
    | Some cond, Some cond' -> eq ("whenCondition" :: path) cond cond'
    | None, None -> ()
    | _ ->
      errorFn ("whenCondition" :: path) (string actual) (string expected)
      ()

    eq ("rhs" :: path) branch.rhs branch'.rhs)
  branches
  branches'
```

StachuDotNet: maybe `let path = $"Case {i} - {branch.pat}" :: path` there would actually be a better compromise, so you have a choice of 'counting' the cases when debugging, or simply referring to whatever pattern is shown

StachuDotNet: We should probably have a "make sure the 2 exprs have the same # of cases` check, right before the NEList.iter2 - see `check path (List.length typeArgs) (List.length typeArgs')` in `EApply` just above in that fn

StachuDotNet (approved):
Looks great to me - just the minor update in TestUtils as noted, otherwise I think this is OK to merge

StachuDotNet (approved):

backend/src/BuiltinExecution/Libs/NoModule.fs
```diff
@@ -152,7 +152,13 @@ and equalsExpr (expr1 : Expr) (expr2 : Expr) : bool =
     equalsExpr target1 target2
     && NEList.length cases1 = NEList.length cases2
     && NEList.forall2
-      (fun (p1, e1) (p2, e2) -> equalsMatchPattern p1 p2 && equalsExpr e1 e2)
+      (fun case1 case2 ->
+        equalsMatchPattern case1.pat case2.pat
+        && match case1.whenCondition, case2.whenCondition with
```
pbiggar, line 157: I don't like having `match` in a `&&` condition - can we do this differently.

backend/src/LibExecution/Interpreter.fs
```diff
@@ -633,14 +633,23 @@ let rec eval
...
+          let! passesWhenCondition =
+            uply {
+              match case.whenCondition with
+              | Some whenCondition ->
+                match! eval state tlid tst newSymtable whenCondition with
+                | DBool b -> return b
+                | _ -> return false
+              | None -> return true
+            }
+          if matchResult = None && passesPattern && passesWhenCondition then
```
pbiggar, line 651: We know that matchResult must be None here so we don't need to check it.

pbiggar: We shouldn't evaluate `whenCondition` unless `passesPattern` is true. Can you add a test for this specific situation as well?

backend/src/LibParser/FSharpToWrittenTypes.fs
```diff
@@ -487,10 +487,14 @@ module Expr =
...
     // | ...
     | SynExpr.Match(_, cond, cases, _, _) ->
-      let convertCase
-        (SynMatchClause(pat, _, expr, _, _, _) : SynMatchClause)
-        : WT.MatchPattern * WT.Expr =
-        (MatchPattern.fromSynPat pat, c expr)
+      let convertCase (clause : SynMatchClause) : WT.MatchCase =
+        match clause with
+        | SynMatchClause(pat, Some whenExpr, expr, _, _, _) ->
+          { pat = MatchPattern.fromSynPat pat
```
pbiggar, line 493: You can combine these into one case (using Option.map on whenExpr).

backend/testfiles/execution/language/ematch.dark
```diff
@@ -645,4 +644,50 @@ module TypeErrors =
...
+
+  (match (5, "hello") with
+   | (x, y) when x > 0 && PACKAGE.Darklang.Stdlib.String.length y == 5 -> "pass"
+   | _ -> "fail") = "pass"
+
+  (match -5 with
+   | x when y > 0 -> true
+   | _ -> false) = Builtin.Test.derrorMessage "There is no variable named: y"
+
+  (match 5 with
```
pbiggar, line 691: These are great tests. Can you also check what happens if the result of whenCondition is non-boolean?

backend/src/LibExecution/Interpreter.fs
```diff
@@ -633,14 +633,23 @@ let rec eval
...
           let newSymtable = Map.mergeFavoringRight st (Map.ofList newDefs)
-          if matchResult = None && passes then
-            let! r = eval state tlid tst newSymtable rhsExpr
+          let! passesWhenCondition =
+            uply {
+              match case.whenCondition with
+              | Some whenCondition ->
+                match! eval state tlid tst newSymtable whenCondition with
+                | DBool b -> return b
+                | _ -> return false
```
pbiggar, line 648: This should be a runtime error, as it's the wrong type (should be a bool). You can raise the same error as `If` statements where the condition has the wrong type.

pbiggar (changes requested):
Looks great! A couple small changes should get this over the line.

backend/src/LibExecution/Interpreter.fs
OceanOak: I've added the following test 
```
  (match 5 with
   | (x, y) when x > 0 -> true
   | _ -> false) = Builtin.Test.derrorMessage
    "Cannot match Int value 5 with a Tuple pattern"
 ```
    
But, I wasn't sure if you meant adding something like this as well
```
    let! passesWhenCondition =
      uply {
        match case.whenCondition with
        | Some whenCondition when passesPattern ->
          match! eval state tlid tst newSymtable whenCondition with
          | DBool b -> return b
          | _ -> return errStr id "When condition should be a boolean"
        | _ -> return true
      }
```
Could you please clarify?

Comment (pbiggar) [reply backend/src/LibExecution/Interpreter.fs]: A test like this:

```fsharp
match 5 with
| 2 when x > 2 -> false
| 3 -> true
```

Above, `x > 2` should never be executed, but I think it will with the current code.

A less contrived example:


```fsharp
match Error 5 with
| Ok x when x > 2 -> false
| Error _ -> true
```

OceanOak: Ah I see, this was returning `there is no variable named x` which is wrong. Now it doesn't evaluate it.
```
match 5 with
| 2 when x > 2 -> false
| 3 -> true
```

pbiggar: Yeah, exactly!

PR #5104, opened 2023-09-28
backpedal a lot of Dval.record and .enum usage

StachuDotNet:
PT2DT and RT2DT don't need to go through the Dval. functions - trying to ensure safe construction of dvals _everywhere_ was causing more pain that it's worth. Maybe we'll circle back some day here.

I was intending for this commit to be part of a larger PR, but that work is taking a bit of time, so I'll merge this in the meantime.


pbiggar:
Nice!

PR #5107, opened 2023-09-30
Finish JSON improvements

pbiggar:
This does the remainder of JSON error handling improvements, esp adding a Json.ParserError type that is returned for errors instead of strings. I believe this handlers all broken things in the Stdlib.Json module.

One open question I have is whether putting a TypeReference in the type is a good idea. This isn't a RuntimeError that will be handled by the runtime, it's a user facing error. So I'm not quite sure what to do with this.

As part of this:
- anything that the typechecker should have caught is now an InternalException
- anything that the user could trigger during parsing is now a parseError
- `serialize` doesn't error (but there is still a typechecker error if the type and data don't match, or if you use an unsupported type)

Unrelated fixes:
- add a special error message for incorrect types to `unwrap` 
- don't default to trying to parse a package constant if package function parsing failed (lost the error message)
- stringify error messages with toRepr to be more readable
- switch to 63 bits for IDs
- Call all Chars "Char"
- if we call a function and it threw a runtimeError with no DvalSource, use the caller's DvalSource
- make sure we always catch testing exceptions using postTestExecutionHook

### DIFF
--- a/CODING-GUIDE.md
+++ b/CODING-GUIDE.md
@@ -21,7 +21,7 @@
   - String (not Str)
   - Float
   - Bool (not Boolean)
-  - Char (not Character)
+  - Char (not Char)
   - Uuid (not UUID)
   - Dict (not Dictionary)
 
--- a/packages/darklang/stdlib/json.dark
+++ b/packages/darklang/stdlib/json.dark
@@ -27,10 +27,18 @@ module Darklang =
             PACKAGE.Darklang.LanguageTools.RuntimeTypes.TypeReference *
             String *
             JsonPath.JsonPath
-          | EnumExtraField of String * JsonPath.JsonPath
+          | EnumExtraField of rawJson: String * JsonPath.JsonPath
           | EnumMissingField of
             PACKAGE.Darklang.LanguageTools.RuntimeTypes.TypeReference *
-            Int *
+            index: Int *
+            JsonPath.JsonPath
+          | EnumInvalidCasename of
+            PACKAGE.Darklang.LanguageTools.RuntimeTypes.TypeReference *
+            caseName: String *
+            JsonPath.JsonPath
+          | EnumTooManyCases of
+            PACKAGE.Darklang.LanguageTools.RuntimeTypes.TypeReference *
+            cases: List<String> *
             JsonPath.JsonPath
           | NotJson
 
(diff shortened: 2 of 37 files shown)


pbiggar:
@StachuDotNet I'm going to merge this, but I'd love some feedback on the TypeReference question.

CODING-GUIDE.md
```diff
@@ -21,7 +21,7 @@
   - String (not Str)
   - Float
   - Bool (not Boolean)
-  - Char (not Character)
+  - Char (not Char)
```
StachuDotNet, line 24: we should revert this change

packages/darklang/stdlib/json.dark
```diff
@@ -0,0 +1,113 @@
...
+            PACKAGE.Darklang.LanguageTools.RuntimeTypes.TypeReference *
+            caseName: String *
+            JsonPath.JsonPath
+          | EnumTooManyCases of
+            PACKAGE.Darklang.LanguageTools.RuntimeTypes.TypeReference *
+            cases: List<String> *
+            JsonPath.JsonPath
+          | RecordDuplicateField of fieldName: String * JsonPath.JsonPath
+          | RecordMissingField of fieldName: String * JsonPath.JsonPath
+          | NotJson
```
StachuDotNet, line 45: > One open question I have is whether putting a TypeReference in the type is a good idea. This isn't a RuntimeError that will be handled by the runtime, it's a user facing error. So I'm not quite sure what to do with this.

Hmm. Is the concern here that users would then be exposed to `TypeReference`, which feels odd?

packages/darklang/stdlib/json.dark
```diff
@@ -0,0 +1,113 @@
...
+
+          // Can't parse JSON with multiple fields (`Enh` and `Extra`) as type `PrettyLikely` at path: `root`
+          | EnumTooManyCases(typ, cases, path) ->
+            let cases = cases |> List.map (fun c -> $"`{c}`") |> String.join ", "
+
+            $"Can't parse JSON with multiple fields ({cases}) as type `"
+            ++ PACKAGE.Darklang.PrettyPrinter.RuntimeTypes.typeReference typ
+            ++ "` at path: `"
+            ++ JsonPath.toString path
+            ++ "`"
```
StachuDotNet, line 94: I don't think this error case is super clear without users knowing how we serialize Enums.
(i.e. "Can't parse JSON with multiple fields" in particular is missing some context)

"For the type `type Option<'x> = Some<'x> | None`, a value `Some 1` is serialized as `{ Some: [1] }` - if an object such as `{ Some: [1], Other: [] }` is presented, we're not sure how to parse it" may be a useful addendum?

packages/darklang/stdlib/json.dark
pbiggar: yes, exactly.

packages/darklang/stdlib/json.dark
pbiggar: I think users will have to know how we serialize Enums or else they won't know what's wrong.

This error isn't a real error. Unlike an RTE or other compiler error, this is something the user can use to print the Error enum at runtime. So looking up types feels outside the scope here, it's really just a pretty print.

It would be nice to do better, but I can't see how with that constraint.

PR #5110, opened 2023-10-02
DvalSource/Caller/tlid updates

pbiggar:
DvalSource has been broken for a while (perhaps a few years). This attempts to untangle it and restore it to its original purpose: a way to identify the Darklang code that was the source of the "problem", and to propagate it appropriately.

As such, I got rid of `state.tlid` and added `state.caller` - those things represent what we're really interested in. It also fixes the signatures of a lot of the interpreter to allow us to consistently get this right.


pbiggar:
Removed the TST changes

PR #5112, opened 2023-10-02
Add warnings

pbiggar:
Enable unused variable warning. Caught a few things (nothing broken in practice).

Also, it seems like the build parallelization thing I added before was being overwritten in each fsproj (and thus didnt work), so this moves OtherFlags to being set in one place (Directory.Build.props) instead of each fsproj (might be something to expand on later).


pbiggar:
Do we want this, @StachuDotNet?

StachuDotNet:
Yes, I really appreciate these warnings, and this setup looks great to me

StachuDotNet (approved):

PR #5127, opened 2023-10-09
Add cloud run

pbiggar:
This is enough to enable BwdServer running on Cloud run.

Doesn't do CI.

Fixes building docker containers, pushing to the cloud, running cloud run (including configuring secrets/IAM), and starting up BwdServer.

### DIFF
--- a/tf/service_env_vars.tf
+++ b/tf/service_env_vars.tf
@@ -12,8 +12,10 @@ variable "service_env_vars" {
     # Ports
 
     ## BwdServer
-    "DARK_CONFIG_BWDSERVER_BACKEND_PORT"    = "11001"
-    "DARK_CONFIG_BWDSERVER_KUBERNETES_PORT" = "11002"
+    "DARK_CONFIG_BWDSERVER_BACKEND_PORT"           = "11001"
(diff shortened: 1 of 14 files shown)


tf/service_env_vars.tf
```diff
@@ -12,8 +12,10 @@ variable "service_env_vars" {
     # Ports
 
     ## BwdServer
-    "DARK_CONFIG_BWDSERVER_BACKEND_PORT"    = "11001"
-    "DARK_CONFIG_BWDSERVER_KUBERNETES_PORT" = "11002"
+    "DARK_CONFIG_BWDSERVER_BACKEND_PORT"           = "11001"
+    "DARK_CONFIG_BWDSERVER_KUBERNETES_PORT"        = "11002"
+    "DARK_CONFIG_BWDSERVER_DANGER_BACKEND_PORT"    = "11001"
+    "DARK_CONFIG_BWDSERVER_DANGER_KUBERNETES_PORT" = "11002"
```
StachuDotNet, line 18: I guess we need to assign these to make Config.fs up the chain not break?

StachuDotNet (approved):

tf/service_env_vars.tf
pbiggar: Right. We can remove them later

PR #5129, opened 2023-10-11
Add a new IntRuntimeError and replace some string errors

OceanOak:
No changelog

This PR:
- Introduces a new IntRuntimeError and uses it for Int.mod, Int.power, and Int.divide
  ```
    type Error =
      | DivideByZeroError
      | OutOfRange
      | NegativeExponent
      | NegativeModulus
      | ZeroModulus
  ```
- Replaces Float.parse string error with a custom IntParseError
- Replaces Int.parse string error with a custom IntParseError
- Updates Int.power to return an `Int` instead of `Result`
- Replaces Uuid.parse string error with a custom UuidParseError

### DIFF
--- a/backend/src/BuiltinExecution/Libs/Float.fs
+++ b/backend/src/BuiltinExecution/Libs/Float.fs
@@ -14,6 +14,18 @@ let constants : List<BuiltInConstant> = []
 
 let fn = fn [ "Float" ]
 
+module FloatParseError =
+  type FloatParseError = | BadFormat
+
+  let toDT (e : FloatParseError) : Dval =
+    let (caseName, fields) =
+      match e with
+      | BadFormat -> "BadFormat", []
+
+    let typeName =
+      TypeName.fqPackage "Darklang" [ "Stdlib"; "Float" ] "FloatParseError" 0
+    DEnum(typeName, typeName, [], caseName, fields)
+
 let fns : List<BuiltInFn> =
   [ { name = fn "ceiling" 0
       typeParams = []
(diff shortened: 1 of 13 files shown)


backend/src/BuiltinExecution/Libs/Int.fs
```diff
@@ -339,20 +356,40 @@ let fns : List<BuiltInFn> =
...
+                @"^[+-]?\d+$"
+              )
+            )
+          then
+            IntParseError.BadFormat |> IntParseError.toDT |> resultError |> Ply
+          else
+            match System.Int64.TryParse(s) with
+            | true, value -> value |> DInt |> resultOk |> Ply
+            | false, _ ->
+              IntParseError.OutOfRange |> IntParseError.toDT |> resultError |> Ply
```
StachuDotNet, line 392: I was imagining this implementation to be different:
- keep the `try` as it was
- rather than one big blanket `with _e`, capture specific kinds of exceptions - see HttpClient implementation

StachuDotNet: for example, the bit in httpClient.fs where we 
```fsharp
with
      | :? TaskCanceledException ->
```

OceanOak: This is much better. Thanks!

backend/src/BuiltinExecution/Libs/Float.fs
```diff
@@ -269,20 +281,31 @@ let fns : List<BuiltInFn> =
...
+        let resultError = Dval.resultError KTFloat (KTCustomType(typeName, []))
         (function
         | _, _, [ DString s ] ->
           try
             float (s) |> DFloat |> resultOk
-          with e ->
-            "Expected a String representation of an IEEE float"
-            |> DString
-            |> resultError
+          with :? System.FormatException ->
```
OceanOak, line 307: I don't think we need to catch the `OverflowException`
here since the result is rounded to infinity, wdyt?

> In .NET Core 3.0 and later, values that are too large to represent are rounded to [PositiveInfinity] or [NegativeInfinity] as required by the IEEE 754 specification. In prior versions, including .NET Framework, parsing a value that was too large to represent resulted in failure.

https://learn.microsoft.com/en-us/dotnet/api/system.double.parse?view=net-7.0

packages/darklang/stdlib/int.dark
```diff
@@ -2,6 +2,10 @@ module Darklang =
   module Stdlib =
     module Int =
 
+      type IntParseError =
```
pbiggar, line 5: Better to call this ParseError

packages/darklang/stdlib/float.dark
```diff
@@ -1,6 +1,9 @@
 module Darklang =
   module Stdlib =
     module Float =
+
+      type FloatParseError = | BadFormat
```
pbiggar, line 5: I would call this ParseError, as the Float is already implied

pbiggar (changes requested):
Looks great! The only change I see is that we should remove the float/int/UUID prefixes from the error names because they're impicit.

PR #5133, opened 2023-10-19
Improve parser errors reported to VS Code extension

StachuDotNet:
- defines new `ParserException`, which inherits our `InternalException` (so that any code capturing `InternalException`s still capture exceptions of this more granular type)
- where possible, updates `LibParser` to extract the _range_ of parser failure and raise the new exception type
- pass this information along to our VS Code exception, through the `Darklang.LanguageTools.LanguageServer.getDiagnostics` function


vscode-extension/server/src/server.ts
```diff
@@ -151,8 +152,30 @@ connection.onDidChangeConfiguration(change => {
...
+  }
+};
+documents.onDidSave(change => {
+  changeToProcessNext = change;
+  processChange();
+});
+documents.onDidChangeContent(change => {
+  changeToProcessNext = change;
+  processChange();
 });
```
StachuDotNet, line 179: maybe this is fancier than it needs to be - in any case, it'll go away when we get rid of this server.ts in favor of a fully Dark-based LSP server

PR #5136, opened 2023-10-24
Initial Treesitter Bindings for Darklang

StachuDotNet:
The goal of this PR is to take Darklang source code as a String, parse it with `tree-sitter`, and have minimal types/functions in place to use the parsed syntax tree _in Darklang_ - this end result is demonstrated by [_this_ test file](https://github.com/darklang/dark/blob/0dbb7723c6bbd00bef32da8a748d0057e4ced9d4/backend/testfiles/execution/stdlib/parser.dark#L24)

- add .NET bindings around `tree-sitter` in a new `LibTreeSitter` C# project (mostly P/Invoke work)
  - forked from [`dotnet-tree-sitter` ](https://github.com/profMagija/dotnet-tree-sitter)
  - to support this, we now clone and build `tree-sitter` to a `tree-sitter.so` file on 'current' architecture, in the `Dockerfile`, and copy this to the appropriate directory just after the container is built
- add .NET bindings around `tree-sitter-darklang` in a new `LibTreeSitter` C# project
  - also ported from `dotnet-tree-sitter`
  - to support this, we now build `tree-sitter-darklang` to a `tree-sitter-darklang.so` shared library and copy this file to the project accordingly, within `scripts/build/build-parser`
  - tested in `backend/tests/Tests/TreeSitter.Tests.fs`
- _start_ to port some of this^ to F# - unused, unreferenced, and abandoned for now -- can circle back later
- on top of the .NET `tree-sitter` and `tree-sitter-darklang` bindings, adds Darklang bindings, via
  - a package type
    ```fsharp
    type ParsedNode =
      { fieldName: String; typ: String; text: String sourceRange: Range; children: List<ParsedNode> }

    // supporting types
    type Point = { row: Int; column: Int }
    type Range = { start: Point; end_: Point }
    ```
  - a builtin function `let Builtins.Parser.parseToSimplifiedTree (sourceCode: String): ParsedNode`
- tests those bindings against a simple single function definition: `let self (i: Int): Int = i`
  (see `backend/testfiles/execution/stdlib/parser`)
- bonus: install Zig for cross-compilation of `tree-sitter-darklang` - unused, but the cross-compilation is working for all 64-bit targets I tried (not working for 32-bit ones but we can circle back later)

Getting this to build and pass in CI depends on an update to `darklang/dockerfile`: https://github.com/darklang/dockerfile/pull/7. I'll need to update our circleci config once more if/when that is merged, to point to the new image.

Some questions:
- [ ] is `_vscode-post-start-command` a reasonable place to copy the `tree-sitter.so` into the `backend` directory? I couldn't find any better place to take care of this
- [ ] related, does the `.circleci/config.yml` update to do the same, seem OK?
- [ ] for now, I've put the `Parser` Builtin module in the `BuiltinLibExecution` library - OK?


parser/tree-sitter-darklang/README.md
```diff
@@ -40,3 +40,11 @@ a wrapper around the `.wasm` file, and the `.wasm` file calls upon language-spec
...
 
 Relevant documentation: https://github.com/tree-sitter/tree-sitter/tree/master/lib/binding_web
+
+### just tossing this somewhere
+
+```
+ gcc -o tree-sitter-darklang.so -shared src/parser.c -Isrc -O3 -fPIC
+```
+
+to get the `.so` file
```
StachuDotNet, line 50: TODO -- redundant, remove

scripts/build/build-parser
```diff
@@ -11,13 +11,19 @@ npm i
...
+
+# copy the WASM bindings to vscode-extension/static
 mkdir -p vscode-extension/static/tree-sitter
-cp parser/tree-sitter-darklang/bindings/tree-sitter-darklang.wasm vscode-extension/static/tree-sitter/
\ No newline at end of file
+cp parser/tree-sitter-darklang/bindings/tree-sitter-darklang.wasm vscode-extension/static/tree-sitter/
+
+# copy .so files for tree-sitter and tree-sitter-darklang
+#   to corresponding .NET projects
+printf "!! do this manually !!\n  get the platform-dependent tree-sitter.so file from\n  https://github.com/davidelettieri/treesitter-bindings/raw/main/tree-sitter-packages/libtreesitter.runtime.linux-x64/libtree-sitter.so\n  and put it in backend/src/LibTreeSitter/\n  , renamed as tree-sitter.so\n"
```
StachuDotNet, line 27: ugly

StachuDotNet: some options:
- just include the .so in our code
- ...

actually, I'll just do that - it's only 200KB, and changes rarely - not sure why I was avoiding that.

scripts/build/build-parser
```diff
@@ -11,13 +11,19 @@ npm i
...
+cp parser/tree-sitter-darklang/tree-sitter-darklang.so backend/src/LibTreeSitter.Darklang
+
```
StachuDotNet, line 29: TODO: also call the generate_tree-sitter_bindings script here - forgot to.

StachuDotNet:
misc fwiw, I haven't even read all of the source in LibTreeSitter - mostly, copied from `dotnet-tree-sitter`, it _worked_, and I didn't look into it much beyond that.

StachuDotNet:
The build-parser script is taking 2mins in CI - something to deal with probably before merging

scripts/generate_tree-sitter_bindings.py
```diff
@@ -0,0 +1,280 @@
+# TODO: port this to a Darklang script
```
pbiggar, line 1: Could you document where this came from? I'm thinking also it doesn't belong in scripts, probably scripts/build?

scripts/devcontainer/_vscode-post-start-command
```diff
@@ -2,6 +2,11 @@
 
 set -euo pipefail
 
+
+echo "Copying tree-sitter.so to backend/src/LibTreeSitter"
```
pbiggar, line 6: I think this is a dependency for building LibTreeSitter in dotnet? I would imagine it would be copied as part of the msbuild fsproj file for that directory?

pbiggar: Definitely this isn't the right place though.

A second place option might be the compile script.

backend/src/LibTreeSitter.Darklang/README.md
```diff
@@ -0,0 +1,8 @@
+This library provides helper functions _specific to Darklang_, wrapping around the
+bindings for [tree-sitter](https://tree-sitter.github.io/tree-sitter/) provided by
+the adjacent `LibTreeSitter` project.
+
+The `DarklangLanguage.cs` file is written manually, and `Generated.cs` file is a set
+of helper functions generated by `./script/generate_tree-sitter_bindings.py`.
```
pbiggar, line 6: You mean ./scripts/ I think

pbiggar (approved):
Looks great!

scripts/generate_tree-sitter_bindings.py
StachuDotNet: it's really not useful to us now; I'm going to delete it and bring a Dark version of it back later if useful

scripts/devcontainer/_vscode-post-start-command
StachuDotNet: > I think this is a dependency for building LibTreeSitter in dotnet? 
this is a dependency for _running_ LibTreeSitter - it's the shared library that we P/Invoke with.

`tree-sitter.so` is only built once, as part of the container being set up, and then copied to that .NET project to be invoked.
Everything in `compile` seems to be done based on some action occurring, so it doesn't quite seem it'll fit.

Additionally, because `tree-sitter` is built outside of the `app` directory, I believe `compile` doesn't have access to the file for copying - I tried in a few other scripts as well, and `_vscode-post-start-command` seemed to be the only one that (1) had access and (2) only runs once.

pbiggar: The copying shouldn't only run once though, it should for example be run again if you delete the dotnet Build directory and build again.

Doing it in a .fsproj file seesm most apppropriate.

PR #5139, opened 2023-10-30
WIP: Add support for Int128

OceanOak:
Changelog:

```
Area of change
- details
```

alternatively:

No changelog


backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -92,6 +92,8 @@ let rec dvalToSql
     | TVariable _, DUInt16 i
     | TUInt16, DUInt16 i -> return Sql.int64 (int i), TUInt16
 
+    // TODO: Int128, UInt128
```
OceanOak, line 95: Not sure what to do here yet

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -562,6 +566,8 @@ let rec lambdaToSql
           let name = randomString 10
           return $"(@{name})", [ name, Sql.int64 (int v) ], TUInt16
 
+        //TODO: Int128, UInt128
```
OceanOak, line 569: Not sure what to do here yet

backend/testfiles/execution/stdlib/json.dark
```diff
@@ -403,6 +403,113 @@ module UInt16 =
...
+      "Can't parse JSON `{ \"key\": 42 }` as type `Int128` at path: `root`"
+
+    (Builtin.Json.parse<Int128> "42.5") |> err = Result.Error
+      "Can't parse JSON `42.5` as type `Int128` at path: `root`"
+
+    (Builtin.Json.parse<Int128> "\"42\n\"") |> err = Result.Error "Not JSON"
+
+    // (-1.7014118346046923E+38) this result is out of range for Int128 and is wrapped around to the min value
+    Builtin.Json.parse<Int128> "-170141183460469231731687303715884105800" = Result.Ok
+      -170141183460469231731687303715884105728Q
```
OceanOak, line 459: Not sure why it wraps around for negative out-of-range values instead of returning an out-of-range error, as it does for positive out-of-range values

OceanOak:
I've tried to add serialization test values, but encountered a compilation error regarding numeric literals
Error:
```
This numeric literal requires that a module 'NumericLiteralQ' defining functions FromZero, FromOne, FromInt32, FromInt64 and FromString be in scope
```

I've added these two modules to address the compilation error 
```
module NumericLiteralQ =
  let FromZero () = System.Int128.Zero
  let FromOne () = System.Int128.One
  let FromInt32 (i : int) = System.Int128.Parse(string i)
  let FromInt64 (i : int64) = System.Int128.Parse(string i)
  let FromString (s : string) = System.Int128.Parse(s)

module NumericLiteralZ =
  let FromZero () = System.UInt128.Zero
  let FromOne () = System.UInt128.One
  let FromInt32 (i : int) = System.UInt128.Parse(string i)
  let FromInt64 (i : int64) = System.UInt128.Parse(string i)
  let FromString (s : string) = System.UInt128.Parse(s)
  ```
  
However, after resolving that, running the tests throws an exception (see screenshot), likely because MessagePack serializer does not know how to handle `System.int128` and `System.uint128`

Screenshot 2023-10-31 at 13 29 26


Here's the snippets causing the exception when the tests are run:

Serialization.TestValues 
```   
   // exprs
     PT.ELet(
       831830078UL,
       PT.LPVariable(7567128UL, "int128"),
       PT.EInt128(7567128UL, 170141183460469231731687303715884105727Q),
       PT.ELet(
         831830079UL,
         PT.LPVariable(7567129UL, "uint128"),
         PT.EUInt128(7567128UL, 340282366920938463463374607431768211455Z),
         e
       )
     )
           
           
  // constValue
  PT.Const.CInt128(1Q)
  PT.Const.CUInt128(1Z)
```


And the relevant part of the SerializationTypes.fs
```
type Expr =
  | EInt of id * int64
    ...
  | EInt128 of id * System.Int128
  | EUInt128 of id * System.UInt128
  ```

backend/tests/TestUtils/TestUtils.fs
```diff
@@ -1261,6 +1277,7 @@ let interestingDvals : List<string * RT.Dval * RT.TypeReference> =
        [ Dval.uint16 16us ]
      ),
      TypeReference.option TUInt16)
+    // TODO: Int128 and UInt128
```
OceanOak, line 1280: Tried adding the following values to the `interestingDvals` List; However, the roundtripping tests failed with `Value in Datastore does not match Datastore expected type` due to the lack of database support for 128-bit ints at the moment

```
    ("option8",
     DEnum(
       Dval.optionType,
       Dval.optionType,
       Dval.ignoreAndUseEmpty [ VT.int128 ],
       "Some",
       [ Dval.int128 12Q ]
     ),
     TypeReference.option TInt128)
    ("option9",
     DEnum(
       Dval.optionType,
       Dval.optionType,
       Dval.ignoreAndUseEmpty [ VT.uint128 ],
       "Some",
       [ Dval.uint128 12Z ]
     ),
     TypeReference.option TUInt128)

```
This also required the `NumericLiteralQ`and `NumericLiteralZ`modules mentioned in this comment https://github.com/darklang/dark/pull/5139#issuecomment-1787177613

StachuDotNet:
I haven't looked into this much but worth reading through https://github.com/MessagePack-CSharp/MessagePack-CSharp#built-in-supported-types, towards how we could extend support for int128 in MessagePack

backend/testfiles/execution/stdlib/json.dark
pbiggar: I recall adding a check for this during parsing for some integers, though I could be wrong?

backend/tests/Tests/Serialization.TestValues.fs
```diff
@@ -21,6 +21,21 @@ let uuid = System.Guid.Parse "31d72f73-0f99-5a9b-949c-b95705ae7c4d"
 let tlid : tlid = 777777928475UL
 let tlids : List<tlid> = [ 1UL; 0UL; uint64 -1L ]
 
+
+module NumericLiteralQ =
```
pbiggar, line 25: What's going on here?

pbiggar (changes requested):
Looks great! A few questions remain.

The main concern I always have with PRs involving ints is have we tested the edge cases well. Is there any more testing that could improve our coverage there?

backend/src/LibBinarySerialization/BinarySerialization.fs
```diff
@@ -14,6 +15,110 @@ module PT2ST = ProgramTypesToSerializedTypes
...
+  let ToBytes (value : System.Int128) =
+    if value = System.Int128.Zero then
+      Array.zeroCreate 16
+    else
+      let low = uint64 value
+      let high = uint64 (value >>> 64)
+      let lowBytes = BitConverter.GetBytes(low)
+      let highBytes = BitConverter.GetBytes(high)
+
+      Array.concat [ lowBytes; highBytes ]
```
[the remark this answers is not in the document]
OceanOak: I was following the GetBytes order low bytes are at the beginning of the array and high bytes are at the end of the array. 
Also when debugging the value `1` I was getting the following, so it made sense to put them in that order
 low bytes :`01-00-00-00-00-00-00-00`
 highBytes :`00-00-00-00-00-00-00-00`
 
 I think we can use a different order if we swap the extraction of the low and high bytes here:
 ```
      let low = uint64 (value >>> 64)
      let high =  uint64 value
 ```
 
 I don't know which order is correct, when looking for an answer I stumbled upon the Endianness concept, I wasn't sure what to use so I followed the GetBytes order.

backend/src/LibBinarySerialization/BinarySerialization.fs
```diff
@@ -14,6 +15,110 @@ module PT2ST = ProgramTypesToSerializedTypes
...
+    if bytes.Length <> 16 then
+      raise (ArgumentException "Byte array must be exactly 16 bytes long.")
+
+    let lowBytes = bytes[0..7]
+    let highBytes = bytes[8..15]
+
+    let low = BitConverter.ToUInt64(lowBytes, 0)
+    let high = BitConverter.ToUInt64(highBytes, 0)
+
+    System.Int128(uint64 high, uint64 low)
```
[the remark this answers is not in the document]
OceanOak: No it is not. Sorry, I forgot to remove it

backend/src/BuiltinExecution/Libs/Int128.fs
```diff
@@ -0,0 +1,371 @@
...
+module Dval = LibExecution.Dval
+
+let types : List<BuiltInType> = []
+
+let modules = [ "Int128" ]
+let fn = fn modules
+let constant = constant modules
+
+let constants : List<BuiltInConstant> =
+  [ { name = constant "zero" 0
```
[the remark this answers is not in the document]
OceanOak: The reason for adding this constant is its use within the `Int128.absoluteValue` function, specifically in the line `if (Builtin.Int128.lessThan a Builtin.Int128.zero) then ...`

**Using 0 instead of Builtin.Int128.zero :** I get an error stating that Int128.lessThan's 2nd argument should be an Int128. However, an Int (0) was passed instead.

**Using 0Q instead of Builtin.Int128.zero :** The loading packages to canvas step fails with error saving function to canvas
details:
  ```
  Type: function; Json: {"body":{"EIf":[5413835516771744658,{"EApply":[5967867546811000111,{"EFnName":[8251259138368260562,{"Ok":[{"BuiltIn":[{"modules":["Int128"],"name":{"FnName":["lessThan"]},"version":0}]}]}]},[],[{"EVariable":[4596083514880703310,"a"]},{"EInt128":[6792095667907226909,0]}]]},{"EApply":[3771221013789687638,{"EFnName":[2615256928638251462,{"Ok":[{"BuiltIn":[{"modules":["Int128"],"name":{"FnName":["negate"]},"version":0}]}]}]},[],[{"EVariable":[727526345080291056,"a"]}]]},{"Some":[{"EVariable":[2686468324135877090,"a"]}]}]},"deprecated":{"NotDeprecated":[]},"description":"","id":"2934c4f0-c135-4d77-a87d-3443a912e516","name":{"modules":["Stdlib","Int128"],"name":{"FnName":["absoluteValue"]},"owner":"Darklang","version":0},"parameters":[{"description":"","name":"a","typ":{"TInt128":[]}}],"returnType":{"TInt128":[]},"tlid":6253735342452977547,"typeParams":[]}
Status: 500
```

File: backend/src/BuiltinExecution/Libs/Json.fs [hunk 9]
```diff
@@ -487,6 +493,57 @@ let parse
       else
         raiseCantMatchWithType TUInt16 j pathSoFar |> Ply
 
+    | TInt128, JsonValueKind.Number ->
+      let mutable i128 = System.Int128.Zero
+      let mutable d = 0.0
+      if
+        System.Int128.TryParse(j.GetRawText(), &i128)
+        && i128 <= System.Int128.MaxValue
```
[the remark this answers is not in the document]
OceanOak: You are right, I believe it can't be larger.

backend/src/BuiltinExecution/Libs/Json.fs
```diff
@@ -487,6 +493,57 @@ let parse
...
+        && i128 >= System.Int128.MinValue
+      then
+        DInt128 i128 |> Ply
+      else if
+        j.TryGetDouble(&d)
+        && d <= (float System.Int128.MaxValue)
```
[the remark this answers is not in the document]
OceanOak: I think we need it here 
Double max value is `1.7976931348623157E+308` which is greater than Int128 max value

backend/testfiles/execution/stdlib/json.dark
OceanOak: I think what you did for int64 is different if you are referring to the following comment

> // dotnet will wrap 9223372036854775808 to be -9223372036854775808 instead, we
> // don't want that and will error instead

This one only happens with negative out-of-range values and it always wraps them around to the Int128 Min value. (I could be misunderstanding something though)

backend/src/BuiltinExecution/Libs/Int128.fs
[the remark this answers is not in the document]
pbiggar: Well we should be able to use 0Q, so it looks like there's something broken there.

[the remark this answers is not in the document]
pbiggar: But yeah, we should remove this.

backend/src/BuiltinExecution/Libs/Json.fs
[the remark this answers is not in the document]
pbiggar: I would add tests for cases like this and then we can be sure. Looks good

packages/darklang/stdlib/int128.dark
```diff
@@ -0,0 +1,137 @@
...
+      let multiply (a: Int128) (b: Int128) : Int128 = Builtin.Int128.multiply a b
+
+
+      /// Divides two 128-bit signed integers
+      let divide (a: Int128) (b: Int128) : Int128 = Builtin.Int128.divide a b
+
+
+      /// Returns the absolute value of <param a> (turning negative inputs into positive outputs)
+      let absoluteValue (a: Int128) : Int128 =
+        if (Builtin.Int128.lessThan a (Builtin.Int128.zero)) then
```
pbiggar, line 48: This should use 0Q/0Z (whichever is int128, I'm not sure)

packages/darklang/stdlib/uint128.dark
```diff
@@ -0,0 +1,115 @@
...
+      let subtract (a: UInt128) (b: UInt128) : UInt128 = Builtin.UInt128.subtract a b
+
+
+      /// Multiplies two 128-bit unsigned integers
+      let multiply (a: UInt128) (b: UInt128) : UInt128 = Builtin.UInt128.multiply a b
+
+
+      /// Divides two 128-bit unsigned integers
+      let divide (a: UInt128) (b: UInt128) : UInt128 = Builtin.UInt128.divide a b
+
```
pbiggar, line 31: Did you remove absoluteValue?

OceanOak: I haven't implemented an absoluteValue function for unsigned integers. Using it on a non-negative number would have no effect

backend/testfiles/execution/stdlib/json.dark
```diff
@@ -403,6 +403,151 @@ module UInt16 =
...
+    Builtin.Json.parse<Int128> "1.7E38" = Result.Ok
+      169999999999999998061923293023115935744Q
+
+    // This is a float that is slightly above the Int128 max
+    (Builtin.Json.parse<Int128> "1.8E38") |> err = Result.Error
+      "Can't parse JSON `1.8E38` as type `Int128` at path: `root`"
+
+    // A large negative number that is within Int128 range
+    Builtin.Json.parse<Int128> "-1.7E38" = Result.Ok
+      -169999999999999998061923293023115935744Q
```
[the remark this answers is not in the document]
OceanOak: We also aren't loosing precision when testing this 
```
 Builtin.Json.parse<Int128> "170000000000000008061923293023115935744" = Result.Ok
      170000000000000008061923293023115935744Q
```

We only loose precision when it hits the double branch (double.ToString)

Here is an example :
```
> let v = System.Double.Parse "-1.7E38";;                            
val v: float = -1.7e+38

> v.ToString("F0", System.Globalization.CultureInfo.InvariantCulture)
- ;;
val it: string = "-169999999999999998061923293023115935744"
```

backend/src/BuiltinExecution/Libs/Int128.fs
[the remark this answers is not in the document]
OceanOak: Please correct me if I'm wrong, but I think the issue is that we cannot currently save Int128 in the database. As a result, we fail when attempting to make a post request to save the absoluteValue function to the PackagesDB?

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -32,6 +32,19 @@ let error2 (msg : string) (v : string) : 'a = error $"{msg}: {v}"
 let error3 (msg : string) (v1 : string) (v2 : string) : 'a =
   error $"{msg}:\n  {v1}\n  {v2}"
 
+let int128 (v : System.Int128) : SqlValue =
+  let typ = NpgsqlTypes.NpgsqlDbType.Numeric
+  let idParam = NpgsqlParameter("tlid", typ)
```
pbiggar, line 37: This param should probably be called `int128` instead of `tlid`

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -32,6 +32,19 @@ let error2 (msg : string) (v : string) : 'a = error $"{msg}: {v}"
...
+  idParam.Value <- System.Numerics.BigInteger.op_Implicit (v)
+  Sql.parameter idParam
+
+let uint128 (v : System.UInt128) : SqlValue =
+  let typ = NpgsqlTypes.NpgsqlDbType.Numeric
+  let idParam = NpgsqlParameter("tlid", typ)
```
pbiggar, line 43: This param should probably be called `uint128` instead of `tlid`

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -774,6 +799,8 @@ let rec lambdaToSql
             | TUInt8 -> "smallint"
             | TInt16 -> "smallint"
             | TUInt16 -> "integer"
+            | TInt128 -> "numeric"
```
pbiggar, line 801: I dont know if it's possible here, but the postgres docs advice specifying the scale here. See https://www.postgresql.org/docs/14/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL

Perhaps these should be `"numeric(0,39)"` (I don't know that 39 is the number; just whatever the right size is).

this might be better to add as a comment rather than getting bogged down on it now.

backend/src/LibBinarySerialization/ProgramTypesToSerializedTypes.fs
```diff
@@ -364,8 +364,9 @@ module MatchPattern =
     | PT.MPUInt8(id, i) -> ST.MPUInt8(id, i)
     | PT.MPInt16(id, i) -> ST.MPInt16(id, i)
     | PT.MPUInt16(id, i) -> ST.MPUInt16(id, i)
-    | PT.MPInt128(id, i) -> ST.MPInt128(id, i)
-    | PT.MPUInt128(id, i) -> ST.MPUInt128(id, i)
+    | PT.MPInt128(id, i) -> ST.MPInt128(id, System.Numerics.BigInteger.op_Implicit i)
```
pbiggar, line 367: Again `bigint` function might work here.

backend/src/LibBinarySerialization/SerializedTypes.fs
```diff
@@ -262,8 +262,8 @@ type MatchPattern =
   | MPUInt8 of id * uint8
   | MPInt16 of id * int16
   | MPUInt16 of id * uint16
-  | MPInt128 of id * System.Int128
-  | MPUInt128 of id * System.UInt128
+  | MPInt128 of id * System.Numerics.BigInteger
```
pbiggar, line 271: F# has a builtin type for this, I think called bigint.

backend/src/Prelude/Json.fs
```diff
@@ -67,6 +67,28 @@ module Vanilla =
       else
         writer.WriteNumberValue(value)
 
+
+  type Int128Converter() =
```
pbiggar, line 71: I think you're right to have these as strings rather than as RawValues, but can you add a comment explaining?

pbiggar (approved):
Looks good in current form, happy to merge. There's a few changes I think we should do, happy to have them done pre- or post-merge.

backend/src/LibCloud/SqlCompiler.fs
```diff
@@ -32,6 +32,19 @@ let error2 (msg : string) (v : string) : 'a = error $"{msg}: {v}"
...
+let int128 (v : System.Int128) : SqlValue =
+  let typ = NpgsqlTypes.NpgsqlDbType.Numeric
+  let idParam = NpgsqlParameter("tlid", typ)
+  idParam.Value <- System.Numerics.BigInteger.op_Implicit (v)
+  Sql.parameter idParam
+
+let uint128 (v : System.UInt128) : SqlValue =
+  let typ = NpgsqlTypes.NpgsqlDbType.Numeric
+  let idParam = NpgsqlParameter("tlid", typ)
+  idParam.Value <- System.Numerics.BigInteger.op_Implicit (v)
```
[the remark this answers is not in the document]
OceanOak: Yes, bigint function doesn't work with int128. I get `No overloads match...` error

backend/src/Prelude/Json.fs
OceanOak: I don't think there's an option to use RawValues here, as `reader` and `writer` don't  have a `GetRawValue()` function.

PR #5144, opened 2023-11-06
Alternative Json parser/serializer

StachuDotNet:
Our current Json serializer does questionable things when serializing `Option`s, and generally doesn't allow users to control how their types are serialized/parsed as JSON. This provides an alternative set of Json serializer/parser that allows such control.

This was written towards a goal of [writing our LSP server in Darklang](https://www.notion.so/darklang/Design-Darklang-LSP-Server-9bba9e04de4f4d21b8b89a7d71e55c0c).

The main ideas here are:
- .NET (the builtins) only have to map from `JsonToken` to/from Json - it does _just_ the Json serialization/parsing
  ```fsharp
  type JsonToken =
    | Null
    | Bool of Bool
    | Number of Float
    | String of String
    | Array of List<JsonToken>
    | Object of List<String * JsonToken>
  ```
  - this is already demonstrated in this WIP PR- 
- we create _tokenizers_ and _detokenizers_ that map from specific types (both base types and custom types) to/from `JsonToken`
  - not yet demonstrated by this PR
  - _but_ we can also provide the ability to 'serialize in the default way'

If we migrated to this path generally, we shouldn't have to touch the Builtins at all as we expand the language


pbiggar (approved):
Looks great. I wouldn't bother with tests for now, since it's so straightforward, apart from one end-to-end test.

We're adding packages to wrap builtin these days, would be worth adding since it's simple.

PR #5145, opened 2023-11-07
More work on the Alt./Configurable Json parser/serializer

StachuDotNet:
needs more thought. Abandoning for now, will revisit in a few days. Happy to take feedback in meantime :)

### DIFF
--- a/backend/testfiles/execution/stdlib/alt-json.dark
+++ b/backend/testfiles/execution/stdlib/alt-json.dark
@@ -200,3 +200,38 @@ module TokenSerialization =
 
       // ..._double_ quotes
       parseToken """{ 'invalidName': 1}""" = tokenParseError TokenParseError.NotJson
+
+
+module Parsing =
+  // The long way - without any helpers
+  (PACKAGE.Darklang.Stdlib.AltJson.serialize<Bool>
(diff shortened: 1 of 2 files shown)


pbiggar:
What's tokenization vs serialization?

StachuDotNet:
> What's tokenization vs serialization?

Dumped some thoughts here:
https://gist.github.com/StachuDotNet/4d9966edd053eef5dd60a17dfaf6f90d

of note,
> In this world,
> - tokenization<'T> is the process of going from 'T (a data of some type) to a JsonToken
> - detokenization<'T> is the process of going from JsonToken to a 'T
> - serializeToken takes a JsonToken and json stringifies it
> - deserializeToken takes a json string and tries to parse out a JsonToken
> - parse<'T> involves deserializing into a token, and then detokenizing
serialize<'T> involves tokenizing some data into a JsonToken,a nd then serializing that token into the json string

pbiggar:
Gotcha.

Some thoughts:
- it seems that just exposing jsontoken is a decent step forward for now. Users can use `Json.tokenize` to get a nested `JsonToken` and write a `JsonToken -> Result<MyType, Whatever>` function using match statements
- I think that "token" is not the right word here, and I think it would be good to use the right words as we build this. Maybe it is the right word, but it doesn't match my understanding of what tokenization is in a parsing context.

PR #5147, opened 2023-11-08
Cli with user input

StachuDotNet:
Supports a really boring CLI app with user input:

```fsharp
Builtin.printLine "Hi there - let's add 2 numbers."

Builtin.print "a: "
let a = Builtin.Stdin.readLine ()

Builtin.print "b: "
let b = Builtin.Stdin.readLine ()

match (PACKAGE.Darklang.Stdlib.Int.parse a, PACKAGE.Darklang.Stdlib.Int.parse b) with
 | Ok a, Ok b ->
   $"{PACKAGE.Darklang.Stdlib.Int.toString a} + {PACKAGE.Darklang.Stdlib.Int.toString b} = {PACKAGE.Darklang.Stdlib.Int.toString (a + b)}"
   |> Builtin.printLine
 | _ -> Builtin.printLine "Invalid input"
```

We'll need this shortly to support `json-rpc`, and then the lsp-server spec

### DIFF
--- a/CODING-GUIDE.md
+++ b/CODING-GUIDE.md
@@ -47,7 +47,7 @@
 - `ignore` should always use a type signature (this should be enforced by the
   compiler)
 
-- use `print` instead of `Console.WriteLine` or similar; the latter deadlocks
+- use `printLine` instead of `Console.WriteLine` or similar; the latter deadlocks
 
 - ensure that `try` do not have a `uply`/`task` in the body unless you know what you
   are doing and provide a comment. Typically, the `uply`/`task` should be on the
--- a/CODING-GUIDE.md
+++ b/CODING-GUIDE.md
@@ -47,7 +47,7 @@
 - `ignore` should always use a type signature (this should be enforced by the
   compiler)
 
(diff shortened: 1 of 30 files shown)


StachuDotNet:
I think this is fine to merge but thought @pbiggar you might not like the `print` -> `printLine` rename.

I did that because I wanted a `print` Builtin that didn't add a newline with the print, for prompting, and `Builtin.printWithoutNewline` didn't quite sounds right. `Builtin.write` might be ok.

pbiggar:
I don't mind the userspace rename, but I don't really like it in our codebase. `print` in our codebase is really internal stuff, so the newline is fine. I dunno if it should match up, maybe I'm just being a curmudgeon.

StachuDotNet:
> I don't mind the userspace rename, but I don't really like it in our codebase. `print` in our codebase is really internal stuff, so the newline is fine. I dunno if it should match up, maybe I'm just being a curmudgeon.

No worries - I suspected you might prefer to keep `print` as-is -- so, I just renamed it back, and the _new_ fn is `printInline`, which feels decent to me :)

PR #5151, opened 2023-11-14
Replace our LSP server, written in Node, with one written in Darklang

StachuDotNet:
supporting work:
- remove `Builtin.File.append` (broken, ambiguous) with `Builtin.File.appendText` (working, clear)
- update `AltJson` module a bit, mostly removing "token" wording
  - rename `type JsonToken` to just `type Json`
  - rename fn `parseToken` to just `parse`
  - rename fn `serialize` to `format`
- `run-cli` now logs stdout+stderr to `rundir/logs/cli.log` (_so_ helpful)

actual improvements:
- added `Builtin.Danger.LanguageServerProtocol.readNextMessage` to read incoming messages
  - because... I'm not sure why - I had issues trying to replicate this with `Builtin.Stdin.readLine` but I'm sure it's solveable :arrow_down: 
  - TODO added to remove this in favor of a new Builtin fn or two
- add `@Darklang.JsonRPC` module corresponding to the json-rpc 2.0 spec. I really like how this turned out.
- add `@Darklang.LanguageTools.LspServer` module with `.runServerCli` to be run via `run-cli` script to act as our VS Code Extension's LSP Server
  - logs to `rundir/logs/lsp.log` 
- updated `vscode-extension` to point to the new Darklang LSP Server, based on a hardcoded boolean toggle

### DIFF
--- a/backend/src/BuiltinCli/Libs/File.fs
+++ b/backend/src/BuiltinCli/Libs/File.fs
@@ -90,9 +90,9 @@ let fns : List<BuiltInFn> =
       deprecated = NotDeprecated }
 
 
-    { name = fn "append" 0
+    { name = fn "appendText" 0
       typeParams = []
-      parameters = [ Param.make "path" TString ""; Param.make "content" TBytes "" ]
+      parameters = [ Param.make "path" TString ""; Param.make "content" TString "" ]
       returnType = TypeReference.result TUnit TString
       description =
         "Appends the given <param content> to the file at the specified <param path>. If the file does not exist, a new file is created with the content. Returns a Result type indicating success or failure."
@@ -100,10 +100,10 @@ let fns : List<BuiltInFn> =
         let resultOk = Dval.resultOk KTUnit KTString
         let resultError = Dval.resultError KTUnit KTString
         (function
-        | _, _, [ DBytes content; DString path ] ->
+        | _, _, [ DString path; DString content ] ->
           uply {
             try
-              do! System.IO.File.WriteAllBytesAsync(path, content)
+              do! System.IO.File.AppendAllTextAsync(path, content)
               return resultOk DUnit
             with e ->
               return resultError (DString e.Message)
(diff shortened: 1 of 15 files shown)


backend/src/BuiltinCli/Libs/File.fs
```diff
@@ -90,20 +90,20 @@ let fns : List<BuiltInFn> =
       deprecated = NotDeprecated }
 
 
-    { name = fn "append" 0
+    { name = fn "appendText" 0
```
pbiggar, line 93: Would be good to start establishing some conventions around this sort of thing (string vs bytes, naming, etc)

backend/src/BuiltinCli/Libs/LanguageServerProtocol.fs
```diff
@@ -0,0 +1,66 @@
...
+module Builtin = LibExecution.Builtin
+open Builtin.Shortcuts
+
+let types : List<BuiltInType> = []
+let constants : List<BuiltInConstant> = []
+
+// TODO reimplement this in Darklang, without any custom built-in function. I had
+// some difficulties around this - I forget what - but this got the job done for now
+let fns : List<BuiltInFn> =
+  [ { name = fn [ "LanguageServerProtocol" ] "readNextMessage" 0
```
pbiggar, line 19: Perhaps for now we should use some names like Private or Danger so that other don't use these.

Another idea I had is that perhaps builtins can only be called from Stdlib functions, which hides them entirely from userspace.

StachuDotNet: I'm hoping to delete this builtin fn before merging - I think I can replace it with a `Builtin.Stdin.todo(...)...` that captures just:
```fsharp
let streamReader = new StreamReader(System.Console.OpenStandardInput(),System.Text.Encoding.UTF8)
let buffer = Array.create length ' '
streamReader.ReadBlock(buffer, 0, length) |> ignore<int>
System.String buffer
``` 

rather than be LSP-specific

StachuDotNet: ended up leaving this for another day, so added a "Danger" module here

StachuDotNet:
fwiw, sample debugging/dev setup
image

PR #5165, opened 2023-11-20
Solidify Darklang LSP Support; handle document sync

StachuDotNet:
- minor fixes to `json-rpc` support
- from the [LSP Spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/) and the [de facto Node LSP server](https://github.com/microsoft/vscode-languageserver-node) that demonstrates the spec, ports many many types we'll need to build out our language server. _most_ of these are still commented out and left in their equivalent TypeScript interface form, but the up-front organization here should help us out a lot. Ones necessary (and a few more) have been fully ported and paired with `toJson`/`fromJson` functions
- specifically, port types needed in the `initialize` request, as well as the`didOpen`, `didClose`, and `didSave` notifications
- reduces our server's advertised capabilities to _only_ include document sync, and add that support to our server
  - involves maintaining a `state` between `runServerCliLoop` iterations

### DIFF
--- a/packages/darklang/json-rpc.dark
+++ b/packages/darklang/json-rpc.dark
@@ -21,7 +21,7 @@ module Darklang =
     let requestIdToJson (id: RequestId) : Json =
       match id with
       | Null -> Json.Null
-      | Int i -> Json.Number i
+      | Int i -> Json.Number(Stdlib.Int64.toFloat i)
       | String s -> Json.String s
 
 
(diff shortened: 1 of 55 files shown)


scripts/build/reload-packages
```diff
@@ -44,7 +44,7 @@ echo -e "Loading packages into DB ${grey}($LOG_DB)${reset}"
 "${LOCALEXEC_EXE}" load-packages > $LOG_DB 2>&1
 
 
-if [[ "$TEST" != "true" ]]; then
+if [[ "$TEST" == "true" ]]; then
```
StachuDotNet, line 47: won't merge - this is just my local trick so I don't have to wait for the `dark-packages` seeding to occur, because it's annoyingly slow.

PR #5171, opened 2023-11-21
Replace Bytes with a List of UInt8

OceanOak:
Changelog:

```
Area of change
- Replace the `Bytes` type with a `List<UInt8>`
```


backend/src/BuiltinExecution/Libs/Base64.fs
```diff
@@ -81,22 +87,54 @@ let fns : List<BuiltInFn> =
...
           System.Convert.ToBase64String(bytes).Replace('+', '-').Replace('/', '_')
           |> DString
           |> Ply
         | _ -> incorrectArgs ())
       sqlSpec = NotYetImplemented
       previewable = Pure
+      deprecated = NotDeprecated }
+
+
+    { name = fn "hexEncode" 0
```
OceanOak, line 110: Not sure if this is the right place for this function. It was originally in 'Bytes.fs' and is used in both crypto and big tests

backend/src/LibExecution/Dval.fs
```diff
@@ -104,3 +104,18 @@ let result
   match dv with
   | Ok dv -> resultOk okType errorType dv
   | Error dv -> resultError okType errorType dv
+
+
+let byteArrayToDvalList (bytes : byte[]) : Dval =
```
OceanOak, line 109: Not sure if this is the right place for these two helper functions

backend/tests/Tests/DvalRepr.Tests.fs
```diff
@@ -99,28 +99,29 @@ let testToDeveloperRepr =
...
+//         Expect.equal actual expected "bad fsharp impl"
+//       }
 
-    testList
-      "hashv2"
-      [ t (NEList.singleton (DBytes [||])) "X1YnxJLFsVg"
-        t (NEList.singleton (DBytes [| 128uy |])) "Hj0nqyrvXis" ]
+//     testList
+//       "hashv2"
+//       [ t (NEList.singleton (DList(VT.uint8, []))) "X1YnxJLFsVg" //DEux3mJnJPs
```
OceanOak, line 121: Is it expected to get different hash values here?

pbiggar: yeah, I'd expect that.

backend/src/LibExecution/Dval.fs
pbiggar: This is where I would put them too

backend/src/BuiltinExecution/Libs/Base64.fs
pbiggar: I'd maybe put these in bytes? It feels like this could go into a package though?

OceanOak: I moved them to Bytes for now, I'll see if I can move it to a package

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -395,14 +394,14 @@ let fns : List<BuiltInFn> =
     { name = fn "toBytes" 0
       typeParams = []
       parameters = [ Param.make "str" TString "" ]
-      returnType = TBytes
+      returnType = (TList(TUInt8))
```
pbiggar, line 397:
```suggestion
      returnType = TList TUInt8
```

backend/src/BuiltinExecution/Libs/String.fs
```diff
@@ -411,13 +410,14 @@ let fns : List<BuiltInFn> =
 
     { name = fn "fromBytesWithReplacement" 0
       typeParams = []
-      parameters = [ Param.make "bytes" TBytes "" ]
+      parameters = [ Param.make "bytes" (TList(TUInt8)) "" ]
```
pbiggar, line 413:
```suggestion
      parameters = [ Param.make "bytes" (TList TUInt8) "" ]
```

pbiggar (approved):
This is great!

pbiggar:
I made two suggestions, and they apply largely across the commit, which is that there are excess parentheses eg `(X(Y))` can always be `(X Y)` and can sometimes be just `X Y`. Feel free to do it as a followup, if not it's not a big deal.

PR #5176, opened 2023-11-22
fix: make docker build work on Windows

xtopherbrandt:
Changelog:

```
Area of change
- dockerfile

alternatively:

No changelog

Docker could not build the development contain in a Windows environment

Run dos2unix on installer scripts that are written to the container filesystem during the contain build to eliminate any DOS line endings.

- add dos2unix to apt install list
- run dos2unix on install-targz-file and install-exe-file
- save yugabyte install script as install-yugabyte and run dos2unix on it before running
- save dotnet8 install script as install-dotnet8 and run dos2unix on it before running`

tested only locally on one machine.

back out the commit

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -138,6 +138,8 @@ RUN DEBIAN_FRONTEND=noninteractive \
       # parser (tree-sitter) dependencies
       build-essential \
       # end parser dependencies
+      # required to clean up line endings for COPY when building on a Windows machine
+      dos2unix \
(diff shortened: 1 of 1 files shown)


pbiggar (approved):
this looks good and i think it's fine to commit as such.

I think these would probably be better as files extracted from the Dockerfile and stored in the repo, and then COPYed in when the dockerfile is run. If that's possible, would make a useful followup.

PR #5177, opened 2023-11-24
Move installer scripts from docker to /scripts

xtopherbrandt:
Changelog:

```
Area of change
- dockerfile
- /scripts directory
```

Simplify the dockerfile by removing script file generation

Move scripts to files in the /scripts/installers folder, then copy them into the docker image during build.

Changes how the image is built and simplifies cross-platform building.

Tested in both Windows and Linux environments.

back out the commit


scripts/installers/install-dotnet8
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Script to install dotnet 8
+
+set -e
+case ${TARGETARCH} in
+  arm64)
+    ARCH=arm64
+    CHECKSUM=3296d2bc15cc433a0ca13c3da83b93a4e1ba00d4f9f626f5addc60e7e398a7acefa7d3df65273f3d0825df9786e029c89457aea1485507b98a4df2a1193cd765
```
pbiggar, line 9: Could we extract the checksums and the versions so that those versions are stored in the dockerfile? Otherwise changing them in the scripts won't cause a dockerfile rebuild.

xtopherbrandt: Good idea. Done.

Dockerfile
```diff
@@ -315,49 +215,8 @@ RUN /home/dark/install-targz-file \
...
-# python on our system is called python3
-sed -i 's|#!/usr/bin/env python|#!/usr/bin/env python3|' ./bin/yugabyted
-cd ..
-EOF
-
-RUN sudo dos2unix /home/dark/install-yugabyte
-
-RUN sudo chown dark:dark /home/dark/install-yugabyte
-RUN chmod +x /home/dark/install-yugabyte
+ENV YUGABYTE_VERSION=2.20.0.0 \
```
pbiggar, line 218: Could we do these as command line arguments, same as they're passed to instal-targz-file?

pbiggar (changes requested):
Looks great! Would love to get the arguments changed, apart from that we're good to go.

pbiggar:
Thanks a million! Note that CI doesn't use this dockerfile by default so i'll fix that up later.

PR #5193, opened 2023-11-27
Provide fake autocomplete ("completions") in our Language Server

StachuDotNet:
- in `initialize` handshake, announce that we support 'completions'
  `"completionProvider":{"resolveProvider":false}`
- handle `textDocument/completion` requests, and respond to them with a single fake autocomplete

https://github.com/darklang/dark/assets/906686/20261594-8371-410b-a5ab-794ad5c895f0

### DIFF
--- a/packages/darklang/languageServerProtocol/README.md
+++ b/packages/darklang/languageServerProtocol/README.md
@@ -36,13 +36,45 @@ project is quite hard to follow, though, so we've reorganized things here.
 - **notification**: a message sent from one LSP party to another, which does not expect a result
 - **message**: a general term for either a notification, request, or response (error or result)
 
-## Some patterns to follow per-file:
+## Some patterns
 
 - try to keep type names the same as the TypeScript ones in the spec
+- per-file organization
+  - supporting types at the top
+  - types around requests, results, and notifications in the middle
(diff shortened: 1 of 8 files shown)


StachuDotNet:
misc note: somehow, I haven't explicitly learned that `ctrl+space` triggers the autocomplete flow in VS Code (at least w/ my setup)

packages/darklang/languageServerProtocol/README.md
```diff
@@ -36,13 +36,45 @@ project is quite hard to follow, though, so we've reorganized things here.
...
+  type CompletionParams = {
+    textDocumentPosition: TextDocumentPositionParams
+    context: Option<CompletionContext>
+  }
+  ```
+
+  The latter is easier to understand, but won't work out of the box if we end up
+  switching to our usage-ignorant Json parser/serializer. I'm going to use the
+  latter for now, and we can re-evaluate if/when needed. That said, for each case
+  where we do this, we should add a comment roughly of `@extends TextDocumentPositionParams`
```
StachuDotNet, line 77: In reflection I think this was a mistake. Going to update the few relevant types to "inline" the relevant fields to match the JSON.

PR #5199, opened 2023-11-30
Add WrittenTypes.dark + Start mapping from parsedNode to WrittenTypes

OceanOak:
Changelog:

```
Area of change
- Add WrittenTypes.dark 
- Start mapping from parsedNode to WrittenTypes
```


StachuDotNet:
Great!

FWIW, some of the source in `packages` could take advantage of our parser and name resolver respecting package hierarchy (i.e. most of the PACKAGE.Darklang.` prefixes could be removed) but totally non-blocking

PR #5205, opened 2023-12-01
Add source ranges to dark writtentypes

StachuDotNet:
I accidentally did a few things at once:
- removed IDs from `writtenTypes.dark` (I don't think we need them)
- removed PT.FnName.Package reference from writtenTypes.dark
- commented out parts of `writtenTypes` that `grammar.js` doesn't yet support (I think it makes sense to keep these roughly matching)
- add source text `Range`s to various nodes in WrittenTypes, to help with Syntax Highlighting, Completions, etc.

We'll probably have to add `Range`s in a few more places shortly to get complete syntax highlighting, but what's here should be enough to get both syntax highlighting and completions working at least _a bit_.


OceanOak (commented):
Looks great, thanks!

PR #5219, opened 2023-12-20
Add a name resolver equivalent in dark

OceanOak:
Changelog:

```
Area of change
- Implementing the name resolver equivalent in dark 
```

This successfully resolves 
- Package functions and types 
- Builtin functions (I don't think we have any builtin types). 

No package and builtin constants (constants are not supported in the grammar yet)
No user functions and types are handled yet


OceanOak:
To followup on:

- I am not sure how to handle resolving user functions and types yet, specifically how to handle running the name resolver twice since we are testing it from a test file.

- In the DarkEditor, we currently have only one function which returns a list of builtins. I am not sure about an appropriate name for this file since it only contains one function

packages/darklang/languageTools/nameResolver.dark
```diff
@@ -0,0 +1,280 @@
...
+
+
+
+      module TypeName =
+        let packageTypeExists (typeName: String) : Bool =
+
+          let typ =
+            Stdlib.HttpClient.request
+              "GET"
+              ("http://dark-packages.dlio.localhost:11003/type/by-name/" ++ typeName)
```
StachuDotNet, line 170: I think it's worth extracting out a `LanguageTools.PackageManager` module that lives on its own, with this sort of usage contained within such. (`LanguageTools.PackageManager.Type.byName urlBase typeName` maybe)

That said, that can wait for another PR.

packages/darklang/languageTools/nameResolver.dark
```diff
@@ -0,0 +1,280 @@
...
+            |> Stdlib.Result.map (fun response -> response.statusCode)
+            |> Builtin.unwrap
+
+          match statusCode with
+          | 200L -> true
+          | e -> false
+
+        let builtinThingExists (typeName: String) : Bool =
+          // TODO: check if builtin type exists
+          true
```
StachuDotNet, line 185: Maybe `false` and replace the TODO with "we don't currently have any Builtin types, and don't expect to)

packages/darklang/languageTools/nameResolver.dark
```diff
@@ -0,0 +1,280 @@
...
+
+        let builtinThingExists (fnName: String) : Bool =
+
+          let fnExists =
+            (Builtin.LanguageTools.allBuiltinFns ())
+            |> Stdlib.List.findFirst (fun f -> f.name == fnName)
+
+          match fnExists with
+          | Some _ -> true
+          | None -> false
```
StachuDotNet, line 247: can replace the last bit with `|> Option.isSome`

packages/darklang/languageTools/parser.dark
```diff
@@ -30,6 +30,12 @@ module Darklang =
       let parseToSimplifiedTree (text: String) : ParsedNode =
         Builtin.Parser.parseToSimplifiedTree text
 
+      let parseName (fnName: String) : Darklang.Stdlib.Result.Result<String> =
+        if (Darklang.Stdlib.String.contains fnName "_v") then
+          // Todo : check if the name is valid
+          Darklang.Stdlib.Result.Result.Ok fnName
+        else
+          Darklang.Stdlib.Result.Result.Ok(fnName ++ "_v0")
```
StachuDotNet, line 38: I'm curious why we need this. Is there something that expects the _v0 suffix to be present?

packages/darklang/languageTools/writtenTypesToProgramTypes.dark
```diff
@@ -16,60 +16,42 @@ module Darklang =
...
+
+            let resolvedName =
+              NameResolver.FnName.resolve
+                resolver
+                modules
+                (WrittenTypes.Name.Unresolved nametoResolve)
+
+            match resolvedName with
+            | Ok name -> name
+            | Error e -> e
```
StachuDotNet, line 54: Hmm this seems a little bit suspicious. Probably OK to keep, but feels like some TODO would be appropriate here

packages/darklang/languageTools/parser.dark
OceanOak: Yes, I believe this doesn't work without the _v0
`http://dark-packages.dlio.localhost:11003/function/by-name/Darklang.Stdlib.String.toBytes_v0`

It fails with 
`PACKAGE.Darklang.Stdlib.Int64.ParseError.BadFormat`

packages/darklang/languageTools/nameResolver.dark
```diff
@@ -0,0 +1,280 @@
+module Darklang =
+  module LanguageTools =
+    module NameResolver =
+
+      type NameResolver = { allowError: Bool }
```
StachuDotNet, line 5: Given the current naming, one might expect to take one of these and call `nr.resolveName` on it.
Maybe we should rename to `NameResolutionSettings`?

scripts/run-backend-tests
```diff
@@ -88,11 +88,16 @@ cd backend && \
...
 cd ..
 
-DARK_CONFIG_TELEMETRY_EXPORTER=none \
-DARK_CONFIG_ROLLBAR_ENABLED=n \
-DARK_CONFIG_QUEUE_PUBSUB_PROJECT_ID=pubsub-test-${RANDOM_VALUE} \
-DARK_CONFIG_TRACE_STORAGE_BUCKET_NAME=trace-test-${RANDOM_VALUE} \
-./scripts/build/reload-packages --test $PUBLISHED_FLAG
+# Reload packages
+
+if [[ "$PUBLISHED" == "true" ]]; then
```
StachuDotNet, line 93: Is there a way this can be 'if running in CI' rather than 'if running the published assets'?
Sometimes, though rarely, we'll run published assets locally, which I think will cause this to fail

packages/darklang/languageTools/writtenTypes.dark
```diff
@@ -5,6 +5,7 @@ module Darklang =
       type SourceRange = Parser.Range
       // <aliases>
 
+      type Name = Unresolved of List<String>
```
StachuDotNet, line 8: Do we need this in WrittenTypes? I don't think it's referenced in the other types here, and realistically (almost) _all_ names in WrittenTypes are sort of unresolved.

Maybe it belongs in PT.dark or WT2PT.dark. Is the intent to capture the fact that a name may be 'unresolved' during the "first pass" of name resolution?

packages/darklang/languageTools/nameResolver.dark
StachuDotNet: extracting that will also given us an opportunity for some caching of those packages, which will be needed before long

StachuDotNet (changes requested):
looking great, just a few questions+comments, noted

packages/darklang/languageTools/writtenTypes.dark
OceanOak: Yes, the only use for it was to capture the unresolved name in the name resolver.

PR #5223, opened 2023-12-26
Preserve CLI build artifacts in CI

StachuDotNet:
I noticed that the CI artifacts of the `build-cli` step were empty - this fixes that.

edit: Ah, maybe this was intentional? Since each executable is ~12MB, I guess these could add up.
@pbiggar what do you think?


pbiggar:
Yeah, this needs another build step to ship them to be part of the release, so that should probably happen in another step during a deploy, in which case you need this.

pbiggar (approved):

PR #5226, opened 2024-01-01
Host darklang executables via github releases (a pre-release)

OceanOak:
Changelog:

```
Area of change
- Host darklang executables via Github releases
```

The `publish-github-release` script:
- Checks for previous releases/pre-releases; if found, it increments the tag by one. (if no tag is found it starts from `v0.0.1`)
- Creates the GitHub release with the new tag
- Uploads the assets if the creation is successful.
- Deletes previous releases.


StachuDotNet (changes requested):
I think, with the way this is set up, a new Tag would be created upon every deploy (i.e. `v0.0.1`), and show up [here](https://github.com/darklang/dark/tags).

I'm wondering if we should
- instead, maintain a consistent `prerelease` tag for now, and delete old releases with a different mechanism (i.e. by SHA)
- or, update the script to purge irrelevant/old tags, too
- maybe we have some other options not coming to mind?

More broadly, I'm a bit wary, for some reason, to assign version numbers to these prereleases.
I'm probably thinking too much about this, though -- I think the idea is that these prereleases are all deleted later anyway, once things are stable soon.
That said, maybe this is some encouragement to think through the self-updating problem more, before merging this?

What do you think?

---

I really like the `jq` usage here. I should use that more often!

---

FYI the `FILES_DIR` is currently set to a personal directory

StachuDotNet:
We continued to chat about this PR in Discord - I think the set-up here looks good. We'll be purging Releases once things are stable anyway, so really I think I was just overcomplicating things here.

The PR is currently marked as WIP - is it ready for review?

StachuDotNet:
FWIW, I was toying with a format like this: `v0.0.1-alpha+[build SHA]` like `v0.0.1-alpha+7a941e342`. (later, `v0.0.1+7a941e342`)

In case that looks appealing.

OceanOak:
> We continued to chat about this PR in Discord - I think the set-up here looks good. We'll be purging Releases once things are stable anyway, so really I think I was just overcomplicating things here.
> 
> The PR is currently marked as WIP - is it ready for review?

Yes

PR #5273, opened 2024-01-23
Improve package manager situation

StachuDotNet:
as a result of this PR,

- reloading packages in local dev is much faster
- canvas-hack is gone, basically folded into local-exec
- rewrites localexec in F# - IMO doing that stuff in Dark is more painful than it's worth
- words towards untangling the awkward "how do we load the canvas, which references packages, before it itself contains packages" problem
- I forget

Needs to be followed up with a terraform dance so we point to the right DB. The old db is 'corrupt' (doesn't match the shape of things any more) and I don't have access to it, so I made my own cluster and will need to follow up on that later so I don't get charged for it.


backend/tests/Tests/Tests.fs
```diff
@@ -31,7 +31,7 @@ let main (args : string array) : int =
 
     let tests =
       [ Tests.AnalysisTypes.tests
-        Tests.BwdServer.tests
+        //Tests.BwdServer.tests
```
StachuDotNet, line 34: need to follow up on these soon, but I think it's OK for these to fail for a few days

StachuDotNet:
Ocean - I wouldn't rebase quite yet

PR #5276, opened 2024-01-24
Implement a self-updating mechanism for Darklang's executable

OceanOak:
Changelog:

```
CLI
- Implement a self-updating mechanism for Darklang's executable
```

Notes / things that needs to be fixed:

 - The CLI should work even if the user has no internet connection available (maybe the self update will just fail and we continue running the commands?)

- On Windows, commands like `uname -s`, which determine the operating system, may not work. 
We might consider something like the following?
   ```
   let os = Stdlib.Process.runCommand "uname -s"
   if os.exitCode != 0L then 
   // try the windows equivalent command
   ```
   
 - code signing for mac


packages/darklang/cli/cli.dark
```diff
@@ -147,5 +147,468 @@ darklang [script path]                            Run a .dark script
...
+        browser_download_url: String }
+
+    type Release =
+      { tag_name: String
+        assets: List<Asset> }
+
+
+    let configPath = ".darklang/bin/darklang/config.json"
+
+    // get list of releases from github api
```
StachuDotNet, line 161:
```suggestion
    /// get tag of most recent release, from github api
    /// (i.e. `v0.0.19`)
```

packages/darklang/cli/cli.dark
```diff
@@ -147,5 +147,468 @@ darklang [script path]                            Run a .dark script
...
+              "x64"
+            else if archName == "arm64" then
+              "arm64"
+            else
+              Builtin.printLine $"Unsupported architecture {archName}"
+              "Unsupported"
+
+          else
+            Builtin.printLine "Failed to get architecture"
+            arch.exitCode
```
StachuDotNet, line 324: I think it's worth tidying this up a bit to represent these unexpected things as Results -- it's easy to lose context of the fact that these strings might be some negative case. I think there are a few places here where this is relevant

StachuDotNet (changes requested):
Overall, nice! I think it's worth
- using Results a bit more, to represent error situations
- probably extracting the functionality here to a sub-module (see adjacent `packageViewer.dark`) so all of the "self-update" stuff is one place

packages/darklang/cli/selfUpdate.dark
```diff
@@ -0,0 +1,454 @@
...
+          | "x86_64" -> Stdlib.Result.Result.Ok "x64"
+          | "aarch64" -> Stdlib.Result.Result.Ok "arm64"
+          | "arm64" -> Stdlib.Result.Result.Ok "arm64"
+          | "arm" -> Stdlib.Result.Result.Ok "arm"
+          | _ -> Stdlib.Result.Result.Error $"Unsupported arch: {archName}"
+
+
+
+      let isLatestVersion () : Bool =
+        let localVersion = (getLocalVersion ()) |> Builtin.unwrap
```
OceanOak, line 168: Does using Builtin.unwrap here defeat the purpose of returning a result?

StachuDotNet (approved):
Great! Ready to merge when you are

Issue #5222, opened 2023-12-26
Bug: List.map Causes Stack Overflow Error on Large Lists


OceanOak:
Preamble: this is probably one of the most pressing limitations of our runtime. It requires some interpreter work, but should be generally accessible to anyone interested.

---

List.map causes a stack overflow when run in Debug mode, but not when running in Release mode

Code to reproduce the issue:
```
((PACKAGE.Darklang.Stdlib.List.repeat_v0 348L 1L)
 |> Builtin.unwrap
 |> PACKAGE.Darklang.Stdlib.List.map (fun f -> f)) = []
 ```
 
After running some tests @StachuDotNet found that 348 is the threshold 
 
The tests pass in CI but fail when run locally, indicating that something is different between Release and Debug builds
 
 Potential reasons for this issue: 
 - Bad implementation of List.map
 
 ```
 let map (list: List<'a>) (fn: 'a -> 'b) : List<'b> =
  (PACKAGE.Darklang.Stdlib.List.fold list [] (fun acc elem ->
    PACKAGE.Darklang.Stdlib.List.push_v0 acc (fn elem)))
  |> PACKAGE.Darklang.Stdlib.List.reverse
  ```
  
  - An issue with our interpreter

StachuDotNet:
some relevant links, from some chatter in discord a bit back

> I found a set of blog posts that I thought you might find interesting, related to the recursion stuff you've been doing in the StdLib
> - https://functional-programming-in-elm.netlify.app/recursion/tail-call-elimination.html
> - https://jfmengels.net/tail-call-optimization/
> - https://blog.janestreet.com/optimizing-list-map/
> - https://jfmengels.net/modulo-cons/

PR #5292, opened 2024-02-06
Prevent the extension from crashing due to unparseable Dark code

OceanOak:
Changelog:

```
Parser
- Update the parser to handle unparseable code 
```

This PR adds :
- a new `Unparseable` type to the `CliScript`  type to handle unparseable code and prevent the extension from crashing.
- a `--skip-self-update` flag for the LSP to prevent running the self-update while using the extension
- a `"parse": "tree-sitter parse"` command to tree-sitter-darklang 

A stack overflow occurs when the number of tokens exceeds ~340, which is consistent with the List.map issue #5222

The issue lies in the `SemanticTokens.toJson` function, specifically in:
```
t.data
 |> Stdlib.List.map (fun d -> Json.Number(Stdlib.UInt64.toFloat d))
 |> Json.Array)
```

### DIFF
--- a/packages/darklang/languageTools/parser.dark
+++ b/packages/darklang/languageTools/parser.dark
@@ -270,32 +270,32 @@ module Darklang =
 
             let defNode =
               match (findNodeByFieldName node "typ") with
-              | Some defNode ->
-                match (parseDefinition defNode) with
-                | Ok def -> def |> Stdlib.Result.Result.Ok
-                | Error e -> Stdlib.Result.Result.Error e
-
+              | Some defNode -> (parseDefinition defNode)
               | None ->
                 Stdlib.Result.Result.Error "No definition node found in type_decl"
 
-            match nameNode, defNode with
-            | Ok name, Ok def ->
+            let keywordTypeNode =
+              match (findNodeByFieldName node "keyword_type") with
+              | Some keywordTypeNode ->
+                (getRange keywordTypeNode) |> Stdlib.Result.Result.Ok
+              | None ->
+                Stdlib.Result.Result.Error "No keyword_type node found in type_decl"
+
+            let symbolEqualsNode =
+              match (findNodeByFieldName node "symbol_equals") with
+              | Some symbolEqualsNode ->
+                (getRange symbolEqualsNode) |> Stdlib.Result.Result.Ok
+              | None ->
+                Stdlib.Result.Result.Error "No symbol_equals node found in type_decl"
+
+            match nameNode, defNode, keywordTypeNode, symbolEqualsNode with
+            | Ok name, Ok def, Ok keywordTypeNode, Ok symbolEqualsNode ->
               (WrittenTypes.TypeDeclaration.TypeDeclaration
                 { range = node.sourceRange
                   name = name
                   definition = def
-                  keywordType =
-                    match (findNodeByFieldName node "keyword_type") with
-                    | Some keywordTypeNode -> getRange keywordTypeNode
-                    | None ->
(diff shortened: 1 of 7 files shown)


packages/darklang/languageTools/semanticTokens.dark
```diff
@@ -167,10 +167,11 @@ module Darklang =
...
               // hello
-              (match contentsMaybe with
-               | Some(contentsRange, _contents) ->
-                 [ makeToken contentsRange TokenType.String ]
-               | None -> [])
+              [ makeToken range TokenType.String ]
+              // (match contentsMaybe with
+              //  | Some(contentsRange, _contents) ->
+              // [ makeToken contentsRange TokenType.String ]
+              //  | None -> [])
```
StachuDotNet, line 174: I think you're right here, and that the commented stuff can be deleted (right?)

OceanOak: Yes, I left it in case it's needed because I didn't understand how it was working. I'll delete it in the next commit

packages/darklang/languageTools/parser.dark
```diff
@@ -193,128 +193,198 @@ module Darklang =
...
+              | Some nameNode ->
+                (Identifiers.parseType nameNode) |> Stdlib.Result.Result.Ok
+              | None -> Stdlib.Result.Result.Error "No name node found in type_decl"
+
+            let defNode =
+              match (findNodeByFieldName node "typ") with
+              | Some defNode ->
+                match (parseDefinition defNode) with
+                | Ok def -> def |> Stdlib.Result.Result.Ok
+                | Error e -> Stdlib.Result.Result.Error e
```
StachuDotNet, line 276: I think this can just be `parseDefinition defNode` ?

packages/darklang/languageTools/parser.dark
```diff
@@ -325,103 +395,170 @@ module Darklang =
...
             let leftArg =
-              (findNodeByFieldName node "left")
-              |> Builtin.unwrap
-              |> Expr.parse
-              |> Builtin.unwrap
+              match (findNodeByFieldName node "left") with
+              | Some leftArgNode ->
+                match (Expr.parse leftArgNode) with
+                | Ok leftArg -> leftArg |> Stdlib.Result.Result.Ok
+                | Error e -> Stdlib.Result.Result.Error e
```
StachuDotNet, line 503: Same as the above - might be OK just as `Expr.parse leftArgNode`

packages/darklang/languageTools/parser.dark
```diff
@@ -498,52 +666,88 @@ module Darklang =
...
+              | None ->
+                Stdlib.Result.Result.Error
+                  "No identifier node found in fn_decl_param"
+
+            let typNode =
+              match (findNodeByFieldName node "typ") with
+              | Some typNode ->
+                match (TypeReference.parse typNode) with
+                | Ok typ -> typ |> Stdlib.Result.Result.Ok
+                | Error e -> Stdlib.Result.Result.Error e
```
StachuDotNet, line 689: here too

StachuDotNet (changes requested):
Looks good! A few cleanups, but then good to merge.

I see there's a few places where we're returning Errors of Strings. I didn't look too closely, and suspect that those are OK -- that their contents are basically ignored, erased by an Unparseable thing, and that it's useful to keep around that string for the time when we include a Reason along with the failures. That said, worth a quick look to make sure we aren't returning a String anywhere we shouldn't

PR #5296, opened 2024-02-09
Refactor cli self-update

OceanOak:
Changelog:

```
CLI: 
- Add a darklang update command to manually update the CLI
- Add bash and zsh modules to the Cli package
- Add some Cli stdlib functions for getting the runtime host and shell
```

This PR refactors the cli's self update code

### DIFF
--- a/packages/darklang/cli/bash.dark
+++ b/packages/darklang/cli/bash.dark
@@ -45,6 +45,10 @@ module Darklang =
         | Ok content ->
           let newContent = content ++ "\n" ++ aliasDefinition
           writeToBashrc newContent
+
+          Builtin.printLine
+            "Changes have been made to your .bashrc file. To apply these changes to your current terminal session, please run the following command:\nsource ~/.bashrc\n\n "
+
           applyBashConfigChanges ()
         | Error e -> Stdlib.Result.Result.Error e
 
--- a/packages/darklang/cli/bash.dark
+++ b/packages/darklang/cli/bash.dark
@@ -16,10 +16,7 @@ module Darklang =
       /// Write content to .bashrc file
       let writeToBashrc (content: String) : Stdlib.Result.Result<String, String> =
         let writeResult = Builtin.fileWrite (Stdlib.String.toBytes content) ".bashrc"
-
-        match writeResult with
-        | Ok _ -> Stdlib.Result.Result.Ok "Wrote to .bashrc"
-        | Error e -> Stdlib.Result.Result.Error e
+        writeResult |> Stdlib.Result.map (fun _ -> "Wrote to .bashrc")
 
 
       /// Delete an alias from .bashrc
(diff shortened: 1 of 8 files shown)


OceanOak:
Notes/ questions:
- Running the self-update manually runs the update twice (because currently we run the self update with every command)
- When a new version of the cli is installed we are not removing the old version; should we remove it? if we should, I think we need to save the previous version's hash in the config file wdyt?

OceanOak:
Notes/ questions:
- Running the self-update manually runs the update twice (because currently we run the self update with every command)
- When a new version of the cli is installed we are not removing the old version; should we remove it? if so, I think we need to save the previous version's hash in the config file wdyt?

packages/darklang/cli/selfUpdate.dark
```diff
@@ -10,105 +17,281 @@ module Darklang =
...
+      let getFromGithub
+        (bearerToken: String)
+        (path: String)
+        : Stdlib.Result.Result<String, String> =
+        let headers =
+          [ ("authorization ", "bearer" ++ bearerToken)
+            ("accept", "application/vnd.github+json")
+            ("user-Agent", "Darklang") ]
+
+        let response = Darklang.Stdlib.HttpClient.request "GET" path headers []
```
OceanOak, line 35: I didn't use the Stdlib.HttpClient.get wrapper function here because it only takes a uri as an arg (headers are empty by default)

packages/darklang/cli/bash.dark
```diff
@@ -0,0 +1,85 @@
...
+        : Stdlib.Result.Result<String, String> =
+        let aliasDefinition = $"alias {alias}='{command}'"
+
+        match (readBashrc ()) with
+        | Ok content ->
+          let newContent = content ++ "\n" ++ aliasDefinition
+          writeToBashrc newContent
+
+          Builtin.printLine
+            "Changes have been made to your .bashrc file. To apply these changes to your current terminal session, please run the following command:\nsource ~/.bashrc\n\n "
```
OceanOak, line 50: Added this message because the apply `applyBashConfigChanges ()` function doesn't seem to work in bash (It works in zsh). The solution is either running  `source ~/.bashrc` manually or the user need to have this line `[[ -f ~/.bashrc ]] && . ~/.bashrc` added to there `bash_profile` file 
any thoughts on how to handle this?

packages/darklang/cli/zsh.dark
```diff
@@ -0,0 +1,81 @@
...
+        | Error e -> Stdlib.Result.Result.Error e
+
+
+      /// Write content to .zshrc file
+      let writeToZshrc (content: String) : Stdlib.Result.Result<String, String> =
+        let writeResult = Builtin.fileWrite (Stdlib.String.toBytes content) ".zshrc"
+
+        match writeResult with
+        | Ok _ -> Stdlib.Result.Result.Ok "Wrote to .zshrc"
+        | Error e -> Stdlib.Result.Result.Error e
```
StachuDotNet, line 23: could be something like
```suggestion
        writeResult |> Result.map (fun _ -> "Wrote to .zshrc")
```

PR #5303, opened 2024-02-16
Add an uninstall command to the CLI

OceanOak:
Changelog:

```
CLI
- Add uninstall command
```

### DIFF
--- a/packages/darklang/cli/UninstallCli.dark
+++ b/packages/darklang/cli/UninstallCli.dark
@@ -1,0 +1,48 @@
+module Darklang =
+  module Cli =
+    module Uninstall =
+      let removeDarklangDirectory () : Stdlib.Result.Result<String, Int64> =
+        Builtin.printLine "Deleting darklang ..."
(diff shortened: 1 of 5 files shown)


packages/darklang/cli/UninstallCli.dark
StachuDotNet, line 1: nitpick - please rename to just `uninstall.dark`, or pull this code into where our install setup is

packages/darklang/cli/bash.dark
```diff
@@ -74,6 +74,25 @@ module Darklang =
           | Error e -> Stdlib.Result.Result.Error e
 
 
+      /// Remove a path from PATH
+      let removeFromPath (path: String) : Stdlib.Result.Result<String, String> =
+        let targetDir = $"$HOME/{path}"
+
+        match (readBashrc ()) with
```
StachuDotNet, line 81:
```suggestion
        match readBashrc () with
```
let's not wrap in parens unless we need to

PR #5312, opened 2024-02-23
Move parsing from semantic tokens to docSync

OceanOak:
No changelog

This PR :
- moves the parsing from semantic tokens to docSync
- prevents the extension from crashing with empty files

### DIFF
--- a/backend/testfiles/execution/stdlib/parser.dark
+++ b/backend/testfiles/execution/stdlib/parser.dark
@@ -77,6 +77,12 @@ module ParseToSimplifiedTree =
                               children = [] } ] } ] } ] }
 
 
+  ("" |> Builtin.parserParseToSimplifiedTree) = ParsedNode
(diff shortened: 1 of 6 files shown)


backend/testfiles/execution/stdlib/parser.dark
```diff
@@ -77,6 +77,12 @@ module ParseToSimplifiedTree =
                               children = [] } ] } ] } ] }
 
 
+  ("" |> Builtin.parserParseToSimplifiedTree) = ParsedNode
```
OceanOak, line 80: Maybe this test isn't necessary?

StachuDotNet: No, I think it's good to have a "what happens if we try to parse an empty string" test -- I see you had to adjust code to deal with the case, I think that warrants a test.

PR #5315, opened 2024-02-28
Expand the grammar to support the remaining signed and unsigned Int types

OceanOak:
Changelog:
```
Tree-sitter-darklang
- Expand the grammar to support the remaining signed and unsigned Int types (Int8, Int16, Int32, Int64, Int128), and update the parser accordingly
```


OceanOak:
Question: 

We currently don't support negative integers (because of `digits: $ => /\d+/,`)

- I've added `positive_digits` and `negative_digits` (because uints will only have positive numbers)
- Made digits `choice($.positive_digits, $.negative_digits)`

 The solution provided in this PR works. but I don't know if it is better to update the rule to something like the following instead. Feedback is appriciated
```
int64_literal: $ =>
  seq(
    field("sign", optional("-")
    field("digits", $.digits),
    field("suffix", alias("L", $.symbol)),
  ),
```

tree-sitter-darklang/grammar.js
StachuDotNet, line 1: Could you please add a single comment somewhere around the Ints, indicating that we're only using .NET suffixes for Ints _temporarily_ until we can get rid of the old parser?

PR #5316, opened 2024-03-05
(wip) More tidying treesitter bindings

StachuDotNet:
I really tried to get rid of the C# here, but I'm running into a bunch of memory-management issues.

https://chat.openai.com/share/cfab34c8-93a7-4465-9b71-4bf9b0535229 represents some of the type of problems I've been dealing with.

In any case, I'm stashing this work, to revisit in a week or so with a clear brain.
In the meantime, I'll merge a smaller directional PR that still works. (and has more C#)


StachuDotNet:
the best path forward here is probably porting as little as possible from C# to F# -- function by function. This effort did too much at once, and it was thus hard to track down the exact place where the memory-management strategies failed.

PR #5323, opened 2024-03-12
Update the grammar to support characters

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support characters 
```


StachuDotNet (changes requested):
One minor change requested.

Also, doesn't need to be part of this PR, but it's worth considering using ".immediate()' more in our grammar. I'm not sure what happens for things like `' a '` right now -- I suspect those spaces are ignored, where they shouldn't be (and this probably applies to other parts of our grammar, like potentially `1 . 2` parsing of floats, etc. Anyway, doesn't need to be addressed in this PR but worth some effort in this space.

OceanOak:
> I'm not sure what happens for things like ' a ' right now -- I suspect those spaces are ignored
Currently, it returns an error, it doesn't ignore the spaces
```(ERROR (UNEXPECTED 'a'))```

PR #5326, opened 2024-03-15
Update the grammar to support lists

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support lists
```


packages/darklang/languageTools/parser.dark
```diff
@@ -229,6 +234,21 @@ module Darklang =
               (WrittenTypes.TypeReference.Builtin.TString node.sourceRange)
               |> Stdlib.Result.Result.Ok
 
+            | text when (Stdlib.String.startsWith text "List") ->
```
OceanOak, line 237: Not sure if this good, if you have a better way in mind (to match directly against `List`) please let me know.
I've tried updating the grammar to `alias ("List", $.list_type)` , but It didn't seem right

backend/testfiles/execution/stdlib/parser.dark
```diff
@@ -212,6 +212,8 @@ module TextToTextRoundtripping =
     ("type MyFloat = Float" |> roundtripCliScript) = "type MyFloat =\n  Float"
     ("type MyChar = Char" |> roundtripCliScript) = "type MyChar =\n  Char"
     ("type MyString = String" |> roundtripCliScript) = "type MyString =\n  String"
+    ("type MyList = List<String>" |> roundtripCliScript) = "type MyList =\n  List<String>"
+    ("type MyList = List<List<String>>" |> roundtripCliScript) = "type MyList =\n  List<List<String>>"
```
OceanOak, line 216: TODO: add a test for a List of a custom type

packages/darklang/languageTools/parser.dark
```diff
@@ -613,6 +636,41 @@ module Darklang =
...
+              |> Stdlib.Option.toResult
+                "No symbol_open_bracket node found in list_literal"
+
+            let contents =
+              node
+              |> findNodeByFieldName "content"
+              |> Stdlib.Option.map (fun contentsNode ->
+                contentsNode.children
+                |> Stdlib.List.filter (fun c -> c.text != ";")
+                |> Stdlib.List.map (fun c -> c |> Expr.parse |> Builtin.unwrap)
```
OceanOak, line 655: Maybe use Result.combine here?

backend/testfiles/execution/stdlib/parser.dark
StachuDotNet: Do you want to add this as a comment in the .dark file?

tree-sitter-darklang/grammar.js
```diff
@@ -311,10 +329,20 @@ module.exports = grammar({
         /Float/,
         /Char/,
         /String/,
+        $.list_type,
         /DateTime/,
         /Uuid/,
       ),
 
+    // List<T>
+    list_type: $ =>
```
StachuDotNet, line 338: let's rename this to list_type_reference or type_reference_list

backend/testfiles/execution/stdlib/parser.dark
OceanOak: I will add the test before the merge, I am just in a different branch currently

tree-sitter-darklang/grammar.js
```diff
@@ -290,6 +291,23 @@ module.exports = grammar({
     // Floats
     float_literal: $ => /[+-]?[0-9]+\.[0-9]+/,
 
+    //
+    // List
```
StachuDotNet, line 295: let's add a TODO here of something like: "allow multi-line lists where a newline is 'interpreted' as a list delimiter (i.e. no `;` needed)"

StachuDotNet: another TODO: support `[1; 2;]` (users ending with a `;` -- as part of being a 'tolerant' parser)

packages/darklang/languageTools/writtenTypes.dark
```diff
@@ -153,6 +154,12 @@ module Darklang =
           symbolOpenQuote: SourceRange *
           symbolCloseQuote: SourceRange
 
+        | EList of
+          SourceRange *
+          contents: Stdlib.Option.Option<SourceRange * List<Expr>> *
```
StachuDotNet, line 159: I think a better 'shape' for this might be
```suggestion
          contents: List<Expr * Option<SourceRange>> *
```
where:
- the Expr is ... just the expr, which already contains its own SourceRange
- the SourceRange after that is where the `;` exists, if it does

Some notes:
- the 'optionalness' of the content is implied by contents just being empty
- this also accounts for `[1; 2;]` noted elsewhere, which I think is a realistic scenario
- this set-up is a bit less nested, and might be easier to work with

packages/darklang/languageTools/writtenTypes.dark
```diff
@@ -50,6 +50,7 @@ module Darklang =
           | TFloat of SourceRange
           | TChar of SourceRange
           | TString of SourceRange
+          | TList of SourceRange * TypeReference.TypeReference
```
StachuDotNet, line 53: I think this is missing:
- the SourceRange of the `List` word itself
- the SourceRange of `<` and of `>`

maybe it's overkill to include those? I think including them would be consistent with our def of `EInt8` for example, though, where we include SourceRanges for symbols

packages/darklang/languageTools/parser.dark
StachuDotNet: You have to wrap that alias in a field to make the grammar.js change have any effect -- mentioned in more detail there.

StachuDotNet (changes requested):
Nice! left some notes

packages/darklang/languageTools/parser.dark
OceanOak: Yes, I remember something about anonymous nodes. I just wasn't sure if it was appropriate to use a field in this case, but this confirms it. Thanks.

tree-sitter-darklang/grammar.js
```diff
@@ -290,6 +291,26 @@ module.exports = grammar({
...
+    // TODO: allow multi-line lists where a newline is 'interpreted' as a list delimiter (i.e. no ; needed)
+    // TODO: support [1; 2;] (lists ending with a ; -- as part of being a 'tolerant' parser)
+    list_literal: $ =>
+      seq(
+        field("symbol_open_bracket", alias("[", $.symbol)),
+        field("content", optional($.list_content)),
+        field("symbol_close_bracket", alias("]", $.symbol)),
+      ),
+
+    list_content: $ =>
```
OceanOak, line 305: Couldn't use the following rule, because this will pass [1 2]
```
    list_content: $ =>
      repeat1(
        seq(
          $.expression,
          optional(field("list_separator", alias(";", $.symbol))),
        ),
      ),
 ```

File: backend/testfiles/execution/stdlib/parser.dark [hunk 9]
```diff
@@ -269,13 +272,22 @@ module TextToTextRoundtripping =
     ("'\\n'" |> roundtripCliScript) = "'\\n'"
     ("'\t'" |> roundtripCliScript) = "'\t'"
 
+    // list literal
+    ("[]" |> roundtripCliScript) = "[]"
+    ("[\"hello\"]" |> roundtripCliScript) = "[\"hello\"]"
+    ("[1L; 2L]" |> roundtripCliScript) = "[1L; 2L]"
+    // TODO: fix this, the trailing semicolon is getting removed
+    ("[1L; 2L; 3L;]" |> roundtripCliScript) = "[1L; 2L; 3L]"
```
StachuDotNet, line 280:
```suggestion
    ("[1L; 2L; 3L;]" |> roundtripCliScript) = "[1L; 2L; 3L]"
```
It's fine that the trailing semicolon is removed here -- it's an "extra" in this context.
(its existence is 'erased' during WT->PT)

tree-sitter-darklang/grammar.js
StachuDotNet: Cool - I like how this turned out, anyway -- better represents the situation we're trying to capture.

PR #5329, opened 2024-03-18
Update the grammar to support tuples

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support tuples
```


packages/darklang/languageTools/parser.dark
```diff
@@ -240,6 +242,83 @@ module Darklang =
...
+                match leftParen, first, asterisk, second, rightParen with
+                | Ok leftParen, Ok first, Ok asterisk, Ok second, Ok rightParen ->
+                  (WrittenTypes.TypeReference.Builtin.TTuple
+                    node.sourceRange
+                    first
+                    asterisk.sourceRange
+                    second
+                    rest
+                    leftParen.sourceRange
+                    rightParen.sourceRange)
```
StachuDotNet, line 315: Do we not need the parts of TTuple to be wrapped in ()? with , separators? not sure how this is working, but we should probably add those anyway.

tree-sitter-darklang/grammar.js
```diff
@@ -310,6 +311,26 @@ module.exports = grammar({
...
+      seq(
+        field("symbol_left_paren", alias("(", $.symbol)),
+        field("first", $.expression),
+        field("symbol_comma", alias(",", $.symbol)),
+        field("second", $.expression),
+        field("rest", optional($.rest_expressions)),
+        field("symbol_right_paren", alias(")", $.symbol)),
+      ),
+
+    rest_expressions: $ =>
```
StachuDotNet, line 326: how about `tuple_literal_the_rest` or something? Each of these names should be really explicit

Issue #5333, opened 2024-03-26
Component: Queues/Workers


StachuDotNet:
This Issue exists to collect many items that relate to "Queues/Workers" in Darklang.
In Darklang-classic, queues only existed in our Cloud runtime (since that was the only runtime), but going forward we should have them supported for the Local/CLI runtime as well.

- [ ] uncomment the `_events.dark` test file. (This was quite flaky in CI, but we decided to put addressing that for a bit, and have uncommented this until we have some focus on queues/workers)
- [ ] bring back Queues, in some minimal capacity
- [ ] add timelimit for queue execution
  - the goal here is to limit the execution of an item on the queue
  - To end computation that runs over (lets say) 3 minutes. I'm not sure it's possible without adding cancellation tokens everywhere - maybe that's what we'll need to do?
- [ ] Workers should be typed
- [ ] `emit`'s arg should be real worker name/reference, not a string
- [ ] worker counts should work
- [ ] it should be clearer if a worker is paused
- [ ] workers should have dead-letter queues
- [ ] review Rollbar: "Deadline Exceeded" (#3815)

PR #5339, opened 2024-03-26
disable flaky queue tests (for now)

StachuDotNet:
When we get back to queues, we can get back to handling the flakiness of these tests.


StachuDotNet:
see #5333

PR #5343, opened 2024-04-02
Update the grammar to support if/else statements

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support if/else statements
- Add an external scanner to handle indentation
```


StachuDotNet:
I wonder if we could write a function in our grammar.js that would abstract over this pattern

```
choice(
  seq(
    $.indent,
    repeat1(field("then_expression", $.expression)),
    $.dedent,
  ),
  field("then_expression", $.paren_expression),
),
```

like
      
```
let maybeWithIndentDedent (thing) = 
  choice(
    seq($.indent, thing, $.dedent),
    thing,
  ),
```

tree-sitter-darklang/grammar.js
```diff
@@ -355,6 +357,34 @@ module.exports = grammar({
...
+        field("keyword_if", alias("if", $.keyword)),
+        field("condition", $.expression),
+        field("keyword_then", alias("then", $.keyword)),
+        choice(
+          seq(
+            $.indent,
+            repeat1(field("then_expression", $.expression)),
+            $.dedent,
+          ),
+          field("then_expression", $.paren_expression),
```
StachuDotNet, line 371: it's neat that you needed a paren_expression here, but not in the equivalent else_expression below. This tree-sitter conflict stuff continues to be a mystery to me.

backend/testfiles/execution/stdlib/parser.dark
```diff
@@ -302,6 +302,107 @@ module TextToTextRoundtripping =
...
+else if true then
+  d"""
+     |> roundtripCliScript) = """if true then
+  a
+else
+  if false then
+    c
+  else
+    if true then
+      d"""
```
StachuDotNet, line 331: Since this PR is just focused on the parser, this is good -- could we follow this PR up with one focused on the pretty-printing of EIfs?
such that:
if an `if` has an else, whose root expression is an `if` as well, we format like:
```
if cond then
  expr
else if cond2 then
  expr2
```
?

StachuDotNet: (it might turn out that _that_ code might need to 'track' indentation to do its job correctly)

OceanOak: Yes, I tried to make it work within this PR, but encountered some issues with the indentations

PR #5349, opened 2024-04-15
Apply self-closing tags only to void elements

OceanOak:
No changelog

### DIFF
--- a/packages/darklang/stdlib/html.dark
+++ b/packages/darklang/stdlib/html.dark
@@ -65,7 +65,7 @@ module Darklang =
           if (Stdlib.List.``member`` voidElements tag.name) then
             match attributesText with
             | "" -> $"<{tag.name}/>"
-            | text -> $"<{tag.name} {text}> </{tag.name}>"
+            | text -> $"<{tag.name} {text}/>"
 
           else
             match tag.children with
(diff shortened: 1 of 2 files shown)


packages/darklang/stdlib/html.dark
```diff
@@ -36,6 +36,24 @@ module Darklang =
...
+              "hr"
+              "img"
+              "input"
+              "keygen"
+              "link"
+              "meta"
+              "param"
+              "source"
+              "track"
+              "wbr" ]
```
StachuDotNet, line 55: let's:
- move this to a constant in the root of the Html node
- put a /// comment to the HTML spec to note what it's about

Feel free to merge as-is and do that in another PR, or follow up in some other PR

packages/darklang/stdlib/html.dark
```diff
@@ -64,8 +65,8 @@ module Darklang =
 
           if (Stdlib.List.``member`` voidElements tag.name) then
             match attributesText with
-            | "" -> $"<{tag.name}/>"
-            | text -> $"<{tag.name} {text}/>"
+            | "" -> $"<{tag.name}>"
```
OceanOak, line 68: I updated this from `<{tag.name}/>` to just `<{tag.name}>` since HTML doesn't recognize self-closing tags, and the parser disregards the slash character [link](https://developer.mozilla.org/en-US/docs/Glossary/Void_element#:~:text=If%20a%20trailing%20/%20(slash)%20character%20is%20present%20in%20the%20start%20tag%20of%20an%20HTML%20element)

Issue #5321, opened 2024-03-11
Grammar updates to make


OceanOak:
This issue gathers the grammar updates that need to be done to match more of our full language.

Expressions:
- [x] unit
- [x] bool
- [x] signed and unsigned Ints (Int8, Int16, Int32, Int64, Int128, UInt8, UInt16, UInt32, UInt64, UInt128)
- [x] float
- [ ] char
- [x] string
- [x] let expression
- [ ] if/else 
- [ ] infix
    - [ ] +, -, *, /, %, ^, >, <, >=, <=, ==, !=, ++
    - [ ] ||, &&
- [ ] lambda
- [ ] field access
- [ ] variable
- [ ] list
- [ ] dict
- [ ] tuple
- [ ] pipe
- [ ] record
- [ ] enum
- [ ] match
- [ ] record update
- [x] paren_expression

Builtin types:
- [ ] Variable
- [x] Unit
- [x] Bool
- [x] Signed and Unsigned Ints (Int8, Int16, Int32, Int64, Int128, UInt8, UInt16, UInt32, UInt64, UInt128)
- [x] Float
- [ ] Char
- [x] String
- [ ] DateTime
- [ ] Uuid
- [ ] List
- [ ] Tuple
- [ ] Dict
- [ ] CustomType
- [ ] DB
- [ ] Fn

Constants:
- [ ] Unit
- [ ] Bool
- [ ] Signed and Unsigned Ints (Int8, Int16, Int32, Int64, Int128, UInt8, UInt16, UInt32, UInt64, UInt128)
- [ ] Float
- [ ] Char
- [ ] String
- [ ] List
- [ ] Tuple
- [ ] Dict
- [ ] Enum

Source file: 
- [x] Type declaration
- [x] Function declaration
- [ ] Constant declaration 

Other:
- [x] Qualified function name
- [x] Qualified type name
- [ ] Comments
- [ ] indentation?
- [ ] generics ('a)?
- [ ] type parameter (GenericThing<'thing>)?
- [ ] string interpolation
- [ ] blank (eg. Error _ ->...)?

OceanOak:
Parser issues to fix: 
- [ ] handle keywords better
- example: 
   ```
    let isEmpty (dict: Dict<'a>) : Bool =
    dict == Dict.empty // Dict is getting parsed as a keyword instead of a module name
   ```
- [ ] Allow function calls with indented arguments in lists without requiring parentheses
- example: 
  ```
    [ Stdlib.Int64.add
[...]

PR #5357, opened 2024-04-22
Tree-sitter grammar: support Enums

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support Enums
```

#5321

### DIFF
--- a/packages/darklang/languageTools/lsp-server/semanticTokens.dark
+++ b/packages/darklang/languageTools/lsp-server/semanticTokens.dark
@@ -26,7 +26,8 @@ module Darklang =
             "parameter" // [7] for function parameter identifiers
             "variable" // [8] for general identifiers
             "function" // [9] for function names/identifiers
-            "property" ] // [10] for field names
+            "property" // [10] for field names
+            "enumCase" ] // [11] for enum case names
 
         let tokenModifiers = []
 
@@ -53,6 +54,7 @@ module Darklang =
             | FunctionName -> 9UL
 
             | Property -> 10UL
+            | EnumCase -> 11UL
 
           let toRelativeTokens
             (tokens: List<LanguageTools.SemanticTokens.SemanticToken>)
--- a/packages/darklang/languageTools/lsp-server/semanticTokens.dark
+++ b/packages/darklang/languageTools/lsp-server/semanticTokens.dark
@@ -27,7 +27,7 @@ module Darklang =
             "variable" // [8] for general identifiers
             "function" // [9] for function names/identifiers
             "property" // [10] for field names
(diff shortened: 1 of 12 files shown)


packages/darklang/languageTools/parser.dark
```diff
@@ -554,6 +554,73 @@ module Darklang =
...
+                        { range = caseNode.sourceRange
+                          name = (caseNameNode.sourceRange, caseNameNode.text)
+                          fields = fields
+                          description = ""
+                          keywordOf = keywordOf })
+                      |> Stdlib.Result.Result.Ok
+                    | None ->
+                      (WrittenTypes.Unparseable { source = caseNode })
+                      |> Stdlib.Result.Result.Error)
+                  |> Stdlib.Result.values
```
OceanOak, line 613: I'm hesitant to use `Stdlib.Result.values` here because it ignores the errored values. Do you have any suggestions for a better approach?

packages/darklang/languageTools/lsp-server/semanticTokens.dark
```diff
@@ -26,7 +26,8 @@ module Darklang =
             "parameter" // [7] for function parameter identifiers
             "variable" // [8] for general identifiers
             "function" // [9] for function names/identifiers
-            "property" ] // [10] for field names
+            "property" // [10] for field names
+            "enumCase" ] // [11] for enum case names
```
StachuDotNet, line 30: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokenTypes I think `enumMember` may be the more universal term. (using the universal term should help various VS Code themes style our code correctly)

packages/darklang/languageTools/writtenTypes.dark
```diff
@@ -217,6 +232,13 @@ module Darklang =
           symbolOpenBrace: SourceRange *
           symbolCloseBrace: SourceRange
 
+        | EEnum of
+          SourceRange *
+          typeName: (SourceRange * List<String>) *
+          caseName: (SourceRange * String) *
+          fields: List<Expr> *
+          symbolDot: SourceRange
```
StachuDotNet, line 240:
```suggestion
          /// (between the typeName and the caseName)
          symbolDot: SourceRange
```
I think this comment would be right?

StachuDotNet (approved):
Good! I answered one question and made a small suggestion. Both of which can be handled in a subsequent PR if that makes it easier (since I imagine you're busy doing the refactor)

packages/darklang/languageTools/parser.dark
OceanOak: Yes, thanks!

PR #5361, opened 2024-04-29
Rename sourceRange to range

OceanOak:
No changelog

### DIFF
--- a/packages/darklang/languageTools/parser/core.dark
+++ b/packages/darklang/languageTools/parser/core.dark
@@ -21,7 +21,7 @@ module Darklang =
 
           /// Where in the source code is this node written/contained
           /// i.e. Line 1, Column 2 to Line 1, Column 5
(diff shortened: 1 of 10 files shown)


packages/darklang/languageTools/parser/core.dark
```diff
@@ -21,7 +21,7 @@ module Darklang =
 
           /// Where in the source code is this node written/contained
           /// i.e. Line 1, Column 2 to Line 1, Column 5
-          sourceRange: Range
+          range: Range
```
OceanOak, line 24: Do we only need to rename this specific `sourceRange` label, or should we update all instances of sourceRanges,
( eg. `type SourceRange = Parser.Range`)?

StachuDotNet: We can rename those SourceRange types to just Range as well, I think

PR #5364, opened 2024-05-02
Tree-sitter grammar: support match expressions

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support match expressions
```

#5321

### DIFF
--- a/tree-sitter-darklang/grammar.js
+++ b/tree-sitter-darklang/grammar.js
@@ -10,6 +10,7 @@ const PREC = {
   SUM: 3,
   PRODUCT: 4,
   EXPONENT: 5,
+  MATCH_EXPR: 6,
   FIELDACCESS: 7,
 };
 
@@ -37,8 +38,9 @@ module.exports = grammar({
         repeat($.expression),
       ),
 
-    //
+    // ---------------------
     // Function declarations
+    // ---------------------
     fn_decl: $ =>
       seq(
         field("keyword_let", alias("let", $.keyword)),
@@ -59,8 +61,9 @@ module.exports = grammar({
         field("symbol_right_paren", alias(")", $.symbol)),
       ),
 
-    //
+    // ---------------------
     // Type declarations
+    // ---------------------
     type_decl: $ =>
       seq(
         field("keyword_type", alias("type", $.keyword)),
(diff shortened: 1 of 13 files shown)


OceanOak:
Questions:
- What should we do about wildcard eg. `Error _ -> ...` or `| _ -> ...`
we can add support for it in the grammar, but I am not sure how it will be captured in writtenTypes. Could you please provide some insights on this?
- why don't we have MPDict?

StachuDotNet:
> * What should we do about wildcard eg. `Error _ -> ...` or `| _ -> ...`
>   we can add support for it in the grammar, but I am not sure how it will be captured in writtenTypes. Could you please provide some insights on this?
It's OK to capture these as Variable match patterns.
Maybe someday we'll have a separate Wildcard MatchPattern, though -- I've thought that for a while - I'm not sure, then, what `_this` would be captured as (wildcard or variable?).

> * why don't we have MPDict?
I'm not sure what that would look like. I don't believe similar languages have one we could use for inspiration (besides's F#'s [record pattern](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/pattern-matching#record-pattern) but I don't think it perfectly applies).
I suppose the answer here is that our other match patterns are used to _deconstruct structures_ and extract values out of them, safely, but the fields of a dict are unstable, so unsafe to deconstruct them
Also, our other patterns can be checked at parse-time, whereas a dict one (if we could imagine how it looks) probably couldn't.

tree-sitter-darklang/grammar.js
```diff
@@ -531,7 +613,64 @@ module.exports = grammar({
       ),
 
     //
+    // Match expression
```
StachuDotNet, line 616: let's add a TODO to allow for one-line `match` exprs (`match myInt with | 1 -> ...`)

StachuDotNet (changes requested):
Looking great! After adding that TODO and a test for `| _ -> ...`, I think this is good to go.

tree-sitter-darklang/grammar.js
```diff
@@ -184,6 +184,7 @@ module.exports = grammar({
         alias($.mp_tuple, $.tuple),
         alias($.mp_enum, $.enum),
         alias($.variable_identifier, $.variable),
+        alias(/_/, $.wildcard),
```
OceanOak, line 187: should we allow `variable_identifier`to start with `_` instead of handling it this way?

StachuDotNet: Hmm, yeah I think so.

I've been thinking, and there's some value to (future) having a specific wildcard pattern: anything with _:
- may not be accessed
- should not be complained about, by the type-checker, for lack of usage (where we might otherwise warn "hey you captured this variable but didn't use it!")

PR #5365, opened 2024-05-03
Tree-sitter grammar: support field access

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support field access
```

#5321

### DIFF
--- a/backend/testfiles/execution/stdlib/parser.dark
+++ b/backend/testfiles/execution/stdlib/parser.dark
@@ -388,6 +388,10 @@ module TextToTextRoundtripping =
 
     // field access
     ("person.name" |> roundtripCliScript) = "person.name"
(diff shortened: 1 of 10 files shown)


backend/testfiles/execution/stdlib/parser.dark
```diff
@@ -386,6 +386,9 @@ module TextToTextRoundtripping =
     // TODO: this is ugly
     ("let x = 1L\n  x" |> roundtripCliScript) = "let x =\n  1L\nx"
 
+    // field access
```
StachuDotNet, line 389: let's also add some others:
- `(Person { name = "Janice" }).name` (just to make sure it works when there's a ()-defined record we're accessing)
- `record.someField.anotherFieldInsideThat` (nested record access)

PR #5367, opened 2024-05-07
Tree-sitter grammar: support Record update

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support Record update
```
#5321


StachuDotNet (approved):
Excellent.

PR #5368, opened 2024-05-10
Tree-sitter grammar: support pipe expression

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support pipe expression
```
#5321

### DIFF
--- a/backend/testfiles/execution/language/epipe.dark
+++ b/backend/testfiles/execution/language/epipe.dark
@@ -42,6 +42,11 @@ type X = { y: Y }
 type MyEnum = A of Int64 * Int64 * Int64
 (33L |> MyEnum.A 21L 42L) = MyEnum.A 33L 21L 42L
 
+// TODO: test this in Parser.Tests
+(33L |> MyEnum.A 21L 42L) = MyEnum.A(33L, 21L, 42L)
+(33L |> MyEnum.A(21L, 42L)) = MyEnum.A 33L 21L 42L
+(33L |> MyEnum.A(21L, 42L)) = MyEnum.A(33L, 21L, 42L)
+
 (3L |> Stdlib.Result.Result.Ok) = Stdlib.Result.Result.Ok 3L
 
 (4L |> (+) 3L |> Stdlib.Option.Option.Some) = Stdlib.Option.Option.Some 7L
--- a/backend/testfiles/execution/language/epipe.dark
+++ b/backend/testfiles/execution/language/epipe.dark
@@ -41,10 +41,6 @@ type X = { y: Y }
 
 type MyEnum = A of Int64 * Int64 * Int64
 (33L |> MyEnum.A 21L 42L) = MyEnum.A 33L 21L 42L
-
-// TODO: test this in Parser.Tests
-(33L |> MyEnum.A 21L 42L) = MyEnum.A(33L, 21L, 42L)
-(33L |> MyEnum.A(21L, 42L)) = MyEnum.A 33L 21L 42L
 (33L |> MyEnum.A(21L, 42L)) = MyEnum.A(33L, 21L, 42L)
 
 (3L |> Stdlib.Result.Result.Ok) = Stdlib.Result.Result.Ok 3L
--- a/packages/darklang/languageTools/semanticTokens.dark
+++ b/packages/darklang/languageTools/semanticTokens.dark
@@ -276,6 +276,7 @@ module Darklang =
             tokenizeDefinition t.definition ]
           |> Stdlib.List.flatten
 
+      // TODO: maybe this should be inside `Expr` module?
       module MatchPattern =
         let tokenize (p: WrittenTypes.MatchPattern) : List<SemanticToken> =
           match p with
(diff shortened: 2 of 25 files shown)


tree-sitter-darklang/grammar.js
```diff
@@ -695,7 +697,16 @@ module.exports = grammar({
...
+          field(
+            "expr",
+            choice(
+              $.variable_identifier,
+              $.record_literal,
+              $.paren_expression,
+              $.field_access,
+              $.apply,
+            ),
+          ),
```
OceanOak, line 709: is this update a bad idea?
    this change was because `1L |> Stdlib.Int64.toString` was parsed as a
    fieldAccess of a pipeEnum i.e `(1L |> Stdlib.Int64).toString` instead of a pipeFnCall `1L |> Stdlib.Int64.toString`

tree-sitter-darklang/grammar.js
```diff
@@ -764,7 +775,7 @@ module.exports = grammar({
     apply: $ =>
       prec.right(
         seq(
-          field("fn", $.qualified_fn_name),
+          field("fn", choice($.qualified_fn_name, $.fn_identifier)),
```
OceanOak, line 778: because I updated `qualified_fn_name` to require one module name I had to add fn_identifier here, not sure about this decision.

Reason behind the `qualified_fn_name`  update:
`1 |> x` was parsed as `pipe_fn_call(qualified_fn_name) ` instead of `pipe_variable_or_fn_call`

tree-sitter-darklang/grammar.js
```diff
@@ -784,6 +795,156 @@ module.exports = grammar({
...
       ),
 
+    // ---------------------
+    // Pipe expressions
+    // ---------------------
+
+    pipe_variable_or_fn_call: $ =>
+      seq(field("variable_or_fn_name", $.variable_or_fn_identifier)),
+
+    pipe_lambda: $ =>
```
OceanOak, line 805: should `$.lambda_expression` be reused here?
i.e.
```
pipe_lambda: $ =>
      choice(
        $.lambda_expression,
        seq(
          field("symbol_open_paren", alias("(", $.symbol)),
          $.lambda_expression,
          field("symbol_close_paren", alias(")", $.symbol)),
        ),
      ),
      ```

File: backend/testfiles/execution/language/epipe.dark [hunk 4]
```diff
@@ -42,6 +42,11 @@ type X = { y: Y }
 type MyEnum = A of Int64 * Int64 * Int64
 (33L |> MyEnum.A 21L 42L) = MyEnum.A 33L 21L 42L
 
+// TODO: test this in Parser.Tests
```
OceanOak, line 45: Note to self: revisit this before the PR merge

packages/darklang/languageTools/writtenTypes.dark
```diff
@@ -340,6 +337,27 @@ module Darklang =
...
+          Range *
+          typeName: (Range * List<String>) *
+          caseName: (Range * String) *
+          fields: List<Expr> *
+          /// between the typeName and the caseName
+          symbolDot: Range
+        | EPipeFnCall of
+          Range *
+          fnName: Name *
+          typeArgs: List<TypeReference.TypeReference> *
```
OceanOak, line 358: Note to self: I remember reading a comment about removing `typeArgs` from this, try to find the comment

tree-sitter-darklang/grammar.js
StachuDotNet: I think it's actually a great solution - clever. Maybe "clever" will end up being _too clever_ (it happens sometimes) but it seems reasonable to me. Luckily, there's composition in play here, so we can always adjust this later if that turns out to be the case.

Anyway, I'd keep it!

tree-sitter-darklang/grammar.js
StachuDotNet: Yeah, I think so. I assume the same thing is reused in both equivalent parts of PT? if so, then yes.

tree-sitter-darklang/grammar.js
StachuDotNet: I'm very hesitant about updating `qualified_fn_name` to require at least one module name - we should avoid that.

> `1 |> x was parsed as pipe_fn_call(qualified_fn_name)`  instead of `pipe_variable_or_fn_call`

Hm at first glance, that seems like the right thing to do, actually?

Ah, I guess the counterargument would be when you allow for inline lambdas / fn declarations, like:
```fsharp
let myLambda = fun x -> x + 1
1 |> myLambda |> etc
```

I suppose, in that case, `myLambda` is more of a variable identifier than a fn name.

So, I think
- WT should be one of those "maybe it's a variable name, maybe it's a fn name" things
- WT2PT should _try_ to find a fn of the name -- if it can't find one, it's an `EPipeVariable` - if it can, it's an `EPipeVariableOrFnCall` - or something like that

hope that makes sense

packages/darklang/languageTools/writtenTypes.dark
StachuDotNet: I recall making that comment somewhere, some time ago. I think I only added the comment to simplify the grammar a bit, until it was complete. Long-term, I think you _should_ be able to `"1" |> Json.parse<Int>`, in which case `typeArgs` here is OK.

backend/testfiles/execution/language/epipe.dark
```diff
@@ -41,6 +41,7 @@ type X = { y: Y }
 
 type MyEnum = A of Int64 * Int64 * Int64
 (33L |> MyEnum.A 21L 42L) = MyEnum.A 33L 21L 42L
+(33L |> MyEnum.A(21L, 42L)) = MyEnum.A(33L, 21L, 42L)
```
StachuDotNet, line 44: Ah, I don't like this. I think you should only be able to "pipe into an enum" when the enum has a _single_ field.
(maybe also with multiple fields, _if you pass in a tuple_)

Can we restrict things as such?

backend/testfiles/execution/language/epipe.dark
```diff
@@ -41,6 +41,7 @@ type X = { y: Y }
 
 type MyEnum = A of Int64 * Int64 * Int64
 (33L |> MyEnum.A 21L 42L) = MyEnum.A 33L 21L 42L
```
StachuDotNet, line 43: same here - not good. let's remove tests that don't align with our actual expectations

backend/testfiles/execution/language/epipe.dark
StachuDotNet: To be clearer,

this are OK/legal:
`type MyEnum = | A of Bool` followed by `true |> MyEnum.A`

this is not:
`type MyEnum = | A of Bool * Bool` followed by `true |> MyEnum.A(false)`

this is maybe-ok, but fine to not implement: 
`type MyEnum = | A of Bool * Bool` followed by `(true, false) |> MyEnum.A`


---

Trying to do the "not ok" thing above is basically like saying:
`Tuple.addField (true,false) false` and expecting `(true, false, false)` which is a bit nonsensical as those are _totally different types_. I tried to do something similarly silly with tuples back in #4236. (enum fields are "tuple-like" in the same way, I think)

packages/darklang/languageTools/semanticTokens.dark
```diff
@@ -276,6 +276,7 @@ module Darklang =
             tokenizeDefinition t.definition ]
           |> Stdlib.List.flatten
 
+      // TODO: maybe this should be inside `Expr` module?
```
StachuDotNet, line 279:
```suggestion
```

I think it's good as-is, unless we refactor other parts of our codebase similarly - more accessible this way

PR #5370, opened 2024-05-15
Create stdlib wrappers for built-in db functions

OceanOak:
No changelog

### DIFF
--- a/packages/darklang/stdlib/db.dark
+++ b/packages/darklang/stdlib/db.dark
@@ -1,0 +1,81 @@
+module Darklang =
+  module Stdlib =
+    module DB =
+      /// Upsert <param val> into <param table>, accessible by <param key>
+      let set (value: 'a) (key: String) (table: 'a) : 'a =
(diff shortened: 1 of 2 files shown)


packages/darklang/stdlib/db.dark
```diff
@@ -0,0 +1,91 @@
+module Darklang =
+  module Stdlib =
+    module DB =
+      /// Upsert <param val> into <param table>, accessible by <param key>
+      let set (value: 'a) (key: String) (table: 'a) : 'a =
```
OceanOak, line 5: Should this be `(table: DB<'a>)` instead of `(table: 'a)`? 
I don't think we have support for this syntax `DB<'a>`

StachuDotNet: If tests are passing, then I think this is OK - can you just add CLEANUPs that they _should_ be `DB<'a>`-style? We can follow up on that when we get rid of LibParser

PR #5372, opened 2024-05-16
Tree-sitter grammar: support DB type reference

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support DB type reference
```

#5321

### DIFF
--- a/packages/darklang/languageTools/parser/identifiers.dark
+++ b/packages/darklang/languageTools/parser/identifiers.dark
@@ -55,30 +55,30 @@ module Darklang =
           : Stdlib.Result.Result<WrittenTypes.QualifiedTypeIdentifier, String> =
           if node.typ == "qualified_type_name" then
             let filteredChildren =
-              node.children
-              |> Stdlib.List.filter (fun n ->
-                n.typ != "type_args"
-                && n.fieldName != Stdlib.Option.Option.Some "symbol_open_angle"
-                && n.fieldName != Stdlib.Option.Option.Some "symbol_close_angle")
+              node.children |> Stdlib.List.filter (fun n -> n.typ != "type_args")
 
             let (modules, typeIdentifierNode) =
               extractModuleIdentifiers filteredChildren
 
-            let typeArgsNodes =
-              node
-              |> findNodeByFieldName "type_args"
-              |> Stdlib.Option.map (fun typeArgsNode ->
-                typeArgsNode.children
-                |> Stdlib.List.chunkBySize 2L
-                |> Builtin.unwrap
-                |> Stdlib.List.map (fun chunk ->
-                  match chunk with
-                  | [ typeArg; _ ] -> TypeReference.parse typeArg
-                  | [ typeArg ] -> TypeReference.parse typeArg
-                  | _ -> createUnparseableError chunk))
+            let args =
+              match findNodeByFieldName node "type_args" with
+              | Some typeArgs ->
+                typeArgs
+                |> findNodeByFieldName "args"
+                |> Stdlib.Option.map (fun typeArgsNode ->
+                  typeArgsNode.children
+                  |> Stdlib.List.chunkBySize 2L
+                  |> Builtin.unwrap
+                  |> Stdlib.List.map (fun chunk ->
+                    match chunk with
+                    | [ typeArg; _ ] -> TypeReference.parse typeArg
+                    | [ typeArg ] -> TypeReference.parse typeArg
+                    | _ -> createUnparseableError chunk))
+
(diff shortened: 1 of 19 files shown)


packages/darklang/languageTools/parser/identifiers.dark
```diff
@@ -54,14 +54,43 @@ module Darklang =
           (node: ParsedNode)
           : Stdlib.Result.Result<WrittenTypes.QualifiedTypeIdentifier, String> =
           if node.typ == "qualified_type_name" then
+            let filteredChildren =
+              node.children
+              |> Stdlib.List.filter (fun n ->
+                n.typ != "type_args"
+                && n.fieldName != Stdlib.Option.Option.Some "symbol_open_angle"
```
OceanOak, line 61: not very happy with this solution, any suggestion of alternatives please?

packages/darklang/prettyPrinter/programTypes.dark
```diff
@@ -883,7 +883,7 @@ module Darklang =
           | [] -> ""
           | _ ->
             p.declaration.typeParams
-            |> Stdlib.List.map (fun typeParam -> $"'{typeParam}")
+            |> Stdlib.List.map (fun typeParam -> $"{typeParam}")
```
StachuDotNet, line 886: Hm - let's update WrittenTypes to record the `'` as a symbol, separate from the actual type variable.
Then, the pretty-printer probably needs the `'` to be included like the source was.

tree-sitter-darklang/grammar.js
```diff
@@ -70,9 +70,26 @@ module.exports = grammar({
       seq(
         field("keyword_type", alias("type", $.keyword)),
         field("name", $.type_identifier),
+        optional(
+          seq(
+            field("symbol_open_angle", alias("<", $.symbol)),
+            field("type_params", $.type_decl_type_params),
+            field("symbol_close_angle", alias(">", $.symbol)),
+          ),
+        ),
```
StachuDotNet, line 79: the syntax to apply type args will be the same for many places (`DB<Int>`, `myFn<Int>`, `MyType<Int>`) - let's extract this out to a `type_args` root-level thing, and share across the use cases.

tree-sitter-darklang/grammar.js
```diff
@@ -874,8 +903,25 @@ module.exports = grammar({
...
+      ),
+
+    type_args: $ =>
+      seq(
+        $.type_reference,
+        repeat(
+          seq(field("symbol_comma", alias(",", $.symbol)), $.type_reference),
+        ),
       ),
```
StachuDotNet, line 925: Separate from my other long comment: I do think that `type_args` should include the `<` symbols.

packages/darklang/languageTools/parser/identifiers.dark
StachuDotNet: I think `type_args` should include the `<` and `>` characters - then, the task seems simpler: make sure that `<` is the first thing in the list, `>` is the last thing in the list, and then extract the other things as you're doing. Then, you just need to look for the one `type_args` node, here.

tree-sitter-darklang/test/corpus/exhaustive/type_refs.txt
StachuDotNet, line 1: the titles of the first 2 tests in this file are reversed (1st title corresponds to 2nd test, and vice versa)

tree-sitter-darklang/grammar.js
```diff
@@ -70,10 +71,28 @@ module.exports = grammar({
...
         field("typ", $.type_decl_def),
       ),
 
+    type_params: $ =>
+      seq(
+        field("symbol_open_angle", alias(token.immediate("<"), $.symbol)),
+        field("params", $.params),
+        field("symbol_close_angle", alias(token.immediate(">"), $.symbol)),
+      ),
+    params: $ =>
```
StachuDotNet, line 85: This name is a bit broad - maybe rename to something like `type_params_items` or inline it in the `type_params` rule?

tree-sitter-darklang/test/corpus/exhaustive/type_refs.txt
OceanOak: `all builtin type references` and` (simple) qualified type identifier` ?
maybe wrong file?

StachuDotNet: Sorry - I meant the lines 263 and 330 in this file should be swapped, I believe

OceanOak: yes, thanks!

PR #5373, opened 2024-05-19
Tree-sitter grammar: support Function Type Reference

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support function type reference
```

#5321

### DIFF
--- a/tree-sitter-darklang/grammar.js
+++ b/tree-sitter-darklang/grammar.js
@@ -811,6 +811,7 @@ module.exports = grammar({
         $.list_type_reference,
         $.tuple_type_reference,
         $.dict_type_reference,
+        $.fn_type_reference,
         $.variable_type_reference,
       ),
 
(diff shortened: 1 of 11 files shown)


tree-sitter-darklang/grammar.js
```diff
@@ -854,6 +855,24 @@ module.exports = grammar({
...
+          repeat1(
+            prec.left(
+              seq(
+                field("symbol_arrow", alias("->", $.symbol)),
+                $.type_reference,
+              ),
+            ),
+          ),
+        ),
+      ),
```
StachuDotNet, line 874: This might fit more naturally like this: 
```suggestion
    fn_type_reference: $ =>
      prec.right(
        seq(
          repeat1(
            prec.left(
              seq(
                $.type_reference,
               field("symbol_arrow", alias("->", $.symbol)),
              ),
            ),
          ),
          $.type_reference,
        ),
      ),
```

Feel free to disregard or leave for another PR, though

StachuDotNet (approved):
I'll merge as-is; if you'd like to refactor to the suggested grammar change, OK to do so in a follow-up PR. Maybe just adding a cleanup note would be good :)

PR #5377, opened 2024-05-28
Tree-sitter grammar: support Constant Declaration

OceanOak:
Changelog:

```
Tree-sitter-darklang
- Update the grammar to support Constant Declaration
```

#5321


StachuDotNet (approved):
Great! Btw, I think 'typesAndFnsAndConsts' could be renamed to 'declarations' (incl when it includes handler definitions)

PR #5379, opened 2024-05-29
Tree-sitter grammar: handle constant expression

OceanOak:
Changelog:

```
Tree-sitter grammar
- add support for constant expression
```

#5321

### DIFF
--- a/tree-sitter-darklang/grammar.js
+++ b/tree-sitter-darklang/grammar.js
@@ -357,6 +357,14 @@ module.exports = grammar({
         $.pipe_expression,
 
         $.record_update,
+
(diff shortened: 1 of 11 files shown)


tree-sitter-darklang/grammar.js
```diff
@@ -357,6 +357,14 @@ module.exports = grammar({
         $.pipe_expression,
 
         $.record_update,
+
+        $.qualified_const_name,
+      ),
+
+    qualified_const_name: $ =>
```
OceanOak, line 364: Not sure if the naming is correct here, please let me know if it should be renamed

Issue #5259, opened 2024-01-14
Component: Parsing, Pretty-Printing


StachuDotNet:
This Issue exists to collect many items, not currently in progress, that relate to Dark's parser(s), pretty-printer(s), name resolution, etc.

- [ ] ...

PR #5381, opened 2024-05-31
Add semantic tokenization tests

OceanOak:
This PR fixes semantic tokenization, and adds tests for it


StachuDotNet:
Those are great comments - I added a link in #5259 to that comment so we don't lose track.
Quick thoughts:
- typeParam point: yeah, probably good to reuse some functionality there
- Dict key thing: yeah I'd say so
- "maybe true and false shouldn't be tokenized as Keyword" - agreed. Just took a look and seems F# tokenizes true and false as `other` or `constant` (depending on which tokenization system). adding a new token type for `constant` might work well for us here. no 'default' available tokens fit super well, besides keyword, and the problem you outline is real

StachuDotNet (approved):
Great work here - shame the auto-formatter does at it does, but those tests will look tidier _soon_. We'll be thankful for these tests.

PR #5388, opened 2024-06-13
Minor cleanup and tidying

OceanOak:
No changelog

### DIFF
--- a/backend/tests/Tests/Tests.fsproj
+++ b/backend/tests/Tests/Tests.fsproj
@@ -11,6 +11,7 @@
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="../../src/Prelude/Prelude.fsproj" />
+    <ProjectReference Include="../../src/LibParser/LibParser.fsproj" />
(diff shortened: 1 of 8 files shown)


backend/tests/Tests/Tests.fsproj
```diff
@@ -11,6 +11,7 @@
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="../../src/Prelude/Prelude.fsproj" />
+    <ProjectReference Include="../../src/LibParser/LibParser.fsproj" />
```
OceanOak, line 14: `Tests.fsproj` references `TestUtils.fsproj`, which in turn references `LibParser.fsproj`. However, since `TestUtils.fsproj ` does not directly use `LibParser.fsproj`, I removed it from `TestUtils.fsproj` and added it directly to `Tests.fsproj`, where it is actually used. Was the previous way of referencing intentional? If so, please let me know, and I'll revert this change

StachuDotNet: No, you made the right move here :)

PR #5389, opened 2024-06-18
wip: include sqlite in darklang cli builds

StachuDotNet:
the artifacts here are in the ~5MB range - more than I anticipated (~1MB, based on docs I read months ago).
Need to play with compiler options more.


StachuDotNet:
superseded by #5446

PR #5390, opened 2024-06-18
wip: cache cli pm on disk

StachuDotNet:
This makes running `dark help`, or anything else, much faster after the first run - no HTTP needed as relevant package stuff is cached on-disk after the first run.

This solution is incomplete -- needs work especially to deal with reasonable cache-invalidation.

I thought I'd jump straight to using sqlite for caching here, but #5389 that turned out to be more MB than I expected, so considering options


StachuDotNet:
Outdated - decided to use sqlite for local storage of PM stuff instead of this weird cache dance.

PR #5392, opened 2024-06-18
Use the new parser when running a function from the cli

OceanOak:
No changelog


backend/src/BuiltinCliHost/Libs/Cli.fs
```diff
@@ -342,6 +342,7 @@ let fns : List<BuiltInFn> =
       deprecated = NotDeprecated }
 
 
+    // TODO: this doesn't seem to be using the new parser
     { name = fn "cliExecuteFunctionWithNewParser" 0
```
OceanOak, line 346: maybe rename `cliExecuteFunctionWithNewParser` to just `cliExecuteFunction` and remove the other `cliExecuteFunction` function?

StachuDotNet: sounds good

OceanOak:
This script was used as a test for this PR 

```
type MyType = Int64

let helloworld () : CliScript.MyType =
  let _ = Builtin.printLine ("Hello, World!")
  0L

let fizzbuzz (n: Int64) : List<String> =
  let list = PACKAGE.Darklang.Stdlib.List.range 1L n

  PACKAGE.Darklang.Stdlib.List.map list (fun i ->
    match ((i % 3L == 0L), (i % 5L == 0L)) with
    | (true, false) -> "Fizz"
    | (false, true) -> "Buzz"
    | (true, true) -> "FizzBuzz"
    | (false, false) -> PACKAGE.Darklang.Stdlib.Int64.toString i)

let wrapperFunction (n: Int64) : Int64 =
  let result = CliScript.fizzbuzz n
  let _ = PACKAGE.Darklang.Stdlib.List.map result (fun x -> Builtin.printLine x)
  0L

CliScript.helloworld ()
CliScript.wrapperFunction 15L
```

Result:
```
Hello, World!
1
2
Fizz
4
Buzz
Fizz
7
8
Fizz
Buzz
11
Fizz
13
14
FizzBuzz
```

TODO:
- [ ] Figure out a way to avoid having to add `CliScript.` when calling a newly declared function in the script

backend/src/BuiltinCliHost/Utils.fs
```diff
@@ -0,0 +1,96 @@
...
+
+open LibExecution.RuntimeTypes
+
+module CliScript =
+  type PTCliScriptModule =
+    { // These will end up in the package manager
+      types : List<PT.PackageType.PackageType>
+      constants : List<PT.PackageConstant.PackageConstant>
+      fns : List<PT.PackageFn.PackageFn>
+      submodules : PT.Packages //TODO
```
OceanOak, line 17: instead of reusing `Packages` type here maybe create a new type

packages/darklang/languageTools/parser/cliScript.dark
```diff
@@ -1,97 +1,258 @@
...
+            submodules:
+              List<LanguageTools.WrittenTypes.ModuleDeclaration.ModuleDeclaration>
+            exprs: List<LanguageTools.WrittenTypes.Expr> }
+
+        type PTCliScriptModule =
+          { types: List<LanguageTools.ProgramTypes.PackageType.PackageType>
+            constants:
+              List<LanguageTools.ProgramTypes.PackageConstant.PackageConstant>
+            fns: List<LanguageTools.ProgramTypes.PackageFn.PackageFn>
+            submodules: List<LanguageTools.ProgramTypes.Definitions>
```
OceanOak, line 21: recheck this

OceanOak:
I am not sure how to avoid adding `CliScript.` when calling a newly declared function in the script. 
Do you have any idea why we didn’t have to do this when we used libParser?

packages/darklang/languageTools/packageManager.dark
```diff
@@ -30,4 +30,14 @@ module Darklang =
           Builtin.packageManagerFindFn name
 
         let get (id: Uuid) : Stdlib.Option.Option<ProgramTypes.PackageFn.PackageFn> =
-          Builtin.packageManagerGetFn id
\ No newline at end of file
+          Builtin.packageManagerGetFn id
+
+
+      let pm () : ProgramTypes.PackageManager.PackageManager =
```
StachuDotNet, line 36: let's add a CLEANUP here for this to be a constant/value, rather than a fn, once we can have constant records

packages/darklang/languageTools/parser/functionDeclaration.dark
```diff
@@ -122,7 +122,24 @@ module Darklang =
             let returnTypeNode =
               findAndParseRequired node "return_type" TypeReference.parse
 
-            let bodyNode = findAndParseRequired node "body" Expr.parse
+            let filterAndParseBody =
```
StachuDotNet, line 125: `bodyExprsMaybe` or just `bodyExprs` might be a better name here.

backend/src/BuiltinCliHost/Libs/Cli.fs
```diff
@@ -101,12 +97,23 @@ let execute
...
+
+    let constants =
+      List.concat
+        [ mod'.constants |> List.map PT2RT.PackageConstant.toRT
+          mod'.submodules.constants |> List.map PT2RT.PackageConstant.toRT ]
+
+    let fns =
+      List.concat
+        [ mod'.fns |> List.map PT2RT.PackageFn.toRT
+          mod'.submodules.fns |> List.map PT2RT.PackageFn.toRT ]
```
StachuDotNet, line 113: Will these still work if there are more than 1-2 layers of modules? I think there should probably be some kinda recursion here instead of this

backend/src/BuiltinCliHost/Utils.fs
```diff
@@ -0,0 +1,147 @@
...
+open LibExecution.RuntimeTypes
+
+module CliScript =
+  type Definitions =
+    { types : List<PT.PackageType.PackageType>
+      constants : List<PT.PackageConstant.PackageConstant>
+      fns : List<PT.PackageFn.PackageFn> }
+
+  type PTCliScriptModule =
+    { // These will end up in the package manager
```
StachuDotNet, line 20: I'm not sure what this comment means

packages/darklang/languageTools/parser/canvas.dark
```diff
@@ -0,0 +1,171 @@
+module Darklang =
+  module LanguageTools =
+    module Parser =
+      module Canvas =
```
StachuDotNet, line 4: I forget where we last stood here - are these used by CliScript parsing? If so, should there be a TODO to stop re-using this "canvas parsing" for cli-script parsing?

StachuDotNet: oh or maybe this is just separately updated but no longer reused - if so, ignore.

OceanOak: No, this whole module isn't used anywhere. I just kept it to use when we parse canvases with the new parser

StachuDotNet: great, sounds good

backend/src/BuiltinCliHost/Utils.fs
OceanOak: me neither, you've added it in https://github.com/darklang/dark/blob/b6a0b2f215ecb116f747f99cb90e634e833c8302/backend/src/LibParser/Canvas.fs#L40-L41

I kept it here just in case it's needed, since Canvas.fs will be deleted at some point

StachuDotNet: whoops. let's remove in both places :)

backend/src/BuiltinCliHost/Utils.fs
```diff
@@ -0,0 +1,147 @@
...
+      let types =
+        match Map.tryFind "types" fields with
+        | Some(DList(_, types)) ->
+          List.map (fun t -> t |> PT2DT.PackageType.fromDT) types
+        | _ -> []
+      let constants =
+        match Map.tryFind "constants" fields with
+        | Some(DList(_, constants)) ->
+          List.map (fun c -> c |> PT2DT.PackageConstant.fromDT) constants
+        | _ -> []
```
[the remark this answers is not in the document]
OceanOak: I think you're right

backend/src/BuiltinCliHost/Libs/Cli.fs
OceanOak: I Think it works, here is a test:
```
module Test =
  module Nested =
    let test1 () : Int64 =
      let _ = Builtin.printLine ("Testing Modules1")
      0L
    module Nested2 =
      let test2 () : Int64 =
        let _ = Builtin.printLine ("Testing Modules2")
        0L
      module Nested3 =
        module Nested4 =
          let test3 () : Int64 =
            let _ = Builtin.printLine ("Testing Modules3")
            0L
            
            
CliScript.Test.Nested.test1 ()
CliScript.Test.Nested.Nested2.test2 ()
CliScript.Test.Nested.Nested2.Nested3.Nested4.test3 ()
```

Result:
Screenshot 2024-07-08 at 18 39 27

StachuDotNet (changes requested):
> I am not sure how to avoid adding CliScript. when calling a newly declared function in the script.
Do you have any idea why we didn’t have to do this when we used libParser?

I think that this is the only thing remaining to change before merging.
I suspect that this is a matter of the `owner` not trickling to all of the places it needs to. 

Take a look at `LibParser/Canvas.fs`, in `toPT`. There are a lot of places there of `(m.owner :: m.name)` being used to front-load the "currentModules` that make it to name-resolution. Unless the name resolver knows we're "currently in CliScript," it'll be unable to resolve the name.

backend/src/BuiltinCliHost/Utils.fs
```diff
@@ -0,0 +1,173 @@
...
+    { types : List<PT.PackageType.PackageType>
+      constants : List<PT.PackageConstant.PackageConstant>
+      fns : List<PT.PackageFn.PackageFn> }
+
+  type PTCliScriptModule =
+    { types : List<PT.PackageType.PackageType>
+      constants : List<PT.PackageConstant.PackageConstant>
+      fns : List<PT.PackageFn.PackageFn>
+      submodules : Definitions
+      exprs : List<PT.Expr> }
```
StachuDotNet, line 24: Maybe this should live in PT eventually, but good here for now

PR #5393, opened 2024-06-18
Remove LibParser usage from Cron.Tests and NewParser

OceanOak:
This PR marks the beginning of our transition from LibParser to the darklang-tree-sitter parser.

### DIFF
--- a/backend/src/LibExecution/ProgramTypesToDarkTypes.fs
+++ b/backend/src/LibExecution/ProgramTypesToDarkTypes.fs
@@ -8,6 +8,7 @@ module PT = ProgramTypes
 module VT = ValueType
 module NRE = LibExecution.NameResolutionError
 module D = LibExecution.DvalDecoder
+module C2DT = LibExecution.CommonToDarkTypes
 
 
 // This isn't in PT but I'm not sure where else to put it...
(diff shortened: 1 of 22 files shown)


backend/src/LibExecution/ProgramTypesToDarkTypes.fs
```diff
@@ -1500,3 +1500,25 @@ module PackageFn =
         deprecated =
           fields |> D.field "deprecated" |> Deprecation.fromDT FQFnName.fromDT }
     | _ -> Exception.raiseInternal "Invalid PackageFn" []
+
+
+module Result =
+  let toDT
+    (nameValueType : KnownType)
```
StachuDotNet, line 1507: maybe `okType`?

StachuDotNet: then the NameResolution fns could wrap around the Result modules' fns

StachuDotNet (changes requested):
Looking good! Just the 3 comments in that one thread, then this is good to go.

backend/tests/Tests/Cron.Tests.fs
```diff
@@ -48,9 +48,8 @@ let testCronSanity =
   testTask "cron sanity" {
     let! canvasID = initializeTestCanvas "cron-sanity"
 
-    let! e = p "5 + 3"
-    let resultExpr =
-      C2DT.Result.fromDT PT2DT.Expr.fromDT e PT2DT.TypeReference.fromDT
+    let! e = p "5 ++ 3"
+    let resultExpr = C2DT.Result.fromDT PT2DT.Expr.fromDT e RT.RuntimeError.fromDT
```
OceanOak, line 52: could you please check this?
I am not sure if `RT.RuntimeError.fromDT` is the right function to send here

backend/tests/Tests/Cron.Tests.fs
```diff
@@ -49,8 +49,7 @@ let testCronSanity =
     let! canvasID = initializeTestCanvas "cron-sanity"
 
     let! e = p "5 + 3"
-    let resultExpr =
-      C2DT.Result.fromDT PT2DT.Expr.fromDT e PT2DT.TypeReference.fromDT
+    let resultExpr = C2DT.Result.fromDT PT2DT.Expr.fromDT e RT.RuntimeError.fromDT
```
OceanOak, line 52: could you check this please?
I am not sure if `RT.RuntimeError.fromDT` is the right error FromDT function to send here

StachuDotNet: considering we're only expecting Ok cases here, even passing in `identity` would be OK here, I think. (just leave it as the Dval that we receive)

Really, the Result that comes back here is only because the `parsePTExpr` returns a result. IN the context of these Cron tests, we really only care to test cases where we properly pass an expr. So I think
- the `p` functino can be changed back to return just a PT.Expr
- adjust accordingly
- , noting that we shouldn't really need `unwrapExecutionResult` in this file - if the fn call fails, or if we can't unwrap to get a PT.Expr, then we should just fail with such a message -- no indiviual tests here should have to deal with that

OceanOak: Thanks!

StachuDotNet (approved):
nice!

PR #5395, opened 2024-06-21
wip: sketching new interpreter

StachuDotNet:
AST -> Bytecode.

In such an early state that it's probably not worth peeking at


StachuDotNet:
further along in another PR

PR #5400, opened 2024-07-09
Tree-sitter-darklang grammar fixes

OceanOak:
No changelog

This PR adds some tree-sitter-darklang grammar tests, and fixes the following issues: 
Issue 1: 
```
[ ErrorSegment.ErrorSegment.String
	"RTETODO typeChecker.toSegments" ] 
```

This was parsed as a list of two elements (an enum and a string literal), rather than a list of one element (an enum with one field)

Issue 2:
```
let x =
  1L
  |> Stdlib.Option.Option.Some
x
```
The body (`x`) of the let_expression was parsed as part of the pipe, as if it was a field of the enum.

### DIFF
--- a/tree-sitter-darklang/test/corpus/exhaustive/exprs/let_expr.txt
+++ b/tree-sitter-darklang/test/corpus/exhaustive/exprs/let_expr.txt
@@ -32,11 +32,9 @@ x
   (expression
     (let_expression
       (keyword) (variable_identifier) (symbol)
-      (indented_block
-        (indent)
-        (expression (simple_expression (int64_literal (digits (positive_digits)) (symbol))))
-        (dedent)
-      )
+      (indent)
+      (expression (simple_expression (int64_literal (digits (positive_digits)) (symbol))))
+      (dedent)
       (expression (simple_expression (variable_identifier)))
     )
   )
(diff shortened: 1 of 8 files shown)


tree-sitter-darklang/test/corpus/exhaustive/exprs/let_expr.txt
```diff
@@ -0,0 +1,170 @@
...
+  (expression
+    (let_expression
+      (keyword) (variable_identifier) (symbol)
+      (indent)
+      (expression (simple_expression (int64_literal (digits (positive_digits)) (symbol))))
+      (dedent)
+      (expression (simple_expression (variable_identifier)))
+    )
+  )
+)
```
StachuDotNet, line 41: This looks wrong to me. I'd expect this I think:

```suggestion
==================
simple let_expression -with indent
==================

let x =
  1L
x

---

(source_file
  (expression
    (let_expression
      (keyword) (variable_identifier) (symbol)
      (indent)
      (expression (simple_expression (int64_literal (digits (positive_digits)) (symbol))))
      (dedent)
    )
    (expression (simple_expression (variable_identifier)))
  )
)
```

OceanOak: Hmm, why? 
`x` is the body of the `let_expression` so it should be included within it, no?

tree-sitter-darklang/test/corpus/exhaustive/exprs/let_expr.txt
```diff
@@ -0,0 +1,170 @@
...
+==================
+let expression with multiple pipes -indented
+==================
+
+let wtTest =
+  testSource
+  |> LanguageTools.Parser.ParserTest.initialParse
+  |> Test.parseTest
+  |> Builtin.unwrap
+wtTest
```
StachuDotNet, line 124: same here - the last line should be outside of the `let_expression`, I believe?

OceanOak: I don't think so, unless I'm misunderstanding something.
here is our grammar before the changes : 
The field `body`(which I don't think is a good name for it) is included in the `let_expression` because we require an expression that follows `let something = something`

```
let_expression: $ =>
      seq(
        field("keyword_let", alias("let", $.keyword)),
        field("identifier", $.variable_identifier),
        field("symbol_equals", alias("=", $.symbol)),
        field("expr", $.expression),
        "\n",
        field("body", $.expression),
      ),
      ```

tree-sitter-darklang/test/corpus/exhaustive/exprs/let_expr.txt
StachuDotNet: you're right, I was confusing myself.

tree-sitter-darklang/test/corpus/exhaustive/exprs/let_expr.txt
StachuDotNet: you're right, I was confusing myself.

PR #5403, opened 2024-07-15
Replace LibParser in SqlCompiler.Tests

OceanOak:
No changelog

### DIFF
--- a/backend/tests/Tests/SqlCompiler.Tests.fs
+++ b/backend/tests/Tests/SqlCompiler.Tests.fs
@@ -9,32 +9,37 @@ open FSharp.Control.Tasks
 
 module PT = LibExecution.ProgramTypes
 open LibExecution.RuntimeTypes
-module NR = LibParser.NameResolver
 
 module C = LibCloud.SqlCompiler
 module S = TestUtils.RTShortcuts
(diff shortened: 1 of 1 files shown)


backend/tests/Tests/SqlCompiler.Tests.fs
```diff
@@ -9,32 +9,37 @@ open FSharp.Control.Tasks
...
-let p (code : string) : Task<Expr> =
-  LibParser.Parser.parseRTExpr
-    builtins
-    pmPT
-    NR.OnMissing.ThrowError
-    "sqlcompiler.tests.fs"
-    code
-  |> Ply.toTask
-
 let parse (fns : List<PT.PackageFn.PackageFn>) (code : string) : Task<Expr> =
```
OceanOak, line 22: Previously, we had two functions `p` and `parse`, the only difference between them was that one of them takes a list of package functions as an arg (for user-defined functions)
I combined the two and am now passing an empty list in the cases where we don't need the list of fns. 

Is it better to keep them separate? and should we also add types and constants as args in case we use them in future tests?

StachuDotNet: This is good. We can add types+consts if/when they come up.

PR #5406, opened 2024-07-20
Big Runtime Rewrite

### LINKED ISSUE #5222: Bug: List.map Causes Stack Overflow Error on Large Lists

StachuDotNet:
This is a big PR that rewrites the Runtime and surrounding bits (esp. `RuntimeTypes`, the `Interpreter`, and `PT2RT`) to not stack-overflow so trivially (#5222). It should also lay the groundwork for improving overall runtime performance.

While working towards that goal, I accidentally handled many other things that have been bugging me for a while.

Here's what's been done towards the stated goal:
- `RuntimeTypes` no longer has a recursive `Expr` concept, but rather has Instructions
- the Interpreter thus is no longer tree-based, but is now a register-based virtual machine bytecode interpreter
  - when calling upon a Package Function or a Lambda, we append a new "call frame" to the virtual machine, to be processed in another iteration of the main interpreter loop
  - so: it's no longer trivial to reach stack overflows (the actual goal) 
- `PT2RT` is now notably more complicated, to map `PT.Exprs` (and fns, etc) to `RT.Instructions`
- I think that the interpreter is generally easier to understand since these changes
- a bunch of lower-level Interpreter and `PT2RT` tests
- temp: `SqlCompiler` and `DB.query` have been commented out - I couldn't figure out a way to make them work in this new world yet, so we'll have to circle back to that

Here's the out-of-scope work that got done:
- `RuntimeErrors` (`RTE`s) are now much better, and always structured -- `OldStringErrorTODO` is **_gone_**
- non-recursive runtime type-checking is now complete[1]
- partial-application works (!) for "applys" to both fns and lambdas
- we reduced a few `Builtin`s in favor of direct impl in package functions
- we no longer build our tree-sitter parser to WASM -- we haven't been using that in a bit. the code is still there for when we need it, just commented
- tidied a _lot_ of test `.dark` files

---

[1] so, this closes #5222, closes #5130, closes #5071

---

There's some follow-up work to be done _after_ this PR is merged:
- [ ] fix AND/OR short-circuiting (maybe make them work via Jump instructions?)
- [ ] bring back `SqlCompiler` and `DB.query`
- [ ] continue to refine handling+reporting of RTEs
  - [ ] especially dev experience (understanding what went wrong)
- [ ] more serious performance-tuning
- [ ] more work around async and concurrent function calls
- [ ] improve our handling of when `incorrectArgs` is hit (this should be _far_ less likely now, but _could_ happen (somehow?) and we don't handle it great)
- [ ] when RTEs occur, point to specific Exprs in the error (involves some fun PT2RT and Interpreter work -- shouldn't be that bad, and will have a lot of benefit)
- [ ] replace constants with values evaluated at dev-time
- [ ] references to Secrets and DBs resolved at dev-time
- [ ] make RTE extensible


StachuDotNet:
thoughts on implemented `match` that I wanted to record in case my current plan doesn't end up working out: https://gist.github.com/StachuDotNet/5a8923721b255ad2a91c1c6dcc3e9b8e

backend/src/LibExecution/Interpreter.fs
```diff
@@ -185,8 +185,14 @@ let execute (exeState : ExecutionState) (vm : VMState) : Ply<Dval> =
 
         | Lambda(parentContext, lambdaID) ->
           let lambda =
-            Map.findUnsafe (parentContext, lambdaID) vm.lambdaInstrCache
+            (match Map.tryFind (parentContext, lambdaID) vm.lambdaInstrCache with
+             | Some l -> l
+             | None ->
+               match Map.tryFind (Source, lambdaID) vm.lambdaInstrCache with
+               | Some l -> l
+               | None -> raiseRTE (RTE.VariableNotFound "lambda not found")) // TODO better error
```
StachuDotNet, line 193: Hm, why do we need this fallback case? (`Source`) - When is that code hit?

backend/src/LibExecution/Interpreter.fs
```diff
@@ -425,7 +431,10 @@ let execute (exeState : ExecutionState) (vm : VMState) : Ply<Dval> =
 
         | CreateLambda(lambdaReg, impl) ->
           vm.lambdaInstrCache <-
-            Map.add (currentFrame.context, impl.exprId) impl vm.lambdaInstrCache
+            vm.lambdaInstrCache
+            |> Map.add (currentFrame.context, impl.exprId) impl
+            |> Map.add (Source, impl.exprId) impl
```
StachuDotNet, line 436: I suppose this is a similar question - why do we need to add two entries?

backend/src/LibExecution/Interpreter.fs
StachuDotNet: I suppose this is for the cases that we were trying to fix for -- but will the solution work if there are multiple nested lambdas? Or if a lambda is returned by a fn and used later on?

OceanOak: This is hit when running functions that take a lambda as an argument (e.g. List.takeWhile, List.sortByComparator, List.sortBy, List.member, List.map, etc.)

The lambda couldn't be found in the `currentFrame.context` which is a `PackageFn`, instead it is found in the `Source`, I wasn't sure if it was the best solution.

As for `will the solution work if there are multiple nested lambdas` not sure yet, as I am encountering other errors in these cases that don't seem to be related to this issue

backend/src/LibExecution/Interpreter.fs
OceanOak: the lambda is created in the `Source` context, but we try to apply it in the `PackageFn` context 
I've added this to make sure that we can find the lambda later regardless of the context in which it's being applied. Again not sure if it is the right thing to do, please let me know if you think there is a better way to do it

PR #5408, opened 2024-07-29
Tree-sitter-darklang: Fix let_expression to handle lp_tuple correctly

OceanOak:
No changelog

### DIFF
--- a/packages/darklang/languageTools/parser/expr.dark
+++ b/packages/darklang/languageTools/parser/expr.dark
@@ -454,14 +454,14 @@ module Darklang =
           (node: ParsedNode)
           : Stdlib.Result.Result<WrittenTypes.Expr, WrittenTypes.Unparseable> =
           let keywordLetNode = findField node "keyword_let"
-          let identifierNode = findField node "identifier"
+          let patternNode = findField node "pattern"
           let symbolEqualsNode = findField node "symbol_equals"
           let expr = findAndParseRequired node "expr" Expr.parse
           let body = findAndParseRequired node "body" Expr.parse
 
-          match keywordLetNode, identifierNode, symbolEqualsNode, expr, body with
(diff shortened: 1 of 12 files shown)


tree-sitter-darklang/grammar.js
```diff
@@ -924,6 +924,7 @@ module.exports = grammar({
     //
     // Tuple
     tuple_type_reference: $ =>
+      // CLEANUP: make the parens optional
```
StachuDotNet, line 927: We can leave this CLEANUP in but I don't hate the idea of requiring `Json.parse<(Int * Bool)>`, not allowing `Json.parse<Int * Bool>` - the former feels clearer. What do you think?

StachuDotNet: actually, "tolerable parser" so nvm - the CLEANUP is good and we should allow the latter too. But maybe I'd prefer to _prettify_ the TypeRef with the parens :)

packages/darklang/languageTools/parser/expr.dark
```diff
@@ -399,15 +399,50 @@ module Darklang =
...
-            (WrittenTypes.Expr.ELet(
-              node.range,
-              WrittenTypes.LetPattern.LPVariable(identifier.range, identifier.text),
-              expr,
-              body,
-              keywordLet.range,
-              symbolEquals.range
-            ))
-            |> Stdlib.Result.Result.Ok
+            match identifier.children with
```
StachuDotNet, line 402: Funny we've had the `let_pattern` in the grammar for a while, not used by `let`s!
Given the fact that let patterns are used by both `ELet` and `ELambda`, this should probably be extracted into a `LetPattern.parse` (or another name - `parseLetPattern` is probably OK based on your judgment)

OceanOak: we had it, I just forgot about it :woman_facepalming: 
Fixed

PR #5410, opened 2024-07-31
Misc tree-sitter grammar fixes

OceanOak:
Misc tree-sitter-darklang grammar fixes
- ignore inline comment (//)
- use variable_identifier instead of expression for dictionaries
- add support for indented arguments in apply

### DIFF
--- a/tree-sitter-darklang/grammar.js
+++ b/tree-sitter-darklang/grammar.js
@@ -128,7 +128,7 @@ module.exports = grammar({
 
     const_dict_literal: $ => dict_literal_base($, $.const_dict_content),
     const_dict_content: $ => dict_content_base($, $.const_dict_pair),
-    const_dict_pair: $ => dict_pair_base($, $.expression, $.consts),
+    const_dict_pair: $ => dict_pair_base($, $.consts),
 
     const_enum_literal: $ => enum_literal_base($, $.const_enum_fields),
     const_enum_fields: $ => enum_fields_base($, $.consts),
@@ -633,7 +633,7 @@ module.exports = grammar({
     // Dict
     dict_literal: $ => dict_literal_base($, $.dict_content),
     dict_content: $ => dict_content_base($, $.dict_pair),
-    dict_pair: $ => dict_pair_base($, $.expression, $.expression),
+    dict_pair: $ => dict_pair_base($, $.expression),
 
     //
     // Tuples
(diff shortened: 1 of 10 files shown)


tree-sitter-darklang/grammar.js
```diff
@@ -28,6 +28,7 @@ module.exports = grammar({
   name: "darklang",
 
   externals: $ => [$.indent, $.dedent],
+  extras: $ => [/[\s\uFEFF\u2060\u200B\u00A0]/, $._inline_comment],
```
StachuDotNet, line 31: What's this?
```suggestion
  extras: $ => [
    /[\s\uFEFF\u2060\u200B\u00A0]/, // some commentary - not sure what this is
    $._inline_comment
  ],
```

tree-sitter-darklang/grammar.js
```diff
@@ -1124,9 +1137,9 @@ function dict_content_base($, dict_pair) {
     repeat(seq(field("dict_separator", alias(";", $.symbol)), dict_pair)),
   );
 }
-function dict_pair_base($, key, value) {
+function dict_pair_base($, value) {
   return seq(
-    field("key", key),
+    field("key", choice($.variable_identifier, $.double_backtick_identifier)),
```
StachuDotNet, line 1142:
```suggestion
    // CLEANUP this should allow for any string (i.e. `This`), without any backticks. may require alternative syntax.
    field("key", choice($.variable_identifier, $.double_backtick_identifier)),
```

StachuDotNet (changes requested):
Two small cleanups, then OK to merge

tree-sitter-darklang/grammar.js
OceanOak: I'll add the commentary.

The default value of extras includes whitespace; when adding comments to it, I believe we should handle whitespace explicitly. Maybe we can just have it as `[/\s/, $._inline_comment]` the others represent unicodes for zero-width spaces and non-breaking spaces, many languages I checked includes them and some don't, should we keep them?

StachuDotNet: I'm not sure - you'd probably be the better judge on this question. :)

OceanOak: The default value in tree-sitter is `[/\s/]`, so I'll just keep it simple. if we notice things breaking we can add those back

PR #5411, opened 2024-08-06
Tree-sitter-darklang: misc grammar fixes

OceanOak:
This PR:
- Fixes mp_list_cons and dictionaries
- Adds support for multiline string
- Allows having type and function declarations before module declarations
- Moves enums from simple_expression to expression


StachuDotNet:
nice work!

PR #5413, opened 2024-08-12
Refactor the grammar to support only tuple-style Enum literals

OceanOak:
No changelog

### DIFF
--- a/packages/darklang/languageTools/parser/core.dark
+++ b/packages/darklang/languageTools/parser/core.dark
@@ -288,7 +288,12 @@ module Darklang =
           (findNodeByFieldName node "enum_fields")
           |> Stdlib.Option.map (fun enumFieldsNode ->
             enumFieldsNode.children
-            |> Stdlib.List.map (fun child -> parseFn child)
+            |> Stdlib.List.chunkBySize 2L
+            |> Builtin.unwrap
+            |> Stdlib.List.map (fun contentSymbolPair ->
+              match contentSymbolPair with
+              | [ contentNode; symbol ] -> parseFn contentNode
+              | [ contentNode ] -> parseFn contentNode)
             |> Stdlib.Result.collect)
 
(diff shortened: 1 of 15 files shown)


OceanOak:
We can probably move enums back to `simple_expressions`

packages/darklang/languageTools/parser/core.dark
```diff
@@ -288,7 +288,12 @@ module Darklang =
           (findNodeByFieldName node "enum_fields")
           |> Stdlib.Option.map (fun enumFieldsNode ->
             enumFieldsNode.children
-            |> Stdlib.List.map (fun child -> parseFn child)
+            |> Stdlib.List.chunkBySize 2L
+            |> Builtin.unwrap
+            |> Stdlib.List.map (fun contentSymbolPair ->
+              match contentSymbolPair with
+              | [ contentNode; symbol ] -> parseFn contentNode
+              | [ contentNode ] -> parseFn contentNode)
```
StachuDotNet, line 296: Hm I'm expecting that this is dealing with: `(field1, field2)` in `MyCase(field1, field2)` - it seems to me that those tokens would come in as `["(", "field1", ",", "field2", ")"]` - but that doesn't seem to be the case. What am I missing?

StachuDotNet: (this is working, so merging, but curious what I'm missing here)

OceanOak: it is only dealing with `field1, field2` because the parentheses are part of `enum_literal_base` not `enum_fields_base`

here is `enum_fields_base`

```
seq(
      fields,
      repeat(
        prec.right(seq(field("symbol_comma", alias(",", $.symbol)), fields)),
      ),
    ),
```

This was just to keep `enum_fields` optional, and not lose the parens when dealing with an enum with empty fields. I can move the parens to `enum_fields` if it would make the grammar clearer

PR #5414, opened 2024-08-20
Misc parser fixes

OceanOak:
No changelog


tree-sitter-darklang/grammar.js
```diff
@@ -523,7 +529,7 @@ module.exports = grammar({
       ),
 
     char_or_string_escape_sequence: _ =>
-      token.immediate(seq("\\", /(\"|\\|\/|b|f|n|r|t|u)/)),
```
OceanOak, line 526: revisit this change. this was added because it was needed in https://github.com/darklang/dark/blob/b8d7b9b4514e9a6da5a59ec6e15d806749c6f33f/backend/testfiles/execution/stdlib/alt-json.dark#L81

PR #5417, opened 2024-08-26
Misc parser fixes

OceanOak:
No changelog

### DIFF
--- a/packages/darklang/languageTools/writtenTypesToProgramTypes.dark
+++ b/packages/darklang/languageTools/writtenTypesToProgramTypes.dark
@@ -503,7 +503,6 @@ module Darklang =
             match constantOrFn with
             | Const c -> ProgramTypes.Expr.EConstant(gid (), c)
             | Fn f -> ProgramTypes.Expr.EFnName(gid (), f)
-            | _ -> ProgramTypes.Expr.EVariable (gid ()) id.constantOrFn.name
 
           | EFieldAccess(_, expr, (_, fieldName), _) ->
             ProgramTypes.Expr.EFieldAccess(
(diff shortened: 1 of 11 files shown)


packages/darklang/languageTools/writtenTypesToProgramTypes.dark
```diff
@@ -472,11 +496,14 @@ module Darklang =
...
-              Identifiers.QualifiedConst.toPT onMissing pm owner currentModule c
+          | EConstantOrFn(_, id) ->
+            let constantOrFn =
+              Identifiers.QualifiedConstOrFn.toPT onMissing pm owner currentModule id
 
-            ProgramTypes.Expr.EConstant(gid (), constant)
+            match constantOrFn with
+            | Const c -> ProgramTypes.Expr.EConstant(gid (), c)
+            | Fn f -> ProgramTypes.Expr.EFnName(gid (), f)
+            | _ -> ProgramTypes.Expr.EVariable (gid ()) id.constantOrFn.name
```
OceanOak, line 506: Am I wrong to think that we wouldn't need this case? It will fail before getting here, right?

```
| _ -> ProgramTypes.Expr.EVariable (gid ()) id.constantOrFn.name
```

StachuDotNet: Yeah, I don't think you need the final branch here. `match constantOrFn` only has the two cases, which you've covered, so I don't think anything else should happen?

OceanOak: I believe so, because we are returning the name resolution error of isConstant in case of an error
https://github.com/OceanOak/dark/blob/8786f2e8ab8cf215cee6e3bf4693ceb32568e544/packages/darklang/languageTools/writtenTypesToProgramTypes.dark#L92

PR #5418, opened 2024-09-02
More parser fixes

OceanOak:
No changelog

Misc parser refactors and fixes

### DIFF
--- a/backend/src/LibExecution/ProgramTypesToDarkTypes.fs
+++ b/backend/src/LibExecution/ProgramTypesToDarkTypes.fs
@@ -921,6 +921,9 @@ module Expr =
             Exception.raiseInternal "Invalid record update" [ "update", update ])
       PT.ERecordUpdate(uint64 id, fromDT record, updates)
 
+    | DEnum(_, _, [], "EConstant", [ DInt64 id; name ]) ->
+      PT.EConstant(uint64 id, NameResolution.fromDT FQConstantName.fromDT name)
+
     | e -> Exception.raiseInternal "Invalid Expr" [ "e", e ]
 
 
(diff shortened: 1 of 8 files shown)


backend/src/LibExecution/ProgramTypesToDarkTypes.fs
```diff
@@ -987,6 +990,7 @@ module Const =
     | DEnum(_, _, [], "CBool", [ DBool b ]) -> PT.Const.CBool b
     | DEnum(_, _, [], "CString", [ DString s ]) -> PT.Const.CString s
     | DEnum(_, _, [], "CChar", [ DChar c ]) -> PT.Const.CChar c
```
OceanOak, line 992: Should we remove this one?

packages/darklang/languageTools/writtenTypesToProgramTypes.dark
```diff
@@ -501,8 +501,10 @@ module Darklang =
               Identifiers.QualifiedConstOrFn.toPT onMissing pm owner currentModule id
 
             match constantOrFn with
-            | Const c -> ProgramTypes.Expr.EConstant(gid (), c)
-            | Fn f -> ProgramTypes.Expr.EFnName(gid (), f)
+            | Ok(Const c) -> ProgramTypes.Expr.EConstant(gid (), c)
+            | Ok(Fn f) -> ProgramTypes.Expr.EFnName(gid (), f)
+            | Error _ ->
+              ProgramTypes.Expr.EVariable (gid ()) constantOrFn.constantOrFn.name
```
OceanOak, line 507: Added this because we shouldn't allow it to fail during the first pass of parsing. Do you think there is a better way to do it?

backend/src/LibExecution/ProgramTypesToDarkTypes.fs
```diff
@@ -987,6 +990,7 @@ module Const =
...
+    | DEnum(_, _, [], "CChar", [ DString c ]) -> PT.Const.CChar c
```
OceanOak, line 993: Should we remove the `| DEnum(_, _, [], "CChar", [ DChar c ]) -> PT.Const.CChar c` case?

StachuDotNet: We should remove the `| DEnum(_, _, [], "CChar", [ DString c ]) -> PT.Const.CChar c` case, right?

packages/darklang/languageTools/writtenTypesToProgramTypes.dark
StachuDotNet: The alternative, I think, is having this bit return a Result instead, and handling that all the way up

backend/src/LibExecution/ProgramTypesToDarkTypes.fs
OceanOak: the `| DEnum(_, _, [], "CChar", [ DString c ])->...` is the one that works
in ProgramTypes we have `| CChar of String`

StachuDotNet: ah yes you're right

PR #5419, opened 2024-09-23
Tidy up the grammar, Add support to db reference, and small fixes 

OceanOak:
No changelog

This PR: 
- Tidies up the tree-sitter Darklang grammar
- Adds support for DB reference (e.g. `Stdlib.DB.set Test { x = "test" } "key" TestDB`)
- Moves `infix_operations` from `expressions` to `simple_expressions` 
- Improves the handling of Record and List literals:
      **Before:** in a Record literal, variable names had to be wrapped in parentheses to avoid being parsed incorrectly (as an `apply`) . Example:
    ```
      Person
        { height = (height)
          name = (name) }
   ```
       
     **Now:** Parentheses are no longer necessary around variable names
       
  ```
    Person
      { height = height
        name = name}
   ```


StachuDotNet (approved):
Really nice!

PR #5422, opened 2024-10-07
Adding LSP Features: Highlight unresolved references, add a basic onHover functionality + Misc fixes

OceanOak:
No changelog

 This PR:
 - Gets the LSP woking again
 - Reimplements `dangerLanguageServerProtocolReadNextMessage` in Darklang
 - Better Semantic Tokenization 
 - Adds a very basic `onHover` functionality
 - `onSave` resend or clear diagnostics
 -  Implements sending diagnostics for unresolved references
 - Updates WT2PT to return unresolved names alongside PT


packages/darklang/languageTools/lsp-server/handleIncomingMessage.dark
```diff
@@ -41,6 +41,42 @@ module Darklang =
...
+      //   LanguageServerProtocol.Diagnostic.Diagnostic
+      //     { range = convertParserRangeToLspRange range
+      //       severity = Stdlib.Option.Option.Some severity
+      //       code = Stdlib.Option.Option.None
+      //       codeDescription = Stdlib.Option.Option.None
+      //       source = Stdlib.Option.Option.None
+      //       message = message
+      //       tags = Stdlib.Option.Option.None
+      //       relatedInformation = Stdlib.Option.Option.None
+      //       data = Stdlib.Option.Option.None }
```
StachuDotNet, line 77: I'd say it's useful. as follow-ups:
- move severity to first param
- add specific fns like `createErrorDiagnostic` that just apply the severity (you might need to make the new fns temporarily take in all 3 params, but once interpreter work is merged we can tidy)

packages/darklang/languageTools/lsp-server/docSync.dark
```diff
@@ -31,31 +31,46 @@ module Darklang =
...
+          let sourceFile =
+            match parsed with
+            | Ok parsedFile ->
+              match parsedFile with
+              | SourceFile sf -> sf
+              | _ -> log "Invalid source file"
+            | Error err -> log err
+
+          let (_ptCliScriptModule, unresolvedNames) =
+            Parser.CliScript.parse
```
OceanOak, line 52: Is it ok to reuse `CliScript.parse` here?

packages/darklang/languageTools/writtenTypesToProgramTypes.dark
```diff
@@ -1416,5 +1416,7 @@ module Darklang =
         : (ProgramTypes.SourceFile.SourceFile *
```
OceanOak, line 1416: why do we have `parsedFileAsSourceFile` as a separate function, and not just as a part of the `sourceFile` function?

packages/darklang/languageTools/writtenTypesToProgramTypes.dark
```diff
@@ -1416,5 +1416,7 @@ module Darklang =
...
           List<Parser.Range * RuntimeErrors.NameResolution.Error>)
         =
+        let pm = LanguageTools.PackageManager.pm ()
```
OceanOak, line 1419: Should pm be sent as a parameter?

packages/darklang/languageTools/parser/cliScript.dark
```diff
@@ -146,10 +158,12 @@ module Darklang =
           (pm: ProgramTypes.PackageManager.PackageManager)
           (filename: String)
```
OceanOak, line 159: Why do we have both `scriptName` and `fileName`? is file name the full path?

packages/darklang/languageTools/lsp-server/docSync.dark
StachuDotNet: Yes - I think that's probably ideal, for now - just focusing on .dark "scripts." Once the LS has a virtual document system set up, where we can be "editing the contents of X package" or something else, we can adjust here accordingly.

packages/darklang/languageTools/writtenTypesToProgramTypes.dark
StachuDotNet: Eventually, yes. I suspect we had just punted that, as it'll require us to trickle the `pm` object in a bunch of places.

packages/darklang/languageTools/writtenTypesToProgramTypes.dark
StachuDotNet: not sure :) can probably tidy.

packages/darklang/languageTools/parser/cliScript.dark
StachuDotNet: Hm, I'm not sure. It doesn't seem like `fileName` is even used?

OceanOak: yes, it isn't used

packages/darklang/languageServerProtocol/language/onHover.dark
StachuDotNet, line 1: Please check out the common.dark in the directory above.
All of the LSP types should already be _somewhere_ in the `languageServerProtocol` directory -- some still in commented typescript, some 'ported' to Dark.
When adding any new types, please make sure to remove the commented typescript source, so we don't end up adding it again later.
Also, when adding those new types, check to see where it's used to see if it should stay in common.dark -- `MarkupKind` for example is used in Completions, OnHover, and SignatureHelp handshakes, so it should probably 'stay' in that file.

packages/darklang/languageTools/lsp-server/docSync.dark
```diff
@@ -31,17 +31,46 @@ module Darklang =
                     ) }
             )
 
+        let parseAndreportDiagnostics
```
StachuDotNet, line 34: picky -- `r` -> `R`

packages/darklang/languageTools/parser/cliScript.dark
StachuDotNet: probably cool to remove it, then, unless I'm missing something

packages/darklang/languageTools/writtenTypes.dark
```diff
@@ -351,11 +351,7 @@ module Darklang =
           fields: List<Expr> *
           /// between the typeName and the caseName
           symbolDot: Range
-        | EPipeFnCall of
-          Range *
-          fnName: Name *
-          typeArgs: List<TypeReference.TypeReference> *
-          args: List<Expr>
+        | EPipeFnCall of Range * fnName: QualifiedFnIdentifier * args: List<Expr>
```
StachuDotNet, line 354: :heart:

StachuDotNet (changes requested):
This is looking good. A few tiny things to fix, but otherwise I think this is OK to merge.

I'll be honest in that I haven't read WT2PT changes in full detail -- maybe once my PR is merged, though, I should sit down and do a full pass-through of the file to provide feedback+ideas. I think it's generally fine, but a round of nitpicking and chatting might be useful :)

PR #5423, opened 2024-10-22
LSP: Add completion support and handle docSync DidChange notifications

OceanOak:
Changelog:

```
LSP
- Handle docSync DidChange notifications
- Improve completions
Extension
- Display script run-time
```

https://github.com/user-attachments/assets/50de707e-6e63-40fd-885e-e24aa27cce57


packages/darklang/languageTools/lsp-server/completions.dark
```diff
@@ -6,49 +6,132 @@ module Darklang =
...
+              filterText = Stdlib.Option.Option.Some label
+              insertText = Stdlib.Option.Option.Some insertText
+              insertTextFormat =
+                Stdlib.Option.Option.Some
+                  LanguageServerProtocol.Completions.InsertTextFormat.InsertTextFormat.PlainText
+              insertTextMode = Stdlib.Option.Option.None
+              // textEdit = Stdlib.Option.Option.None
+              textEditText = Stdlib.Option.Option.None
+              commitCharacters = Stdlib.Option.Option.Some [ " " ]
+              data = Stdlib.Option.Option.Some(Json.String kind) }
```
StachuDotNet, line 34: The `CompletionItem` type has a `kind` field in it, which will affect which icon is shown in the list. If you set up that type, incl the Enum-mapping, you should be able to get "better" (a :package: rather than a :wrench:) icons, rather than simply tossing it in data.

StachuDotNet (approved):
Great! The suggestion I made around `CompletionKind` can wait for a follow-up PR.

PR #5425, opened 2024-10-29
LSP: Improve Hover Support to Show Helpful Information

OceanOak:
No changelog

This PR builds on the work started in #5422 to add Hover support to the language server, along with miscellaneous WT changes and parser fixes.


OceanOak:
TODOs that can be tackled in this PR or in a follow-up
- [ ] show better hover information for (dictionary keys, record field names, enum case names, let patterns, and Builtin functions)
- [ ] Retrieve the underlying type for alias chains
- [ ] Escape early by checking if the range is outside the bounds of the full range
- [ ] extract more helper functions

backend/src/BuiltinExecution/Libs/Packages.fs
```diff
@@ -41,8 +41,8 @@ let parseGenericName (name : string) : GenericName =
     match List.rev modulesAndName with
     | name :: modulesInReverse ->
       { owner = owner; modules = List.rev modulesInReverse; name = name }
-    | [] -> failwithf "Invalid name (no name): %s" name
-  | [] -> failwithf "Invalid name (no owner): %s" name
+    | [] -> Exception.raiseInternal $"Invalid name (no name): {name}" []
+  | [] -> Exception.raiseInternal $"Invalid name (no owner): {name}" []
```
StachuDotNet, line 45: When using `.raiseInternal` it's better to include pieces of data in the metadata [] rather than use string interpolation, for consistency (and also because then you can, when looking at all logs, more easily collect all logs of the same type.

packages/darklang/keywordDesc.dark
```diff
@@ -0,0 +1,65 @@
+module Darklang =
+  module KeywordDescription =
```
StachuDotNet, line 2: Maybe put this this module in Darklang.LanguageTools?

StachuDotNet (approved):
I think this is good to merge as-is.
That said, I was expecting something a bit different:

right now this does a good job of:
- looking through the tree to find whatever node is under position the hover
- returning all the data needed to display for the OnHover thing

, while I was expecting:
- a new type (enum) definition to support the different _kinds_ of thing that could be under the cursor (something like `type ThingAtPosition = | ModuleName of List<String> | Keyword of keywordName: String |  | ...` (this might be a long type, and I'm not totally sure what all data would be required here)
- a function, existing _outside_ of `hoverInformation.dark` (rather, in some _more generic_ location) that: 
  - given a position, returns that ThingAtPosition
- `hoverInformation.dark`'s responsibility is then just to map the `ThingAtPosition` to whatever string(s) need to be returned (maybe that same `Info` type)

There will be multiple contexts in which we care about "what's at the position." Right now, this is all about what we're _hovering_ on, but other similar use cases might be useful for autocomplete of your _cursor position_. To that end, the ThingAtPosition might be paired with a # of how many characters _into_ the thing we're in. i.e. that generic fn might return `(ThingAtPosition, Int)` rather than just `ThingAtPosition`. That way, when you go to autocomplete, we know the `.` you just typed is here `List.Submodule.|` rather than here `List.|Submodule.`, as that may affect what suggestions are returned

Anyway, I think this is worth merging soon, but might be worth mulling on this and refactoring when you think it's appropriate. Let me know if this makes sense and what you think.

OceanOak:
Thanks for the feedback!

Some initial thoughts

> a new type (enum) definition to support the different kinds of thing that could be under the cursor (something like type ThingAtPosition = | ModuleName of List<String> | Keyword of keywordName: String |  | ... (this might be a long type, and I'm not totally sure what all data would be required here)

This might work for hover (still figuring out how), but for completion, we don’t really have the context of thingAtPosition since it’s incomplete and we’re not parsing at that stage


> That way, when you go to autocomplete, we know the . you just typed is here List.Submodule.| rather than here List.|Submodule., as that may affect what suggestions are returned

I believe the current autocomplete implementation is working well for the case you mentioned, as it lets us accurately determine the cursor position and what precedes it

Here’s how I was thinking about it, and why I couldn’t come up with something that works for both:
- in the hover case we only need the one word that is under the cursor (the full word), for example if we are hovering on map in `Stdlib.List.ma|p` we display the information for the function (signature, fqname, etc.)
- For autocomplete, we need to identify the current thing under the cursor and whatever comes before it, stopping when we encounter a space

- For hover, we rely heavily on ranges and use them to get context from WT
- For autocomplete, we focus solely on the cursor position and what comes before it (no context available) 

- For hover, the cursor can be at any position within the word (start, middle, or end), and we need the full word, unlike in the completion case where we only care about what comes before the cursor
- For autocomplete, if we put the cursor at the middle of the word I don't think we expect the autocomplete box to show (at least based on what I've tested) 

There might be a way to unify the position logic for both cases, but I’ll need to experiment to see how it plays out

StachuDotNet:
You're right - the overlap is more with other LSP handshakes, but not autocomplete.
Those overlapping functionalities:
- go to: definition, type definition, implementation, declaration, etc
- find references
- call hierarchy stuff, type hierarchy stuff
- rename stuff

PR #5431, opened 2024-11-22
LSP:  Add support for a File System Provider

OceanOak:
This PR:
- Adds a File System Provider:
  - Implements `ReadFile` (works for types, function, and constants).
  - Starts sketching out `WriteFile`, `Delete`, and `Rename`.
- Adds support for `window/showDocument`.
- Improves some error handling.
- Adds a new `Test Remote File` command to test both the File System Provider (readFile) and window/showDocument.

Current state: 

https://github.com/user-attachments/assets/4ffb1fd6-47a4-4c9e-bd97-9628a57d2396


OceanOak:
Questions: 
1- In the lsp-server logs: when showing a document, we get the following:
```
From client: {"jsonrpc":"2.0","id":1,"result":{"success":true}}
Error parsing incoming message as json-rpc:
{"jsonrpc":"2.0","id":1,"result":{"success":true}}
Error: 'method' field must be provided
To client: {"jsonrpc": "2.0", "error": {"code": -32700, "message": "Parse error"}, "id": null}
```
The client sends a `{"jsonrpc":"2.0","id":1,"result":{"success":true}}` after successfully displaying a document, but it doesn’t include a method field. Since we don’t control this response and the server doesn’t crash, should we just ignore the error message (`Error: 'method' field must be provided`)? Or do you think there’s a way to handle it?

2 - Currently, in the file system provider, we only support reading functions, types, and constants because the package manager provides functions that return these three. What should we do about viewing full modules and submodules? Are we planning to implement something like getAll[Fns/Types/Consts]InModule functions? Any ideas or suggestions?

3 - Regarding the `delete` functionality in the file system provider:
Does deleting a file imply deleting the type, function, or constant we read?
The URI we have is virtual—it doesn’t actually point to a specific entity but is used to display the content we searched for.
Same with Rename: are we renaming the type, function, or constant we read? And what happens if we read a module?
Maybe these functions (delete, rename) are intended only for user code and not for package code?

vscode-extension/client/src/extension.ts
```diff
@@ -12,6 +21,89 @@ import {
...
+      );
+
+      if (!response || typeof response.content !== "string") {
+        throw new Error("Invalid response from LSP server");
+      }
+
+      // Convert response content to Uint8Array and cache it
+      const content = new TextEncoder().encode(response.content);
+      SimpleRemoteFileProvider.fileContents.set(uri.toString(), content);
+      return content;
```
StachuDotNet, line 51: Maybe we don't know but:
why do we need both of these lines? they seem to do the same thing.
Maybe the first line sets the content for any _opened_ documents, and the `return` saves the content for any future openings?

packages/darklang/languageTools/lsp-server/fileSystemProvider.dark
```diff
@@ -0,0 +1,828 @@
...
+              |> Stdlib.Option.values
+              |> Json.Object
+
+
+          let handleReadFileRequest
+            (state: LspState)
+            (requestId: JsonRPC.RequestId)
+            (params: ReadFileParams.ReadFileParams)
+            : LspState =
+            // Remove the darklang:// prefix and the .dark suffix from the URI, (e.g.darklang://type/Darklang/Stdlib/Option/Option.dark)
```
StachuDotNet, line 73: let's add a CLEANUP to make this 'remove scheme part of URI' a bit more universal.
In fact, something like that might be useful in our Stdlib -- like `Stdlib.Uri.removeScheme" or something.

Probably not worth any effort as part of this PR, but this feels a tad precarious as is, so worth the CLEANUP.

packages/darklang/languageTools/lsp-server/fileSystemProvider.dark
```diff
@@ -0,0 +1,828 @@
...
+
+            let nameForLookup =
+              parts |> Stdlib.List.tail |> Builtin.unwrap |> Stdlib.String.join "."
+
+            let pm = LanguageTools.PackageManager.pm ()
+            // e.g.type/Darklang/Stdlib/Option/Option
+            match elementType with
+            | "type" ->
+              let findType = pm.findType
+              let getType = pm.getType
```
StachuDotNet, line 89: nitpick: don't think we need these helpers -- they're both only used once

packages/darklang/languageTools/lsp-server/fileSystemProvider.dark
```diff
@@ -0,0 +1,828 @@
...
+          module ReadFileResponse =
+            type ReadFileResponse = { content: String }
+
+            let toJson (r: ReadFileResponse) : Json =
+              [ Stdlib.Option.Option.Some(("content", Json.String r.content)) ]
+              |> Stdlib.Option.values
+              |> Json.Object
+
+
+          let handleReadFileRequest
```
StachuDotNet, line 68: This could use a bit of tidying:
- there are a few redundant branches here -- the lines [see bottom] are repeated throughout many branches. maybe just have a bit of code at the top that is meant to _just_ get the text to return, and then everything feeds into the "actually return it" bit of code?
- CLEANUP although this isn't part of the official LSP spec, these file system provider handshakes could be of use outside of _our_ language server. maybe we can abstract some functionality out to another module? (`Darklang/lsp-extensions/fileSystemProvider` or something?). At least, the JSON types and mappings could be there.

(redundant code:
```
let responseJson =
                  (ReadFileResponse.ReadFileResponse { content = prettyPrinted })
                  |> ReadFileResponse.toJson
                  |> (fun r ->
                    JsonRPC.Response.Ok.make
                      (Stdlib.Option.Option.Some(requestId))
                      r)
                  |> Stdlib.AltJson.format

                logAndSendToClient responseJson

                // send a showDocument request
                let showDocRequestString =
                  ShowDocument.makeShowDocumentRequest params.uri requestId

                logAndSendToClient showDocRequestString
                state
```
)

StachuDotNet (changes requested):
Looking great - just a few tiny tidyings then this should be good to merge.

packages/darklang/languageTools/lsp-server/fileSystemProvider.dark
OceanOak: I believe these were added because we don't allow direct fieldAccess function calls (because of a parsing error)

StachuDotNet: ah right -- OK, then maybe a CLEANUP to fix that when we have that interpreter/parsing stuff working.

vscode-extension/client/src/extension.ts
OceanOak: I got the caching process to work. Now we only make LSP requests when reading new files

packages/darklang/languageTools/lsp-server/fileSystemProvider.dark
OceanOak: > CLEANUP although this isn't part of the official LSP spec, these file system provider handshakes could be of use outside of our language server. maybe we can abstract some functionality out to another module? (Darklang/lsp-extensions/fileSystemProvider or something?). At least, the JSON types and mappings could be there.

I kept the changes for this ^ in a separate [commit](https://github.com/darklang/dark/pull/5431/commits/06e587ed92e8863174d8448bea90ca3953a91e6). Please let me know if I misunderstood what you meant

packages/darklang/languageTools/lsp-server/fileSystemProvider.dark
```diff
@@ -0,0 +1,115 @@
...
+            : LspState =
+            let responseJson =
+              // (ReadFileResponse.ReadFileResponse { content = content })
+              (LspExtension.FileSystemProvider.ReadFile.ReadFileResponse.ReadFileResponse
+                { content = content })
+              |> LspExtension.FileSystemProvider.ReadFile.ReadFileResponse.toJson
+              // |> ReadFileResponse.toJson
+              |> (fun r ->
+                JsonRPC.Response.Ok.make (Stdlib.Option.Option.Some(requestId)) r)
+              |> Stdlib.AltJson.format
```
StachuDotNet, line 19:
```suggestion
            let responseJson =
              (LspExtension.FileSystemProvider.ReadFile.ReadFileResponse.ReadFileResponse
                { content = content })
              |> LspExtension.FileSystemProvider.ReadFile.ReadFileResponse.toJson
              |> (fun r ->
                JsonRPC.Response.Ok.make (Stdlib.Option.Option.Some(requestId)) r)
              |> Stdlib.AltJson.format
```

PR #5433, opened 2024-12-09
Build the darklang exe for Windows

OceanOak:
No changelog

### DIFF
--- a/backend/src/LibTreeSitter/LibTreeSitter.fsproj
+++ b/backend/src/LibTreeSitter/LibTreeSitter.fsproj
@@ -20,7 +20,7 @@
     the correct native library for the relevant runtime.
   -->
   <PropertyGroup>
-    <RuntimeIdentifiers>linux-x64;linux-musl-x64;linux-arm64;linux-arm;osx-x64;osx-arm64</RuntimeIdentifiers>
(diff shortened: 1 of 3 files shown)


backend/src/LibTreeSitter/LibTreeSitter.fsproj
```diff
@@ -55,6 +55,14 @@
     <EmbeddedResource Include="lib/tree-sitter-macos-arm64.dylib" LogicalName="tree-sitter.dylib" />
     <EmbeddedResource Include="lib/tree-sitter-darklang-macos-arm64.dylib" LogicalName="tree-sitter-darklang.so" />
```
OceanOak, line 56: was it intentional to use `.so` instead of `.dylib` for the logical name of the embedded resource on macOS

OceanOak:
We need this to be merged so I can continue working on getting the CLI to work on Windows. It tries to download it from GitHub Releases and fails because the Windows version isn't included

backend/src/LibTreeSitter/LibTreeSitter.fsproj
StachuDotNet: no

PR #5448, opened 2025-01-27
Add support for Estatement

OceanOak:
Changelog:

```
- Add support for EStatement
```


backend/tests/Tests/Interpreter.Tests.fs
```diff
@@ -632,6 +632,24 @@ module Fns =
 
   let tests = testList "Fns" [ Builtin.tests; Package.tests ]
 
+module Statement =
+  let simple = t "()\n true" E.Statements.statement (RT.DInt64(1L))
+  let simple2 =
+    t "Builtin.printLine(\"hello\")\n 3L" E.Statements.statement2 (RT.DInt64(3L))
+  let nested =
+    t
+      "Builtin.printLine(\"1st statement 1st expr\")\nBuiltin.printLine(\"1st statement 2nd expr\")\nBuiltin.printLine(\"2nd statement 1st expr\")\n0L"
```
OceanOak, line 641: TODO: fix this

I don't think using `Builtin.printLine` in these tests is a good idea, as it creates noise when we run the tests
example:
image

Alternatives use another function that returns a unit --either create it, or use a package function. Hmm, maybe not, I don't think we have an easy way to access package functions from here

packages/darklang/prettyPrinter/programTypes.dark
```diff
@@ -919,7 +924,9 @@ module Darklang =
...
             f.parameters
-            |> Stdlib.List.map (fun p -> $"({p.name} : {p.``type``})")
+            |> Stdlib.List.map (fun p ->
+                let params = PrettyPrinter.RuntimeTypes.typeReference p.``type``
+                $"({p.name} : {params})")
             |> Stdlib.String.join ", "
 
-          parameters ++ " -> " ++ f.returnType
\ No newline at end of file
+          parameters ++ " -> " ++ (f.returnType |> PrettyPrinter.RuntimeTypes.typeReference)
```
OceanOak, line 932: TODO: re-check this
It seems weird to use the runtimeTypes pretty-printer here

tree-sitter-darklang/grammar.js
```diff
@@ -433,6 +436,12 @@ module.exports = grammar({
         field("symbol_right_paren", alias(")", $.symbol)),
       ),
 
+    statement: $ =>
+      prec.left(
+        PREC.STATEMENT,
```
OceanOak, line 441: TODO: (in case needed) Explain, why we used left associativity and why `EStatement` was given a higher priority than `ELambda`

example
```
Builtin.printline "hello"
Builtin.printline "world"
0L
```
Left associativity: **((** Builtin.printline "hello" \n Builtin.printline "world"**)** \n0L **)**

```
(source_file
  (expression
    (statement
      (expression
        (statement
          (expression (apply...))
          (expression (apply...))
        )
      )
      (expression (int64_literal...))
    )
  )
)
```

Right associativity:**(** Builtin.printline "hello"\n **(** Builtin.printline "world"\n 0L **))**
```
(source_file
  (expression
    (statement
      (expression (apply...))
      (expression
        (statement
          (expression (apply...))
          (expression (int64_literal...))
        )
      )
    )
  )
)
```

backend/tests/Tests/Interpreter.Tests.fs
```diff
@@ -632,6 +632,24 @@ module Fns =
 
   let tests = testList "Fns" [ Builtin.tests; Package.tests ]
 
+module Statement =
+  let simple = t "()\n true" E.Statements.simple (RT.DInt64(1L))
+
+  let nested =
+    t
+      "myFnThatReturnsUnit()\nmyFnThatReturnsUnit()\nmyFnThatReturnsUnit()\n0L"
+      E.Statements.nested
```
OceanOak, line 641: The test was something like 
```
Builtin.printLine("1st statement 1st expr")
Builtin.printLine("1st statement 2nd expr")
Builtin.printLine("2nd statement 1st expr")
0L
```
but it created some noise when running tests
example:
image

So, I updated the test values to create a function that returns a Unit. I'm not sure if this is the best thing to do here

https://github.com/darklang/dark/blob/5bb0a03d221c52174ad29201e3fee324c04f9eb3/backend/tests/Tests/TestValues.fs#L499C1-L514C57

https://github.com/darklang/dark/blob/5bb0a03d221c52174ad29201e3fee324c04f9eb3/backend/tests/Tests/TestValues.fs#L670C1-L678C41

backend/src/LibExecution/Interpreter.fs
```diff
@@ -796,6 +796,18 @@ let execute (exeState : ExecutionState) (vm : VMState) : Ply<Dval> =
 
         | RaiseNRE nre -> raiseRTE (RTE.ParseTimeNameResolution nre)
 
+        | CheckIfFirstExprIsUnit reg ->
```
StachuDotNet, line 799: Out of context, this DU case name isn't totally clear. Maybe just `RequireExprToReturnUnit` or something like that?

StachuDotNet: It's probably fine - maybe a CLEANUP to think about this would be useful tho

backend/tests/Tests/Interpreter.Tests.fs
StachuDotNet: This seems good to me

packages/darklang/prettyPrinter/programTypes.dark
StachuDotNet: Hmm, it does feel a bit odd.

The only alternative I can think of would be updating RuntimeTypes (.fs) such that `BuiltInFn`'s  `returnType` (as well as `parameters`' type) had _PT_ `TypeReference`s.

The main blocker to _that_ is that "RT comes before PT" in `LibExecution` and we've had that line drawn in the sand for ages. I question that line pretty often, but it'd take some wrangling to get it all compiling again if we did that.

I'd put a CLEANUP here noting the weirdness and potential solution^ but otherwise fine for now.

---

Oh, another alternative would be introducing a RT2PT module _just_ for TypeReference, but that seems like a can of worms / slippery slope / some other idiom.

StachuDotNet (changes requested):
Looks great! Some minor notes (CLEANUPs to consider), otherwise this looks good to merge.

PR #5449, opened 2025-01-30
Implement multi-pattern matching in match expressions

OceanOak:
Changelog:

```
Area of change
- details
```


OceanOak:
### These tests fail in F# but are currently passing in Darklang
  // All patterns should have the same variables.
  (match (1L,2L) with
   | (x,2L) | (2L,y) when x == y -> "pass"
   | _ -> "fail") = "pass"

  // The two sides bind different sets of variables
  (match (1L,2L) with
    | (x,y) | (x,_) when y > 0L -> "pass"
    | _ -> "fail") = "pass"

  // incompatible pattern structures: List vs Cons
  (match [1L;2L] with
   | [x;y] | x::y -> "pass"  // List vs Cons pattern
   | _ -> "fail") = "pass"
   
  // mixing different types
  (match (1L, "test") with
   | (x, "test") | ("test", x) -> x
   | _ -> "fail") = 1L
   
 ### Unrelated to pattern grouping : 
 // the type should match. can't match an int with a string
  (match (1L, "test") with
    | ("test", x) -> x
    | (x, "test") -> x
    | _ -> "fail") = "fail

//  all branches of a pattern match expression must return values of the same type
 (match (1L,2L) with
    | (x,2L) | (2L,x) when x = 1L -> "pass"
    | _ -> 0L) = "pass"
    
 Do we want to be as strict?

backend/src/LibExecution/ProgramTypes.fs
```diff
@@ -192,6 +192,8 @@ type MatchPattern =
 
   | MPVariable of id * string
 
+  | MPOr of id * List<MatchPattern>
```
StachuDotNet, line 195:
```suggestion
  // CLEANUP consider making this `List` an `NEList`
  | MPOr of id * List<MatchPattern>
```

I recognize that you experimented with making `MatchCase` allow for an `NEList<MP>`, which is similar to this, but what do you think about this^?

My concern with the current approach is that it allows for an `MPOr []` -- so there can be a match case without _any_ real patterns on the LHS. That feels like it should be illegal. We could either tackle that now, or later. Would the NEList being _here_ vs 'higher up' (i.e. `MatchCase`) make any difference?

backend/testfiles/execution/stdlib/language-tools/semanticTokenization.dark
```diff
@@ -833,6 +833,29 @@ module TokenizeMatchExpression =
...
+
+
+  ("match x with\n| 1L | 2L | 3L -> true \n| _ -> false" |> tokenize) =
+    [ (TokenType.Keyword, (0L, 0L), (0L, 5L))
+      (TokenType.VariableName, (0L, 6L), (0L, 7L))
+      (TokenType.Keyword, (0L, 8L), (0L, 12L))
+      (TokenType.Symbol, (1L, 0L), (1L, 1L))
+      (TokenType.Number, (1L, 2L), (1L, 3L))
+      (TokenType.Symbol, (1L, 3L), (1L, 4L))
+      //TODO missing symbol
```
StachuDotNet, line 845: what symbol is missing here?

backend/src/LibExecution/ProgramTypesToRuntimeTypes.fs
```diff
@@ -234,6 +234,50 @@ module MatchPattern =
...
+      // - the symbols used in the pattern
+      // - the register count needed by the pattern
+      let patternsWithSymbols =
+        patterns
+        |> List.map (fun pat ->
+          let pat, patSymbols, patRc = toRT Map.empty rc pat
+          (pat, patSymbols, patRc))
+
+      // find the highest register count needed across all patterns
+      let maxRc = patternsWithSymbols |> List.map (fun (_, _, rc) -> rc) |> List.max
```
StachuDotNet, line 249: this as-is is fine, but FYI there's also `List.maxBy` if you wanted this to be a tiny bit shorter

StachuDotNet (approved):
Great job. Had a quick q/thought about an NEList possibility, but OK to merge like this depending on response.

backend/src/LibExecution/ProgramTypesToRuntimeTypes.fs
OceanOak: I think it turned out a little longer
```
let maxRc =
        patternsWithSymbols
        |> List.maxBy (fun (_, _, rc) -> rc) // it returns a (MatchPattern * Map<string, Register> * int)
        |> (fun (_, _, rc) -> rc)
```
I never really used it, thanks for mentioning it. Should we add it to the list of Stdlib functions that we want to add to Darklang?

backend/src/LibExecution/ProgramTypes.fs
OceanOak: I think it will fail during the parsing phase before getting to this point, but it's probably worth updating it.
I don’t think there is much difference, but I still wonder if it would be cleaner to make the changes in MatchCase

backend/testfiles/execution/stdlib/language-tools/semanticTokenization.dark
StachuDotNet: Hmm, for  `| MPOr of Range * List<MatchPattern>`, I thought that `Range` would correspond to the whole thing, not the `|` symbol.

Here over in `MatchCase`,
```
      type MatchCase =
        { pat: Range * MatchPattern * Range
          whenCondition: Stdlib.Option.Option<Range * Expr>
          rhs: Expr }
```
, I thought that the 3rd part of `pat` would maybe capture the `|` part. Not sure how that fits into things.

In any case, two thoughts here:
- it's not that important to capture the `|` at this time, so probably ok to just add a CLEANUP and ignore this
- we should put a `///` comment on that `pat` to clarify what the tuple parts correspond to
- (bonus: that second bullet's point probably applies to a few more `Range`s, but I haven't looked to see)

backend/src/LibExecution/Interpreter.fs
```diff
@@ -101,6 +101,13 @@ let rec checkAndExtractMatchPattern
   | MPEnum(caseName, fields), DEnum(_, _, _, caseNameActual, fieldsActual) ->
     if caseName = caseNameActual then rList fields fieldsActual else false, []
 
+  | MPOr patterns, dv ->
+    patterns
+    |> NEList.toList
+    |> List.map (fun p -> r p dv)
+    |> List.tryFind (fun (matches, _) -> matches)
+    |> Option.defaultValue (false, [])
```
StachuDotNet, line 109:
```suggestion
  | MPOr patterns, dv ->
    patterns
    |> NEList.map (fun p -> r p dv)
    |> NEList.tryFind (fun (matches, _) -> matches)
```

backend/testfiles/execution/stdlib/language-tools/semanticTokenization.dark
OceanOak: > Hmm, for | MPOr of Range * List<MatchPattern>, I thought that Range would correspond to the whole thing, not the | symbol.

That's correct, it does correspond to the whole thing (There are no pipe symbols captured in an `MPOr` currently)


And here  `pat: Range * MatchPattern * Range`
- the first range is for the `|`
- the second one is for the  `->`

PR #5455, opened 2025-02-25
Performance improvement for dictSet function

hummy123:
No changelog

There was a `CLEANUP: terrible perf` comment for the dictSet function inside the `backend/src/BuiltInExecution/Libs/Dict.fs` file.

The reason for the poor performance was that `Map.toList` was called, and then the `@` operator (shorthand for the `List.concat` function) was called immediately afterwards to put a key-value pair at the end of the just-created list.

While cons (`(1, 3) :: list`) is an O(1) operation, List.concat/@ is an O(n) operation because it needs to traverse the whole list and put the new tuple at the end.

We can:
1. Delete the call to Map.toList
2. Delete the use of the @ operator
3. Cons a list of all elements in the map starting from the last element to the first element, *passing the tuple we want at the end as the initial list* (which erases the need for the @ operator/List.concat).

Code before changes:

```
o
|> Map.toList
|> (@) [ (k, v) ]
|> TypeChecker.DvalCreator.dict vm.threadID vt
|> Ply
```

Code after changes:

```
Map.foldBack (fun key value acc -> (key, value) :: acc) o [ (k, v) ]
|> TypeChecker.DvalCreator.dict vm.threadID vt
|> Ply
```

I think the anonymous function is a bit terse to be readable, so I would probably do

```
let consKV key value acc = (key, value) :: acc
Map.foldBack consKV o [ (k, v) ]
```

instead, if it was my own code, but I didn't see that style anywhere else in this file, so I didn't.

I kept the `CLEANUP: terrible perf` comment since one other place in the Dict.fs file (line 112 as of this commit) mentions performance regarding Map.toList, but I did take the word "terrible" out of the comment since it's better than before at least.

It's a simple change and `dotnet fsi` can prove it works as intended.

```
let map = Map.empty.Add(1, 3).Add(2, 5).Add(3, 7).Add(4, 9)

// output: [(1, 3); (2, 5); (3, 7); (4, 9)]
Map.toList map

// same output as Map.toList: we fold through from last element to first, starting with an empty list
Map.foldBack (fun key value acc -> (key, value) :: acc) map []

// output: [(1, 3); (2, 5); (3, 7); (4, 9); (5, 11)]
// We pass in an initial list containing (5, 11), so (5, 11) is at the end
Map.foldBack (fun key value acc -> (key, value) :: acc) map [(5, 11)]
```

It's a small change in purely functional code with equal chance to failing as the current implementation.

I didn't explain the anonymous function in the comments or our initial list, but I will if it's preferred that I do.

### DIFF
--- a/backend/src/BuiltinExecution/Libs/Dict.fs
+++ b/backend/src/BuiltinExecution/Libs/Dict.fs
@@ -230,10 +230,8 @@ let fns : List<BuiltInFn> =
       fn =
         (function
         | _, vm, _, [ DDict(vt, o); DString k; v ] ->
-          // CLEANUP terrible perf
-          o
-          |> Map.toList
-          |> (@) [ (k, v) ]
+          // CLEANUP perf
+          Map.foldBack (fun key value acc -> (key, value) :: acc) o [ (k, v) ]
           |> TypeChecker.DvalCreator.dict vm.threadID vt
           |> Ply
         | _ -> incorrectArgs ())


StachuDotNet:
Really nice!

Appreciate your explanations, but something more terse would have been plenty fine :)
e.g.:

> Fixed perf issue noted by a `CLEANUP` in `Builtin.dictSet`:
> Previously, this builtin used `Map.toList` and `@` to append the new entry, but with this PR is uses `Map.foldBack` to do this more efficiently.

No complaints here, just noting we don't require formality.
More important are making sure that `backend/testfiles/execution/dict.dark` has tests covering this scenario. As far as I can think, I think we're OK.

---

FYI: there were two things bad about the performance here, that I'm aware of:
- the `Map.toList` followed by `@` you fixed - thanks!
- calling `TypeChecker.DvalCreator.dict`, which will end up doing way too much double-checking of type-safety. There are some directional notes in that module of `TypeChecker` to improve performance. Any work in that space would be greatly appreciated, if you're into that sort of thing! (I probably have a lot of CLEANUPs in the space of type-checking and such since a recent runtime rewrite concluded.)

StachuDotNet:
I actually forgot about the PR template - the PR description template exists as a general guideline, but lately we've been less structured as there isn't a "live environment" with dark-next running where peoples' live programs will break, etc, unlike -classic where that pattern was established. So, I think Ocean and I typically just wipe it out and write a good short description

PR #5456, opened 2025-02-26
Perform minimal type checking for dictSet function,

hummy123:

There's a performance problem with the `dictSet` function: the dictionary is a map, which we convert to a list, only for the dictionary type-checking function to convert it back to a map.

Suboptimal performance comes from two things:
1. Converting a map to a list, only to convert the list back to a map (and adding one key).
2. Type checking every single element in the dictionary, which is unnecessary because we accept that the map is type safe as a prerequisite for the `dictSet` function.

I think we can perform fewer operations while upholding type safety.

1. Get any key/value pair from the map (preferably without searching the whole map).
2. Compare the types of the old key/value pairvwith the new key/value pair we wish to insert
3. If the old pair and the new pair have the same types, insert them into the existing map (else, raise an error)

No need to check the type of every single element, or convert the map to a list, which are both O(n) operations.

Commit only contains changes to a bit of purely functional code, and all the Dict.set tests which passed before still pass now.

I also added two new tests, adding different key/value pairs to empty dictionaries.

The general idea is commented.

## Other notes

I wanted to add a test to check if a type error is raised when we:

1. Remove everything from a (string, int) dictionary
2. Set a key/value pair of some different type like (string, string)

but I wasn't sure how to write multi-line tests as there didn't seem to be any in the files I looked at. I think my code won't introduce any regression in that regard because it passes the exact same arguments to the type-checking function, except the list to type-check is shorter.

As an alternative to this approach, if we want to type check N elements but not to covert the whole map to a list, we can call Maq.toSeq and iterate through a number of elements smaller than the length of the map, but I'm not sure what we gain with that approach.

### DIFF
--- a/scripts/build/_build-server
+++ b/scripts/build/_build-server
@@ -10,7 +10,7 @@ import sys
 import threading
 import signal
 
-run_tests = True
+run_tests = False
 
 # When compiling code, use the optimized version of the build. This causes
 # --optimize to be passed to the compile script.
--- a/scripts/build/_build-server
+++ b/scripts/build/_build-server
@@ -10,7 +10,7 @@ import sys
 import threading
 import signal
 
-run_tests = False
+run_tests = True
 
 # When compiling code, use the optimized version of the build. This causes
 # --optimize to be passed to the compile script.
--- a/backend/src/BuiltinExecution/Libs/Dict.fs
+++ b/backend/src/BuiltinExecution/Libs/Dict.fs
@@ -230,10 +230,35 @@ let fns : List<BuiltInFn> =
       fn =
         (function
         | _, vm, _, [ DDict(vt, o); DString k; v ] ->
-          // CLEANUP perf
-          Map.foldBack (fun key value acc -> (key, value) :: acc) o [ (k, v) ]
-          |> TypeChecker.DvalCreator.dict vm.threadID vt
-          |> Ply
+          // Our algorithm to avoid excessive type checking for this 'set' function:
+          // First, we get any random key which already exists in the map.
+          // Secondly, because the values already existing in this map are already
+          // guaranteed to be the same type (or else we would have no map here),
+          // we create a list of the old random (key, value) pair with the new
+          // (key, value) pair and check if this new list is type safe.
+          // If so, we add the new (key, value) pair to the map.
+          let alwaysTrue _ _ = true
+          // Get random key in map, without searching the whole map
+          match Map.tryFindKey alwaysTrue o with
+          | Some oldKey ->
+            // Get value belonging to old key
+            // Note: The Prelude module shadows the normal 'find'
+            // function which raises an exception if there's no matching key
+            // so we have to perform unnecessary pattern matching, or edit Prelude
+            // to have a `Map.findUnsafe` function.
+            match Map.find oldKey o with
+            | Some oldVal ->
+              let typeList = [ (oldKey, oldVal); (k, v) ]
+              // Implicit: below function call will raise an exception if not type safe.
+              let _ = TypeChecker.DvalCreator.dict vm.threadID vt typeList
+              let map = DDict(vt, Map.add k v o)
+              Ply map
+            | None ->
+              // This case is impossible
+              TypeChecker.DvalCreator.dict vm.threadID vt [ (k, v) ] |> Ply
+          | None ->
+            // The None case means the existing map is empty
+            TypeChecker.DvalCreator.dict vm.threadID vt [ (k, v) ] |> Ply
         | _ -> incorrectArgs ())
       sqlSpec = NotYetImplemented
       previewable = Pure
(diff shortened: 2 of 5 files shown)


OceanOak:
I like the idea behind this change! I think we might be able to simplify it a bit while keeping the performance improvements you introduced.
For non-empty dictionaries: create a new dictionary with just the new key-value pair
For non-empty dictionaries:
- Get the first existing key-value pair to use as a type reference(instead of a random one)
   Something like 
   ```
      let firstKey = Map.keys o |> Seq.head
      let firstVal = Map.find firstKey o |> Option.get
    ```
- Create a minimal test list with both the existing pair and the new pair (as you did)
- Use this minimal list to verify the type
- If type matches add the pair

What do you think? 

(Stachu will be doing the actual review since he knows this best)

hummy123:
I think that's a good idea! I wasn't sure of how to clean the code up, but I like your suggestion. It gave the following idea for implementation:

```fsharp
if Map.isEmpty map then
  // create from [(key, value)] list as existing commit does
else
  let (oldKey, oldVal) = Map.minKeyValue map
  // use oldKey and oldVal same way existing commit does next
```

I think that would simplify it by getting rid of the pattern matching, and we would only need one line to find the (key, value) pair instead of making two separate queries to the map as the commit currently does.

If Stachu is fine with the general idea of this commit, I would be happy to update the commit to incorporate that suggestion.

scripts/build/_build-server
```diff
@@ -10,7 +10,7 @@ import sys
 import threading
 import signal
 
-run_tests = True
+run_tests = False
```
OceanOak, line 13: in case you didn't know, you can run tests from the command line using:
- `./scripts/run-backend-tests` to run all backend tests
-  `./scripts/run-backend-tests --filter-test-list X` to run only test lists with X in their names
e.g. `./scripts/run-backend-tests --filter-test-list dict`

hummy123: Thanks. I didn't know but that's helpful and a lot better than the way I tried doing it before. Appreciate it.

hummy123:
> I think this solution puts a lot of responsibility on the specific builtin function. Rather, I'd love to see a solution that could be easily re-used, with the majority of updates likely in `TypeChecker.fs`.

I think this is a good idea and the latest commit follows it. (I have some tolerance to messy code 😅 but don't want to make things harder for others to maintain, so I appreciate the advice from you and Ocean on how to simplify the code further.)

There's a regression which I've added a test for and it's easily resolved, but I'm not sure what the maintainers' preferences are.

The first of these tests passes fine. The second one would have before, but now the second one doesn't. We're only checking the minimum key/value pair instead of every element which is why the first passes while the second doesn't.

```fsharp
  Stdlib.Dict.set_v0 (Dict { key1 = "val1before" }) "key1" "val1after" =
    Builtin.testDerrorMessage "Cannot add two dictionary entries with the same key `key1`"

  Stdlib.Dict.set_v0 (Dict { key1 = "val1"; key2 = "val2" }) "key2" "newVal2" =
    Builtin.testDerrorMessage "Cannot add two dictionary entries with the same key `key2`"
```

The fix is easy: we can move up (or duplicate) the call to `Map.contains` in `TypeChecker.DvalCreator.dict`, checking either in the builtin function or the `dictAdd` function if the key we wish to insert already exists in the map.

```fsharp
// exception-raising code in TypeChecker.DvalCreator.dict function
if Map.containsKey k entries then
  RTE.Dicts.Error.TriedToAddKeyAfterAlreadyPresent k
  |> RTE.Error.Dict
  |> raiseRTE threadID
```

I would personally prefer to put the if statement in the builtin (maybe put the exception-raising code in a new function so it doesn't need to be copy-pasted repeatedly).

The reason for my preference is that it's easier to see and understand control flow at a glance by looking at the specific builtin without digging into the exact behaviour of functions that are called by the builtin. There are also some functions that aren't meant to disallow "reinsertion of the same key" (like `Dict.merge` or `Dict.fromListOverwritingDuplicates`).

That's just my preference though. I'd appreciate a second opinion on where to put the duplicate-check since Chesterton's fence is a thing and I'm not as familiar with the code.

hummy123:
That looks like it would be perfect actually, and cleaner than the way I previously thought of distinguishing duplicates. I think that kind of approach didn't come to mind before because of a misunderstanding I had of your original comment.

I'll try implementing your suggestions tomorrow and seeing if there's anything to improve. (I think the only thing that sticks out to me is the new bool parameter when the Darklang coding guidelines, if I remember, prefer two-case discriminated unions.)

StachuDotNet:
a 2-case DU would be fine. I was mostly lazy to come up with names :)
Alternatively, have 1 private function that takes in that bool, and 2 tiny public wrapper fns with explicit names (`dictAddEntryMerging` and `dictAddEntryNonMerging` for example)

PR #5464, opened 2025-04-01
Address CLEANUP note for List.fs: use custom order type in comparison

hummy123:
Small commit which addresses a CLEANUP note in `src/BuiltinExecution/Libs/List.fs`.

The note was: 

```
// CLEANUP something like type ComparatorResult = Higher | Lower | Same
// rather than 0/1/-2
```

in the context of comparing values.

This commit introduces a new `type order = Less | Equal | Greater` in List.fs and refactors the comparison functions to return it.

I thought a bit about other functional languages for the case names. Haskell uses `LT | EQ | GT`, Standard ML uses `LESS | EQUAL | GREATER` and OCaml uses integers instead of a custom order type.

Maybe the original proposal `Higher | Lower | Same` would be a bit better because the average word is less verbose? I thought it made sense to follow the conventions of another functional language but can see the other perspective too.

It might be worth putting a bit more thought into naming since, although this order type is local to List.fs, I don't think it's uncommon to pattern match on an order type (and typing all those different cases repeatedly because of it).

I questioned the purpose of the change in this commit at first because this comparison is only used for sorting the list using the F# standard library function `List.sortWith` which takes an int. Then I realised, it might be considered clearer/more readable to use the custom order type internally and convert it into an int when we finally want to pass the result to `List.sortWith`.

There [are three old tests](https://github.com/darklang/dark/blob/b10c0d3a61a49f9a238f201a3cf9c84eeb107b58/backend/testfiles/execution/stdlib/list.dark#L268) for the `List.sort` builtin provided by Dark's standard library, and these still pass.

### DIFF
--- a/backend/src/BuiltinExecution/Libs/List.fs
+++ b/backend/src/BuiltinExecution/Libs/List.fs
@@ -135,7 +135,7 @@ module DvalComparator =
       | Less -> Less
       | Equal -> compareLists t1 t2
 
-  and compareMaps (o1 : List<string * Dval>) (o2 : List<string * Dval>) : order =
+  and compareMaps (o1 : List<string * Dval>) (o2 : List<string * Dval>) : Order =
     match o1, o2 with
     | [], [] -> Equal
     | [], _ -> Less
--- a/backend/src/BuiltinExecution/Libs/List.fs
+++ b/backend/src/BuiltinExecution/Libs/List.fs
@@ -11,7 +11,7 @@ module TypeChecker = LibExecution.TypeChecker
 
 
 module DvalComparator =
-  type order =
+  type Order =
     | Less
     | Equal
     | Greater
--- a/backend/src/BuiltinExecution/Libs/List.fs
+++ b/backend/src/BuiltinExecution/Libs/List.fs
@@ -23,7 +23,7 @@ module DvalComparator =
     else Equal
 
   // should this take a vmstate?
-  let rec compareDval (dv1 : Dval) (dv2 : Dval) : order =
+  let rec compareDval (dv1 : Dval) (dv2 : Dval) : Order =
     match dv1, dv2 with
     | DUnit, DUnit -> Equal
 
--- a/backend/src/BuiltinExecution/Libs/List.fs
+++ b/backend/src/BuiltinExecution/Libs/List.fs
@@ -124,7 +124,7 @@ module DvalComparator =
 
 
 
-  and compareLists (l1 : List<Dval>) (l2 : List<Dval>) : order =
+  and compareLists (l1 : List<Dval>) (l2 : List<Dval>) : Order =
     match l1, l2 with
     | [], [] -> Equal
     | [], _ -> Less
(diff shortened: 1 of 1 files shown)


backend/src/BuiltinExecution/Libs/List.fs
```diff
@@ -10,36 +10,44 @@ module Interpreter = LibExecution.Interpreter
 module TypeChecker = LibExecution.TypeChecker
 
 
-// CLEANUP something like type ComparatorResult = Higher | Lower | Same
-// rather than 0/1/-2
-
 module DvalComparator =
+  type order =
```
StachuDotNet, line 14: style nitpick
```suggestion
  type Order =
```

StachuDotNet:
Nice work! Just one request to upper-case Order, otherwise this is good to merge. Thanks!

backend/src/BuiltinExecution/Libs/List.fs
```diff
@@ -10,36 +10,44 @@ module Interpreter = LibExecution.Interpreter
...
+
+  let order v1 v2 =
+    let result = compare v1 v2
+    if result < 0 then Less
+    elif result > 0 then Greater
+    else Equal
+
   // should this take a vmstate?
-  let rec compareDval (dv1 : Dval) (dv2 : Dval) : int =
+  let rec compareDval (dv1 : Dval) (dv2 : Dval) : order =
```
StachuDotNet, line 26:
```suggestion
  let rec compareDval (dv1 : Dval) (dv2 : Dval) : Order =
```

backend/src/BuiltinExecution/Libs/List.fs
```diff
@@ -112,29 +124,37 @@ module DvalComparator =
...
     | h1 :: t1, h2 :: t2 ->
-      let c = compareDval h1 h2
-      if c = 0 then compareLists t1 t2 else c
+      match compareDval h1 h2 with
+      | Greater -> Greater
+      | Less -> Less
+      | Equal -> compareLists t1 t2
 
-  and compareMaps (o1 : List<string * Dval>) (o2 : List<string * Dval>) : int =
+  and compareMaps (o1 : List<string * Dval>) (o2 : List<string * Dval>) : order =
```
StachuDotNet, line 138:
```suggestion
  and compareMaps (o1 : List<string * Dval>) (o2 : List<string * Dval>) : Order =
```

backend/src/BuiltinExecution/Libs/List.fs
```diff
@@ -112,29 +124,37 @@ module DvalComparator =
 
 
 
-  and compareLists (l1 : List<Dval>) (l2 : List<Dval>) : int =
+  and compareLists (l1 : List<Dval>) (l2 : List<Dval>) : order =
```
StachuDotNet, line 127:
```suggestion
  and compareLists (l1 : List<Dval>) (l2 : List<Dval>) : Order =
```

PR #5478, opened 2025-06-16
Fix broken link

simonw:
The README was linking to https://github.com/darklang/classic-darklang which is a 404 - the correct link is https://github.com/darklang/classic-dark.


StachuDotNet:
thanks!

PR #5479, opened 2025-06-27
CLI: Fix view command

OceanOak:
No changelog


packages/darklang/languageTools/nameResolver.dark
```diff
@@ -103,9 +103,6 @@ module Darklang =
           (name: WrittenTypes.Name)
           : ProgramTypes.NameResolution<ProgramTypes.FQTypeName.FQTypeName> =
           match name with
-          // TODO remodel things appropriately so this is not needed
-          | KnownBuiltin(_name, _version) -> "Builtin types don't exist" // TODO: error
```
StachuDotNet, line 107: Ah I guess we got rid of this case a while back and forgot to delete this code?

OceanOak: Yes, it seems like it

PR #5481, opened 2025-07-04
Consolidate CLI efforts

StachuDotNet:
We used to have a tiny CLI 'app' with a few commands, and the experience was wholly uninteractive. Later, we added a 'repl' experience with more commands (and many more still to go...), which could be called either interactively or noninteractively. This PR consolidates those efforts. Some functionality is lost for now, incl. self-update of the exe.


StachuDotNet:
I had spent a few hours here debugging what I _thought_ was a runtime type-checking error. It turned out to be that I had moved the `ExecutionError` type from the `Darklang.CLI` module into an `ExecutionError` submodule, while not updating `PackageIDs`.

Two TODOs to come out of this:
- [ ] update the package loading stuff to complain if any of the referneced ID's aren't found during a full parse
- [ ] update `TC.fs` to fail more explicitly when something is not found

Issue #5246, opened 2024-01-09
Component: Internal things to improve


StachuDotNet:
This Issue exists to collect many items, not currently in progress, that don't quite effect users, but we still want to address at some point.

(If an issue is urgent, or important enough, then it should have its own issue.)

- [ ] Everything marked TODO in the codebase
- [ ] Everything marked CLEANUP in the codebase (was #3938)
- [ ] Somehow report on types allow<>'d to be serialize that are unused (see #4589)
- [ ] Prevent Timeout exceptions from being rollbar'd (see #4092)
- [ ] Decide between `init`, `lazy`, and passing values for startup values and singletons (see #3944)
- [ ] Lock dotnet tool versions in Dockerfile (see #3896)
- [ ] Warm up BwdServer when it starts (see #3782)
- [ ] Run F# backend against coverage tool (see #3688)
- [ ] Document/Fix F# Debugging (see #3634)
- [ ] Consider banning `string` function (see #3567)
- [ ] improve F# build performance (see #2952)
- [ ] Include reference to user in rollbars created by queueworker (see #3905)

Once we start on any of these,
- extract the bullet into its own issue, copying any relevant notes
- , and delete the bullet here

PR #5483, opened 2025-07-08
Add CLI integration tests

OceanOak:
No changelog

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -157,10 +157,9 @@ RUN DEBIAN_FRONTEND=noninteractive \
       && apt clean \
       && rm -rf /var/lib/apt/lists/*
 
-# Install expect for the CLI integration tests.
+# Install expect for the CLI integration tests
 RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
-  DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends expect || \
-  echo "Warning: expect installation failed, continuing without it" && \
+  DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends expect && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,6 +76,11 @@ RUN echo "deb [arch=${TARGETARCH}] https://download.docker.com/linux/ubuntu $(ls
 
 RUN echo "deb https://apt.releases.hashicorp.com $(lsb_release -cs) main" > /etc/apt/sources.list.d/hashicorp.list
 
+# Charm
+RUN mkdir -p /etc/apt/keyrings
+RUN curl -fsSL https://repo.charm.sh/apt/gpg.key | gpg --dearmor -o /etc/apt/keyrings/charm.gpg
+RUN echo "deb [signed-by=/etc/apt/keyrings/charm.gpg] https://repo.charm.sh/apt/ * *" | tee /etc/apt/sources.list.d/charm.list
+
 # Mostly, we use the generic version. However, for things in production we want
 # to pin the exact package version so that we don't have any surprises.  As a
 # result, sometimes the versions upgrade from under us and break the build. To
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -138,6 +143,10 @@ RUN DEBIAN_FRONTEND=noninteractive \
       libstdc++6 \
       zlib1g \
       # end .NET dependencies
+      # charm dependencies
+      ffmpeg \
+      ttyd \
+      vhs \
```
StachuDotNet, line 149: I imagine these packages are pretty small? Just resisting making devcontainer much bigger. I think we're fine, but might be nice (add to #5246?) long-term to keep track of our devcontainer size, install time; and in PRs that update it, record how much of an impact it makes. I suspect these deps are fine and we'd be OK to just add a relevant TODO or 2 to that Issue

Dockerfile
```diff
@@ -148,6 +157,12 @@ RUN DEBIAN_FRONTEND=noninteractive \
       && apt clean \
       && rm -rf /var/lib/apt/lists/*
 
+# Install expect for the CLI integration tests.
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
+  DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends expect || \
+  echo "Warning: expect installation failed, continuing without it" && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/*
```
StachuDotNet, line 165: is there no simpler way to install expect? this doing an rm -rf on lists, and _another_ apt-get clean seem like a bit much

packages/darklang/cli/integrationTests/test-framework.dark
```diff
@@ -0,0 +1,283 @@
...
+          match TestUtils.getFileContent filePath with
+          | Ok content -> Stdlib.Bool.not (Stdlib.String.isEmpty content)
+          | Error _ -> false
+
+        let filesAreEqual (file1: String) (file2: String): Bool =
+          let result = Builtin.cliExecute $"diff -q {file1} {file2}"
+          result.exitCode == 0L
+
+        let generateDiff (expectedFile: String) (actualFile: String): String =
+          let result = Builtin.cliExecute $"diff --unified=3 --color=always {expectedFile} {actualFile}"
```
StachuDotNet, line 85: won't work on Windows. We're not on Windows, so that's fine for now, but we should be careful in a few ways:
- if something only works on UNIX systems, note so (e.g. in the `integrationTests/README.md?`)
- generally avoid using `Builtin.cliExecute` in code outside of stdlib -- and use it as little as possible.
  - to that end, this may be a good time to add a `Darklang.Stdlib.CLI.diff` -- which (for now) can work on Unix doing what you did, and fail/RTE on Windows

StachuDotNet: rethinking, some of this feedback doesn't make sense since we'll likely always run the tests in the devcontainer, but it's probably a good habit to get into?

packages/darklang/cli/integrationTests/test-framework.dark
```diff
@@ -0,0 +1,283 @@
...
+                | _ ->
+                  Builtin.printLine "No output captured"
+                TestResult.Failure $"CLI command failed: {e}"
+              | Ok _ ->
+                if Stdlib.Bool.not (isFileEmpty actualFile) then
+                  TestUtils.printlnColored CliColors.red "✗ Output file is empty."
+                  TestResult.Failure "Output file is empty"
+                else
+                  if Stdlib.Bool.not (TestUtils.fileExists expectedFile) then
+                    TestUtils.printlnColored CliColors.yellow $"⚠ Expected output file not found: {expectedFile}"
```
StachuDotNet, line 236: might be good to add a `printlnWarning` shortcut, or similar? just to be a bit less verbose

Dockerfile
OceanOak: Adding ffmpeg, ttyd, vhs, and expect increased the image size from 4.94 GB to 5.09 GB (~154 MB increase)
will add a TODO

OceanOak: is the TODO just to track the dev container size and install time...?

PR #5486, opened 2025-07-14
Add syntax highlighting functionality to `view function/type/constant` command

OceanOak:
image

Should syntax highlighting be under a flag? Something like:
`--highlight` or `--style=highlighted` and `--style=plain`


StachuDotNet:
no flag needed yet. some day we can add `--plain` as an option

PR #5538, opened 2025-09-10
notes: fancy self and arg refs

StachuDotNet:
(notes from a chat with Ocean for her use)


StachuDotNet:
ocean has since implemented this properly, elsewhere

PR #5640, opened 2026-04-23
Interpreter: move lambda + package fn instruction caches onto ExecutionState

OceanOak:
Passing an inline lambda as an `httpServerServe` handler failed before the server even started with `Uncaught exception: IncorrectArgs`. 
                                                                                                                                                                                                                                     
- Root cause: httpServerServe's pattern match only accepted `AppNamedFn`. Even if it hadn't, `lambdaInstrCache` lived on `VMState`, so every cross-VM invocation (e.g. HTTP handler) had to copy the caller's lambdas into each new VM by hand (O(N) per call)
- Moved `lambdaInstrCache` and `packageFnInstrCache` off `VMState` onto `ExecutionState`. Every VM spawned under a single execution now shares the caches, so lambda lookup just works across VMs without any copying.  
-  Inline-lambda handlers now work:                                                                                                                                                                                                   
    `./scripts/run-cli eval 'Stdlib.HttpServer.serve 9099L (fun req -> ...)'`.

### DIFF
--- a/backend/src/BuiltinCloudExecution/Libs/DB.fs
+++ b/backend/src/BuiltinCloudExecution/Libs/DB.fs
@@ -70,30 +70,22 @@ let resolveLoadValues
   }
 
 
-/// Look up a lambda from the VM cache by its exprId
-let lookupLambdaImpl (vm : VMState) (exprId : id) : LambdaImpl =
-  match
-    Map.tryFind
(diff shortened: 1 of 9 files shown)


StachuDotNet:
Great work here - I struggled for this a while some weeks back.

PR #5644, opened 2026-05-01
Replace BwdServer with CLI Http.serve; remove more Cloud-specific branding of code

StachuDotNet:
Replaces BwdServer (ASP.NET-hosted, multi-canvas) with the CLI's `Stdlib.HttpServer.serve` builtin on `System.Net.HttpListener`:
- Drops `Microsoft.AspNetCore.*` from the runtime dep graph.
- Rebrands a semi-working 'http-server' CLI command to `darklang serve <fn-path>`
- ^ supports per-request logging, body-size limits, graceful SIGINT shutdown, multi-handler conflict detection, X-Forwarded-Proto canonicalization, opt-in `Stdlib.Http.faviconResponse`.
- Routing is in Dark (`Stdlib.HttpServer.routeRequest`)
- F# only does accept + parse + write.
- Along with deleting `BwdServer`, `LibService` and `LibCloudExecution` is now gone. `LibHttpMiddleware` has been folded into `BuiltinHttpServer`.
- `HttpListener`, as a lower-level thing that relying on ASP.NET, is a step in the right direction. Eventually we might lower further to a 'raw' `TcpListener` for various reasons, but nontrivial so we can tackle that if/when relevant.
- `darklang serve` runs at ~10-15 ms/request with HTTP/1.1 keepalive (browsers, fetch, axios, requests - the typical case). Short-lived clients (curl-in-a-loop) see ~130 ms/req, ~80 ms of which is HttpListener-internal blocking, not Dark interpreter (verified via `dotnet-trace`). `serve --help` and `docs for-ai-internal` document this. Not amazing perf, but good enough for our needs.

Blocker for merge: tracing regression.

---

BwdServer was the last real cloud-branded exe. Now that it's gone, did some work to trim the cloud branding a bit.
- renamed `LibDB` to `LibSqlite`
- renamed `LibPackageManager` to `LibDB` - now manages all DB-accessing stuff. (UserDB, Tracing, etc.)
- `BuiltinDB` now exposes all builltins that are DB-bound

---

Context for why this was done: I wanted to have AI build some http-bound projects, but the `http-server` project was finnicky and incomplete. Have been meaning to get this work done for a bit, so closed the loop.


StachuDotNet:
replacing with larger, better PR shortly

PR #5645, opened 2026-05-01
Make package dependencies location-aware

OceanOak:
This PR makes package dependencies location-aware instead of hash-only.

Package hashes are content-addressed, so different package items can share the same hash. Treating the hash as the dependency identity could make reverse dependency lookup, delete checks, propagation, and AST rewrites affect the wrong package item.

  ## What changed
  - Track `PackageLocation` in `NameResolution`.
  - Store dependency edges with hash, item kind, and package location.
  - Query reverse dependencies by `(location, itemKind)`, with hash fallback.
  - Make propagation, SCC hash stabilization, AST rewriting, and canonical hashing use location-aware substitutions.
  - Preserve content-addressability: location guides lookup/substitution, but is not included in canonical hash bytes.
  - Fix package update refresh by compacting stale WIP ops and evaluating new package values immediately.
  - Update Dark/CLI callers and tests.

## Note
Deprecations are still hash-keyed and have a cleanup note for a future location-aware pass.


backend/src/BuiltinCliHost/Libs/Canvas.fs
```diff
@@ -62,6 +62,7 @@ let fns () : List<BuiltInFn> =
                   typ =
                     PT.TypeReference.TCustomType(
                       { originalName = []
+                        location = None
                         resolved = Ok(PT.FQTypeName.Package typeHash) },
```
StachuDotNet, line 66: maybe we should rename `resolved` to `hash`. Or rename both to `resolvedLocation` and `resolvedHash`. what do you think?

StachuDotNet: or could be

```fsharp
TCustomType
  { source = []
    resolved = Ok { hash = PT.FQTypeName.Package [hash]; location = [location] } }
```

Feels noisy, but structurally correct?

Fine as-is, just thinking out loud.

StachuDotNet:
This all feels positive, worth merging. That said, I feel like the mental/technical weight of the 'propogation' feature continues to feel heavier than it strictly needs to be, but can't think of a small refactor. Maybe we need to somehow greatly simplify our internal state management - there's a lot of dancing around I/O between Ops and Sql, and the CLI, and PM state supporting the Builtins, etc. i.e. maybe our lives would be easier if we removed the majority of sql tables, leaving just ops and caches, managing state in a background daemon accessed by the CLI app? Just an idea.

backend/src/BuiltinCliHost/Libs/Canvas.fs
OceanOak: Hmm, sometimes resolved isn't a hash (Builtins)
maybe something like 
```
  type ResolvedName<'name> =
    { name : 'name
      location : Option<PackageLocation> }

  type NameResolution<'name> =
    { source : List<string>
      resolved : Result<ResolvedName<'name>, NameResolutionError> }
```
wdyt?

StachuDotNet: Feel structurally great. A bit noisy, but I'd rather be noisy than wrong :)

PR #5672, opened 2026-06-08
Make numeric operators polymorphic over all numeric types

OceanOak:
Infix operators (+ - * / % ^ < > <= >=) used to lower to Int64-specific builtins, so 1.0 + 2.0 failed at runtime with an int64Add type error and / only did float division. 

This PR:
- points them at new polymorphic builtins that pick the right numeric type at runtime like equals already does.
Now `1 + 2`, `1.0 + 2.0`, `1y * 2y`, `5L >= 5L`, etc. all just work. Operands must be the same numeric type, mixing types gives a `NumericOperationOnIncompatibleTypes` error
 
- covers unary `-` (negate) and `^` (power), which were Int64 only before too, so -x and a ^ b now work for every integer type, not just Int64. And / does integer division on ints now (truncates toward zero)

- Integer overflow now raises instead of silently wrapping: `127y + 1y`,` Int64.MinValue negated`, `2 ^ 64`, etc. all give an out-of-range error.

### DIFF
--- a/backend/src/Builtins/Builtins.Pure/Libs/Int32.fs
+++ b/backend/src/Builtins/Builtins.Pure/Libs/Int32.fs
@@ -98,7 +98,11 @@ let fns () : List<BuiltInFn> =
       description = "Adds two 32-bit signed integers together"
       fn =
         (function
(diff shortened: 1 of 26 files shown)


StachuDotNet:
Lovely -- and a long time coming!

PR #5673, opened 2026-06-09
dark apps — a registry + lifecycle for Dark apps and daemons

StachuDotNet:
One place to see, curate, run, and manage the Dark apps an instance cares about — interactive tools and
background daemons. Each instance keeps its own curated list (so people sharing a codebase have different
working sets), and you can see at a glance what's running.

## The surface

```
$ dark apps --help
dark apps
  See, curate, run, and manage the Dark apps this instance cares about.

  apps                      Interactive browser: review the catalog, install, start/stop
  apps list                 Apps you've installed (empty by default)
  apps list-available       The full catalog (+ which you've installed)
  apps add <slug>           Install an app (foreground apps get a shell shim)
  apps remove <slug>        Uninstall it
  apps inspect <slug>       Detail: target, installed?, status + logs

  Foreground apps
  apps run <slug>           Run attached in your terminal

  Daemons
  apps start <slug>         Launch detached (alias: spawn)
  apps stop <slug>          Stop a running daemon
  apps status [slug]        Running / stopped / stale (+ pid)
  apps logs <slug>          Recent log lines
  apps enable <slug>        Auto-start on login (OS service)
  apps disable <slug>       Remove its auto-start
```

The no-arg **interactive browser** does it all on one screen. The catalog is grouped into **Daemons** and
**Foreground apps**; each daemon shows a colored status dot (green ● running · yellow ● stale · dim ○
stopped) plus its live state and bound port inline (`running (pid …) :9090`), and rows truncate to one
line. Enter is the primary action — it **runs** the selected foreground app (handing it the screen) or
**toggles** the selected daemon; `a` add, `x` remove, `i` detail (slug + target, and a daemon's recent
logs), `esc` exits.

## Model

An **App** is a launchable Dark entrypoint; its `target` says how it launches:

```
/// How an app launches, and the reference it carries — one axis, no invalid states.
type Target =
  | Foreground of command: String    // a CLI command, dispatched in-process — run attached
  | Daemon of entrypoint: String     // a fully-qualified fn, eval'd detached — start/stop/status/logs

type App =
  { name: String          // pretty label, e.g. "Dark Packages HTTP Server"
    slug: String          // kebab id used by `apps <verb>`, the pidfile, the shim, config keys
    description: String
    target: Target }
```

Daemons are managed (`start`/`stop`/`status`/`logs`/`inspect`/`enable`); foreground apps are `run`. Two
lists: the **catalog** (a hardcoded seed plus any `App` value discovered in packages by type) and the
per-instance **installed** list (slugs in `cli-config.json`, empty until you `add`).

## How it works

- **`Darklang.Stdlib.Cli.Daemon`** (generic, reusable): a daemon is a detached `dark` process tracked by
  a pidfile at `~/.darklang/run/<slug>.{pid,log}`. `start` launches it via the running exe
  (`getCurrentExecutablePath`, no env var) and verifies it came up before reporting the pid; `stop`
  SIGTERMs; liveness is signal-0; `tailLog` reads the log. Pure POSIX, no new builtins.
- **`Darklang.Cli.Apps`** (`cli/apps/`): `model` / `registry` / `install` / `service` / `examples` /
  `servers` / `command` / `editor`. Foreground `run` dispatches the command in-process; a daemon `start`
  evals `<entrypoint> "<slug>"` detached. Daemons take their config (interval, port, …) from
  `cli-config.json` keyed `apps.<slug>.<key>`. HTTP-server daemons and `dark serve` both go through
  `Stdlib.HttpServer.serve`, configured by `HttpServer.Config.Config` (build with `Config.defaults port`,
  override any field via record-update); it blocks until SIGTERM.
- **`Darklang.Stdlib.Cli.Shim`** (generic, reusable — sibling to `Daemon`): installs a command shim, a
  tiny executable in `~/.darklang/bin/<name>` that re-execs the running `dark` binary with a fixed command
  (`exec <exe> <command> "$@"`), so `<name>` runs from the shell. `Cli.Apps.Install` is a thin wrapper —
  the OS mechanics (shim script, bin dir, exec perms) live in the stdlib layer, not the app.
- **`Darklang.Cli.Component`**: a small generic adapter that turns a self-contained component (a `render`
  + `onKey` over some state) into a full-screen app — no per-component SubApp boilerplate. The `text-editor`
  example uses it to surface the outliner's `TextEditor` widget standalone in a few lines. (Pass *named*
  fns, not inline lambdas — lambda values don't currently survive being stored in `AppState` and
  re-invoked across loop iterations.)
- **SubApp hand-off**: a new `SubAppAction.LaunchCommand` lets one full-screen app replace itself with
  another (the browser uses it so Enter on a foreground app launches that app in place).

## Auto-start & platform

`apps enable <slug>` registers a daemon as an OS service so it relaunches on login/boot — a **systemd user
unit** on Linux, a **launchd agent** on macOS; `apps disable` removes it. Without `enable`, an installed
daemon runs only while an `apps start` process is alive (`nohup` survives logout, not a reboot).

The daemon layer is plain POSIX (`nohup`, `kill -0`, `SIGTERM`, pidfiles, `getCurrentExecutablePath`) —
identical on Linux and macOS, no per-OS branching. On a platform without it (Windows), the daemon
commands report that cleanly instead of failing.

## Unrelated progress

Wins that rode along while building this:

- **One-shot TUIs actually run now.** `dark apps` / `dark outliner` / etc. were silent no-ops as one-shot
  commands — they set up a SubApp page and exited. They now hand off to the interactive loop and return to
  the shell on exit.
- **`run-cli` no longer breaks TUIs.** It piped through `tee` (making stdout a non-tty), so full-screen
  apps couldn't render; it now skips the `tee` when stdout is a tty.
- **`dark serve` goes through `Stdlib.HttpServer.serve`**, not `Builtin.httpServerServe` directly.
- **Config-write bug fixed.** `Dict.set` throws on an existing key; the config writer used it, so
  `dark config set <existing-key>` (and the installed-apps list) silently failed — switched to
  `setOverridingDuplicates`.
- **Divide-by-zero guard in the main loop.** `runInteractiveLoop` divided by `Terminal.getWidth()`,
  which can return 0 at startup and crash any TUI — now falls back to 80.
- **TUIs redraw in place.** A couple of full-screen renders emitted the clear/home sequence without the
  real `ESC` byte, so frames stacked and scrolled instead of clearing — fixed to a true kill-and-fill.
- **Daemons start on a fresh machine.** `Stdlib.Cli.Daemon.launchDetached` relied on `~/.darklang/run`
  already existing (the detached log redirect writes there); it now creates the run dir before launching.
- **HTTP daemons default to forwarded, host-reachable ports.** Defaults moved into the devcontainer's
  reserved serve range (`dark-packages`→9090, `demo-http`→9091), the port dropped from descriptions, and
  the devcontainer (`devcontainer.json` + `scripts/builder`) now publishes 9090–9099 so daemons are
  reachable from the host — not only under VS Code's port-forwarding. The live port shows in the browser
  row / `status` / `inspect` (read from the daemon's actual log line).


StachuDotNet:
I'm gonna merge to help make some of my other work easier, but would appreciate an async set of feedback at your convienience. Happy to follow up in another PR

StachuDotNet:
accidentally hit close instead of merge :)

OceanOak:
I like where this is going. Having one app surface for both foreground apps and daemons feels pretty clean.

Some observations:

1. I added an app by creating a .dark file declaring an App value with slug myApp. It correctly shows up in apps list-available but apps add mine rejects it with "no app 'myApp'". 
The reason is that add resolves the slug through findBySlug, which only searches the six hardcoded available () apps instead of the full catalog, so right now you can only add those hardcoded ones and any discovered app like myApp is unaddable.

---
2. `config set apps.dark-packages.port 9095`
Set apps.dark-packages.port = 9095
nice to show a url?

---
3. `apps` command
some apps have a checkmark, some a green dot, and some an outlined dot, but there's no indication of what those states mean. After using it for a while you can figure it out, but it may not be obvious to new users

---
4. Shell-injection through a slug
apps disable <slug> interpolates the unvalidated slug straight into a /bin/sh -c command, so any slug containing shell metacharacters executes arbitrary commands 

example:
 ```
  ./scripts/run-cli apps disable "zz; touch /tmp/dark_inj_test; #"
  ✓ disabled auto-start for 'zz; touch /tmp/dark_inj_test; #'
  ls -l /tmp/dark_inj_test
  -rw-r--r-- 1 dark dark 0 Jun 10 09:24 /tmp/dark_inj_test
  ```

---
6. clicking tab after `apps add` shows:
```
list
list-available
add
remove
run
start
spawn
stop
status
logs
```
I think this is the wrong completion

---
7. running `apps list` command
we display:
`No apps installed. `apps list-available` shows the catalog; `apps add <slug>` adds one.`

apps installed? are we really installing apps?

PR #5682, opened 2026-06-25
Preserve the declared type name in mismatch errors

OceanOak:
This PR fixes type-mismatch errors that showed the wrong expected type name. The expected type was rendered from its ValueType, which keeps only the package content hash; for structurally-identical package types, such as the `Int*.ParseError` family, that could display a canonical sibling instead of the referenced package type.

The fix carries the resolved package location from PT into RT custom type references and stores it on `FnParameterNotExpectedType /FnResultNotExpectedType`. The pretty-printer uses that location when rendering the expected type, while preserving the normal ValueType rendering for type arguments and fallback cases.

To test it:
`./scripts/run-cli fn '/Darklang.Stdlib.Test.ppErrTest' '(e: Stdlib.Int16.ParseError): String = "ok"'`
`./scripts/run-cli eval 'Stdlib.Test.ppErrTest 5L'`

Before: `expects Darklang.Stdlib.Int128.ParseError, but got Int64 (5)`
After:  `expects Stdlib.Int16.ParseError, but got Int64 (5)`

It also fixes the same problem for function names.
To test it:
`./scripts/run-cli fn '/Darklang.Stdlib.Int128.ppErrTest' '(e: Stdlib.Int16.ParseError): String = "ok"'`
`./scripts/run-cli fn '/Darklang.Stdlib.Test.ppErrTest' '(e: Stdlib.Int16.ParseError): String = "ok"'`
`./scripts/run-cli eval 'Stdlib.Test.ppErrTest 5L'`
(identical bodies → same hash, so the two names collapse to one function)

Before: `Darklang.Stdlib.Int128.ppErrTest's 1st parameter ...`
After: `Darklang.Stdlib.Test.ppErrTest's 1st parameter ...`

### DIFF
--- a/backend/src/LibExecution/RuntimeTypes.fs
+++ b/backend/src/LibExecution/RuntimeTypes.fs
@@ -597,6 +597,7 @@ type Instruction =
   | CreateRecord of
     createTo : Register *
     typeName : FQTypeName.FQTypeName *
+    typeReferenceName : List<string> *
     typeArgs : List<TypeReference> *
     fields : List<string * Register>
 
@@ -614,6 +615,7 @@ type Instruction =
   | CreateEnum of
     createTo : Register *
     typeName : FQTypeName.FQTypeName *
+    typeReferenceName : List<string> *
     typeArgs : List<TypeReference> *
     caseName : string *
     fields : List<Register>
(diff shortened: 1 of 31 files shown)


backend/src/LibExecution/RuntimeTypes.fs
```diff
@@ -116,12 +116,24 @@ type NameResolutionError =
...
 type NameResolution<'a> =
-  { originalName : List<string>; resolved : Result<'a, NameResolutionError> }
+  {
+    originalName : List<string>
+    resolved : Result<'a, NameResolutionError>
+    /// The resolved item's location, when known (only meaningful when
+    /// `resolved` is `Ok`). Preserved from PT to disambiguate the rendered
+    /// name of structurally-identical siblings.
+    location : Option<PackageLocation>
+  }
```
StachuDotNet: I think "only meaningful when `resolved` is `Ok`" indicates that a more natural shape might be:
```fsharp
{
    originalName : List<string>
    resolved : Result<('a * PackageLocation), NameResolutionError>
  }
```
What do you think?

PR #5683, opened 2026-07-04
Replace tree-sitter and FCS with a single hand-written parser

OceanOak:
Darklang used to parse code two different ways: a tree-sitter grammar for tooling, and F# Compiler Services for execution, kept in sync by hand. This PR throws both out and replaces them with a single hand-written recursive-descent parser. Everything now parses through one path.

#### The new parser:
- LibParser/Lexer.fs, a tokenizer.
- LibParser/Parser.fs, the recursive-descent parser. It recovers from errors and emits WrittenTypes with full source ranges (keyword and symbol spans are kept, so highlighting and hover still work).
- LibParser/WrittenTypes.fs, one written-types tree.
- LibParser/GRAMMAR.md, the grammar written down.

It runs in two stages: the lexer turns source into tokens, then the recursive-descent parser builds the WrittenTypes tree. It recovers from parse errors (so tooling still gets a partial tree plus diagnostics instead of nothing). See GRAMMAR.md for the full grammar.

 
#### What got deleted
- The whole tree-sitter-darklang/ grammar, generated tables, native .so, and its test corpus.
- The FCS bridge (FSharpToWrittenTypes.fs) plus the F# compiler services parse path.
- The old Dark-side parser.
- The FSharp.Compiler.Service nuget dependency and the LibTreeSitter project.

#### Name resolution
- NameResolver is now branch-aware.

#### CLI 
- CLI script and expression parsing run through the new parser end to end and return precise diagnostics instead of the old __parseError__ magic string.

#### LSP and syntax highlighting
The parser is F#, but the highlighter and language server are written in Dark. The bridge between them is two builtins that surface the parser's output as Dark values:
- parserParseToWrittenTypes, the full range-complete WrittenTypes tree. Every node carries its exact source range, which is what highlighting and hover run on. Almost everything calls this one: CLI highlighting, the CLI package editor (viewing and creating fns, types, values), the execution parse path, and the LSP all call it.
- parserParseDiagnostics, the parse errors as (range, message) pairs. This drives error reporting on both surfaces: the CLI flags parse errors when you create a package item, and the LSP turns them into red squiggles.

#### Syntax cleanups
- Dropped the `I` suffix on int literals and the _v0 suffix at call sites.
- Statement sequences now lower to EStatement instead of a let _ = ... wildcard.

#### Testing
- Corpus gate: every .dark package file has to parse cleanly, so any regression fails the build.
- Structural, fuzz, and range-invariant tests over the parser output and diagnostics.
- Round-trip tests: source to PT and back to pretty-printed source.
- Differential parity gate: There's one WrittenTypes to ProgramTypes lowering per side (F# for execution, Dark for tooling) with a differential test that fails if the two ever disagree.


Dockerfile
```diff
@@ -113,9 +113,8 @@ RUN DEBIAN_FRONTEND=noninteractive \
       libstdc++6 \
       zlib1g \
       # end .NET dependencies
-      # parser (tree-sitter) dependencies
+      # for compiling native dependencies
       build-essential \
```
StachuDotNet: Do we still need build-essential? I suspect something related to sqlite embedding will fail if we remove it, but worth an attempt at removal, as iirc it's a somewhat heavy dependency.

backend/paket.dependencies
```diff
@@ -17,7 +17,6 @@ nuget Fumble = 1.0.3
 nuget SQLitePCLRaw.bundle_e_sqlite3 = 2.1.11
 
 // Parser
```
StachuDotNet:
```suggestion
```
(nitpick, remove section)

Dockerfile
OceanOak: Yes, it is still used, but only by the native CLI build (`./scripts/build/build-release-cli-exes.sh --aot`). We can replace it with GCC, though, since it's smaller

backend/src/Prelude/NEList.fs
```diff
@@ -76,6 +76,12 @@ let ofSeq (head : 'a) (seq : seq<'a>) : NEList<'a> =
 
 let singleton (head : 'a) : NEList<'a> = { head = head; tail = [] }
 
+/// A `List` to an `NEList`, falling back to `[dflt]` when the list is empty.
+let ofListWithDefault (dflt : 'a) (l : List<'a>) : NEList<'a> =
```
StachuDotNet, line 80: FYI this is unused

StachuDotNet (approved):
Great. I'm going to merge as-is, but a few quick thoughts/comments:
- looking forward to adjusting our syntax as we've discussed
- I suspect some of Builtins.CliHost/Libs/Cli.fs's `declarationsToModule` might be better in LibParser or something, but not strongly held

backend/src/Prelude/NEList.fs
OceanOak: It is used in three places, all in WrittenTypes.fs (lines 645, 653, 677)

PR #5685, opened 2026-07-13
Sync Packages Across Dark Instances

StachuDotNet:
# Sync packages across Dark instances

*Use Dark across all your devices: author a function on the laptop, it shows up on the desktop — the instances
agree on the same code, no central server.*

## Summary

Adds **package sync**: run more than one Dark instance — two of your own machines, or you and a collaborator —
connect them, and branch structure, package edits, and conflict resolutions flow between them over HTTP,
converging to the same state everywhere. Opt-in, and once set up automatic; local and offline stay the default.

It's built on storage we already have. Darklang stores packages as an **op log** (`package_ops`), and every
queryable table (functions, types, values) is a **projection** folded from that log. So sync is just: a peer
serves its new ops, you append them, and they fold in exactly like your own edits. No merge engine, no separate
wire format — the op log *is* what travels.

Two scoping calls worth flagging: sync is **wholesale, not per-branch** (a `private`/`follow` control leaked
through the content-addressed blob channel, so every branch syncs and proper scoped sync is a future PR); and
transport is **HTTP-only** (a peer is a running instance, not a `.db` file).

## At a glance

- **The commands.** `dark sync` pulls your peers' new ops over HTTP; `dark sync serve` lets them pull yours;
  `dark sync daemon` runs both on a timer. Content-addressed ops make a re-pull a no-op; an `origin_ts`
  last-writer-wins stamp makes concurrent edits converge regardless of arrival order.
- **The op log, adjusted for sync.** Each op now carries `origin_ts` (its author's stamp, kept across sync);
  `package_ops`'s key becomes `(id, branch_id)`, fixing a ghost-function bug. Projections are unchanged.
- **`resolutions` — a new synced log.** Two instances editing the same name auto-resolve by last-writer-wins,
  but it's *recorded*, never silent; overriding it mints a `Resolution` that syncs, so every peer converges on
  your pick.
- **Branches sync wholesale + consistently** — a branch `whatever` is the same id + name on every instance.
- **Meaning-stable hashing** — an op id hashes a function's *meaning* (binder names are alpha-normalized), so
  ids survive an op-format change and content-addressing keeps working.
- **Synced values self-heal.** A value arrives from a peer *unevaluated* (`rt_dval` is NULL until it's run);
  the store now evaluates such values on the next startup, and the read is null-safe — so a not-yet-evaluated
  synced value returns cleanly instead of crashing a command with an internal NULL.
- **Perf, bundled (general wins).** `List.append` O(n²)→O(n) (every large `List.map` in Dark was quadratic);
  native log reads took a 500-op read from 57s to ~1s.

---

### Quick start

Two machines on a tailnet, syncing in under a minute — author a value on one, watch it show up on the other:

```text
# feriel is reachable:
feriel$ dark sync serve                       # serving on :9000 (foreground)

# you connect to feriel, then turn on automatic sync:
you$ dark sync connect http://feriel.tailnet:9000
you$ dark sync daemon start                  # first run: a short settings dialog, then pull+serve in background

# feriel authors and commits:
feriel$ dark val Feriel.Demo.greeting = 99
feriel$ dark commit "add Demo.greeting"

# …within the poll interval it lands on yours, attributed to Feriel — no manual step:
you$ dark view Feriel.Demo.greeting           # → val greeting = 99
```

The whole surface (`dark sync` with no args syncs now; `dark sync help` prints this):

```text
dark sync — keep your packages in step with your other instances, over HTTP.

  dark sync                    sync now from every instance you're connected to
  dark sync connect <url>      start syncing with an instance (e.g. http://feriel.tailnet:9000)
  dark sync disconnect <url>   stop syncing with an instance
  dark sync serve              let peers pull from you (add --port N; default 9000)
  dark sync status             who you're syncing with, and how current each is
  dark sync fetch <url>        sync once from a single instance

Automatic (background):
  dark sync daemon start       pull from + serve to your peers on an interval
  dark sync daemon status      whether it's running, and with what settings
  dark sync daemon config      view/adjust interval, port, mode, on-boot
  dark sync daemon stop        turn automatic sync off

When two instances edit the same name, sync auto-resolves (latest wins) but records it:
  dark conflicts               review + override sync divergences
```

### How it works

```text
   Instance A                       the wire (HTTP, JSON)                    Instance B
 +-------------+   GET /sync/events?package=&branch=&resolution=   +-------------+
 |   op log    | ------------------------------------------------> |   op log    |
 | packageOps  |     one bounded batch per log, since a cursor     | append,     |
 | branchOps   |                                                   | then FOLD   |
 | resolutions | <------------------------------------------------ | into the    |
 |  -> FOLD    |          (B serves its own new ops back)          | projections |
 | projections |                                                   | = CONVERGED |
 +-------------+                                                   +-------------+
```

Read a peer's new events since your cursor, append them to your log, where they fold in exactly like your own
edits (same F# playback a local edit uses). Two properties make it converge:

- **Idempotent** — ops are content-addressed, so a re-pull inserts nothing and folds nothing.
- **Order-independent** — the fold resolves each name by `origin_ts` (last-writer-wins), so concurrent edits
  reach the same state everywhere regardless of arrival order.

The one subtle bit: a received op keeps its *author's* `origin_ts` (a fresh local stamp would be wrong for
replication), and on receive I reconcile that stamp to the earliest (MIN) across instances — so every store
ends up with one canonical timestamp per op, and LWW picks the same winner everywhere.

### Connecting, pulling, serving

```text
$ dark sync connect http://feriel.tailnet:9000
Now syncing with http://feriel.tailnet:9000 — reachable (10843 ops)   # probes /sync/health up front
$ dark sync
Pulled from 1 instance: new value Demo.Sync.hello
$ dark sync
Already up to date with 1 instance
$ dark sync
Pulled from 1 of 2 instances: new fn Demo.Lib.greet   # the other peer was offline; it says so above this
```

A pull says what it *did*, not a raw op count (authoring one value emits two ops, so a count would lie). It
names the changes when there are a few, counts them by action when there are many. `connect` probes the peer
right away, so a dead port or typo'd URL is caught there, not later on a silent pull. And `dark sync serve`
prints the address peers actually use and only claims "Listening" once the port is really bound.

### Branches

Branch structure is just another synced log: a peer's `branchOps` (create/commit/merge/rebase) travel *before*
its package ops, so an incoming op's `branch_id` always resolves. Every branch syncs **wholesale** — `whatever`
is the same id + name everywhere, never scoped by peer or account.

```text
$ dark sync status
Syncing with 2 instances (run `dark sync` to pull the latest):
  feriel   http://feriel.tailnet:9000   (pulled)
  framey   http://framey.tailnet:9000   (not pulled yet)
  3 branches: main, feature-parse, experiment
```

Each peer shows a name derived from its URL (`feriel`, or the whole host for a bare IP), so you deal in
`feriel`, not the URL — `--from feriel` / `fetch feriel` / `disconnect feriel` all take it. Concurrent rebases
of the same branch converge (LWW on a synced `origin_ts`), and `dark rebase --status` tells you plainly whether
a branch is up to date, a clean rebase behind, or blocked by a real name conflict.

### Conflicts

Two instances can bind the **same name** to different content at once. LWW picks a deterministic winner — but
dropping the other edit silently isn't OK, so every such divergence is *recorded* in `sync_conflicts` (a local
review log, never synced) with both candidates and the winner.

```text
$ dark conflicts show Feriel.Util.parse
Feriel.Util.parse — two versions (fn):

  yours   4d8e1f0a
      let parse (s: String): Int = Stdlib.Int.parse s |> unwrapOr 0
  theirs  9f3a1c2b  (active — auto:last-writer-wins)
      let parse (s: String): Int = Stdlib.String.trim s |> Stdlib.Int.parse |> unwrapOr -1

$ dark conflicts keep-mine Feriel.Util.parse
Kept yours (4d8e1f0a) at Feriel.Util.parse. This resolution will sync to your peers.
```

`show` renders each side's *actual source*, so keep-mine vs keep-theirs is an informed call. `ok` accepts the
auto pick; `keep-mine`/`keep-theirs` override it and sync that decision so every peer converges on it. (An
override is a thin `Resolution` overlay, applied after the fold — not a new op, which would collide on content
hash.)

### The daemon

`dark sync daemon start | stop | status | config` — automatic sync as one toggle over both the pull and serve
sides. First `start` runs a short dialog (EOF-safe: piped/no-TTY falls through to defaults); later starts are a
one-liner.

```text
$ dark sync daemon status
Automatic sync:
  pull    sync: running (pid 4821)
  serve   sync-server: running (pid 4822) :9000
  mode: pull+serve, every 30s, port 9000, on boot, settings: dark sync daemon config
```

---

## Internals (the short version)

The F# core stays small; most of the surface is `.dark`.

- **The op logs are plain SQLite tables.** `package_ops`/`branch_ops` already existed; `resolutions` is new. No
  bespoke "event-log" type — I tried one and a perf spike killed it (`Stdlib.Sqlite` already reads the tables
  into Dark at ~native speed, so the eventual Dark-managed fold is "plain tables + Sqlite"). Reads are native
  builtins so serving thousands of ops is milliseconds.
- **One convergence rule.** `Lww.isStale` is the single last-writer-wins predicate the fold, conflict
  detection, and the resolution overlay all share, so they can't drift apart.
- **One name, one item.** A location is `(owner, modules, name)` — no kind; `item_type` is a lookup hint, not
  identity. The fold used to key on it, so a fn bound over a value left both live. Everything keys by name now;
  authoring refuses a second kind at a name (the sync fold still replaces — it has to converge). This retired
  `dark conflicts … --kind`.
- **Meaning-stable hashing.** An op id hashes a canonical form with binders alpha-normalized to `$0`,`$1`,…, so
  two functions identical up to a binder rename hash the same and ids survive an op-format re-serialize.
- **Trace-detail now defaults OFF** (and OFF is a true no-op). Traces have no GC yet and a single row can hit
  ~1 GB, so a long-running `serve`/daemon would fill the disk. `config/dev` + CI still opt in with
  `DARK_CONFIG_TRACE_DETAIL=on`; the shipped CLI gets the safe default.
- **One unrelated fix, in its own commit.** `main` currently calls `Stdlib.List.collect`, which doesn't exist,
  so `dark view` errors on plain `main`; the last commit swaps it for `List.map |> List.flatten`. Happy to
  split it into its own PR.
- **Tests:** three sync layers — op-replay convergence (`SyncScenarios`), true two-store in-process sync
  (`MultiInstance`, incl. a `dark conflicts` render regression), and a flag-gated real-process-over-HTTP tier
  (`SyncE2E`, `DARK_E2E=1`) — plus migration + meaning-stable-hashing unit coverage.

## Not in this PR (known limits, none blocking a local-first v1)

- **Trust model.** Connecting a peer is a full trust boundary for now — no op-integrity check, synced values
  auto-evaluate, `serve` is unauthenticated. Fine for your own machines / a trusted collaborator on a private
  tailnet; hardening (verify op ids, authenticated serving, localhost-bind) is the main thing before untrusted
  peers. (Already done here: the `sqlite*` builtins are capability-gated, and synced blobs are re-hashed.)
- **Two structural edges** (documented, not blocking): a merge racing a same-branch commit can diverge `main`
  (the merge op doesn't record *what* it merged); and two independently-created same-name branches jam sync
  (same name, different ids → an FK abort wedges the cursor until reset). Both want the per-op-isolation +
  quarantine work — a focused follow-up.
- **A re-version reads as a "conflict" on peers that were just behind.** Conflict detection has no ancestry
  check, so a plain catch-up looks like a divergence. Data still converges; it's the framing. Needs causal info
  the op log doesn't carry yet.
- **Deprecation is content-keyed, not name-keyed** — deprecating a value also deprecates other names holding
  the same content (a guard now warns you first). Per-name deprecation is the follow-up.
- **Sync is wholesale** — every branch's ops travel. Fine for a personal tailnet; scoped sync is a future PR.


StachuDotNet:
Still more to tighten, but ready for some initial review.

OceanOak:
Really excited to see this coming together. We've been waiting for sync for a while!

A few sync UX questions and a couple of bugs I hit while testing.

1. When sync prints:
Pulled 5 changes from 1 instance
what exactly did we pull?  A short breakdown would be nice

2. Also, if the sync/pull detects conflicts, we should mention it, something like:
`Pulled 5 changes; 2 conflicts need review. Run 'dark conflicts'`

3. Are there any risks to LWW? it can feel like silent data loss unless sync clearly reports the conflict and winner.

4. We should have more control over sync
   - Why does sync pull all branches by default?
       maybe we can do something like:
       dark sync               # current branch
       dark sync --all         # every branch
   - We shouldn't automatically sync every account’s unmerged branches
   - Maybe the default should be the current account's branch plus anything merged into main

5. the term pull could get confusing since this isn't equivalent to git pull, I think calling it "pull" may set the wrong expectation

6. maybe we should sync WIP as well

7. Bug: Same location, two item kinds
Resolving one conflict hides both when a function and value share the same location.
to reproduce:
we have Machine A and B
```
A
dark fn  SyncTest.Kind.clash "(x: Int): Int = x + 1"
dark val SyncTest.Kind.clash = 10
dark commit "A fn+value clash" --yes

B  
dark fn  SyncTest.Kind.clash "(x: Int): Int = x + 2"
dark val SyncTest.Kind.clash = 20
dark commit "B fn+value clash" --yes

A 
dark sync
B
dark sync

A
dark conflicts                        # Two lines: "SyncTest.Kind.clash (fn)" and "(value)"
dark conflicts keep-mine SyncTest.Kind.clash   # you can only name the location, not the kind
dark conflicts                        # Bug: both gone. one resolve marked both kinds overridden
```
I think resolution commands need either the item kind or conflict ID:
dark conflicts keep-mine SyncTest.Kind.clash --kind fn
If the location is ambiguous, the current command should refuse and ask which conflict to resolve.


8. `dark ops` There are two ops for each dark val. One stores the value and one binds the name, but the output doesn't explain that. Could dark ops show the op kind so this is obvious?
```
# A
dark val SyncTest.Update.test = 10
dark commit "before update" --yes

dark val SyncTest.Update.test = 11
dark commit "after update" --yes

# B
dark sync
dark ops

Output:

10663 ops in this instance's log — counted + read from SQLite via Stdlib.Sqlite in 4ms
  id        branch    commit/WIP  size   message / origin_ts
  dd6c5458  89282547  40554d4e  97B  after update
  0d3bd8ba  89282547  40554d4e  87B  after update
  60340e02  89282547  8dcf62ea  97B  before update
  561c7cc7  89282547  8dcf62ea  87B  before update
```

9. Rebase conflicts can leave a branch permanently stuck
Once the same name is committed on a branch and its parent, rebase refuses the branch, but there’s no command that can resolve the conflict.
dark conflicts handles sync conflicts, not rebase conflicts. discard only clears WIP, and merge requires the rebase first. The CLI says to fix the conflicts on the branch, but there doesn't appear to be a way to do that.
```
# Create a branch and commit a value
dark branch create demo
DARK_BRANCH=demo dark val MyTest.Demo.x = 1
DARK_BRANCH=demo dark commit "branch" --yes

# Commit the same name on main
dark val MyTest.Demo.x = 2
dark commit "main" --yes

# Rebase fails
DARK_BRANCH=demo dark rebase
# Rebase failed - conflicts found: MyTest.Demo.x

DARK_BRANCH=demo dark merge
# Must rebase first.
```
This leaves the branch stuck forever. We need a rebase-resolution flow: maybe keep-mine, keep-parent, or an explicit edit/remove workflow that the next rebase recognizes.

10. Related: rebase --status currently reports conflicts, but not whether the branch actually needs rebasing. Should it report both?

11. Unrelated to sync (maybe we can add cleanups for these)

- We don't a command that shows which account is currently logged in. It would be useful when commits and sync involve multiple authors.

- This should fail because demo starts lowercase
    `dark val MyTest.demo.x = 1`

- In MyTest.Conf.a, MyTest is treated as the owner. I don't think users should be able to create definitions under arbitrary owners, we should only be able to add definitions to the currently logged-in account's namespace.

PR #5686, opened 2026-07-13
Parser improvements

OceanOak:
This PR:

- Added a shared validation gate between parsing and WT2PT.
  - Script, package, and test files share one syntax parse.
  - File-purpose rules are applied afterwards.
  - Execution paths only accept a ValidatedSourceFile.

- Migrated value declarations to val syntax
- Kept let for functions and local bindings.
- Fixed arguments and locals shadowing package names.
- Improved interpolation scanning, error recovery, and source ranges.

- Removed parser recovery nodes from ProgramTypes.


backend/src/LibSerialization/Binary/Serializers/PT/Expr.fs
```diff
@@ -765,8 +762,8 @@ module Expr =
       let id = r.ReadUInt64()
       ESelf id
     | 36uy ->
-      let id = r.ReadUInt64()
-      EError id
+      raiseFormatError
+        "ProgramTypes Expr tag 36 is a legacy parse-recovery hole and cannot be loaded"
```
StachuDotNet, line 766: don't care about anything 'legacy' probably OK to remove code?

StachuDotNet (commented):
Great. Love seeing EError gone (so soon!). one small note, nonblocking

PR #5692, opened 2026-07-27
Migrate SCM+Sync into package code, harden it, stand up an internal relay

StachuDotNet:
## Ownership

Sync, branches, conflicts, review and propagation were F#. They're Dark now. I deleted the F# rather than
porting it -- `LibDB/{Merge,Rebase,Conflicts,Lww,Resolutions,ChangeSummary,BranchOpPlayback}.fs` and the
matching builtins -- because keeping both would leave two places deciding what a merge means.

`packages/darklang/scm/` is a silo per table, each owning every statement against its own. F# keeps the
fold, authoring, the merge mechanism, the cascade, and schema changes to canonical tables. The rule I held
to: decisions in Dark, mechanism in F#. `canMerge` is Dark; flipping the frontier is F#.

## A branch is an overlay now

`BranchOp` and its playback are gone, as are `package_ops.branch_id` and the `branch_ops` table. A branch's
ops sit in `package_ops` at `effective = 0`, tagged in `op_branches`. It used to be a column value; it's a
set of ops layered over main.

`effective` splitting stored from applied is what lets a branch, a review queue and a relay be one
mechanism instead of three.

The cost, and I'd flag it in review: `locations` lost its `branch_id`, so a direct read answers about main
while you're on a branch, and answers plausibly. The schema used to protect you there and now discipline
does. There's a test that fails if any Dark file outside the silos reads it.

## Decisions travel

New op cases: `Deprecate`, `Undeprecate`, `Resolve`, `Decide`, `BranchEvent`. `PropagateUpdate` and
`RevertPropagation` are gone.

That's the substance behind "decisions travel". A resolution was a row in `resolutions`, a conflict a row
in `sync_conflicts`, and rows stay on one machine. As ops, a conflict resolution, a propagation pin, a
constraint ack and a merge all sync and hold.

`Resolve` has to be its own case because ops are content-addressed: re-authoring `SetName(name -> old hash)`
produces an op that already exists, dedups, and folds nothing. There's no way to say "the same binding, but
now I mean it" as a plain op.

`SetName` gained `previous`, the hash it replaced. Without it, "we both invented this name" and "I edited
your version" look identical, and only the first is a conflict.

## Commands

Newly at top level: `push` `pull` `connect` `whoami` `identity` `switch` `branches` `diff` `log` `resolve`
`propagate` `constraints` `ack` `undeprecate`. Nothing removed.

Less new than that list looks. Propagation already ran on every edit -- `cli/packages/propagate.dark` was
internal helpers; what's new is a policy you can set. `Constraints` was already named in `ProgramTypes.fs`;
this implements it. `sync` existed and now means push-then-pull against a relay.

Actually new: the relay, per-instance identity, and `resolve`/`ack` as durable syncable answers instead of
local state.

Branch names are aliases over uuids. You and I can each hold a `fix-auth`; pulling mine imports it as
`fix-auth~<shortid>`.

## Propagation

Cascade unchanged, policy is new. `propagate pin <name>` / `follow <name>` write a `Decide` op. Both are
retroactive -- by the time you pin, the repoint already happened, so pin undoes it and follow catches you
up. Most-specific-first (item, module, parent modules, owner); no row means follow.

Un-propagated usages are the steady state in a distributed store, not a backlog: a caller on your machine
can't be propagated to from mine, and mustn't silently repoint when we sync. So staleness is
`dark constraints`, not a line at commit that reappears forever.

**A pin travels.** You pinning a caller rolls it back on my machine too. That's the design working as
written, and it's the piece I'd want us to agree on before we lean on it.

## What it's like to use

Conflicts don't block. They're auto-resolved by LWW and recorded, so the loser is kept and reviewable, and
the record names the person rather than "yours"/"theirs".

Failure paths say what to do: a network failure tells you the work is local and safe, a read-only store or
full disk says which and where, a wrong relay secret reads differently from a missing one, and a command
given a target it can't find names the target. Tab completion covers the SCM commands, and anything that
answers a question answers `--json`.

## The relay

`packages/darklang/matter.dark` is a relay and a package browser, served by
`dark serve Darklang.Matter.router`, with the image and `fly.toml` under `deploy/`. It's deployed and
running for internal use. The address isn't in the repo: its read endpoints are unauthenticated, so treat
anything pushed there as public. Ask me for the host and the write secret.

It stores and hands back ops and never folds them into its own package set, which is the load-bearing part:
pushing code to a relay can't change what the relay runs.

The browser reads the relay's own store -- module tree, source, doc comments, and what-calls-this from the
fold's call graph, plus search. `/p?name=Darklang.Stdlib.List.map` is the one that shows it off.

Only the relay's own package set is browsable, since only it is folded. Showing the versions clients have
pushed needs a second projection folded for display.

Separately, and not in this PR: darklang.com's package pages were pointed at endpoints that no longer
exist. Fixed in that repo as one local unpushed commit, so nobody redoes it.

## Ready to use?

For the two of us plus the relay, yes for what's below, with one gap I'd close first.

Upgrade is newest-stamp-wins, not "local edits win". Edit a stdlib function today, upgrade to a build cut
tomorrow, and the build wins silently. `locations.op_id` records which op wrote each binding, so the fix is
available; it didn't block the relay, which never authors.

Also open: a branch rename doesn't travel, though merge and archive do. Hosted ops aren't browsable. Op ids
are a 128-bit truncation of the content hash.

## Testing

The suite covers the fold, the overlay, conflict detection and the CLI surface. Past that I ran it for real
against the deployed relay with fresh instances, which is how the worst bug turned up:

- two fresh peers, an edit on top of an edit resolving with no false conflict, and two divergent edits
  producing a real one that's auto-resolved, kept, and overridable back across
- a branch-name collision in both directions
- three machines on one branch through the relay, with the merge landing on main everywhere
- archive travelling and hitting the right branch, including one renamed on import
- interrupted mid-chunk push, relay restarted mid-pull, read-only store, full disk, unopenable store, wrong
  secret, no secret, dead relay, older client against the upgraded relay

The bugs I found were all silent -- no error, just a plausible wrong answer. That's the failure mode
content addressing makes easy and the thing worth looking for. The worst: a Release CLI embedded my working
store, so a fresh install carried the relay write secret in plaintext, claimed to be my machine, and went
quiet against a relay holding everything it was missing. `export-seed` already stripped all of that; only
the self-contained release script ran it. Fixed as wiring, the build now refuses without a seed, and I've
rotated the secret.

## Where to look

`LibDB/PackageOpPlayback.fs` for the fold, `LibDB/Branches.fs` for the overlay and the trap above,
`cli/sync.dark` with `scm/packageOps.dark`'s transport section for the wire, `scm/branches.dark` for what a
merge is allowed to do.

One thing I chose not to do: `scm/packageOps.dark` spans identity, decode, preview, reconciliation, draft,
wire and relay storage, and wants splitting. It's a rename across a lot of call sites and I didn't want to
land that unreviewed.


StachuDotNet:
splitting this PR up. Too big. Ignore :)

PR #5707, opened 2026-08-13
Add an at-rest type checker

OceanOak:
This PR adds a type checker that validates serialized packages (ProgramTypes) without executing them, then wires it through the whole workflow: save, commit, CLI, and editor.

 The checker
  - Every item gets one of three outcomes:
    - Checked (fully proven against a closed type environment)
    - Failed (at least one definite type error)
    - Incomplete (couldn't prove safety: missing deps, unresolved names, unsupported constructs)
  - Bidirectional type inference: the checker works in two directions. When it already knows what type an expression should be (e.g. a function parameter declared as Int), it checks the expression against that. When it doesn't (e.g. let x = ...), it infers the type from the expression itself.
  - Unification: when the checker walks the code, it often doesn't know a type yet, so it uses a placeholder, an "unknown" (type variable). The code itself generates equations between types, and unification is just solving those equations.
  - Alias expansion: the checker sees through aliases when comparing types, and detects aliases that point at themselves in a circle.
  - Generalization of immutable values: a helper like let id = fun x -> x gets a generic type, so it can be used with an Int in one place and a String in another without conflict.
  - Error recovery: when a pattern is invalid, its variables are still brought into scope, so the checker reports one error for the pattern itself, not a pile of "unknown variable" errors for everything that used them.
  - Signatures are declared before bodies are checked, so mutually recursive functions work regardless of storage order.

  Integration
  - On save: saving a package item runs the at-rest check via a new authoring adapter that loads only the candidate's transitive dependency closure.
  - On update: after propagation finishes, all visible transitive dependents are checked against the final graph. Multi-item updates produce one consolidated, location-aware report.
  - On commit: commit is blocked while the branch has items with definite type errors  (commit --allow-type-errors bypasses the check)
  - In the CLI: new typecheck / at-rest check commands
  - In the editor: LSP diagnostics now surface at-rest findings in-editor


StachuDotNet:
Hey -- I had a long review message drafted, and ended up having an agent address the feedback rather than dump a wall of thoughts here. About to merge, but please LMK if I've done anything negative :)

Notable changes:
- Moved some pretty-printing/diagnostics to package-land, and structured some data along the way. Benefit: instead of printing "Type #a1b2c3d4",  you get things like "In argument 3: expected Int, got Int64".
- `AtRestTypeChecker.fs` was really long, so I split it up.

Longer term I imaginea lot of this living in package-land rather than F#. The set of at-rest checks is fixed right now, and I think people should be able to pick which ones apply to their packages and write their own, which pretty much means they need to be ordinary Darklang.

PR #5720, opened 2026-08-24
Replaces the capability system with effects and permissions 

OceanOak:
This PR replaces the capability system with effects and permissions

### Permission model

  - Added runtime checks for HTTP, file, process, and other host effects.
  - Access is denied by default.
  - Checks match the actual URL, file path, port, or process being used.
  - When possible, permission errors include the exact permissions allow ... command needed.

### Effect ceilings

  - Functions can declare a ceiling, for example :{Http, Clock}.
  - A ceiling limits what the function and its callees may do; it never grants access.

### Access propagation

  - Function values and closures keep the permissions they had when created.
  - This also applies to functions nested inside lists, records, tuples, and other values.

### Package and function control

  - Function requirements include the permissions needed by their calls, lambdas, and deferred code
  - The functions that ship with Darklang are trusted by the CLI.
  - User code cannot write packages that pretend to belong to Darklang or have unfinished hashes

### Instance and run policies

  - Added machine-wide instance policies and per-run access.
  - New installations get a safe default policy.
  - Policy changes and package approvals are restricted to the host.
  - Host operations can optionally be written to an audit log.

### Host boundary

  - Builtins that touch the filesystem, HTTP, processes, POSIX, or similar resources use the checked host boundary.
  - Each builtin checks the exact path, URL, port, or process it is about to use.
  - Operations that cannot be safely scoped, such as arbitrary native database access, require the entire effect.
  
### Approvals and pins

  - Added approvals for packages and pins for logical function names.
  - Each approval applies to one immutable function version.
  - Revoking or changing an approval updates related dependencies safely.
  - Added permissions commands to view and manage approvals and pins.


StachuDotNet:
Great changes, appreciate the new model. I pushed one small commit: `docs/permissions-todos.md` now exists (AGENTS.md and effects.md referenced it but it wasn't in the tree; content is collected from what those docs already call deferred -- sanity-check it matches what you intended), effects.md names it, budget.json's note is back to one line (history.md carries the cost story), and a few inline notes, listed below.

remaining asks/thoughts, none blocking:

- the "remove tests that hang the suite" commit deletes the registry-driven bogus-arg sweep (`everyCommandRefusesABogusArgument`), and `missingTargetsAreNamed` doesn't cover `permissions`, so new commands aren't swept. was something hanging? I'd prefer the offender in `notSweepable` with a reason. same commit drops two control assertions in BranchOps.Tests; those look independently restorable
- caps profiles / TUI editors: is there a place for them in the new world? the editors maybe not (typed grammar + denial hint is a better loop imo), but the profiles were a nice on-ramp between deny-all and `allow all`
- Host/ inside LibExecution vs its own assembly below it: deliberate? fine either way, just curious
- fwiw I tried to get guest code to widen its own permissions and couldn't; the edges match the doc (interactive prompt, Process/Native, owner column). lmk if there's a path I missed
- TODOs and CLEANUPs I inlined:
  - HostTypes.fs: CLEANUP consider FailureKind.PermissionDenied -> OsAccessDenied (three different things share that name)
  - Cli.fs: doc comment on ExecutionError.Denied saying why it's not a RuntimeError case, + TODO on exposing the structured denial as a Dark value
  - CallGraph.fs: TODO consider specializing requirements when a resource arg is a literal (`file-read /that/path` at approve time instead of bare `file-read`)
  - Blob.fs: CLEANUP empty callEffects vs the persistent blob-store read -- deliberate ambient local storage? if so say it in effects.md
  - Interpreter.Tests.fs: CLEANUP split PermissionsGate (~810 lines) into its own file; happy to do that one myself after this lands

PR #5753, opened 2026-09-07
Migrate SCM and sync to Dark; stabilize op identity; branches as overlays; propagation without marker ops; commits that travel; a hosted relay

StachuDotNet:
Main already stores packages as an append-only log of content-addressed ops with projections folded from it. This PR keeps that idea, reworks nearly every mechanism on top of it, moves most SCM and sync deciding from F# into Dark package code, and deploys a hosted relay.

- **Op identity stabilized**: an op's id no longer depends on parse artifacts, so two reloads, or two machines, mint identical ids. The enabler for everything below. A doc comment is deliberately NOT part of that identity -- editing one repoints nobody -- so prose travels on an op of its own (`UpdateDoc`), scoped to the NAME rather than to the body.
- **Branches became overlays**: one flat log; a branch is a set of tags over it, not rows in side tables.
- **Propagation became a feature, not an op kind**: the marker ops are gone; the cascade writes ordinary ops at edit time; pin/follow are synced decisions; a constraints subsystem watches what propagation must not touch.
- **Sync went peer-to-peer -> relay**: committed ops travel with their commit rows, so history reads identically everywhere; the relay stores and forwards but never runs what it hosts.
- **The logic moved to Dark**: twenty-six `scm*` builtins (plus the conflict, dependency, blob and op-log sets) deleted; merge, rebase, conflicts, drafts, commits and propagation decide in `packages/darklang/scm/`; F# keeps the fold, hashing, serialization and the boot overlay.
- **Misc**: a full CLI SCM surface (`--json` contracts and exit codes for agents), the workbench absorbed the old review TUI, backups and store-health checks, the builtin allowlist 38 -> 0.

---

## The op log

- The turn this PR makes: a store held the current state; from here it holds the history that produced it. The op log IS the program; a branch is a view of that history, not a copy.
- Op ids were content hashes already, but an `Add*` hashed its serialization, parser node ids included, so every reload re-minted them and each binary upgrade grew every store by ~5,500 "new" ops. Now an `Add*` hashes the item's content hash; a test pins one op's hash to a literal so a hashing change forces a deliberate migration.
- The log went flat: one row per op ever (PK was `(id, branch_id)`), branch membership a tag in `op_branches`.
- One new column carries three features: `effective = 0` means stored-but-inert, which is branch isolation, review queues, and relay hosting (a pushed op can never rebind the relay's own code) as one mechanism.
- Refold is asserted byte-identical per name binding, not per row count, so a fold that binds a name to the wrong hash is a red test.
- Time is a designed component: one process-monotonic authoring clock, advanced past every received stamp, so recency is a logical clock and a fast wall clock cannot win every future last-writer-wins.
- New op kinds: `Unbind` (a name ends; content stays; a stamped tombstone defeats late arrivals), `Decision` (a human judgment -- override, ack, pin -- as one synced op), `BranchEvent` (a merge or archive travelling). `SetName` records the hash it replaced: that is how a merge tells "your edit of mine" from "we both invented this name".
- Gone with them: the fold's rename heuristic, which auto-unlisted every other name sharing a hash -- identical content is one item, so it was taking out colleagues' names. Retiring a name is an explicit `Unbind` now.
- The F#/Dark line, stated once: F# keeps parse, hash, serialize, the fold, and atomic or hot-path SQL; Dark decides everything else and prints every sentence. Last-writer-wins, the pin order and the wire shape exist in both languages on purpose, pinned by shared test tables; `PTConformance` fails the build if the type mirror drifts.

## Branches and merges

- A branch is a stable uuid; its name is only an alias, non-unique on purpose, so two machines both minting `fix-auth` stay two branches. Rename is one UPDATE.

```bash
$ dark switch fix-metrics
started branch "fix-metrics".
$ dark fn Demo.Metrics.rate "(hits: Int64) : Int64 = hits * 3L"
✓ Created function: Demo.Metrics.rate
$ dark eval 'Demo.Metrics.rate 5L'               # live on the branch, no build step
15
$ dark --branch main eval 'Demo.Metrics.rate 5L'
Error: Demo.Metrics.rate not found
$ dark commit "triple the rate" -y && dark merge fix-metrics
commit aa5eab45 -- 2 ops.
Merged 2 op(s) from branch "fix-metrics" into its parent.
```

- Its ops sit in the one log at `effective = 0`, overlaid nearest-binding-first; branches off branches inherit the whole chain; a branch's `Unbind` masks main's binding underneath.
- A branch resolves through COMMITTED work only: a parent's draft, main's included, is invisible downstream. WIP is private to where it was authored, in both directions.
- `locations` is main-only now (it was branch-scoped). A direct read answers about main even standing on a branch -- the recurring bug class while building this -- so a tripwire test fails any new read that does not mark itself main-scoped.
- Merge stopped gating on WORKFLOW: the NotRebased blocker is gone, and having WIP elsewhere in the tree never blocks. Into main it flips the frontier effective and folds; into a parent branch it retags. It still refuses two things, both of which commit refuses too: uncommitted work on the branch being merged (commit it or discard it -- otherwise it arrives as somebody else's history with no message explaining it), and definite at-rest type errors, with the same typed-out `--allow-type-errors` way past. Merging around commit's gate was landing exactly what that gate exists to stop.
- Rebase is base-moving, not replay: it accepts the parent's state and repoints the branch's own callers of what moved, reporting "you both changed X" apart from "your callers followed".
- Per-name fork bases (captured at first touch) replace the whole-branch base commit; merge, rebase and conflict detection all compare against them.
- A merge or archive commits the `BranchEvent` it authors, so the fact travels; an event arriving before its bundle parks and re-arms.
- Conflicts are not new here, and a separate PR is reworking that space; this one changed mechanics only: one base-agnostic detector for merges AND sync imports, lineage detection with no recorded base, content-derived conflict ids (idempotent re-detection), the recency loser recorded instead of dropped, and "keep mine" authoring a NEW op rather than re-stamping a travelled one.
- Commit ids are minted in Dark, content-derived over message, author, stamp, parent and op ids: the same work on the same history gets the same id on two machines. A commit is a stamp, not a gate; authoring stays live-on-write.

## Propagation rework

Main baked the cascade into the log: `PropagateUpdate` froze one store's repoints under a random uuid, and `RevertPropagation` referenced those uuids. None of that can sync -- same cascade, two machines, two ops that never dedupe; a frozen repoint list replays the wrong blast radius elsewhere. Both cases are gone; propagation writes only ordinary ops.

- The cascade runs at the edit, honors pins, and lands its repoints in your draft tagged as propagation's rather than yours -- blessed or discarded together at commit:

```bash
$ dark fn Demo.Trip.legMiles "(km: Int64) : Int64 = km * 5L"
Updated function: Demo.Trip.legMiles
  2 dependents repointed: Demo.Trip.cost, Demo.Trip.summary
  (`dark propagate pin <name>` undoes one and keeps it put)

$ dark eval 'Demo.Trip.summary 10L'   # two fns from the edit; no build, no stage
151                                    # was 61
```

- If a repoint breaks a dependent, the save says so: `note: 1 repointed dependent now fails typecheck`.
- `dark propagate pin` writes a `Decision`: it commits, syncs, survives a reload, and a colleague's machine honors it. Policy resolves item -> module -> owner, explicit rows only; a branch inherits its parent chain's pins.
- A pin is retroactive: pinning after a repoint rolls the item back and un-stages the repoint. `dark undo` walks authoring history; every version stays reachable.
- The constraints subsystem (`dark constraints` / `ack`) reports the three situations needing a human: a stale usage that must not auto-repoint (the declared steady state, not an error), a name that became a different kind of thing, and a decision of yours a peer's later change displaced (authorship-checked). Findings have stable ids, so acks stick.
- The commit gate rides the same rails: `commit` refuses definite type errors or unresolved names on live items -- typed overrides must be spelled out; `-y` alone never waves them through.

## Sync, and the relay

Setup is one command, once:

```bash
$ dark sync setup
  This instance's name [inst-4f21]: ada
  Relay url: https://your-relay.example
  Write secret (not shown; Enter to skip -- reading needs none): ········
Set up. `dark sync` from now on; `dark sync setup` again to change any of it.
```

```mermaid
sequenceDiagram
    participant A as machine A
    participant R as relay
    participant B as machine B
    A->>A: author + eval (live on write)
    A->>A: dark commit
    A->>R: dark sync (committed ops + their commit rows)
    Note over A,R: a draft stays local, and push says so
    B->>R: dark sync
    R-->>B: same ops, same commit id, same message
    B->>B: eval the new fn, seconds after A wrote it
    B->>B: branch, fix, commit, dark branch push
    A->>R: dark sync status
    R-->>A: a branch from B sits unpulled, with the pull command
    A->>A: pull, review, amend, push back
```

- The old sync was symmetric peer-to-peer (unauthenticated servers, per-peer cursors, poll daemons); all deleted. One relay, connected once, that stores and forwards and never folds -- the worst a bad relay can do is withhold.
- Push sends committed ops only, each travelling with its commit, so the receiver files work under the author's commit: same id, same message, everywhere. Commits are rows, not a chain; a missing parent breaks nothing.
- Push asks what the relay holds (a per-relay SET of pushed ops), not where a cursor stopped; relay paging is by arrival order, never origin stamp. A wiped relay is detected and clients repopulate it -- losing the relay loses availability, not history.
- The write secret never reaches Dark: F# attaches it per configured origin, the config getter raises on credential keys, a native scan refuses any outbound bundle containing it, and the relay compares in constant time, failing closed when unset. Branch reads are gated like writes.
- First sync got cheap: brotli compression (a real page: 2.9 MB -> 280 KB) plus prefetch pipelining took first pulls from hundreds of MB to tens.
- Discovery is built in: `sync status` names colleagues' unpulled branches with the exact command; a pull prints the arriving commits' subjects. `dark review pull` stages a peer's ops inert; approve folds, reject archives. And set-and-forget exists: `dark apps enable sync` runs an auto-starting daemon that syncs every 30s (a pull landing mid-edit is proven safe), and `dark config set sync.branches all` widens every sync to branches too -- yours pushed and pulled, colleagues' pulled.
- The relay is Dark on this same substrate (`packages/darklang/sync/relay/`), deployed from this branch: fly.io, one machine, one volume. It also serves a public package browser (`/m`, `/p`, `/search`, `/stats`).

## What a person sees

    dark switch <name>            start or move to a branch; author and eval there without touching main
    dark diff / log <branch>      what it changes vs its parent; the sequence that got it there
    dark merge / rebase           into the parent; onto the parent. Neither blocks
    dark conflicts / resolve      step through what a merge, rebase or pull auto-resolved; override
    dark connect / push / pull    a relay, by url and write secret; main's committed ops move
    dark branch push / pull       a branch moves as its own bundle, same uuid on both machines
    dark propagate pin / follow   whether an edit repoints what calls it; retroactive
    dark constraints / ack        what needs a human: stale usages, kind changes, displaced decisions
    dark discard [name]           drop a draft, or one name from it
    dark status / commit          what changed since your last commit; seal it, whole or --include=
    dark commits / show / ops     the history, a commit's contents, the log itself
    dark review                   a peer's ops staged inert on a review branch; approve lands them
    dark permissions approve      review a version's effect requirements and approve it; the name
                                  resolves to that version until you review a newer one
    dark backups                  copies of the store: take one, list, prune, restore (WAL-safe)
    dark apps enable sync         set-and-forget: an auto-starting daemon syncs every 30s
    dark                          the workbench: home, status, conflicts, commits, on a real TTY

**Existing commands, lateral changes**: `fn`/`type`/`val`/`module` author from an argument, a file, or stdin; `dark branch <typo>` no longer silently creates a branch; a dash-led argument is never a name; `dark version --local` answers from the binary alone (the update check dominates the bare command, ~11s against a dropping firewall). Every destructive verb lists what would go and takes a y/N a piped stdin cannot answer; the refusal teaches `-y`.

**SCM**: `--include=` commits part of a draft with its dependency closure computed and reported. Editing a colleague's version prints a heads-up naming them, their commit, and its age -- computed before the save. `dark commits` hides archive markers behind `--all` and says how many. The at-rest checker audits every declaration on a branch, checking the branch; the old at-save dependents report was branch-blind and is retired, its one lost moment restored as the typecheck note above. Branch choice is decided once at boot (`--branch`, `DARK_BRANCH`, stored); parallel agents on one machine use the first two.

**The workbench**: absorbed the deleted review TUI -- home is the SCM summary; changes, commits, branches and conflicts panes off one cached read; live-updating header as another machine's work arrives; text clips instead of overrunning panes.

**For agents**: `--json` on every question-answering command, declared add-only-stable, with refusal reasons as tokens; nonzero exits where a loop branches; `docs for-ai` addresses the agent as a named audience; the LSP's branch surface rewritten on the new API and starting on the process's branch.

```bash
$ dark status --json
{"branch":"main","changed":[],"conflicts":[],"constraints":0,"draftOps":0,"leftBehind":[],"propagates":[],"removed":[]}
```

## What review changed

Ocean reviewed this branch twice and filed twenty-eight findings, each with a reproduction; all twenty-eight are fixed here and the reproductions are the checks. A third round is my own, from questions Stachu asked reading the diff. The ones that changed BEHAVIOUR rather than repairing it:

**Round one**

- **Merge is gated again, on the two things commit gates.** It refuses uncommitted branch work (commit it or discard it, so it arrives under a message somebody wrote), and it runs the same at-rest type check over the committed frontier, with the same typed-out `--allow-type-errors` way past. Merging around commit's gate was landing exactly what that gate exists to stop.
- **A rename is two ops.** `dark rename <name> <new-name>` exists, and the workbench uses the same helper: the old name ends, the same content binds at the new one. Content and callers are untouched. It used to be one op leaning on a fold heuristic this PR deleted, so it added a second name and kept the first.
- **Branch isolation got three real fixes.** Propagation on a branch no longer discovers dependents through main's version of a name the branch rebound (and then overwrites the branch's own work with main's). Rebase moves bases onto the parent's *committed* state, not its live draft. A `Deprecate` authored on a branch is now visible on that branch, as an overlay -- the same shape a branch's names already had.
- **Sync bookkeeping stopped lying twice.** A branch pull records what the bundle carried rather than everything the branch holds, so a local-only commit is no longer skipped by the next push. And a main op that arrives while an unmerged branch holds the same content is promoted rather than ignored, so `pull` stops reporting 0 for work it just received.

**Round two**

- **Ops now settle by when a thing was SAID, not when it arrived.** Deprecations and propagation policies both ordered by arrival, so two machines folding the same two ops in different orders disagreed about whether a function is harmful. Both order by the op's own stamp now, and an unset policy is a tombstone rather than a delete -- deleting the row deletes the evidence a later-arriving older decision has to lose to.
- **An op can say a thing again.** Ops are content-addressed, so "put this name back to the version it held" produces the op that already exists, which dedupes and folds nothing. A revert is unsayable as a `SetName`; it is authored as the `Decision` it is, carrying a stamp so 611 -> 622 -> 611 is three decisions rather than one.
- **A lambda cannot be stored, so we stopped storing them.** `DApplicable` split: a named fn persists, a lambda does not. A serialized lambda carried only its expression id, which resolves against the execution state that made it -- so it came back dangling on the next process.
- **The relay keeps the newer copy.** A branch push replaced whatever the relay held, so an older client could overwrite a newer bundle. It compares stamp and op count now, and an identity survives the round trip.

**Round three (from reading the diff)**

- **A doc comment belongs to a NAME, not to a body.** Ten stdlib `ParseError` types share one hash -- one declaration, so one item -- and under content-keying, documenting `Int64.ParseError` documented `UInt64.ParseError` too. `UpdateDoc` is scoped to a location, and reaches a record field, an enum case or a parameter (by position, since a parameter's name is not part of identity). The declaration's own `///` stays in the item and is what a lookup by hash shows. Two people editing one doc comment now produce a recorded conflict, and `conflicts show` prints the two sentences so `override A|B` can take one.
- **An approved version is what a run resolves.** `permissions approve` recorded a version, listed it and showed it, and nothing narrowed name resolution by it -- in every release since the feature was introduced, because every caller passed `None`. It is wired into the lowering a run does, which is where a name becomes a hash.
- **One schema directory, and one answer for an existing store.** `schema.sql` became `migrations/schema/*.sql`, nine files by subsystem. More importantly there were two mechanisms for changing an existing store's shape and no written rule for choosing; the same migration got written both ways and failed both ways. `LibDB.Releases` is the only one now, and the reason is written down: a step is code, so it can look at the store before acting. A schema change is still two things -- the shape a NEW store is born with and how the change reaches one that already exists -- and forgetting the second is silent, because `CREATE TABLE IF NOT EXISTS` no-ops and the column simply is not there on a colleague's machine. So `scripts/migrations/new` scaffolds BOTH halves from one timestamp, and each names the other in a comment, so the pair greps.
- **Four files past a thousand lines were split by subject**, and the tests were put where their subject is rather than where they were written: `CliScmRound2.Tests.fs` named a review round, which tells a later reader nothing. Dark lets several files contribute to one module, so the `packageOps` and `sync` splits are physical only -- one module, one owner of `package_ops`, no call site changed. Each gate now also says in two lines why it is a gate rather than an F# or Dark test, because "it needs its own store" does not separate them and a PROCESS boundary does.
- **Committing after authoring the same declaration twice deleted the name.** Collapse keeps one naming per name at commit and kept the last, but a rebind to the hash a name already holds writes no row -- so dropping the run's first deleted the only row there was. Reachable from an editor's save button.

Three fixes went into the effect/permission system that landed on main while this branch was in review, and they are flagged for its author rather than slipped in: the bundled-caller check refused the CLI itself (it reaches the transport through a lambda of its own entry expression); raw SQLite being `Native` would have put `dark status` behind `permissions allow native`, since the SCM is Dark now and reads its own store; and the policy file was keyed on `$HOME`, so every clone, gate rundir, relay and throwaway instance shared one -- and any instance that could not write its home fell back to deny-all and then refused its own work.

One number is worse and I could not say why. The reference workload's allocation on a PUBLISHED build moved 9.1 -> 9.7 MB, steady across runs, while the debug build barely moved -- so it is not the one-off projection rebuild a schema change causes. Attributing it wants a bisect with a published build per commit, which the pre-round-3 binary cannot do (its embedded package refs predate the op vocabulary change), and `alloc-profile` sampled too little to name a caller. The budget is raised so CI is honest about where the branch is, with that gap written into the budget's own note rather than left for someone to rediscover.

## Deleted

- Tables `branch_ops`, `sync_conflicts`, `resolutions`; op kinds `PropagateUpdate`, `RevertPropagation`; the whole F#-side `BranchOp` universe.
- Twenty-six `scm*` builtins plus the conflict, dependency, blob and op-log sets; the `Darklang.Sync` namespace; the sync daemon apps; the changeset review TUI (the review queue replaces its workflow; its browsing view has no successor yet, by choice); the release-number migrator.
- The builtin allowlist, 38 entries to 0: every exemption was a wrapper that already existed with a caller going around it.

## How it was tested

- Tests moved with the code: Dark decisions tested in Dark (nine scm testfiles); every deleted suite's concerns traceable to a successor.
- New F# suites pin the model: overlay isolation both directions, refold-byte-identical bindings, two real stores converging in any arrival order, every op case round-tripping both serializers, malformed wire ops never throwing.
- A registry-driven sweep runs every command four ways (bare, `--help`, valid, plausibly wrong), on main and on a branch.
- Ten gates behind one entry point (`scripts/testing/gates`, one CI line) cover what in-process tests cannot: a live relay, a hostile relay, a published binary's first day, the TUI, the LSP, reload reproducibility, plus a meta-gate proving no gate touches the shared dev store.
- And two real machines against the deployed relay: concurrent edits converging on the same conflict id, branch round-trips, a relay wipe under live clients, an agent authoring from the CLI docs alone.

## Implementation details

- Branch projections are IN-MEMORY overlays, not tables: a PM for any branch from nothing but its id, no branch_id column anywhere, isolation by construction. Costs: invalidation discipline (every fold drops every cache) and on-demand rebuilds -- fine at ~2k branch ops, unmeasured past that; per-branch materialized projections are the known next step if overlays get hot.
- Why `branchId` left the call graph: a package manager is BUILT for a branch, so an id beside it was a second, disagreeable way to say the same thing -- and the id-taking variants answered about main while claiming otherwise more than once. `ExecutionState` carries the branch for printers; parsing from files takes none, since source text always lowers against main.
- The wire format: 64-hex hashes as tagged raw bytes (halves the log's largest category; anything else keeps the string form), a version-dispatched deserializer hook, and ONE Option-returning decoder -- a store legitimately holds ops from newer builds, so readers tolerate and writers never delete what they could not decode.
- Draft rewrites (discard, unstage, collapse-at-commit) are one transaction -- delete, re-insert, re-fold, stamps preserved; a crash mid-rewrite used to lose the draft. `sqliteExecBatch` (many statements, one transaction) is what lets Dark-decided policy write atomically.
- A binary upgrade tops up the store from its embedded seed and re-asserts your local edits as `Decision` ops when the new seed would take a name back; the release-number migrator became named idempotent schema steps. A release build refuses to ship a seed that cannot produce the binary's pinned refs.
- `Decision`/`BranchEvent` are named for travel-worthiness, not elegance; the constraints/resolutions rework in a sibling branch is expected to revisit this vocabulary, and nothing here entrenches it.
- Two tiny secret builtins earn their keep: `cryptoConstantTimeEquals` (Dark string equality short-circuits, so timing-safe compare cannot be package code) and the Sqlite builtins being capability-gated is what keeps raw SQL from reading credential rows under a narrowed grant; scoping them tighter is a named follow-up.

---

Review commit by commit: fifteen of them carry the design, each one facet in reading order -- transport prerequisites, the storage model, op identity, the fold, overlays, the F#-leaves/Dark-arrives exchange, sync and the relay, the two CLI surfaces, the workbench, the LSP, both test tiers, gates + CI, docs. The commits after them are the rebase onto the effect system and the three rounds of answers to review, kept separate so what changed since each read is visible. Then run the verbs: `dark switch x`, author, `status`, `commit`, `merge x`, `conflicts`. The remaining gaps are written down with estimates and none loses data; the two a reviewer will meet first are the content-keyed cascade (identical bodies share a hash, so editing one repoints the other's callers) and the relay browser not showing hosted ops (it correctly never folds them).

---

One outstanding annoyance, now with a plan rather than a complaint: the things that get a person TO a working instance (`install`, `sync setup`, `config set`) are command-line shaped, and bare `dark` opens the workbench -- so the experience we route everyone into cannot set itself up. The `:` bar now hands the terminal off by registry flag rather than a hardcoded list, which unbreaks running them; the rest is `notes/merge-plan/interactive-setup-2026-09-09.md`, about eleven hours, six of which are the part with a user-visible payoff.


OceanOak:
First round of review:

1. On a branch, dark discard <name> removes all operations for the name, including committed ones, so the last committed version is lost
  to reproduce:
  ./scripts/run-cli branch create repro
  ./scripts/run-cli fn /Repro.T.f '() : Int64 = 1L'
  ./scripts/run-cli commit "f v1 on branch" -y
  ./scripts/run-cli eval 'Repro.T.f ()'   // 1
  
    ./scripts/run-cli fn /Repro.T.f '() : Int64 = 2L'
    ./scripts/run-cli eval 'Repro.T.f ()'  // 2
    ./scripts/run-cli status     # draft: 1 item changed
    
    ./scripts/run-cli discard Repro.T.f -y # dropped Repro.T.f from your draft (2 ops)
    ./scripts/run-cli eval 'Repro.T.f ()'  # Error: Repro.T.f not found
    ./scripts/run-cli status     # clean: nothing changed since your last commit

2. extension broke 
The extension sends {} for dark/canMerge, dark/merge, dark/rebase and dark/rebaseStatus, where the server now requires { branchId }; it sends newName for dark/renameBranch, where the server wants name; and it still calls dark/scm/commit, dark/scm/discard, dark/scm/getCommits, dark/scm/getCommitOps and dark/scm/getWipOps, which the server no longer handles at all

3. Editing only a function's doc comment reports "Updated" and puts an op in the draft, but the stored description never changes, because the AddFn op id is the content hash, which excludes descriptions, so the edit is dropped as a duplicate.
  
    ./scripts/run-cli fn /Doc.T.f - <<'EOF'
    /// first description
    let f () : Int64 = 1L
    EOF
    ./scripts/run-cli commit "doc v1" -y
    
    ./scripts/run-cli fn /Doc.T.f - <<'EOF'
    /// second description
    let f () : Int64 = 1L
    EOF
    ./scripts/run-cli view Doc.T.f       # still "/// first description"

4.The workbench rename is broken, it doesn't really rename the function
5. Missing a rename command from the cli
6. sync setup doesn't work from the workbench
7. Main imports silently skip ops already present on an unmerged branch
   Reproduction: 
         Two instances on one relay
   A: dark branch create feat
       dark --branch feat fn /SM.T.two '() : Int64 = 2L'
       dark --branch feat commit "on feat" -y
       dark switch main
  
  B: dark fn /SM.T.two '() : Int64 = 2L'
     dark commit "on main" -y
     dark push                            # Pushed 2 ops

  A: dark pull                            # Pulled 0 new op(s)
     dark eval 'SM.T.two ()'      # Error: SM.T.two not found
     dark --branch feat eval 'SM.T.two ()'   # 2
     dark sync status                # in sync with the relay
      

8. branch pull marks local-only branch ops as already on the relay
reproduction:
  
    A: dark branch create work
       dark --branch work fn /SM.T.first '() : Int64 = 1L'
       dark --branch work commit "first" -y
       dark branch push work
       dark sync status                # no line for "work" (correct)

     dark --branch work fn /SM.T.second '() : Int64 = 2L'
     dark --branch work commit "second" -y
     dark sync status                # branch "work": 2 op(s) not on the relay   (correct)

     dark branch pull                # branch "work" is up to date.
     dark sync status                # the "work" line is gone                   (wrong)
     sqlite3 <A rundir>/data.db "select count(*) from sync_pushed where relay like '%/branch'" // 4 the relay holds 2 for this branch
  
    B: dark branch pull --from <A's identity>
       dark --branch work eval 'SM.T.second ()'    # Error: SM.T.second not found
  
    A: dark branch push work                       # sends all 4
    B: dark branch pull --from <A's identity>      # imported branch "work" (1 op(s))
       dark --branch work eval 'SM.T.second ()'    # 2

9. Propagation overwrites branch-specific code with main’s version
  repro
  dark fn /PR.T.base '() : Int64 = 20L'
  dark fn /PR.T.caller '() : Int64 = PR.T.base () + 1L'
  dark commit "main" -y
  dark branch create feat
  dark --branch feat fn /PR.T.caller '() : Int64 = 777L'
  dark --branch feat commit "caller 777" -y
  dark --branch feat eval 'PR.T.caller ()'       # 777
    
    dark --branch feat fn /PR.T.base '() : Int64 = 30L' # "1 dependent repointed: PR.T.caller"
    dark --branch feat eval 'PR.T.caller ()'       # 31, branch edit gone
    dark --branch feat view PR.T.caller            # base () + 1L
    dark --branch main eval 'PR.T.caller ()'       # 21, main untouched

10. rebase imports main’s uncommitted code into a branch
  // main
  dark fn /RX.T.root '() : Int64 = 111L'
  dark fn /RX.T.user '() : Int64 = RX.T.root ()'
  dark commit "rx v1" -y
    
    dark branch create rx
    dark --branch rx fn /RX.T.branchUser '() : Int64 = RX.T.root ()'
    dark --branch rx commit "branch user" -y
    dark switch main
    
    // main
    dark fn /RX.T.root '() : Int64 = 222L'
    dark status                                   # shows the draft, not committed
    
    dark --branch rx eval 'RX.T.root ()'          # 111
    dark --branch rx eval 'RX.T.branchUser ()'    # 111
    
    // rebase
    dark --branch rx rebase                       # "1 caller(s) now point at the parent's version: RX.T.branchUser"
    
    // the branch is now inconsistent with itself
    dark --branch rx eval 'RX.T.root ()'          # 111
    dark --branch rx eval 'RX.T.branchUser ()'    # 222
  

11. Archiving a parent strands its active child
  repro
  dark branch create parent
  dark --branch parent fn /AR.T.p '() : Int64 = 1L'
  dark --branch parent commit "p" -y
  dark branch create child                       # while on parent, so a child of it
  dark --branch child fn /AR.T.c '() : Int64 = AR.T.p () + 1L'
  dark --branch child commit "c" -y
  dark switch main
  dark --branch child eval 'AR.T.p ()'           # 1
  dark branch archive parent -y                  # archived branch "parent"
  dark --branch child eval 'AR.T.p ()'           # Error: AR.T.p not found
  dark branches                                  # child <- <id> (a branch not in this store ...)
  dark merge child                               # Merged 2 op(s) ... into its parent
  dark merge parent                              # no branch "parent"
  dark eval 'AR.T.c ()'                          # not found on main

12. Merge bypasses commit’s type-error protection
  repro
  dark branch create te
  dark --branch te fn /TE.T.bad '() : Int64 = "oops"'
  dark --branch te commit "bad" -y          # refused: at-rest type check found definite errors
  dark switch main
  dark merge te                             # Merged 2 op(s) from branch "te" into its parent.
  dark status                               # clean
  dark commits 2                            # merged branch "te"
  dark eval 'TE.T.bad ()'                   # return value expects Int64, but got String
  
  Why are we able to merge uncommitted stuff in the first place

13. Merging a child branch into its parent moves the child's ops but leaves its propagation pins recorded under the child's id, so the parent cannot see them and its next edit repoints a caller the child had pinned.
  dark branch create par
  dark --branch par fn /PN.T.base '() : Int64 = 10L'
  dark --branch par fn /PN.T.caller '() : Int64 = PN.T.base () + 1L'
  dark --branch par commit "par" -y
  dark branch create kid                      # while on par, so a child of it
  dark --branch kid propagate pin PN.T.caller
  dark --branch kid commit "pin" -y
  dark switch main
  dark merge kid                              # Merged 1 op(s) from branch "kid" into its parent.
  sqlite3 <rundir>/data.db "select substr(branch_id,1,8), name, policy from propagation_policy" # row is under kid's id, not par's
  dark --branch par fn /PN.T.base '() : Int64 = 20L'    # 1 dependent repointed: PN.T.caller
  dark --branch par eval 'PN.T.caller ()'     # 21; a pinned caller should still give 11

14. A partial commit selects namings by content hash, so --include= of one function also commits any unrelated name that happens to have the identical body, reporting it as "needed"
  repro
  dark fn /PC.T.public '() : Int64 = 987L'
  dark fn /PC.Private.unfinished '() : Int64 = 987L'
  dark commit "public only" --include=PC.T.public -y  # also committed (needed by what you named): PC.Private.unfinished
  dark status                                             # clean; nothing left in the draft

15. After dark undo, running dark discard <name> says it dropped the change but does nothing, because it does not recognise the kind of op that undo writes.
  repro
  dark fn /DU.T.f '() : Int64 = 1L'
  dark commit "v1" -y
  dark fn /DU.T.f '() : Int64 = 2L'
  dark commit "v2" -y
  dark undo DU.T.f                       # back on v1, as a draft
  dark eval 'DU.T.f ()'                   # 1
  dark discard DU.T.f -y              # dropped DU.T.f from your draft (0 ops).
  dark eval 'DU.T.f ()'                   # still 1
  dark status                                # still 1 item changed
  dark discard -y                         # discarded 1 op;
  dark eval 'DU.T.f ()'                  # gives 2

16. dark discard decides the draft is empty by counting changed name bindings, so a draft holding only a pin or a deprecation is reported as "empty" and left in place.
  repro
  dark fn /DP.T.f '() : Int64 = 1L'
  dark commit "f" -y
  dark propagate pin DP.T.f                # pinned DP.T.f
  dark status                              # clean
  sqlite3 <rundir>/data.db "select count(*) from package_ops where commit_hash is null and effective=1"   # 1
  dark discard -y                          # nothing to discard -- your draft on main is empty.
  dark propagate                           # pin DP.T.f still there; the op count is still 1

Not tested:
- A relay run from a published dark binary fails every pull with HTTP 500, because Queries.fs:84 encodes commit hashes with reflection-based JSON and release builds strip reflection out (Not tested)
- If someone else pushes while you are pushing, dark push saves a relay head that includes their ops, and your next dark pull says "Nothing new" while you are missing their work.
- A merge that fails mid-way leaves the branch's ops flagged as effective, so the next dark command's startup fold lands them on main with no merge commit while the branch still shows as unmerged.
- When two bindings of one name share a timestamp, the branch overlay lets the later-arrived op win while main's fold uses the hash tie-break from Lww.fs, so a branch can show one version and show the other the moment it is merged.

backend/src/Builtins/Builtins.Matter/Libs/PM/PackageOps.fs
```diff
@@ -524,51 +617,390 @@ let fns (pm : PT.PackageManager) : List<BuiltInFn> =
...
+      previewable = Impure
+      callEffects = set [ Effect.PackageRead; Effect.PackageWrite ]
+      deprecated = NotDeprecated }
+
+
+    // IMPORT a branch (from a portable bundle): register it, store its ops effective=0 + tag the
+    // frontier (NOT folded into main), and re-derive the per-name bases against THIS instance's main
+    // (recordNameBases -- the base is the destination's fork point). Cross-instance "branches follow
+    // me". Returns count stored.
+    { name = fn "scmImportBranchOps" 0
```
OceanOak, line 886: This builtin is behind dark branch pull and dark branch import, it writes code into the store but is labelled read-only

backend/src/Builtins/Builtins.Matter/Libs/PM/Packages.fs
```diff
@@ -603,116 +618,136 @@ let fns (pm : PT.PackageManager) : List<BuiltInFn> =
...
+    // implementation detail. Ops are content-addressed, so `SetName(name -> the old hash)` is
+    // byte-identical to the op that first bound it: it INSERT-OR-IGNOREs and folds NOTHING, so the
+    // rollback silently doesn't happen. `Override` exists precisely to say "this binding again, but now I
+    // mean it", and a pin is the same act as a conflict override -- a human overruling what the machine
+    // picked.
+    //
+    // Emitting an op rather than touching `locations` is what makes the undo sync, audit and conflict like
+    // any other authoring. Deleting ops instead would do none of that, and would be unsafe besides: a pinned
+    // dependent may be the only thing still referencing the version being deleted.
+    { name = fn "pmSetName" 0
```
OceanOak, line 619: pmSetName writes a new binding into the package store but is labelled read-only

backend/src/LibDB/PackageOpPlayback.fs
```diff
@@ -284,109 +284,100 @@ let private applyAddFn (ctx : Ctx) (fn : PT.PackageFn.PackageFn) : Task<unit> =
...
-      // this name): keep the EARLIEST origin_ts so the binding's stamp is identical on every instance
-      // regardless of fold/arrival order. Skip unless this op is strictly earlier (then re-bind to lower it).
-      // Without this the fold would re-stamp the binding with a LATER equal-hash op, and a different-hash op
-      // stamped between the two could then win on one instance and lose on another → divergence. (receiveOps'
-      // MIN-reconcile already keeps ops from arriving with a raised stamp; this makes the fold correct on its
-      // own, for any caller.)
+      // Same content re-applied: keep the EARLIEST origin_ts, so the binding's stamp is identical on every
+      // instance regardless of arrival order. Re-stamping with a later equal-hash op would let a
+      // different-hash op stamped between the two win on one instance and lose on another.
       | Some(curHash, Some curTs), Some t when curHash = itemHashStr -> t >= curTs
```
OceanOak, line 378: Setting a name back to its current value keeps the old timestamp, so a machine that gets the ops in a different order can end up on a different version? 
example:
 f = 101 (t1), f = 202 (t2), f = 101 (t3)
in order   t1, t2, t3  ->  101   (t3 beats t2)
reordered (relay served them in arrival order, e.g. two people pushed at once)  t1, t3, t2  ->  202   (t3 dropped as "same as t1", stamp stays t1, t2 beats it)

backend/src/LibDB/PackageOpPlayback.fs
```diff
@@ -454,294 +451,592 @@ let private serializeAnnotation
...
+    | PT.DecisionKind.Override _ ->
+      // Folded as a binding, not here. Kept explicit so adding a case to `DecisionKind` is a compile
+      // error in both places rather than a silent no-op in one of them.
+      ()
+
+    | PT.DecisionKind.Propagation PT.PropagationPolicy.Unset ->
+      // Clearing is a decision like any other, so it's an op -- but it's the one that removes the row rather
+      // than writing it. Still guarded by origin_ts, so a stale unset can't wipe a newer pin.
+      do!
+        exec ctx "DELETE FROM propagation_policy
```
OceanOak: why does unset delete the row? we should keep the row on unset, with the op's time and a state of "unset"

OceanOak:
Round 2:
1. Changing a function back to an earlier version does nothing: the store thinks it has already seen that change and drops it, but the CLI still says "Updated"
  dark branch create rb
  dark --branch rb fn /RA.T.b '() : Int64 = 611L'; dark --branch rb commit "611" -y
  dark --branch rb fn /RA.T.b '() : Int64 = 622L'; dark --branch rb commit "622" -y
  dark --branch rb fn /RA.T.b '() : Int64 = 611L'        # Updated function: RA.T.b
  dark --branch rb eval 'RA.T.b ()'                      # 622 **I expected to get 611 here**
  dark --branch rb status                                # clean

2. Deprecations: last to arrive wins, not newest
Marking a function harmful and later undoing it are two ops. When they reach a machine, whichever comes last wins, even if it is the older one.

3. Partial commit on a branch commits a caller without its dependency's name
commit --include=<caller> on a branch commits the caller but leaves its dependency's name in the draft, so anyone downstream can run the caller but cannot find the dependency

./scripts/run-cli branch create pb
./scripts/run-cli --branch pb fn /PB.T.dep '() : Int64 = 5L'
./scripts/run-cli --branch pb fn /PB.T.caller '() : Int64 = PB.T.dep () + 1L'
./scripts/run-cli --branch pb commit "caller" --include=PB.T.caller -y     # 3 ops; 1 op (1 item) still in the draft
./scripts/run-cli --branch pb status                                       # draft: 1 item changed  (dep's name)
./scripts/run-cli branch create pbchild                                  # while on pb, so a child of it
./scripts/run-cli --branch pbchild eval 'PB.T.caller ()'          # 6
./scripts/run-cli --branch pbchild eval 'PB.T.dep ()'             # Error: PB.T.dep not found

4. When you accept someone's work through dark review, it shows up in your history as your own commit, "merged branch review", and their commit, with their name and message, is lost

5. If you write A then B, commit B first and then A, your next commit will point at B as its parent instead of A, so A drops out of the history chain
repro
  dark fn /PL.T.a '() : Int64 = 1L'
  dark fn /PL.T.b '() : Int64 = 2L'
  dark commit "B" --include=PL.T.b -y          # commit 4d942e4a
  dark commit "A" --include=PL.T.a -y          # commit 552dd70f
  dark fn /PL.T.c '() : Int64 = 3L'
  dark commit "C" -y                           # commit d7385b91
  sqlite3 ~/.darklang/data.db "select substr(hash,1,8), message, substr(parent,1,8) from commits order by rowid desc limit 3"
      // d7385b91  C  -> 4d942e4a (B)     expected A
      // 552dd70f  A  -> 4d942e4a (B)
      // 4d942e4a  B  -> b2fad22a

  C's parent is B; it should be A. A also points at B, so A is on nobody's chain


6. dark propagate with no arguments crashes in a terminal with  "lambda not found"


7. If you pin something on a branch and then work on a child of that branch, the pin still applies there, but dark propagate on the child says nothing is pinned
repro
dark branch create par
dark --branch par fn /IP.T.base '() : Int64 = 10L'
dark --branch par fn /IP.T.caller '() : Int64 = IP.T.base () + 1L'
dark --branch par propagate pin IP.T.caller
dark --branch par commit "par" -y
dark branch create kid                          # child of par
dark --branch kid propagate                     # "no propagation choices"
dark --branch kid fn /IP.T.base '() : Int64 = 20L'
dark --branch kid eval 'IP.T.caller ()'         # 11: the pin was applied anyway

8. Resolving a conflict by name can hit an old, already-resolved one
 If a name has diverged twice, the first conflict resolved and the second still pending, conflicts override <name> picks the old resolved row, says it succeeded, and leaves the real one open.
repro
  A switch main
  A fn /CF.T.same '() : Int64 = 100L'; A commit "base" -y; A push; B pull

 // first conflict: A says 200, B says 300
  A fn /CF.T.same '() : Int64 = 200L'; A commit "A200" -y; A push
  B fn /CF.T.same '() : Int64 = 300L'; B commit "B300" -y; B pull  # 1 conflict(s) auto-resolved by recency and recorded
  B conflicts list # 1 pending: #3bc638f4  CF.T.same   A (yours, kept) 300L   B (lost) 200L
  B conflicts override CF.T.same B # overrode #3bc638f4: CF.T.same now binds 55e3ab54

  // second conflict on the same function: A says 400, B says 500
  A fn /CF.T.same '() : Int64 = 400L'; A commit "A400" -y; A push
  B fn /CF.T.same '() : Int64 = 500L'; B commit "B500" -y; B pull # 1 conflict(s) auto-resolved by recency and recorded
  B conflicts list # 1 pending: #f9b43c03  CF.T.same   A (yours, kept) 500L   B (lost) 400L
  B conflicts override CF.T.same B # overrode #3bc638f4: CF.T.same now binds 55e3ab54     <- the OLD conflict, and the 200 version
  B conflicts list # 1 pending: #f9b43c03                                  <- the real one, still open

Not tested:
- An older branch push overwrites a newer one on the relay
The relay keeps one bundle per branch and replaces it on every push, without checking which is newer. Push a branch, add to it, push again, then push the old copy: the relay serves the old copy and peers lose the newest commit.
- Identities with a space or & break sync quietly
You can name your instance anything. A space makes the relay's owner list unparseable, so automatic branch sync on every peer skips you. An & gets read as a second query parameter, so your ops are recorded under two owners and your branches cannot be pulled by name.
- Resolving a conflict by name can resolve the wrong one
  When the same function has conflicted twice, and the first conflict was already resolved, dark conflicts override CF.T.same ... picks the old resolved conflict
  instead of the new open one. It says "overrode", but the open conflict is still there and the function still runs the auto-picked version. The list only shows open conflicts; the command that acts on them does not use the same filter, so it finds the oldest row for that name, resolved or not.

PR #5755, opened 2026-09-08
Effect system follow-ups

OceanOak:


backend/src/Builtins/Builtins.Cli/Libs/File.fs
```diff
@@ -35,7 +35,7 @@ module FileError =
     let (caseName, fields) =
       match failure.kind with
       | Host.FailureKind.NotFound -> "NotFound", []
-      | Host.FailureKind.PermissionDenied -> "PermissionDenied", []
+      | Host.FailureKind.OsAccessDenied -> "PermissionDenied", []
```
StachuDotNet: Hmm, did the .dark side of this not change case name?

OceanOak: Fixed

PR #5773, opened 2026-09-17
Consistent and correct float pretty printing

StachuDotNet:
Consistent and correct float pretty printing

This PR was prompted by a reported diff between our interpreter and [the compiler](https://github.com/darklang/compiler):

> Float.toString intentionally uses correct shortest-roundtrip formatting rather than the interpreter's lossy G12 rendering (docs/compatibility/stdlib/floats-and-math.md:72).

The compiler had it right, so this brings the interpreter in line.

We were printing floats a few different ways: `Float.toString` and the user-DB writer used .NET's `"G12"` (twelve significant digits, then round), `Json.serialize` used `"G16"`. Both throw information away. `0.1 + 0.2` printed as `0.3`, two different floats could print identically, and printing then parsing didn't give you the number back. For the user DB that's data loss on disk; for the REPL it meant `let x = 0.1 + 0.2` captured `0.3` as the binding's source.

Every float-to-text path now prints the shortest digits that parse back to the same double, laid out the same way as the compiler's `Float.toString` so the two agree byte for byte: fixed notation from 1e-4 up to 1e16, scientific outside that with no zero padding (`1e-5`, `1e+16`), `.0` only on whole numbers in fixed notation. The 1e16 bound is where a double stops holding every integer exactly, so fixed notation never pads with digits the value doesn't have; Python, Swift and Rust's Debug stop at the same place. The compiler was at 1e12 (the old `G12` boundary) and moves to match.

---

- One formatter, `Prelude.floatToShortestString`, behind `Float.toString`, `Json.serialize` and the user-DB writer.
- No `.0` suffix on scientific forms. `Float.toString 1e17` gave `1e+17.0`, which `Float.parse` refuses.
- The ProgramTypes pretty-printer no longer prints the literal `0.00001` as `0.-05.0`.
- Tests pin the compiler's own vectors (both notation boundaries, the smallest subnormal, a 17-digit value) plus `toString |> parse` round-trips.

REPL output for float noise gets uglier (`0.30000000000000004` instead of `0.3`), but that's the actual number. Tidy display is a separate fn with an explicit precision, not a lossy default. Old stored user-DB floats aren't migrated; nothing depends on pre-existing data right now.


pbiggar:
Here's the range chatgpt gave me:


```
                 fixed-point range

Ada             [ never: always scientific ]
Java            1e-3  ─────────────── 1e7
C-style %g      1e-4  ────────────── 1e6       (p=6)
Go default      1e-4  ────────────── 1e6
Ruby*           1e-4  ───────────────────────── 1e15
PHP default     1e-4  ────────────────────────── 1e14
Python          1e-4  ─────────────────────────── 1e16
Swift           1e-4  ─────────────────────────── 1e16
Rust Debug      1e-4  ─────────────────────────── 1e16
Delphi          1e-5  ───────────────────────── 1e15
JavaScript      1e-6  ───────────────────────────────── 1e21
Scratch         1e-6  ───────────────────────────────── 1e21
Rust Display    [ never: always fixed ]

```

Feels like somewhere in 16-21 is right.

PR #5776, opened 2026-09-18
Add Dark-native testing

OceanOak:
This PR adds Dark-native testing. Tests are functions taking Unit and returning Stdlib.Test.T, discovered automatically

Follow-ups:
  - Run only tests affected by code changes
  - Add isolated execution and timeouts
  - Migrate existing stdlib tests to the new system

Run tests with `dark test <module-or-function>`


StachuDotNet:
FYI needs rebase

backend/src/Builtins/Builtins.Pure/Libs/Test.fs
```diff
@@ -0,0 +1,156 @@
+/// Runtime support for Dark test helpers.
+module Builtins.Pure.Libs.Test
```
StachuDotNet: I don't think any of the builtins in this module are Pure, should likely move

StachuDotNet:
This feels like this is halfway between worlds. I think we should either:
- have this not add F# (probably, represent the Tests as fns, and extend our search-for-Values-by-Type so we also support find-Fns-by-Sig or something (i.e. by some special Return type, or maybe the whole Signature block fits some shape, not sure))
- or represent Tests as a new special thing in ProgramTypes, with an AddTest Op, etc., and broadly invest in it a bit more. Fold in the at-rest checker (probably build such wholly in Dark, on top of broader Tests infra etc)

The former gets a PR merged faster, but I think tests _are_ special in our system, so let's respect their weight and invest a bit more here.

Some quick directional notes:
- I see you renamed a Tests module to Suite -- maybe TestSuite would be best?
- CliTest similarly isn't ideal -- unless these are CLI-only tests (which I think they aren't -- they should be tests across potential runtimes, I think)
- see Dev vault 05.Implementation/Constraints/dl-constraints.md for some old notes on direction, around potential modeling of constraint timing, 
- maybe we need DB tracking of when last the tests have been run, make sure that projections track that appropriately for quick querying of "tests that need to run" etc. not sure, worth thinking through
- tests involving no effects don't need to run again (unless something actually changes, like via propogation -- but in that case, the test really is changing)
- Long-term, I still feel as though "tests" should live under a broader umbrella of Constraints, where some Constraints are parse-time/dev-time, some are at-rest (these, the at-rest type checker, and all sorts of other stuff), and potentially Runtime constraints (I used to think that these might be useful, but currently they don't make sense to me. That said, that's probably overenginering for this PR -- focusing only on at-rest tests, and just calling them "tests," should be plenty.
- in case unclear, I think the whole at-rest type checker work could migrate to package land.

Once we model the PT stuff, and invest bottom-to-top in a Test idea, I think a lot of this code will actually simplify / fall away, especially on the F# side.

Right now these are run when you explicitly call the CLI command. How should the fit into the dev UX more broadly? Should they be required at commit-time? For sync? re-run in the background as you do code? How does propogation apply to tests?

https://package.elm-lang.org/packages/elm-explorations/test/2.2.1/ Elm is always good inspiration.

StachuDotNet:
One floating negative thing I don't love about "tests being their own fancy package items" is that package items always have _names_ -- but it'd be really annoying/absurd to name a bunch of tiny tests like `1 + 2 = 3`. Not sure what to do about that. Maybe we throw those into some test suite/bucket/group, and that's what a package item is. I don't love that lack of granularity, though, in terms of identifying/referencing a failure. So it's definitely a bit of an awkward fit. Anyway, I think this is worth iterating on design for, and thinking through these problems :)

PR #5778, opened 2026-09-21
Add Traits

StachuDotNet:
## The problem

Dark had no way to say "any type that can do X". `+` was a builtin typed `'a -> 'b -> 'a` that switched on the two values at runtime; the checker carried a hand-written table of which operator works on which of the fourteen numeric types; `Stdlib.Int64.max` existed fourteen times because nothing could be written over "any type with a `<`"; a user type could not have a `+`, a `==` of its own, or a `toString`.

A trait names one or more functions over one type that is left open. An implementation fills them in for one concrete type. A call picks the implementation by the type of the value it is handed. The operators, equality and `toString` are now traits with implementations for every builtin type, so nothing anyone wrote changes meaning, and a type of your own joins with one `impl`.

Perf, up front: the operators cost what they did. Two operands of one builtin numeric type, `-x` on a signed one, and `==` on anything but a record or an enum are answered before dispatch. The allocation gate reads 9.6 to 9.7 MB published against a 9.7 budget; debug 9.5 against a budget re-pinned at that value here (1.8% over the old one, paid once per process for the trait-name set and the trait refs). A dispatched call is 77 to 121 bytes over a direct call once its selection is remembered, about 3 KB the first time.

## What it looks like

```
type Point = { x: Int64; y: Int64 }

impl ToString for Point =
  let toString (p: Point) : String =
    $"({Stdlib.Int64.toString p.x}, {Stdlib.Int64.toString p.y})"

let p = Point { x = 1L; y = 2L }
ToString.toString p    // "(1, 2)"
p.toString             // the same call

let describe<'a: ToString> (v: 'a) : String =
  "value: " ++ v.toString
```

`'a: ToString` on `describe` is a bound: `describe` accepts any type with a `ToString` implementation, and refuses at the call, before its body runs, when the type has none.

Our own numerics:

```
type Vec = { x: Int; y: Int }

impl Add for Vec =
  let add (a: Vec) (b: Vec) : Vec =
    Vec { x = a.x + b.x; y = a.y + b.y }

let v1 = Vec { x = 1; y = 2 }
let v2 = Vec { x = 10; y = 20 }
v1 + v2                // Vec { x = 11, y = 22 }
```

`true + false` is now "No implementation of Stdlib.Add for Bool. Write `impl Add for <type> = ...`", which is what was true all along. The checker's table is gone; it asks whether an implementation exists.

## UX

**Syntax.** `trait Name<'a> =` with method signatures, one per line; a method may carry an effect row (`let fetch (u: 'a) :{Http} String`) that every implementation must fit under. `impl[<'a: B>] Trait for Type =` with method fns, or `let m = Existing.fn` to name a fn that already exists (no wrapper is generated). Bounds go on fns, types, traits and impls: `'a: ToString + Ord`. Rust's shape, chosen because Dark's declarations already looked like it.

**Calls.** `ToString.toString p`, or `p.toString`. A field always wins over a method of the same name. `a.add b` is a partial application of a two-argument method.

**Resolution.** A bare trait name is looked up from the current module outward, and if nothing is found, in `Stdlib`. So `impl Add for Point` anywhere means `Stdlib.Add`, unless a closer module has its own `Add`. Two traits with identical methods are one trait, whatever they are named.

**Where an implementation lives, and where it is seen.** An implementation is a package item, named for the type and the trait under the module it was written in. `impl ToString for Point` written in `Acme` is the item `Acme.Point.ToString`, and its method fn is `Acme.Point.ToString.toString`. The stdlib's `impl Add for Int64`, written in the module `Stdlib.Int64`, is `Stdlib.Int64.Add` (the type segment is not repeated when the module is already named for the type). The module is only the name: an implementation is visible to dispatch on the whole branch it is bound on, so a call anywhere finds it without opening anything, and a second one for the same trait and type anywhere on the branch is the ambiguity below.

**Authoring.** `dark trait Acme.Area "<'a> = let area (s: 'a) : Float"` and `dark impl Acme "Area for Circle = let area (c: Circle) : Float = ..."`, shaped like `fn`/`type`/`val` (inline, a file, or `-`); `dark module` for several at once. The command is `impl` because that is the keyword you write; the prose everywhere says "implementation". `rename`, `delete`, `deprecate` and `edit` know both kinds.

**Listings.** `ls` has Traits and Implementations sections; `view Stdlib.ToString` prints the trait back and `view Stdlib.Int64.Add` the implementation; `tree`, `search --trait` / `--impl`, completion, the workbench and the relay browser show both; `dark impls Add` lists every implementation as dispatch sees them.

**Coherence.** Two visible implementations of one trait for one type is legal to store and an error to call. It is an error at the call, a checker finding on save, and a standing `ambiguous-implementation` finding in `dark constraints` (counted by `status`, in `--json`). Deprecating one resolves it: a deprecated implementation is not a dispatch candidate.

**Errors.** Mixed operand types under an operator say what they did before: "Cannot perform numeric operation on Int64 and Float". A missing implementation says which trait and type and what to write. "More than one implementation of X for T is visible" points at `dark impls X`. "Cannot tell which type to dispatch Zero.zero on" asks for a type arg. In a script, an impl whose trait or method target does not resolve is a parse-time diagnostic.

**Docs.** `dark docs traits` is the one story (implement a trait you have, declare your own, bounds, where they live, the stdlib traits, ambiguity, errors); `operators`, `syntax`, `types` and `for-ai` carry one example each and point at it. Hover, semantic tokens and completions know both keywords.

**The stdlib.** `Add Sub Mul Div Mod Pow Neg` are the operators, `Ord` the comparisons, `Eq` is `==` and `!=`, `ToString` is behind every builtin type's `toString`, `Zero` and `One` are the identities. Every numeric type implements the operators (one alias line per method, next to its own fns); String implements `Add` (`"a" + "b"` is `"ab"`; `++` still works) and `ToString`; Bool, Char, DateTime and Uuid implement `ToString`. Written once over the traits: `Stdlib.max`, `min`, `clamp` (over `Ord`), `sum` and `product` (with a seed, so `[]` never has to ask which type it is), `toString`. `==` compares every value structurally unless its type implements `Eq`; builtin types are never overridden, and `List.sort` and dict keys stay structural.

**Builtins.** Seven new ones in Matter: `pmFindTrait`, `pmGetTrait`, `pmGetLocationsByTrait`, `pmFindTraitImpl`, `pmGetTraitImpl`, `pmGetLocationsByTraitImpl` and `pmImplCandidates`, each with one Dark wrapper; `pmSearchNames` and `pmSearchNamesAndHashes` return six lists instead of four. They exist because the store is F#'s and Dark reads it only through builtins; the first six are the same find/get/locations trio every item kind has (`pmFindType`, `pmGetType`, `pmGetLocationsByType`), so the Dark package manager, `view`, `rename` and the rest can treat a trait like a type. One generic `pmFindItem kind` would cut them to three, but every other kind is spelled out and a one-off saves nothing worth the exception. `pmImplCandidates` is the one genuinely new thing: it hands `dark impls` the same candidate list dispatch reads, so what is listed is what a call can reach.

## Internals

**ProgramTypes.** `PT.Trait` (type params, the first being self; supertrait bounds; methods as signatures with an optional ceiling) and `PT.TraitImpl` (the trait, its type args, the self type, own type params and bounds, methods as `(name, fn)` pairs). Both are content-addressed, named by `SetName`, carried by `AddTrait` and `AddTraitImpl`, with `FQTraitName`, `Reference` and `ItemKind` cases. `FQFnName.TraitMethod (trait, method)` is what `ToString.toString` resolves to; `bounds` sit on fns and types.

**RuntimeTypes.** `Trait.Trait` (the shape dispatch needs: the self-argument position per method) and `ImplCandidate` (trait, self type, method fns, source hash). The execution state carries `traits`, and the package manager `getTrait`, `implCandidates`, `implCandidatesByMethod` and the selection memo.

**Dispatch.** Applying a `TraitMethod` resolves the self type: explicit type args, then the self-positioned argument's runtime type, then the caller's bound through its type symbol table. It picks the implementation by the self type's head (type args break ties; a blanket `for 'a` loses to a specific one) and calls the impl fn exactly as a direct call would, so traces, ceilings and permissions see the fn that ran. Bounds are checked at fn entry, where parameter types are. Selection is memoised per (branch, trait, method, self type) under the store's cache generation; a side-loaded manager (a script's own impls) has its own memo. A candidate counts only while a name binds it on the branch asked and it is not deprecated. `Zero.zero<Int64> ()` hands the impl fn none of the trait's type args.

**Fast paths and lowering.** The operators lower at PT to RT time to their trait method, so no stored hash moved; `Builtin.negate` (what the parser stores for `-x`) lowers to `Neg.negate`; `!=` to `boolNot (Eq.equals a b)`, so one method serves both operators. `FastOps` answers two operands of one builtin numeric type, `-x` on a signed one, `"a" + "b"`, and `==` on anything but a record or an enum without dispatch; `Eq` with no implementation falls back to structural equality and the memo remembers "none". With no package refs generated yet (a fresh CI tree) the operators lower to the builtins as before.

**Checker.** A `TraitMethod` call takes its signature from the trait's method; a bounded call, an operator or a unary minus records a constraint, discharged at the end of the item (`MissingImpl`, `AmbiguousImpl`, `UnboundTypeParameter`, or a blocker while the type is unknown). A conditional impl owes its own bounds at the type it matched, so `Area List<Circle>` owes `Area Circle` at the call. Traits and impls are validated as items: the method set, each method fn unified with the trait's signature at the self type, and the impl fn's ceiling within the method's. `==` records no constraint, since every type is comparable.

**Serialization and hashing.** Binary format v3: two item encodings, op tags 15 and 16, reference tags 3 and 4. v2 blobs still read, and no existing hash moved. A trait carries its own kind tag in its canonical hash, so a trait and a same-shaped record differ. An older binary that pulls a bundle holding these ops stores them unapplied and keeps working, the designed behaviour for an op it cannot read; this is the first language feature to rely on it.

**Package manager and store.** Tables `package_traits` and `package_trait_impls(hash, trait_hash)` with an index on the trait; the fold writes them and `Impls.fs` reads the index. Dependency edges: an impl depends on its trait, its self type and its method fns; a trait on the types in its signatures; a bounded fn on its traits. Every package-manager layer (store, branch overlay, in-memory, combine, hide, ephemeral) carries the two kinds; the deferred resolver re-resolves traits, impls and bounds on every author, as it does fns. The trait-first check in name resolution is gated on the set of live trait names, a handful, not the type names.

**SQL.** An operator trait method pushes down as the SQL operator its builtin carried, `!=` as `NOT (a = b)`.

**Everywhere else.** Every switch over item kinds has five arms: the CLI, the LSP (hover, tokens, diagnostics, the file provider), the workbench, the relay browser, the Dark mirrors of PT, RT and WT, and both printers. The impl item prints one module up from its own path, as `impl Add for Int64 = let add = Stdlib.Int64.add`, and its method fns print as those lines.

**Tests.** The language testfile walks dispatch on user and builtin types, conditional impls, bounds at entry, receiver calls, `Vec + Vec`, `Eq`, `Zero`/`One`, `sum`, `ToString` and the error texts; the checker suite has one test per code and one for a conditional impl's bounds at the call; serializer and hashing suites cover the new items and pin that no existing hash moved; one CLI test walks the shell story end to end (author with `trait` and `impl`, dispatch, a rival implementation, the error, the finding, deprecate, rename, delete). Three bugs were found only by running scripts and authoring by hand, none by the suite; reviewing that way is worth an hour.

## What it means elsewhere

**SCM.** Nothing new to learn. A trait and an implementation are named, content-addressed items, so branches, drafts, commits, `discard`, `undo`, rename and deprecation treat them like a fn. A branch's implementations are its own until merged; an edited implementation on a branch stops being a candidate there while main keeps the old one.

**Sync.** The two ops travel as bytes like every other op. A peer on this build applies them; a peer on an older build stores them unapplied and reads the rest of the bundle, and applies them after upgrading. Not exercised end to end against an older binary (follow-ups).

**Conflicts and resolution.** Two branches binding the same implementation name (`Acme.Point.ToString`) to different content is an ordinary name conflict with the ordinary resolutions. Two implementations under different names for the same trait and type is not a conflict but the `ambiguous-implementation` finding, resolved by deprecating one; `ack` leaves it.

**Propagation.** A method fn is an ordinary fn, so editing it repoints the implementation's method reference and every caller follows as today. Editing a trait repoints its implementations and every bounded fn (`AffectedTrait`, `AffectedTraitImpl`).

**Effects and permissions.** A trait method's effect row is a ceiling; the checker refuses an implementation whose fn may do more (`ImplExceedsCeiling`). At runtime the impl fn runs as itself, with its own ceiling and package policy, so dispatch neither widens nor narrows what a call may do. A method with no row is pure.

## Follow-ups

**`Eq` inside containers.** Today `[u1] == [u2]` compares the two Users structurally even when `impl Eq for User` exists; only the outermost value dispatches. The structural walk (`Dval.equals`) has no access to the package manager, which is what keeps it cheap and synchronous. Three options: leave it and document it (done); consult the implementation per element type inside the walk, one memoised lookup per distinct element type, which turns a pure function into one that reads the store and adds a lookup to every list, dict and record comparison; or reject `==` on a container whose element type has an `Eq` implementation, which is a surprise of a different kind. My advice: leave it. A custom `Eq` is rare, containers of such types rarer, and a type that needs it can compare elements explicitly; if it bites, the second option is a day and I would take the cost only then.

**A warning when a record field shadows a trait method.** `p.show` means the field when `Point` has one and the method when it does not, so adding a field named `show` to `Point` silently changes what `p.show` means. The checker could flag a record that gains a field named like a method of a trait it implements. The checker has no warning severity today, only `Failed` (blocks commit) and `Incomplete`; a warning is a new severity through the checker, its Dark mirror and the CLI report, so this waits for that.

**Pulling these ops into an older binary.** The rule is that an op the deserializer rejects is stored and left unapplied, and every reader tolerates it; the serializer tests cover the reading side. What is not tested is the whole path: a v2 binary pulling a bundle with `AddTrait` ops, continuing to work, and applying them after an upgrade. It needs an older binary built and two stores; half a day when one is to hand.

**`$"{p}"` calling `ToString`.** Interpolation is one synchronous instruction; dispatching from inside it means lowering each non-String segment as a call. Half a day; `Stdlib.toString p` works today.

**The wasm client.** The browser REPL lowers declarations itself and reports `trait` and `impl` as unsupported rather than dropping them; running code that uses the stdlib traits works, since dispatch is in the shared interpreter. Porting means teaching `Repl.fs` the two item kinds the way the CLI host was taught: half a day.

**Deprecating the polymorphic builtins.** `Builtin.add` and friends stay: the fast path restates them, `Float.mod` wraps one, and the checker uses them when called by name. Removing them is a day and not user-facing.

**Out of scope by decision.** Trait objects (`dyn`), method-level generics, `where` clauses, and the compiler's side of dispatch (bounds as dictionary parameters, monomorphisation resolving through the impl index), which is a separate effort in the compiler clone.


pbiggar:
This looks great. One question I have for traits (and I did read this but I didnt understand if it was covered) is how to deal with modules, as modules don't meaningfully "open" like they do in eg C++.

For example:

```
trait Show<'a> =
  let show (value: 'a) : String
```

```
module Stdlib.Int64

type Int64
```

```
module Paul.Internal

impl Show for Stdlib.Int64
   ...
```

```
module Darklang

let x = show 56L
```

In the last one, how does Darklang know that Show is implemented for Int64? to my knowledge, Paul.Internal is not open (and how do we know to use Paul.Internal and not Stachu.Internal?).

StachuDotNet:
> how does Darklang know that Show is implemented for Int64? to my knowledge, Paul.Internal is not open

Right now, the name resolver works it out ([F#](https://github.com/StachuDotNet/dark/blob/60f79cebf111f933ae4f29304194f70fb036ec9d/backend/src/LibParser/NameResolver.fs#L268), [Dark](https://github.com/StachuDotNet/dark/blob/60f79cebf111f933ae4f29304194f70fb036ec9d/packages/darklang/languageTools/nameResolver.dark#L369); both do this).

```fsharp
module Paul.Internal

impl Show for Int64 =
  let show (n: Int64) : String =
    "int " ++ Stdlib.Int64.toString n
```

```fsharp
module Darklang

let x = Show.show 56L    // "int 56"

// `Show.show` is in a function position, so the fn resolver takes it, and it
// tries two readings in order:
//   1. a trait method. Is any trait named `Show`? Resolve `Show` the way a type
//      name resolves: this module outward, then as written, then `Stdlib.Show`.
//      Does that trait have a method `show`? Yes to all three, so the name is
//      TraitMethod(Show#a1b2, "show", None)
//   2. otherwise a plain fn `show` in a module `Show`, same walk.
//
// when you save: looks through every impl of Show on this branch, keeps the one
//                written for Int64, and stores it as the third part of the name
//   TraitMethod(Show#a1b2, "show", Some Paul.Internal.Int64.Show.show#c3d4)
```

It finds them in an index of every impl, kept by trait ([`Impls.fs`](https://github.com/StachuDotNet/dark/blob/60f79cebf111f933ae4f29304194f70fb036ec9d/backend/src/LibDB/Impls.fs#L21)). The module only names the impl.

---

> how do we know to use Paul.Internal and not Stachu.Internal?

The system expects one implementation of a trait per type. Today a second one stops you.

```fsharp
module Paul.Internal

impl Show for Int64 =
  let show (n: Int64) : String =
    "int " ++ Stdlib.Int64.toString n


module Stachu.Internal

impl Show for Int64 =
  let show (n: Int64) : String =
    Stdlib.Int64.toString n


Show.show 56L
// Error: More than one implementation of Show for Int64 is visible (2)
```

Based on this thread, I'm making it a conflict with one side auto-chosen, so a sync never blocks you.

- the newer `AddTraitImpl` op wins, the rule the SCM already uses for two writes to one name
- it stays chosen until you ack it, override it, or deprecate one
- the choice is stored with the call when you save, so a newer implementation reaches you through propagation, like any other update
- an override that says "always mine on this machine" comes later

```fsharp
Show.show 56L    // "int 56", Paul.Internal wrote his later
                 // stored: TraitMethod(Show#a1b2, "show", Some ...Paul...show#c3d4)

// dark constraints
// #a1b2c3d4  Stachu.Internal.Int64.Show
//            and Paul.Internal.Int64.Show both implement Show for Int64;
//            a call runs Paul.Internal.Int64.Show, written later
```

